PR #43, opened 2021-02-11
Make replay work properly, add more testing

Sushisource:
Replay wasn't tested or working properly yet. This adds a basic replay test and makes it work, as well as an integ test with parallel timers.

The test code is quite verbose. @vitarb and I discussed a desire to make a nice testing API that can reduce verbosity and allow us to author tests quickly.


vitarb (approved):

Sushisource:
Github completely hosed this up when we took it public

PR #44, opened 2021-02-11
Add shutdown API

Sushisource:
Allows the lang SDK to ensure server will no longer be polled while performing orderly shutdown.

Depends on https://github.com/temporalio/sdk-core/pull/43 being merged first.


Sushisource:
Also hosed by Github when we took repo public

PR #47, opened 2021-02-12
Add shutdown API

Sushisource:
Allows the lang SDK to ensure server will no longer be polled while performing orderly shutdown.

Depends on #46 being merged first.


src/lib.rs
```diff
@@ -58,6 +65,13 @@ pub trait Core: Send + Sync {
 
     /// Returns an instance of ServerGateway.
     fn server_gateway(&self) -> Result<Arc<dyn ServerGatewayApis>>;
+
+    /// Eventually ceases all polling of the server. [Core::poll_task] should be called until it
+    /// returns [CoreError::ShuttingDown] to ensure that any workflows which are still undergoing
+    /// replay have an opportunity to finish. This means that the lang sdk will need to call
+    /// [Core::complete_task] for those workflows until they are done. At that point, the lang
+    /// SDK can end the process, or drop the [Core] instance, which will close the connection.
+    fn shutdown(&self) -> Result<()>;
```
mfateev, line 74: In Java we have two ways to shutdown: shutdown and shutdown now. The former is a clean shutdown and the latter is an immediate shutdown without waiting for tasks to complete. 
There is also `awaitTermination` that allows waiting for shutdown for a specified timeout. 

I think we need to support all the above features. But I'm not saying the API should be the same.

Sushisource: @mfateev Not sure either of those actually belong in core. Lang SDK can decide all by itself if it wants to exit abruptly (it can just terminate the process) -- same with the timeout. If it has not polled core enough to drain the queue completely, it can know that and decide to terminate then as well.

vitarb (approved):

src/lib.rs
Sushisource: Gonna merge this so I can use it in some of the existing PRs, but we can definitely come back to this if needed.

PR #48, opened 2021-02-15
Added randomness seed update on workflow reset

mfateev:
On workflow reset a `WorkflowTaskFailed` with `WORKFLOW_TASK_FAILED_CAUSE_RESET_WORKFLOW` cause is written to the history. It contains a new run-id. The new run-id must be used for random seed at this point to avoid producing duplicated UUIDs and random values.

[Here is ](https://github.com/temporalio/sdk-java/blob/30420212a26283cf7cbe6339512058a3186074e8/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowTaskStateMachine.java#L124) the relevant Java code.

### DIFF
--- a/src/machines/timer_state_machine.rs
+++ b/src/machines/timer_state_machine.rs
@@ -227,9 +227,7 @@ impl WFMachinesAdapter for TimerMachine {
                 timer_id: self.shared_state.timer_attributes.timer_id.clone(),
             }
             .into()]),
-            TimerMachineCommand::AddCommand(_) => {
-                unreachable!()
-            }
+            TimerMachineCommand::AddCommand(_) => unreachable!(),
         }
     }
 }
--- a/src/machines/timer_state_machine.rs
+++ b/src/machines/timer_state_machine.rs
@@ -413,8 +413,6 @@ mod test {
         let _enter = s.enter();
 
         let (t, mut state_machines) = workflow_reset_hist;
-        let initial_seed = state_machines.randomness_seed;
-        assert!(initial_seed > 0);
         let commands = t
             .handle_workflow_task_take_cmds(&mut state_machines, Some(2))
             .unwrap();
@@ -424,7 +422,6 @@ mod test {
             commands[0].command_type,
             CommandType::CompleteWorkflowExecution as i32
         );
-        assert_ne!(initial_seed, state_machines.randomness_seed);
     }
 
     #[rstest]
@@ -433,7 +430,6 @@ mod test {
         let _enter = s.enter();
 
         let (t, mut state_machines) = workflow_reset_hist;
-        let initial_seed = state_machines.randomness_seed;
         let commands = t
             .handle_workflow_task_take_cmds(&mut state_machines, Some(1))
             .unwrap();
@@ -452,6 +448,5 @@ mod test {
             commands[0].command_type,
             CommandType::CompleteWorkflowExecution as i32
         );
-        assert_ne!(initial_seed, state_machines.randomness_seed);
     }
 }
--- a/src/machines/timer_state_machine.rs
+++ b/src/machines/timer_state_machine.rs
@@ -287,8 +287,11 @@ mod test {
         });
 
         let mut t = TestHistoryBuilder::default();
-        let mut state_machines =
-            WorkflowMachines::new("wfid".to_string(), "runid".to_string(), Box::new(twd));
+        let mut state_machines = WorkflowMachines::new(
+            "wfid".to_string(),
+            "54619485-18AC-4970-9BF2-D24F875F7816".to_string(),
+            Box::new(twd),
+        );
 
         t.add_by_type(EventType::WorkflowExecutionStarted);
         t.add_workflow_task();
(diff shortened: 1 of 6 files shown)


mfateev:
The update path is ready for review. But before this PR is merged I need to change the WorkflowMachines to keep both runIds and send the update to the language SDK.

src/machines/timer_state_machine.rs
```diff
@@ -271,13 +272,8 @@ mod test {
...
             We have two versions of this test, one which processes the history in two calls,
-            and one which replays all of it in one go. The former will run the event loop three
-            times total, and the latter two.
-
-            There are two workflow tasks, so it seems we should only loop two times, but the reason
-            for the extra iteration in the incremental version is that we need to "wait" for the
-            timer to fire. In the all-in-one-go test, the timer is created and resolved in the same
-            task, hence no extra loop.
+            and one which replays all of it in one go. Both versions must produce the same number
+            of activations.
```
Sushisource, line 276: I mentioned in the SDK channel that Java in fact has this same behavior in the timer test, where the event loop runs three times - but that doesn't necessarily mean there are three activations. I will look into this more and confirm that there are only two activations, but three instances of driving the machines.

src/machines/workflow_task_state_machine.rs
```diff
@@ -99,7 +106,29 @@ impl TryFrom<HistoryEvent> for WorkflowTaskMachineEvents {
             }),
             Some(EventType::WorkflowTaskTimedOut) => Self::WorkflowTaskTimedOut,
             Some(EventType::WorkflowTaskCompleted) => Self::WorkflowTaskCompleted,
-            Some(EventType::WorkflowTaskFailed) => Self::WorkflowTaskFailed,
+            Some(EventType::WorkflowTaskFailed) => Self::WorkflowTaskFailed(WFTFailedDat {
+                // TODO(maxim): How to avoid clone of attributes. We need to borrow e for the
+                // MalformedEvent down there. But forcing clone just for that doesn't make sense.
```
Sushisource, line 111: The key here is that malformed event doesn't borrow `e`, it consumes it, so you somehow need to tell the compiler "if this struct has the data I need, then consume it and take `new_run_id` out, but if it doesn't, then give it to `MalformedEvent` which will also consume it."

Actually doing that exact thing is probably not worth the effort over just cloning, but it doesn't need to be all of the attributes. You only care about the run id, so the thing to do here would be to attempt to extract that first, clone it, and if it didn't work feed `e` to `MalformedEvent`. That's similar to what I do above with `WorkflowTaskStarted`. Cloning a run id is very cheap in the grand scheme of things compared to complicating the code to avoid it -- but you might want to give it a shot anyway. Puzzles like this are a great way to learn Rust. I will say that Rust often puts you in this mindset of trying to never clone things or `Box` them or whatever because it's super fun to play the optimization game, but you have to remember Java is cloning and `Box`ing things *all over the place*, so it's often worth stepping back and remembering it's probably not worth optimizing prematurely.

Sushisource (commented):
Cool! Glad you got to get your hands dirty in here. We should explore the three-event-loops thing, as I mentioned that behavior exists in Java.

src/machines/workflow_machines.rs
```diff
@@ -441,3 +460,13 @@ impl WorkflowMachines {
         }
     }
 }
+
+pub(super) fn uuid_to_randomness_seed(run_id: &str) -> i64 {
+    let uuid = Uuid::parse_str(run_id).unwrap();
```
Sushisource, line 465: We don't want to unwrap - see other comment.

src/machines/workflow_machines.rs
```diff
@@ -44,7 +48,7 @@ pub(crate) struct WorkflowMachines {
     replaying: bool,
     /// Workflow identifier
     pub workflow_id: String,
-    /// Identifies the current run and is used as a seed for faux-randomness.
+    /// Identifies the current run
     pub run_id: String,
```
Sushisource, line 52: So, since we now have an expectation later on that this is in fact a well formed UUID, we should probably represent it as one here and make the `WorkflowMachines` constructor do it. The constructor is already fallible, so that deals with the parsing unwrap problem

mfateev: I've updated uuid_to_randomness_seed to return Result.

Sushisource: @mfateev I think it'd be nice to have `WorkflowMachines.run_id` be a `UUID` -- and handle the error any time it's being set or in the constructor. Then `uuid_to_randomness_seed` can be infallible.

We might want to make an error type specific to this as well

mfateev: It would be weird to parse and stringify it all the time. The run_id is just passed around as a string.

Sushisource: The reason would be that we're just delaying the inevitable -- if we know that `run_id` is supposed to be a well-formed UUID, then we should check that sooner rather than later. Crashing when we need to use it as a seed is incidental, if it was supposed to be well-formed all along. If it *doesn't* actually need to be well-formed, then that's maybe different, but it doesn't sound like that's true.

src/machines/timer_state_machine.rs
Sushisource: @mfateev I did confirm this was true, here and in Java, would appreciate your take on this and we may want to restore some of the original comment.

Sushisource (approved):

PR #59, opened 2021-03-03
Make shutdown interrupt long poll

Sushisource:
This fixes a problem where if polling was already ongoing the shutdown flag would not interrupt it properly.

Also fixes a possible panic if shutdown is called twice.


bergundy (approved):
LGTM, verified shutdown with sdk-node.

bergundy:
@Sushisource one thing that I have noticed is that requesting shutdown twice causes core to not emit the `Core::shutdown` error.

Sushisource:
> @Sushisource one thing that I have noticed is that requesting shutdown twice causes core to not emit the `Core::shutdown` error.

I'll take a look at that

Sushisource:
@bergundy Yeah so that caused a panic... which seemingly should've been the more obvious thing than it no longer returning shutdown. Are you somehow auto-re-initializing core if it crashes? Probably want to avoid that and just crash as well, otherwise we might end up with weird undefined behavior.

bergundy:
@Sushisource I'm not re-initializing core, I accidentally encountered the double shutdown panic while trying to figure out shutdown in `sdk-node`.
I had to add some state of in the node consumer to prevent calling shutdown twice which can happen since we expose `shutdown()` in the worker API.

vitarb (approved):

PR #63, opened 2021-03-08
Activity

vitarb:
WIP implementation of activities


src/lib.rs
```diff
@@ -62,17 +70,24 @@ pub type Result<T, E = CoreError> = std::result::Result<T, E>;
...
 /// worker(s) using it.
 pub trait Core: Send + Sync {
-    /// Ask the core for some work, returning a [Task], which will eventually contain either a
-    /// [protos::coresdk::WfActivation] or an [protos::coresdk::ActivityTask]. It is then the
-    /// language SDK's responsibility to call the appropriate code with the provided inputs.
+    /// Ask the core for some work, returning a [Task], which will contain a [protos::coresdk::WfActivation].
+    /// It is then the language SDK's responsibility to call the appropriate code with the provided inputs.
     ///
     /// TODO: Examples
+    /// TODO: rename to poll_workflow_task and change result type to WfActivation
```
Sushisource, line 77: Might as well do the rename here

src/lib.rs
```diff
@@ -62,17 +70,24 @@ pub type Result<T, E = CoreError> = std::result::Result<T, E>;
...
     fn poll_task(&self, task_queue: &str) -> Result<Task>;
 
+    /// Ask the core for some work, returning a [protos::coresdk::Task], which will contain a [protos::coresdk::ActivityTask].
```
Sushisource, line 80: We should get rid of top-level Task -- though we can do that in the "refactor proto" PR if desired

Sushisource: You might wanna do it now though just because it removes a lot of pointless matching in this PR... up to you

src/lib.rs
```diff
@@ -380,6 +445,9 @@ pub enum CoreError {
     /// When thrown from complete_task, it means you should poll for a new task, receive a new
     /// task token, and complete that task.
     UnhandledCommandWhenCompleting,
+    /// Indicates that underlying function returned Ok, but result type was incorrect.
+    /// This is likely a result of a bug and should never happen.
+    UnexpectedResult,
```
Sushisource, line 450: Yeah I think this goes away if we get rid of top-level task type and only accept appropriate completions, so probably worth fixing those in this PR

src/machines/activity_state_machine.rs
```diff
@@ -42,17 +58,155 @@ fsm! {
...
-pub(super) enum ActivityMachineError {}
+#[derive(Debug, derive_more::Display)]
+pub(super) enum ActivityMachineCommand {
+    #[display(fmt = "Complete")]
+    Complete(Option<Payloads>),
+}
+
+#[derive(Debug, Clone, derive_more::Display)]
+pub(super) enum ActivityCancellationType {
+    /**
```
Sushisource, line 76: Make these actual docstrings

src/machines/activity_state_machine.rs
```diff
@@ -42,17 +58,155 @@ fsm! {
...
+}
 
-pub(super) enum ActivityCommand {}
+impl Default for ActivityCancellationType {
+    fn default() -> Self {
+        ActivityCancellationType::TryCancel
+    }
+}
+
+/// Creates a new, scheduled, activity as a [CancellableCommand]
```
Sushisource, line 98: Comment is wrong now, I probably didn't update place you copied it from so that needs to be fixed too

src/machines/activity_state_machine.rs
```diff
@@ -42,17 +58,155 @@ fsm! {
...
+        &self,
+        event: &HistoryEvent,
+        has_next_event: bool,
+        my_command: ActivityMachineCommand,
+    ) -> Result<Vec<MachineResponse>, WFMachinesError> {
+        Ok(match my_command {
+            ActivityMachineCommand::Complete(result) => vec![ResolveActivity {
+                activity_id: self.shared_state.attrs.activity_id.clone(),
+                result: Some(ActivityResult {
+                    status: Some(activity_result::Status::Completed(ActivityTaskSuccess {
```
Sushisource, line 185: I assume we're intentionally only handling success case right now?

src/machines/activity_state_machine.rs
```diff
@@ -163,7 +323,10 @@ impl StartedActivityCancelCommandCreated {
 pub(super) struct StartedActivityCancelEventRecorded {}
 
 impl StartedActivityCancelEventRecorded {
-    pub(super) fn on_activity_task_completed(self) -> ActivityMachineTransition {
+    pub(super) fn on_activity_task_completed(
+        self,
+        attrs: ActivityTaskCompletedEventAttributes,
+    ) -> ActivityMachineTransition {
         // notify_completed
         ActivityMachineTransition::default::<Completed>()
```
Sushisource, line 331: I believe this needs to issue the `Complete` command as well, given in Java the same thing  happens here as from `Started`->`Completed`

src/machines/test_help/workflow_driver.rs
```diff
@@ -37,6 +39,9 @@ pub(in crate::machines) struct TestWfDriverCache {
     /// means keeping the workflow suspended in a future somewhere. I tried this and it was hard,
     /// but ultimately it's how real workflows will need to work.
     unblocked_timers: DashMap<String, bool>,
+
+    // Holds a mapping of activity id -> is completed
+    completed_activities: DashMap<String, bool>,
```
Sushisource, line 44: Was hoping you wouldn't need to do this but looks like you got it working which is nice. I should be able to get rid of all this soon

src/machines/workflow_machines.rs
```diff
@@ -57,6 +59,9 @@ pub(crate) struct WorkflowMachines {
     /// TODO: Make this apply to *all* cancellable things, once we've added more. Key can be enum.
     timer_id_to_machine: HashMap<String, MachineKey>,
 
+    /// TODO document
+    activity_id_to_machine: HashMap<String, MachineKey>,
```
Sushisource, line 63: We should combine these ID maps by using an enum as the key, mentioned in above TODO

src/pollers/mod.rs
```diff
@@ -100,8 +104,13 @@ pub trait ServerGatewayApis {
     ) -> Result<StartWorkflowExecutionResponse>;
 
     /// Fetch new work. Should block indefinitely if there is no work.
-    async fn poll_workflow_task(&self, task_queue: String)
-        -> Result<PollWorkflowTaskQueueResponse>;
+    async fn poll_task(&self, req: PollTaskRequest) -> Result<PollTaskResponse>;
+
+    /// Fetch new work. Should block indefinitely if there is no work.
```
Sushisource, line 109: These docstrings should be more specific -- and I think we should get rid of the generic `poll_task`

src/lib.rs
vitarb: Yes, I thought about doing that, but didn't want to bloat the scope of this change. Also making result type WfActivation would require moving task token from Task to WfActivation level.

src/lib.rs
vitarb: Makes sense, we can move task token inside of each result type.

src/machines/activity_state_machine.rs
vitarb: Yes, I'll add failure and cancelation soon as well.

src/machines/activity_state_machine.rs
vitarb: Yep, it should be the same thing as in the Started->Completed.

src/machines/workflow_machines.rs
vitarb: Are we sure that activity and timer IDs would never clash?

Sushisource: They can't. Enum as key. The key is:

```rust
enum CommandID {
    Timer(String),
    Activity(String)
}
```

src/machines/activity_state_machine.rs
Sushisource: This is resolved but the code isn't changed -- not sure if you meant to do that?

src/lib.rs
```diff
@@ -526,7 +533,47 @@ mod test {
         .unwrap();
     }
 
-    #[rstest(hist_batches, case::incremental(&[1, 2]), case::replay(&[2]))]
+    #[rstest(core,
+    case::incremental(single_activity_setup(& [1, 2])),
+    case::replay(single_activity_setup(& [2]))
```
Sushisource, line 538: The weirdo formatting here with the space after `&` and lack of indentation would be nice to fix

src/pollers/mod.rs
```diff
@@ -113,26 +113,35 @@ pub trait ServerGatewayApis {
     async fn poll_activity_task(&self, task_queue: String) -> Result<PollTaskResponse>;
 
     /// Complete a task by sending it to the server. `task_token` is the task token that would've
-    /// been received from [PollWorkflowTaskQueueApi::poll]. `commands` is a list of new commands
+    /// been received from [poll_workflow_task_queue] API. `commands` is a list of new commands
```
Sushisource, line 116: Should just be `[poll_workflow_task]`

Sushisource: Looks like the other added comments have the same extra `_queue`

src/lib.rs
```diff
@@ -62,17 +67,24 @@ pub type Result<T, E = CoreError> = std::result::Result<T, E>;
...
+    /// Ask the core for some work, returning a [Task], which will contain a [protos::coresdk::WfActivation].
+    /// It is then the language SDK's responsibility to call the appropriate code with the provided inputs.
     ///
     /// TODO: Examples
+    /// TODO: rename to poll_workflow_task and change result type to WfActivation
     fn poll_task(&self, task_queue: &str) -> Result<Task>;
 
+    /// Ask the core for some work, returning a [protos::coresdk::Task], which will contain a [protos::coresdk::ActivityTask].
+    /// It is then the language SDK's responsibility to call the completion API.
+    fn poll_activity_task(&self, task_queue: &str) -> Result<Task>;
```
Sushisource, line 79: We need some tests for this API -- ideally we can test it in the existing test you made, simulating what the worker would need to do in order for the activity result to actually end up in the history (which would only apply to the incremental version, not the full replay version, like we discussed)

vitarb: The thing is that if we do it with the fake core, these API interactions won't be triggering state machine changes and hence all we'll be testing is that core calls the server on poll/complete. Which is a legitimate use case to test, I just don't think it should be a part of that history replay flow.

src/lib.rs
vitarb: Fixed.

src/machines/activity_state_machine.rs
vitarb: Changed to:
/// Creates a new activity state machine and a command to schedule it on the server.

src/machines/test_help/workflow_driver.rs
vitarb: We can ditch it together with the test in the `activity_state_machines.rs`, since we've agreed to use the style similar to what is in the `lib.rs`

Sushisource: As mentioned in slack, merge master and see TODO in new async workflow driver

src/machines/workflow_machines.rs
vitarb: Added CommandID on top of your recent async changes.

src/lib.rs
vitarb: Let's do it in a follow up, there are a few things we should discuss, so it's more of a stand-alone refactor.

src/machines/activity_state_machine.rs
vitarb: Sorry, that's my (probably bad) habit of replying right away and pushing changes later.

src/machines/test_help/workflow_driver.rs
vitarb: Done.

src/machines/workflow_machines.rs
vitarb: Done, same as above.

src/lib.rs
vitarb: Added an integ test for it.

src/machines/test_help/async_workflow_driver.rs
```diff
@@ -36,23 +38,23 @@ struct TestWfDriverCache {
...
-        if let Some(t) = bc.issued_commands.remove(id) {
+        if let Some(t) = bc.issued_commands.remove(&id) {
             t.unblocker.send(()).unwrap()
         };
     }
 
     /// Cancel a timer by ID. Timers get some special handling here since they are always
     /// removed from the "lang" side without needing a response from core.
-    fn cancel_timer(&self, id: &str) {
+    fn cancel_timer(&self, id: CommandID) {
```
Sushisource, line 50: This should probably only accept a &str and then turn it into a `CommandID::Timer`

src/machines/test_help/workflow_driver.rs
```diff
@@ -0,0 +1 @@
+
```
Sushisource, line 1: This file shouldn't be here any more

Sushisource (approved):
Let's address the few small remaining comments and LGTM! :tada:

src/lib.rs
vitarb: Makes sense, as we've agreed, merging without that part, and we'll follow up with the refactor.

src/machines/test_help/async_workflow_driver.rs
vitarb: Done

src/machines/test_help/workflow_driver.rs
vitarb: Came up after a merge. Deleted.

src/pollers/mod.rs
vitarb: I agree, but let's do it as a follow up. As that would break lang side by doing that and it's better to coordinate that change with Roey.

src/pollers/mod.rs
vitarb: It's the temporal API name, so I though we want to use it as is to avoid confusion.

PR #67, opened 2021-03-10
Make activity task usable

bergundy:
StartActivity attributes were pasted from https://github.com/temporalio/api/blob/dfdf839e185089b000dd8613bfea57375088f88a/temporal/api/workflowservice/v1/request_response.proto#L236

I'm not sure if this was the right decision but it's convenient to use.

### DIFF
--- a/protos/local/core_interface.proto
+++ b/protos/local/core_interface.proto
@@ -73,6 +73,8 @@ message WFActivationJob {
         CancelWorkflow cancel_workflow = 6;
         // A request to signal the workflow was received.
         SignalWorkflow signal_workflow = 7;
+        // An activity was resolved with, result could be completed, failed or cancelled
+        ResolveActivity resolve_activity = 8;
     }
 }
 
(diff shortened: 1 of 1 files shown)


protos/local/core_interface.proto
```diff
@@ -116,19 +117,42 @@ message SignalWorkflow {
...
+    google.protobuf.Duration schedule_to_close_timeout = 12 [(gogoproto.stdduration) = true];
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: "to" is used to indicate interval. --)
+    google.protobuf.Duration start_to_close_timeout = 13 [(gogoproto.stdduration) = true];
+    google.protobuf.Duration heartbeat_timeout = 14 [(gogoproto.stdduration) = true];
+    // This is an actual retry policy the service uses.
+    // It can be different from the one provided (or not) during activity scheduling
+    // as the service can override the provided one in case its values are not specified
+    // or exceed configured system limits.
+    temporal.api.common.v1.RetryPolicy retry_policy = 15;
```
Sushisource, line 142: Can we start instead we just the things you know you need? We may very well end up adding back all of this but we can at least not have these api linter comments etc.

I'd rather start small and add as needed than end up stuck with a bunch of cruft.

Sushisource (approved):

PR #69, opened 2021-03-18
Redo proto definitions

bergundy:
# !!WIP!!
@Sushisource I didn't know you also started working on this before I started, let's sync and see if this is the right direction.


Sushisource:
We used the other PR

PR #70, opened 2021-03-18
Eliminate upstream protos

Sushisource:


protos/local/core_interface.proto
```diff
@@ -50,7 +44,7 @@ message Task {
 // a cached state.
 message WFActivation {
     // The current time as understood by the workflow, which is set by workflow task started events
-    google.protobuf.Timestamp timestamp = 1 [(gogoproto.stdtime) = true];
```
bergundy, line 53: Can we put `WFActivation` in a separate file with all of the activation jobs?

protos/local/workflow_commands.proto
```diff
@@ -0,0 +1,87 @@
...
+    // Activities are provided by a default retry policy controlled through the service dynamic
+    // configuration. Retries are happening up to schedule_to_close_timeout. To disable retries set
+    // retry_policy.maximum_attempts to 1.
+    common.RetryPolicy retry_policy = 11;
+}
+
+message QueryResult {
+    oneof variant {
+        QuerySuccess succeeded = 1;
+        string failed_with_message = 2;
```
bergundy, line 65: Should we have a full failure object here for e.g stacktrace?

protos/local/workflow_commands.proto
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+message QuerySuccess {
+    repeated common.Payload responses = 1;
```
bergundy, line 70: How many responses do we expect?
A function typically returns a single value (or a tuple for multiple values).

protos/local/workflow_commands.proto
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/// Issued when the workflow completes successfully
+message CompleteWorkflowExecution {
+    repeated common.Payload result = 1;
+}
+
+/// Issued when the workflow errors out
+message FailWorkflowExecution {
+
```
bergundy, line 86: Should have an exception (failure) attribute

protos/local/workflow_commands.proto
```diff
@@ -0,0 +1,87 @@
...
+message ScheduleActivity {
+    string activity_id = 1;
+    string activity_type = 2;
+    string namespace = 3;
+    // The name of the task queue to place this activity request in
+    //  TODO: Do we care about sticky/not sticky?
+    string task_queue = 4;
+    map<string, common.Payload> header_fields = 5;
+    repeated common.Payload input = 6;
```
bergundy, line 41: Maybe let's call this `arguments`?

protos/local/workflow_commands.proto
```diff
@@ -0,0 +1,87 @@
...
+        //  UpsertWorkflowSearchAttributes upsert_workflow_search_attributes_command_attributes = 14;
+    }
+}
+
+message StartTimer {
+    string timer_id = 1;
+    google.protobuf.Duration start_to_fire_timeout = 2;
+}
+
+message ScheduleActivity {
```
bergundy, line 33: This will probably change after we talk about local vs. non-local activities.

protos/local/core_interface.proto
```diff
@@ -183,68 +173,48 @@ message WFActivationCompletion {
...
 message WFActivationFailure {
-    temporal.api.enums.v1.WorkflowTaskFailedCause cause = 1;
-    temporal.api.failure.v1.Failure failure = 2;
-
-    // Other bits from RespondWorkflowTaskFailedRequest as needed
+    common.UserCodeFailure failure = 1;
 }
 
 /// Used in ActivityResult to report cancellation
 message ActivityTaskCancelation {
```
bergundy, line 200: I prefer to put all of the activity result variants in a [separate file](https://github.com/temporalio/sdk-core/blob/739f20dd249e6b6d3ef95c86f02f6861d2938aeb/protos/local/activity_result.proto).

protos/local/core_interface.proto
```diff
@@ -102,67 +94,65 @@ message ResolveActivity {
...
-    string timer_id = 1;
-}
-
 message UpdateRandomSeed {
     uint64 randomness_seed = 1;
 }
 
 message QueryWorkflow {
-    temporal.api.query.v1.WorkflowQuery query = 1;
+    string query_type = 1;
```
bergundy, line 102:
```suggestion
    string method = 1;
```

protos/local/core_interface.proto
```diff
@@ -102,67 +94,65 @@ message ResolveActivity {
...
-    // This is an actual retry policy the service uses.
-    // It can be different from the one provided (or not) during activity scheduling
-    // as the service can override the provided one in case its values are not specified
-    // or exceed configured system limits.
-    temporal.api.common.v1.RetryPolicy retry_policy = 15;
-}
-
-message CancelActivity {
-    // TODO: add attributes
+    string signal_name = 1;
```
bergundy, line 111:
```suggestion
    string method = 1;
```

bergundy (commented):
I'd rather split all of the WF activations into their own file and package, same as the commands.

Maybe put the WF activation completion in a new file? [see here](https://github.com/temporalio/sdk-core/blob/739f20dd249e6b6d3ef95c86f02f6861d2938aeb/protos/local/workflow_activation_completion.proto)

See also [PollTaskReq](https://github.com/temporalio/sdk-core/blob/739f20dd249e6b6d3ef95c86f02f6861d2938aeb/protos/local/core_interface.proto#L20) from my branch.

protos/local/core_interface.proto
```diff
@@ -183,68 +173,48 @@ message WFActivationCompletion {
...
 message ActivityTaskSuccess {
-    temporal.api.common.v1.Payloads result = 1;
+    repeated common.Payload result = 1;
     // Other bits from RespondActivityTaskCompletedRequest as needed
 }
 
 /// Used in ActivityResult to report failure
 message ActivityTaskFailure {
-    temporal.api.failure.v1.Failure failure = 1;
+    repeated common.Payload failure = 1;
```
bergundy, line 212: This should probably be a failure struct so we can pass in an exception

protos/local/workflow_commands.proto
Sushisource: Good question. I think a full exception would probably just end up as an activation failure, but I'm not sure. We might have to learn more about this

protos/local/workflow_commands.proto
Sushisource: Yeah, I'm a little torn about the stuff like this. I included it since it's possible to return to the server, but I'd probably prefer the simpler version as well. I'll do that until we have a reason not to.

protos/local/core_interface.proto
Sushisource: I'd like to keep it to type. `method` generally refers to functions with some `self` receiver type -- IE they belong to some class. Different languages might represent different handlers as free-floating functions, whole classes, even maybe a constant (contrived example) so I think type is probably the best thing here.

protos/local/workflow_commands.proto
Sushisource: We can add the local option once we get to implementing it.

vitarb (approved):

PR #72, opened 2021-03-23
Return fewer and more specific errors / API Changes

Sushisource:
## Description, Motivation and Context

What was changed:
Errors simplified and better segregated for more obvious handling. Unified polling and completion APIs

Motivation & Context:
Simpler usage from lang side.

Closes issue: 

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### DIFF
--- a/protos/local/core_interface.proto
+++ b/protos/local/core_interface.proto
@@ -16,18 +16,15 @@ import "google/protobuf/timestamp.proto";
 import "google/protobuf/duration.proto";
 import "google/protobuf/empty.proto";
 
-// A request as given to [crate::Core::poll_task]
+// A request as given to [crate::Core::poll_task]. Either `workflows`, `activities`, or both must be
+// set to true. To set neither will result in an error.
 message PollTaskReq {
-    // What type of task to poll for
-    enum TaskType {
-        // Poll for workflows
-        WORKFLOWS = 0;
-        // Poll for activities
-        ACTIVITIES = 1;
-    }
-
-    // A list of task types to poll for
-    repeated TaskType types = 1;
+    // Should poll_task be allowed to return workflow tasks?
+    bool workflows = 1;
+    // Should poll_task be allowed to return activity tasks?
+    bool activities = 2;
+    // What task queue are we polling on?
+    string task_queue = 3;
 }
 
 // An instruction to perform work from core->lang sdk
--- a/protos/local/core_interface.proto
+++ b/protos/local/core_interface.proto
@@ -16,42 +16,6 @@ import "google/protobuf/timestamp.proto";
 import "google/protobuf/duration.proto";
 import "google/protobuf/empty.proto";
 
-// A request as given to [crate::Core::poll_task]. Either `workflows`, `activities`, or both must be
-// set to true. To set neither will result in an error.
-message PollTaskReq {
-    // Should poll_task be allowed to return workflow tasks?
-    bool workflows = 1;
-    // Should poll_task be allowed to return activity tasks?
-    bool activities = 2;
-    // What task queue are we polling on?
-    string task_queue = 3;
-}
-
-// An instruction to perform work from core->lang sdk
-message Task {
-    // A unique identifier for this task
-    bytes task_token = 1;
-    // The type of task to be performed
-    oneof variant {
-        // Wake up a workflow
-        workflow_activation.WFActivation workflow = 2;
-        // Run an activity
-        activity_task.ActivityTask activity = 3;
-    }
-}
-
-// Sent from lang side to core when calling [crate::Core::complete_task]
-message TaskCompletion {
-    // The id from the [Task] being completed
-    bytes task_token = 1;
-    oneof variant {
-        // Complete a workflow task
-        workflow_completion.WFActivationCompletion workflow = 2;
-        // Complete an activity task
-        activity_result.ActivityResult activity = 3;
-    }
-}
-
 // A request as given to [crate::Core::send_activity_heartbeat]
 message ActivityHeartbeat {
     string activity_id = 1;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,5 @@
 #![warn(missing_docs)] // error if there are missing docs
+#![allow(clippy::upper_case_acronyms)]
 
 //! This crate provides a basis for creating new Temporal SDKs without completely starting from
 //! scratch
@@ -232,14 +233,15 @@ where
         }
     }
 
+    #[instrument(skip(self))]
     fn poll_activity_task(&self, task_queue: &str) -> Result<ActivityTask, CoreError> {
-        return match abort_on_shutdown!(self, poll_activity_task, task_queue.to_owned()) {
+        match abort_on_shutdown!(self, poll_activity_task, task_queue.to_owned()) {
             Ok(work) => {
                 let task_token = work.task_token.clone();
                 Ok(ActivityTask::start_from_poll_resp(work, task_token))
             }
             Err(e) => Err(e),
-        };
+        }
     }
 
     #[instrument(skip(self))]
@@ -398,7 +400,7 @@ impl<WP: ServerGatewayApis> CoreSDK<WP> {
     /// Feed commands from the lang sdk into appropriate workflow manager which will iterate
     /// the state machines and return commands ready to be sent to the server
     fn push_lang_commands(&self, run_id: &str, cmds: Vec<WFCommand>) -> Result<Vec<ProtoCommand>> {
-        Ok(self.access_wf_machine(run_id, move |mgr| mgr.push_commands(cmds))?)
+        self.access_wf_machine(run_id, move |mgr| mgr.push_commands(cmds))
     }
 
     /// Remove a workflow run from the cache entirely
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -242,7 +242,13 @@ where
                     .workflow_task_tokens
                     .get(&task_token)
                     .map(|x| x.value().clone())
-                    .ok_or_else(|| CoreError::NothingFoundForTaskToken(task_token.clone()))?;
+                    .ok_or_else(|| CoreError::MalformedCompletion {
+                        reason: format!(
+                            "Task token {} had no workflow run associated with it",
+                            fmt_task_token(&task_token)
+                        ),
+                        completion: None,
+                    })?;
                 match wfstatus {
                     wf_activation_completion::Status::Successful(success) => {
                         let commands = self.push_lang_commands(&run_id, success)?;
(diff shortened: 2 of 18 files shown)


src/workflow/mod.rs
```diff
@@ -7,15 +7,31 @@ pub(crate) use concurrency_manager::WorkflowConcurrencyManager;
...
+// NOTE: Docstrings take the place of #[error("xxxx")] here b/c of displaydoc
+pub enum WorkflowError {
+    /// The machine with `run_id` was not found in memory
+    MissingMachine { run_id: String },
+    /// Underlying error in state machines: {0:?}
+    UnderlyingMachinesError(#[from] WFMachinesError),
+    /// There was an error in the history associated with the workflow: {0:?}
+    HistoryError(#[from] HistoryInfoError),
+    /// Error buffering commands
+    CantSendCommands(#[from] SendError<Vec<WFCommand>>),
```
vitarb, line 33: Should we name it more imperatively, putting subject (Command) first? (e.g. CommandDispatchError)

src/workflow/mod.rs
```diff
@@ -7,15 +7,31 @@ pub(crate) use concurrency_manager::WorkflowConcurrencyManager;
...
-    protosext::HistoryInfo,
-    CoreError, Result,
+    protosext::{HistoryInfo, HistoryInfoError},
 };
-use std::sync::mpsc::Sender;
+use std::sync::mpsc::{SendError, Sender};
+
+type Result<T, E = WorkflowError> = std::result::Result<T, E>;
+
+/// Errors relating to workflow management and machine logic
```
vitarb, line 21: Can we mention here in the comment when this error is returned and who is expected to be on the receiving end?

vitarb (approved):
LGTM, with a small comment.

src/workflow/mod.rs
Sushisource: Yep, makes sense.

src/workflow/mod.rs
Sushisource: Will do. I think I can simplify more of the errors in general as well.

protos/local/core_interface.proto
```diff
@@ -16,18 +16,15 @@ import "google/protobuf/timestamp.proto";
...
-    }
-
-    // A list of task types to poll for
-    repeated TaskType types = 1;
+    // Should poll_task be allowed to return workflow tasks?
+    bool workflows = 1;
+    // Should poll_task be allowed to return activity tasks?
+    bool activities = 2;
+    // What task queue are we polling on?
+    string task_queue = 3;
```
mfateev, line 27: Why do we care about task queue at the SDK level? Does it mean that multiple simultaneous polls are expected if there is more than one task queue?

src/lib.rs
```diff
@@ -142,80 +135,112 @@ struct CoreSDK<WP> {
...
 
-    #[instrument(skip(self))]
-    fn poll_activity_task(&self, task_queue: &str) -> Result<Task> {
-        if self.shutdown_requested.load(Ordering::SeqCst) {
-            return Err(CoreError::ShuttingDown);
-        }
+            // TODO: There's an interesting question here about whether activities or workflows
+            //  should take priority in the event that the caller is OK with either, and we
+            //  could potentially poll both and just proceed with whichever returns first... but
+            //  that starts getting tricky. Perhaps we should not allow specifying both.
```
bergundy, line 198: From lang side we need to be able to request both in order to not spawn multiple threads.
I don't think there should be precedence of either workflows or activities, better to poll on both concurrently.

src/lib.rs
```diff
@@ -142,80 +135,112 @@ struct CoreSDK<WP> {
...
-                        task_token: task_token.clone(),
+    fn poll_task(&self, req: PollTaskReq) -> Result<Task> {
+        let task_queue = req.task_queue;
+        // The poll needs to be in a loop because we can't guarantee tail call optimization in Rust
+        // (simply) and we really, really need that for long-poll retries.
+        loop {
+            if req.workflows {
+                // We must first check if there are pending workflow tasks for workflows that are
+                // currently replaying, and issue those tasks before bothering the server.
+                if let Some(pa) = self.pending_activations.pop() {
```
bergundy, line 175: Can we pull checking for pending activations out of this loop?
I think we only need to loop for retrying the long poll, right?

src/lib.rs
bergundy: The way it's built right now, we can't poll for both which is problematic.

src/lib.rs
```diff
@@ -142,80 +135,112 @@ struct CoreSDK<WP> {
...
+            } else {
+                return match abort_on_shutdown!(self, poll_activity_task, task_queue.to_owned()) {
+                    Ok(work) => {
+                        let task_token = work.task_token.clone();
+                        Ok(Task {
+                            task_token,
+                            variant: Some(task::Variant::Activity(work.into())),
+                        })
+                    }
+                    Err(e) => Err(e),
```
bergundy, line 241: There may be pending activations here as well if lang only polls on activities.

bergundy:
@Sushisource @vitarb @mfateev after thinking about it some more I think this unified poll API doesn't make sense given the current service API.
How would polling for both activity and workflow tasks work? We can only return a single value and we shouldn't buffer activity tasks since they're considered started as soon as poll returns.
I also think we should change the API to be async (probably using callbacks), otherwise the lang SDK will redundantly need to spawn multiple threads for each poll type and possible for completions as well as we don't want to block the lang thread on communication to the server.

bergundy:
I think we still need more work on the error definitions, it's unclear what action should be taken for each error on the lang side.

src/workflow/mod.rs
```diff
@@ -7,15 +7,34 @@ pub(crate) use concurrency_manager::WorkflowConcurrencyManager;
...
+
+type Result<T, E = WorkflowError> = std::result::Result<T, E>;
+
+/// Errors relating to workflow management and machine logic. These are going to be passed up and
+/// out to the lang SDK where it will need to handle them. Generally that will usually mean
+/// showing an error to the user and/or invalidating the workflow cache.
+#[derive(thiserror::Error, Debug, displaydoc::Display)]
+#[allow(clippy::large_enum_variant)]
+// NOTE: Docstrings take the place of #[error("xxxx")] here b/c of displaydoc
+pub enum WorkflowError {
```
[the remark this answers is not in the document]
Sushisource: Both are possible

src/lib.rs
```diff
@@ -142,80 +135,112 @@ struct CoreSDK<WP> {
     shutdown_requested: AtomicBool,
 }
 
+/// Can be used inside the CoreSDK impl to block on any method that polls the server until it
+/// responds, or until the shutdown flag is set (aborting the poll)
+macro_rules! abort_on_shutdown {
+    ($self:ident, $gateway_fn:tt, $poll_arg:expr) => {
+        $self.runtime.block_on(async {
+            let shutdownfut = async {
+                loop {
```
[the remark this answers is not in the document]
Sushisource: Not with a channel, since shutdown can and will be read multiple times. I could use a condition variable here but it just didn't seem terribly important to optimize at the moment. I can fix this in another PR

src/lib.rs
Sushisource: Not at all. Before this change it recursed to clear pending activations in the event of shutdown. The loop still serves that purpose

src/lib.rs
Sushisource: Heh, yes, hence the TODO. Polling for both is problematic as I discovered here. See my other reply

Sushisource:
> @Sushisource @vitarb @mfateev after thinking about it some more I think this unified poll API doesn't make sense given the current service API.
> How would polling for both activity and workflow tasks work? We can only return a single value and we shouldn't buffer activity tasks since they're considered started as soon as poll returns.
> I also think we should change the API to be async (probably using callbacks), otherwise the lang SDK will redundantly need to spawn multiple threads for each poll type and possible for completions as well as we don't want to block the lang thread on communication to the server.

Yeah I agree I think. We've gone in circles a bit on this one, and the main reason it felt nice on this side to keep them unified was to avoid duplicating `task_token` in a bunch of different messages. That I'm less concerned about now than the whole "it turns out it doesn't make any sense to try to poll for both at the same time" thing, which makes the unified API pretty pointless. It also allows us to again return more specific error types from each API and avoid some matching/unwrapping.

Changing it to be async is... a bigger question. Callbacks aren't terribly idiomatic in this case IMO. I'd prefer to just actually have an `async` api at that point. Alternatively, each lang's glue layer can kinda wrap Core however it feels is appropriate, which I think makes good sense here. The question is what should the default be, `async` seems reasonable to me and fixes some slightly ugly things we need to do right now internally. Glue layers can pretty trivially turn that into non-async blocking, or even a callback based API.

src/workflow/mod.rs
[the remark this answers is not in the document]
bergundy: How do we distinguish which errors can be returned for each endpoint?

bergundy (commented):
We've already discussed this F2F, we need to rewrite the API to be async and split the poll endpoints for activities and workflows.

src/workflow/mod.rs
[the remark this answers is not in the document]
Sushisource: They all can, since both endpoints update workflow machines.

src/pollers/mod.rs
```diff
@@ -52,34 +51,38 @@ pub struct ServerGatewayOptions {
...
+        // TODO: Only apply this to long poll requests
+        metadata.insert(
+            "grpc-timeout",
+            timeout_str
+                .parse()
+                .expect("Timeout string construction cannot fail"),
+        );
+        metadata.insert(
+            "client-name",
+            "core-sdk".parse().expect("Static value is parsable"),
```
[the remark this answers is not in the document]
Sushisource: Perhaps, I think out of scope for this PR tho

src/pollers/mod.rs
```diff
@@ -52,34 +51,38 @@ pub struct ServerGatewayOptions {
...
-        "core-sdk".parse().expect("Static value is parsable"),
-    );
-    Ok(req)
+/// This function will get called on each outbound request. Returning a `Status` here will cancel
+/// the request and have that status returned to the caller.
+fn intercept(opts: &ServerGatewayOptions) -> impl Fn(Request<()>) -> Result<Request<()>, Status> {
+    let timeout_str = format!("{}m", opts.long_poll_timeout.as_millis());
+    move |mut req: Request<()>| {
+        let metadata = req.metadata_mut();
+        // TODO: Only apply this to long poll requests
```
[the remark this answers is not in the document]
Sushisource: Yep. Out of scope here too though I think.

PR #73, opened 2021-03-24
Adding state machine tests for activity

vitarb:
Simple activity state machine tests, similar to those that we have for timer.

### DIFF
--- a/src/machines/test_help/async_workflow_driver.rs
+++ b/src/machines/test_help/async_workflow_driver.rs
@@ -131,26 +131,20 @@ impl CommandSender {
 
     /// Request to create a timer
     pub fn timer(&mut self, a: StartTimer) -> impl Future {
-        let tid = a.timer_id.clone();
-        let c = WFCommand::AddTimer(a);
-        self.send(c);
-        let rx = self.twd_cache.add_sent_cmd(Timer(tid.clone()));
-        let cache_clone = self.twd_cache.clone();
-        async move {
-            cache_clone.set_cmd_blocked(Timer(tid));
-            rx.await
-        }
+        self.send_blocking_cmd(Timer(a.timer_id.clone()), WFCommand::AddTimer(a))
     }
 
     /// Request to run an activity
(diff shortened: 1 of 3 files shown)


src/machines/test_help/async_workflow_driver.rs
```diff
@@ -138,6 +142,19 @@ impl CommandSender {
...
+    /// Request to run an activity
+    pub fn activity(&mut self, a: ScheduleActivity) -> impl Future {
+        let aid = a.activity_id.clone();
+        let c = WFCommand::AddActivity(a);
+        self.send(c);
+        let rx = self.twd_cache.add_sent_cmd(Activity(aid.clone()));
+        let cache_clone = self.twd_cache.clone();
+        async move {
+            cache_clone.set_cmd_blocked(Activity(aid));
+            rx.await
```
Sushisource, line 154: This code and the `timer` code are almost totally the same except the actual map key part. Let's factor it out.

src/machines/test_help/async_workflow_driver.rs
```diff
@@ -127,13 +131,20 @@ impl CommandSender {
...
-        let c = WFCommand::AddTimer(a);
+        self.send_blocking_cmd(Timer(a.timer_id.clone()), WFCommand::AddTimer(a))
+    }
+
+    /// Request to run an activity
+    pub fn activity(&mut self, a: ScheduleActivity) -> impl Future {
+        self.send_blocking_cmd(Activity(a.activity_id.clone()), WFCommand::AddActivity(a))
+    }
+
+    fn send_blocking_cmd(&mut self, id: CommandID, c: WFCommand) -> impl Future {
```
Sushisource, line 142: We really only need the second parameter here I think.

Sushisource (approved):

src/machines/test_help/async_workflow_driver.rs
Sushisource: Oh, nevermind, I see why. We don't know how to access the ID.

Issue #75, opened 2021-03-25
Avoid polling on shutdown_requested


bergundy:
Isn't there a way to avoid this polling? e.g. with a [channel](https://docs.rs/tokio/1.4.0/tokio/sync/index.html#watch-channel)?

_Originally posted by @bergundy in https://github.com/temporalio/sdk-core/pull/72#r601039530_

PR #77, opened 2021-03-26
Asyncify public APIs

Sushisource:
## What was changed:
Public Core APIs are now async

## Why?
We wanted to have separate APIs for workflow and activity polling, and in order to support single-threaded language SDKs, as well as for ergonomics reasons, using an async API makes sense.

## Closes issue: 
Fixes https://github.com/temporalio/sdk-core/issues/75

## How has this been tested?
Existing tests updated and pass

### DIFF
--- a/.buildkite/docker/Dockerfile
+++ b/.buildkite/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.50
+FROM rust:1.51
 
 RUN rustup component add rustfmt && \
 	rustup component add clippy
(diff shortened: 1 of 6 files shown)


.buildkite/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.50
+FROM rust:1.51
```
bergundy, line 1: 🎊

bergundy (approved):
Seems like it was an easy transition, nice!

vitarb (approved):

Issue #61, opened 2021-03-04
Differentiating fatal and per-workflow / other non-fatal errors


Sushisource:
As of right now, all errors returned from the Core interface are in one enum, listed here:

```rust
pub enum CoreError {
    /// [Core::shutdown] was called, and there are no more replay tasks to be handled. You must
    /// call [Core::complete_task] for any remaining tasks, and then may exit.
    ShuttingDown,
    /// Poll response from server was malformed: {0:?}
    BadDataFromWorkProvider(PollWorkflowTaskQueueResponse),
    /// Lang SDK sent us a malformed completion: {0:?}
    MalformedCompletion(TaskCompletion),
    /// Error buffering commands
    CantSendCommands(#[from] SendError<Vec<WFCommand>>),
    /// Couldn't interpret command from <lang>
    UninterpretableCommand(#[from] InconvertibleCommandError),
    /// Underlying error in history processing
    UnderlyingHistError(#[from] HistoryInfoError),
    /// Underlying error in state machines: {0:?}
    UnderlyingMachinesError(#[from] WFMachinesError),
    /// Task token had nothing associated with it: {0:?}
    NothingFoundForTaskToken(Vec<u8>),
    /// Error calling the service: {0:?}
    TonicError(#[from] tonic::Status),
    /// Server connection error: {0:?}
    TonicTransportError(#[from] tonic::transport::Error),
    /// Failed to initialize tokio runtime: {0:?}
    TokioInitError(std::io::Error),
    /// Invalid URI: {0:?}
    InvalidUri(#[from] InvalidUri),
    /// State machines are missing for the workflow with run id {0}!
    MissingMachines(String),
}
```

Most of these are not fatal. In fact, probably the only fatal error here that should cause lang sdk to give up entirely is `TokioInitError`, and maaaybe `CantSendCommands`, which is specific to a workflow, but if it happens something very wrong has happened (other workflows might continue working but that one never will). All other errors are either transient network problems (ex tonic errors), expected (`ShuttingDown`), bad data from the server or the lang sdk, or are specific to some individual workflow.

So, we need a way to make that more explicit. A few options:

## Nested enums
This is probably the most natural and obvious option

Break up `CoreError` into this top level enum, and create new sub-enums
```rust
pub enum CoreError {
    Fatal(FatalError),
    WorkflowError {
        run_id: String,
        error: WorkflowError
    }
    ServerError(ServerError), // Bad data received from server
[...]

PR #78, opened 2021-03-26
Further simplify public errors

Sushisource:
## What was changed:
Make errors returned by public api be specific to each method for better ergonomics and clarity

## Closes issue: 
https://github.com/temporalio/sdk-core/issues/61

## How has this been tested?
Existing tests


vitarb (approved):

src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+pub(crate) struct WorkflowUpdateError {
+    /// Underlying workflow error
+    pub source: WorkflowError,
+    /// The run id of the erring workflow
+    pub run_id: String,
+}
+
+/// Errors thrown during initialization of [crate::Core]
+#[derive(thiserror::Error, Debug, displaydoc::Display)]
+pub enum CoreInitError {
```
bergundy, line 19: I'd like us to have clearer instructions for lang for what needs to happen in each of these cases.
e.g. `CoreInitError` -> `InvalidUri` is fatal, we should create a new core, `TonicTransportError`, when is this retry-able who should retry (lang / core)?

src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+    InvalidUri(#[from] InvalidUri),
+    /// Server connection error: {0:?}
+    TonicTransportError(#[from] tonic::transport::Error),
+}
+
+/// Errors thrown by [crate::Core::poll_workflow_task]
+#[derive(thiserror::Error, Debug, displaydoc::Display)]
+pub enum PollWfError {
+    /// There was an error specific to a workflow instance with id ({run_id}): {source:?}
+    WorkflowUpdateError {
```
bergundy, line 30: I'm assuming this means kill the workflow cache for given `run_id`? If so, can we move this to an activation (e.g. `DeleteWorkflow{ reason }`.

src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+        /// Underlying workflow error
+        source: WorkflowError,
+        /// The run id of the erring workflow
+        run_id: String,
+    },
+    /// Poll workflow response from server was malformed: {0:?}
+    BadPollResponseFromServer(PollWorkflowTaskQueueResponse),
```
bergundy, line 37: Is this fatal? What should we do in this case? Doesn't seem like we can recover from this..

src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+    /** [crate::Core::shutdown] was called, and there are no more replay tasks to be handled. You
+     * must call [crate::Core::complete_workflow_task] for any remaining tasks, and then may
+     * exit.*/
+    ShuttingDown,
```
bergundy, line 41: Just to clarify, does "no more replay tasks" mean all pending completions have been received from lang and it's safe to stop polling? If this is the case, we should rename to `ShutDown`.

src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Unhandled error when calling the temporal server: {0:?}
+    TonicError(#[from] tonic::Status),
```
bergundy, line 43: When is this retry-able and who should initiate the retry (lang / core)?

src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn from(_: ShutdownErr) -> Self {
+        Self::ShuttingDown
+    }
+}
+
+/// Errors thrown by [crate::Core::poll_activity_task]
+#[derive(thiserror::Error, Debug, displaydoc::Display)]
+pub enum PollActivityError {
+    /// [crate::Core::shutdown] was called, we will no longer fetch new activity tasks
+    ShuttingDown,
```
bergundy, line 65: What about if there are pending activity task completions? Will we get this error or will core block the poll request until no more pending tasks?

src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Unhandled error when calling the temporal server: {0:?}
+    TonicError(#[from] tonic::Status),
```
bergundy, line 67: Same as above , what should lang do in this case?

src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+        /// The completion, which may not be included to avoid unnecessary copies.
+        completion: Option<WfActivationCompletion>,
+    },
+    /// There was an error specific to a workflow instance with id ({run_id}): {source:?}
+    WorkflowUpdateError {
+        /// Underlying workflow error
+        source: WorkflowError,
+        /// The run id of the erring workflow
+        run_id: String,
+    },
```
bergundy, line 93: Both require deleting the workflow instance on lang side for `run_id`, right?

src/errors.rs
```diff
@@ -0,0 +1,123 @@
...
+    /** There exists a pending command in this workflow's history which has not yet been handled.
+     * When thrown from [crate::Core::complete_workflow_task], it means you should poll for a new
+     * task, receive a new task token, and complete that new task. */
+    UnhandledCommandWhenCompleting,
```
bergundy, line 97: When would this happen?

bergundy (commented):
I like this new structure, definitely an improvement.

Sushisource:
@bergundy So the answer to many of these "what should lang do?" questions is simply "I don't know yet". I will update the docstrings for every case where it's clear, but many of them are semi ambiguous and we'll have to figure it out as we go, particularly underlying network errors.

src/errors.rs
Sushisource: Initialization errors might as well always be considered fatal. It's so early on that just throwing everything out and trying again seems easiest and simplest

src/errors.rs
Sushisource: I don't like it so much. This call already needs to return a `Result`, so it's not like we really gain anything by moving the error into the activation, and we can't make activations with `DeleteWorkflow` in them be the only way of forcing that action because problems can happen when you call the completion API too. So, the path of "delete the workflow because of an error from an API call" always needs to exist, and that feedback is immediate and obvious vs the somewhat indirect way of stuffing it in an activation.

Sushisource: You can always translate this error into something you stuff inside an activation if you want to do that for your control flow purposes

src/errors.rs
Sushisource: Not necessarily. The server maybe just burps or something. This one is TBD

src/errors.rs
Sushisource: Yes, it means that

src/errors.rs
Sushisource: I really don't know, it depends, there's a bunch of different things that could happen here. Whatever makes sense to retry we will retry within core but that's not gonna all be in this PR

src/errors.rs
Sushisource: Not necessarily the first -- the first probably indicates a bug in lang. The second, definitely.

src/errors.rs
Sushisource: For example WF receives a signal or query after it thought it was done, but before it reported being done to server.

Sushisource:
@bergundy Merging this one since I've got work going on top of it but as usual feel free to lemme know if you have more comments. I think we'll just need to iterate here as we learn more.

bergundy:
Of course @Sushisource makes sense.
I'll make sure we [handle the errors](https://github.com/temporalio/sdk-node/issues/51) to the best of our current understanding on the node SDK before the alpha release.

PR #80, opened 2021-03-30
Add ActivityTaskCompletion message

bergundy:
## Why?
I found myself needing this when trying to work with the new interface in the node SDK.

## How has this been tested?
Only with lang integration.

## Any docs updates needed?
Updated where applicable.

### DIFF
--- a/protos/local/core_interface.proto
+++ b/protos/local/core_interface.proto
@@ -21,3 +21,9 @@ message ActivityHeartbeat {
     string activity_id = 1;
     repeated common.Payload details = 2;
 }
+
+// A request as given to [crate::Core::complete_activity_task]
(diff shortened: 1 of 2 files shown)


bergundy:
@Sushisource I needed it because it's easier for me to transfer a single proto struct from JS to rust.
This is consistent with the workflow completion API.

Sushisource:
> @Sushisource I needed it because it's easier for me to transfer a single proto struct from JS to rust.
> This is consistent with the workflow completion API.

Seems you could just define it in the glue, but, I suppose it's worth doing for the consistency. I'll fix your build errors and merge this.

Sushisource (approved):

PR #84, opened 2021-04-07
Activity timeout

vitarb:
Preview of activity timeout handling in state machines.
There is a bug when timeout happens for canceled activity in try cancel state, it would result in two activations to the lang side (one for cancellation and one for timeout).
We need to discuss and address this issue, ideally as part of this PR.
Also I think I've encountered an issue with test framework and was unable to run activity timeout after cancel test incrementally.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -825,7 +825,7 @@ mod test {
                                         )),
                                     }
                                 ] => {
-                                    assert_eq!(failure.message, "Activity task timedOut".to_string());
+                                    assert_eq!(failure.message, "Activity task timed out".to_string());
                                     assert_eq!(aid, &activity_id.to_string());
                                 }
                             );
@@ -878,7 +878,7 @@ mod test {
                                         )),
                                     }
                                 ] => {
-                                    assert_eq!(failure.message, "Activity task timedOut".to_string());
+                                    assert_eq!(failure.message, "Activity task timed out".to_string());
                                     assert_eq!(aid, &activity_id.to_string());
                                 }
                             );
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -847,14 +847,15 @@ mod test {
         .await;
     }
 
+    #[rstest(hist_batches, case::incremental(&[1, 3]), case::replay(&[3]))]
     #[tokio::test]
-    async fn cancelled_activity_timeout() {
+    async fn cancelled_activity_timeout(hist_batches: &[usize]) {
         let wfid = "fake_wf_id";
         let activity_id = "fake_activity";
         let signal_id = "signal";
 
         let mut t = canned_histories::scheduled_cancelled_activity_timeout(activity_id, signal_id);
-        let core = build_fake_core(wfid, &mut t, &[3]);
+        let core = build_fake_core(wfid, &mut t, hist_batches);
 
         poll_and_reply(
             &core,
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -783,6 +783,113 @@ mod test {
...
+                    &job_assert!(wf_activation_job::Variant::StartWorkflow(_)),
+                    vec![ScheduleActivity {
+                        activity_id: activity_id.to_string(),
+                        ..Default::default()
+                    }
+                    .into()],
+                ),
+                // Activity is getting resolved right away as it has been timed out.
+                gen_assert_and_reply(
+                    &job_assert!(wf_activation_job::Variant::ResolveActivity(_)),
```
Sushisource, line 810: We should probably match on the specific type of resolution here, make sure it's a timeout

src/lib.rs
```diff
@@ -783,6 +783,113 @@ mod test {
...
+                        activity_id: activity_id.to_string(),
+                        ..Default::default()
+                    }
+                    .into()],
+                ),
+                // Activity is getting resolved right away as it has been timed out.
+                gen_assert_and_reply(
+                    &job_assert!(
+                        wf_activation_job::Variant::ResolveActivity(_),
+                        wf_activation_job::Variant::ResolveActivity(_)
```
Sushisource, line 884: Just to confirm this  is the bug we're going to fix, right?

vitarb: Yes, fixed it using shared state and checking cancellation type. Also addressed similar issue on the cancellation path for try cancel.

src/lib.rs
vitarb: Yes, going to address it.

src/lib.rs
```diff
@@ -807,7 +808,28 @@ mod test {
...
+                                                result: Some(ActivityResult {
+                                                    status: Some(activity_result::Status::Failed(ar::Failure {
+                                                        failure: Some(failure)
+                                                    })),
+                                                })
+                                            }
+                                        )),
+                                    }
+                                ] => {
+                                    assert_eq!(failure.message, "Activity task timedOut".to_string());
```
Sushisource, line 828: "timedOut" instead of "timed out" is a bit odd

Sushisource: We could also potentially just add a `TimedOut` status, since we control this now, and that is a lot more clear. I vote we go for that.

src/machines/activity_state_machine.rs
```diff
@@ -735,6 +723,16 @@ fn new_timeout_failure(dat: &SharedState, attrs: ActivityTaskTimedOutEventAttrib
     }
 }
 
+fn convert_payloads(
```
Sushisource, line 726: This is general enough that it doesn't belong here. Probably in proto or protosext module

Sushisource (approved):
LGTM, comments not urgent.

src/lib.rs
vitarb: I agree, copied from java, I think it might be a good time to change it.

src/machines/activity_state_machine.rs
vitarb: It is, the only thing that concerns me is WFMachinesError, which is specific to the state machine code. Should we generalize it too, or what do you think?

src/lib.rs
vitarb: Addressed

src/lib.rs
vitarb: Addressed

src/machines/activity_state_machine.rs
Sushisource: The function could just go in the machines module

PR #86, opened 2021-04-08
Show fsm coverage

Sushisource:
## What was changed:
Only when compiled for test, we collect coverage info about which state machine transitions have been hit and turn it into nice little state diagrams with colored lines.

## Why?
Much easier to tell where we have coverage gaps while testing state machines

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/fsm/state_machine_trait/src/lib.rs
+++ b/fsm/state_machine_trait/src/lib.rs
@@ -1,5 +1,7 @@
-use std::error::Error;
-use std::fmt::{Debug, Display, Formatter};
+use std::{
+    error::Error,
+    fmt::{Debug, Display, Formatter},
+};
 
 /// This trait defines a state machine (more formally, a [finite state
 /// transducer](https://en.wikipedia.org/wiki/Finite-state_transducer)) which accepts events (the
@@ -70,6 +72,9 @@ pub trait StateMachine: Sized {
 
     /// Given the shared data and new state, create a new instance.
     fn from_parts(shared: Self::SharedState, state: Self::State) -> Self;
+
+    /// Return a PlantUML definition of the fsm that can be used to visualize it
+    fn visualizer() -> &'static str;
 }
 
 /// The error returned by [StateMachine]s when handling events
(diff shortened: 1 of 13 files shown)


fsm/state_machine_trait/src/lib.rs
```diff
@@ -70,6 +72,9 @@ pub trait StateMachine: Sized {
 
     /// Given the shared data and new state, create a new instance.
     fn from_parts(shared: Self::SharedState, state: Self::State) -> Self;
+
+    /// Return a PlantUML definition of the fsm that can be used to visualize it
+    fn visualizer() -> &'static str;
```
vitarb, line 77: Why not return a String here instead?

src/machines/mod.rs
```diff
@@ -251,6 +262,46 @@ trait Cancellable: StateMachine {
...
+        event: Self::Event,
+    ) -> Result<Vec<Self::Command>, MachineError<Self::Error>> {
+        #[cfg(test)]
+        let from_state = self.state().to_string();
+        #[cfg(test)]
+        let converted_event_str = event.to_string();
+
+        let res = StateMachine::on_event_mut(self, event);
+        if res.is_ok() {
+            #[cfg(test)]
```
vitarb, line 284: Should this be one line above and gate entire if statement?

vitarb (approved):
LGTM, couple small comments

fsm/state_machine_trait/src/lib.rs
Sushisource: Because the viz string is generated at compile time -- this is more efficient than passing an owned string around.

src/machines/mod.rs
Sushisource: Yeah, in principle it probably should but in practice that makes clippy complain about this unnecessary `let` binding and this seemed an easier way to deal with that than a clippy allow. Compiler will optimize it out anyway.

PR #87, opened 2021-04-09
Add missing tests for the Started state in activity state machine

vitarb:
Added tests for untested transitions and removed some unneeded code from the state machine.

### DIFF
--- a/src/machines/activity_state_machine.rs
+++ b/src/machines/activity_state_machine.rs
@@ -48,34 +48,45 @@ fsm! {
     Created --(Schedule, on_schedule)--> ScheduleCommandCreated;
 
     ScheduleCommandCreated --(CommandScheduleActivityTask) --> ScheduleCommandCreated;
-    ScheduleCommandCreated --(ActivityTaskScheduled(i64), shared on_activity_task_scheduled) --> ScheduledEventRecorded;
+    ScheduleCommandCreated --(ActivityTaskScheduled(i64),
+        shared on_activity_task_scheduled) --> ScheduledEventRecorded;
     ScheduleCommandCreated --(Cancel, shared on_canceled) --> Canceled;
 
     ScheduledEventRecorded --(ActivityTaskStarted(i64), shared on_task_started) --> Started;
-    ScheduledEventRecorded --(ActivityTaskTimedOut(ActivityTaskTimedOutEventAttributes), shared on_task_timed_out) --> TimedOut;
+    ScheduledEventRecorded --(ActivityTaskTimedOut(ActivityTaskTimedOutEventAttributes),
+        shared on_task_timed_out) --> TimedOut;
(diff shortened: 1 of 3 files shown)


src/machines/activity_state_machine.rs
```diff
@@ -50,8 +50,7 @@ fsm! {
     Created --(Schedule, on_schedule)--> ScheduleCommandCreated;
 
     ScheduleCommandCreated --(CommandScheduleActivityTask) --> ScheduleCommandCreated;
-    ScheduleCommandCreated
-      --(ActivityTaskScheduled(i64), shared on_activity_task_scheduled) --> ScheduledEventRecorded;
+    ScheduleCommandCreated --(ActivityTaskScheduled(i64), shared on_activity_task_scheduled) --> ScheduledEventRecorded;
```
Sushisource, line 53: Why change these all back to one line? Enforcing the 100 char wrap is nice, not a big fan of having to scroll around horizontally to read these.

src/test_help/canned_histories.rs
```diff
@@ -567,6 +614,94 @@ pub fn cancel_scheduled_activity_with_signal_and_activity_task_cancel(
...
+/// 7: EVENT_TYPE_WORKFLOW_EXECUTION_SIGNALED
+/// 8: EVENT_TYPE_WORKFLOW_TASK_SCHEDULED
+/// 9: EVENT_TYPE_WORKFLOW_TASK_STARTED
+/// 10: EVENT_TYPE_WORKFLOW_TASK_COMPLETED
+/// 11: EVENT_TYPE_ACTIVITY_TASK_CANCEL_REQUESTED
+/// 12: EVENT_TYPE_WORKFLOW_EXECUTION_SIGNALED
+/// 13: EVENT_TYPE_WORKFLOW_TASK_SCHEDULED
+/// 14: EVENT_TYPE_WORKFLOW_TASK_STARTED
+/// 15: EVENT_TYPE_WORKFLOW_TASK_COMPLETED
+/// 16: EVENT_TYPE_ACTIVITY_TASK_CANCELED
```
Sushisource, line 632: I think maybe you don't need the signal here and can to AT_Canceled right after AT_Cancel_Requested and cover the same path? Not a big deal but might be able to shorten a bit.

Sushisource: Oh, I guess that's the history of the other one you added. What different paths do they cover? Maybe worth adding a note about that.

Sushisource (approved):
Just some minor stuff

src/test_help/canned_histories.rs
Sushisource: Which is scheduled vs started? If they cover identical paths maybe we don't need this signal version -- but it sounds like you're saying they cover different paths?

vitarb: They do cover different paths in the state machine due to activity started event.

PR #89, opened 2021-04-11
Make pending activations interrupt long poll

Sushisource:
## What was changed:
Pending activations need to take priority over any currently in-progress but not yet completed poll. This fixes that.

## Why?
Cancels of activities were getting lost because of the long poll

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -363,11 +363,10 @@ impl<WP: ServerGatewayApis> CoreSDK<WP> {
         task_token: Vec<u8>,
     ) -> WfActivation {
         if next_a.more_activations_needed {
-            self.pending_activations.push(PendingActivation {
+            self.add_pending_activation(PendingActivation {
                 run_id: next_a.get_run_id().to_owned(),
                 task_token: task_token.clone(),
             });
-            self.pending_activation_notify.notify_one();
         }
         next_a.finalize(task_token)
     }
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -193,7 +195,34 @@ where
...
+                }
+            };
+            let poll_result_future = self
+                .server_gateway
+                .poll_workflow_task(task_queue.to_owned());
+            let selected_f = tokio::select! {
+                _ = shutdownfut => {
+                    Err(ShutdownErr.into())
+                }
+                _ = pa_fut => {
```
bergundy, line 219: What happens to the poll request in this case?
Is it cancelled? Is it ignored?
We probably don't want to create a new request if we got interrputed.

Sushisource: Yeah it's a good question, was thinking about it after I wrote this. I'll see if I can figure out a way to test that while I finish this. Ideally, it doesn't really matter, because we spit out the PA so fast that when we go back to requesting it'll kinda be like nothing ever happened... but perhaps we need to keep the polling future around in order to ensure that behavior. I'll play with it.

Sushisource: I think this opens up too big of a can of worms to deal with in this PR. Ideally the request should actually be cancelled I think. Keeping it around raises too many questions like "what if the next poll is for a different task queue" or "how long should we buffer the response" and so on. I opened https://github.com/temporalio/sdk-core/issues/91

vitarb: I agree, we should discard initial poll as from the lang's perspective it polled and got an activation, it shouldn't care if core made a request to the server or not in order to get it.

vitarb (approved):

PR #92, opened 2021-04-13
Activity heartbeat

vitarb:
Implementation of activity heartbeat in the core SDK, which allows lang-side activities to send heartbeat requests rapidly without hitting server on every request. This is possibly because core would aggregate heartbeat requests for up to 1/2 of heartbeat timeout before sending them to the server. Note that first heartbeat request will be sent immediately if core hasn't seen request for the same task token within 1/2 of the heartbet timeout, this means that first heartbeat request for each activity will be sent immediately. Upon shutdown new heartbeat requests will not be recorded and errors will be returned to the core. Outstanding requests will be drained and sent to the server.

### DIFF
--- a/protos/local/core_interface.proto
+++ b/protos/local/core_interface.proto
@@ -16,7 +16,7 @@ import "google/protobuf/timestamp.proto";
 import "google/protobuf/duration.proto";
 import "google/protobuf/empty.proto";
 
-// A request as given to [crate::Core::send_activity_heartbeat]
+// A request as given to [crate::Core::record_activity_heartbeat]
 message ActivityHeartbeat {
     ActivityId activity_id = 1;
     repeated common.Payload details = 2;
--- a/protos/local/core_interface.proto
+++ b/protos/local/core_interface.proto
@@ -18,24 +18,8 @@ import "google/protobuf/empty.proto";
 
 // A request as given to [crate::Core::record_activity_heartbeat]
 message ActivityHeartbeat {
-    ActivityId activity_id = 1;
+    bytes task_token = 1;
     repeated common.Payload details = 2;
-}
-
-message ActivityId {
-    oneof id {
-        bytes task_token = 1;
-        ManualCompletionActivityTaskToken manual_completion_task_token = 2;
-    }
-}
-
-message ManualCompletionActivityTaskToken {
-    common.WorkflowExecution workflow_execution = 1;
-    string activity_id = 2;
-}
-
-message RecordActivityHeartbeatResponse {
-    bool cancel_requested = 1;
 }
 
 // A request as given to [crate::Core::complete_activity_task]
--- a/protos/local/core_interface.proto
+++ b/protos/local/core_interface.proto
@@ -20,6 +20,7 @@ import "google/protobuf/empty.proto";
 message ActivityHeartbeat {
     bytes task_token = 1;
     repeated common.Payload details = 2;
+    google.protobuf.Duration heartbeat_timeout = 3;
 }
 
 // A request as given to [crate::Core::complete_activity_task]
--- a/protos/local/core_interface.proto
+++ b/protos/local/core_interface.proto
@@ -18,8 +18,24 @@ import "google/protobuf/empty.proto";
 
 // A request as given to [crate::Core::send_activity_heartbeat]
 message ActivityHeartbeat {
-    string activity_id = 1;
+    ActivityId activity_id = 1;
     repeated common.Payload details = 2;
+}
+
+message ActivityId {
+    oneof id {
+        bytes task_token = 1;
+        ManualCompletionActivityTaskToken manual_completion_task_token = 2;
+    }
+}
+
+message ManualCompletionActivityTaskToken {
+    common.WorkflowExecution workflow_execution = 1;
+    string activity_id = 2;
+}
+
+message RecordActivityHeartbeatResponse {
+    bool cancel_requested = 1;
 }
 
 // A request as given to [crate::Core::complete_activity_task]
--- a/src/activity/activity_heartbeat_manager.rs
+++ b/src/activity/activity_heartbeat_manager.rs
@@ -80,8 +80,7 @@ impl ActivityHeartbeatManager {
         // 2. join the futures and for any future that becomes ready:
         // - send heartbeat details to the server (if any)
         // - clear details from the local cache
-        // - schedule next task
-        //in case if heartbeat has been sent
+        // - schedule next task in case if heartbeat has been sent, otherwise clear task token from scheduled_heartbeats
         unimplemented!()
     }
 }
--- a/src/activity/activity_heartbeat_manager.rs
+++ b/src/activity/activity_heartbeat_manager.rs
@@ -292,7 +292,7 @@ mod test {
             heartbeat_timeout: Some(Duration::from_millis(1000).into()),
         }) {
             Ok(_) => {
-                assert!(false, "heartbeat should not be recorded after the shutdown")
+                unreachable!("heartbeat should not be recorded after the shutdown")
             }
             Err(e) => {
                 matches!(e, ActivityHeartbeatError::ShuttingDown);
@@ -319,7 +319,7 @@ mod test {
             heartbeat_timeout: None,
         }) {
             Ok(_) => {
-                assert!(false, "heartbeat should not be recorded without timeout")
+                unreachable!("heartbeat should not be recorded without timeout")
             }
             Err(e) => {
                 matches!(e, ActivityHeartbeatError::HeartbeatTimeoutNotSet);
--- a/src/activity/activity_heartbeat_manager.rs
+++ b/src/activity/activity_heartbeat_manager.rs
@@ -106,7 +106,8 @@ impl<SG: ServerGatewayApis + Send + Sync + 'static> ActivityHeartbeatProcessor<S
         let details = self.heartbeat_rx.borrow().clone();
         let _ = self
             .server_gateway
-            .record_activity_heartbeat(self.task_token.clone(), details.into_payloads());
+            .record_activity_heartbeat(self.task_token.clone(), details.into_payloads())
+            .await;
         loop {
             sleep(self.delay).await;
             let stop = select! {
@@ -123,10 +124,9 @@ impl<SG: ServerGatewayApis + Send + Sync + 'static> ActivityHeartbeatProcessor<S
                 }
                 _ = self.heartbeat_rx.changed() => {
                     // Received new heartbeat details.
-                    let details = self.heartbeat_rx.borrow();
-                    // TODO see if we can get rid of cloning details.
+                    let details = self.heartbeat_rx.borrow().clone();
                     let _ = self.server_gateway
-                        .record_activity_heartbeat(self.task_token.clone(), details.clone().into_payloads());
+                        .record_activity_heartbeat(self.task_token.clone(), details.into_payloads()).await;
                     false
                 }
             };
--- a/src/activity/activity_heartbeat_manager.rs
+++ b/src/activity/activity_heartbeat_manager.rs
@@ -46,7 +46,10 @@ impl ActivityHeartbeatManager {
         self.heartbeat_details
             .insert(heartbeat.task_token.clone(), heartbeat.details);
 
-        if !self.scheduled_heartbeats.contains(&heartbeat.task_token) {
+        if self
+            .scheduled_heartbeats
+            .insert(heartbeat.task_token.clone())
+        {
             let initial_delay = time::Duration::from_millis(0);
             let heartbeat_timeout: time::Duration = heartbeat
                 .heartbeat_timeout
--- a/src/activity/activity_heartbeat_manager.rs
+++ b/src/activity/activity_heartbeat_manager.rs
@@ -145,16 +145,10 @@ impl<SG: ServerGatewayApis + Send + Sync + 'static> ActivityHeartbeatManager<SG>
         self.shutdown_tx
             .send(true)
             .expect("shutdown channel can't be dropped before shutdown is complete");
-        let mut pending_handles = vec![];
-        for v in self.heartbeat_processors.iter() {
-            self.heartbeat_processors.remove(v.key()).map(|v| {
-                pending_handles.push(v.1.join_handle);
-            });
+        for v in self.heartbeat_processors.drain() {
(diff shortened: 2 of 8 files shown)


protos/local/core_interface.proto
```diff
@@ -16,12 +16,28 @@ import "google/protobuf/timestamp.proto";
...
+    }
+}
+
+message ManualCompletionActivityTaskToken {
+    common.WorkflowExecution workflow_execution = 1;
+    string activity_id = 2;
+}
+
+message RecordActivityHeartbeatResponse {
+    bool cancel_requested = 1;
```
bergundy, line 38: I thought that we're notifying lang of activity cancellation in `ActivityTask`, that's at least what we've talked about.

protos/local/core_interface.proto
```diff
@@ -16,12 +16,28 @@ import "google/protobuf/timestamp.proto";
...
 }
 
+message ActivityId {
+    oneof id {
+        bytes task_token = 1;
+        ManualCompletionActivityTaskToken manual_completion_task_token = 2;
+    }
+}
+
+message ManualCompletionActivityTaskToken {
```
bergundy, line 32: I thought we could just use a service client to manually complete the activity since we might want to complete manually from a process that's not running a worker.

src/lib.rs
```diff
@@ -303,7 +326,17 @@ where
...
 
-    async fn send_activity_heartbeat(&self, _task_token: ActivityHeartbeat) -> Result<(), ()> {
+    async fn record_activity_heartbeat(
+        &self,
+        _details: ActivityHeartbeat,
+    ) -> Result<RecordActivityHeartbeatResponse, ActivityHeartbeatError> {
+        unimplemented!()
+    }
+
+    async fn get_last_activity_heartbeat(
```
bergundy, line 336: What is this for?
When is lang supposed to call this?

bergundy (commented):
Looks like we went in different directions.
I implemented things differently in the node SDK.

protos/local/core_interface.proto
Sushisource: The existing SDKs let you catch errors thrown by heartbeating which can include being cancelled. I think we need that, so then you'd have to block on sending a heartbeat until you get an `ActivityTask` that says it worked or not, which feels like a kinda roundabout way of doing it

src/errors.rs
```diff
@@ -143,3 +143,9 @@ pub enum CompleteActivityError {
     #[error("Unhandled error when calling the temporal server: {0:?}")]
     TonicError(#[from] tonic::Status),
 }
+
+/// Errors thrown by [crate::Core::record_activity_heartbeat] and [crate::Core::get_last_activity_heartbeat]
+#[derive(thiserror::Error, Debug)]
+pub enum ActivityHeartbeatError {
```
Sushisource, line 149: Our existing errors from lib functions aren't protos. I dunno that this one needs to be either

src/lib.rs
```diff
@@ -100,8 +102,29 @@ pub trait Core: Send + Sync {
...
 
-    /// Indicate that a long running activity is still making progress
-    async fn send_activity_heartbeat(&self, task_token: ActivityHeartbeat) -> Result<(), ()>;
+    /// Notify workflow that activity is still alive. Long running activities that take longer than
+    /// `activity_heartbeat_timeout` to finish must call this function in order to report progress,
+    /// otherwise activity will timeout and new attempt will be scheduled.
+    /// `result` contains latest known activity cancelation status.
+    /// Note that heartbeat requests are getting batched and are sent to the server periodically,
+    /// this function is going to return immediately and cancelation status will be taken from the server response
+    /// to the most recently sent batch.
```
Sushisource, line 111: Should we even bother with this? Do the existing SDKs batch? In thinking about it, it seems totally reasonable to expect the activity code to block briefly while it does this. It also gives you a reason *not* to do it in a tight loop, etc, which is probably a good thing. We get to avoid the complexity of this mechanism too.

protos/local/core_interface.proto
vitarb: I think we should discuss this, let me schedule a meeting to clear everything out.

src/lib.rs
vitarb: When you want details of the last heartbeat back, technically you have it, but that way you would need to implement the logic on the lang side. (e.g. latest heartbeat details sent or heartbeat details from the activity task)

protos/local/core_interface.proto
vitarb: Update: after offline conversation we've decided to send cancellation as an activity task to lang. This would allow us innovate more in the area of activity cancellations and to dispatch cancellations without heartbeats in the future.

protos/local/core_interface.proto
vitarb: That's a fair point, let's simplify it for now and see if we need manual completion in the core later.

src/lib.rs
vitarb: Discussed offline, summary: we are going to decouple heartbeats and cancellation in the long run. Heartbeat reporting should be fast and can be batched by the core.

src/activity/activity_heartbeat_manager.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::convert::TryInto;
+use std::future::Future;
+use std::ops::Div;
+use std::sync::Arc;
+use std::{thread, time};
+use tokio::time::{sleep, Sleep};
+
+pub(crate) struct ActivityHeartbeatManager {
+    /// Core will aggregate activity heartbeats and send them to the server periodically.
+    /// This map contains latest heartbeat details for each activity that reported them.
```
Sushisource, line 16: Always good to say what the key is in maps (and the set below). We should probably just make a task token type at this point, but that can be another PR

src/activity/activity_heartbeat_manager.rs
```diff
@@ -0,0 +1,86 @@
...
+    task_token: Vec<u8>,
+    next_delay: time::Duration,
+    trigger: Sleep,
+}
+
+impl ActivityHeartbeatManager {
+    pub fn new() -> Self {
+        let (heartbeat_tx, heartbeat_rx) = unbounded::<ActivityHeartbeatTask>();
+        let (shutdown_tx, shutdown_rx) = bounded(1);
+        thread::spawn(move || Self::processor(heartbeat_rx, shutdown_rx));
```
Sushisource, line 36: This should very likely be a `tokio::spawn` rather than a dedicated thread

src/pollers/mod.rs
```diff
@@ -132,6 +136,27 @@ pub trait ServerGatewayApis {
...
+        &self,
+        task_token: Vec<u8>,
+        details: Option<Payloads>,
+    ) -> Result<RecordActivityTaskHeartbeatResponse>;
+
+    /// Report activity task heartbeat by sending details to the server. This function should only be used
+    /// when task token is not available, such as in case of manual completion, otherwise use `record_activity_heartbeat`.
+    /// `activity_id`, `workflow_id` and `run_id` must be set to respective IDs of activity, workflow and workflow run.
+    /// `result` contains `cancel_requested` flag, which if set to true indicates that activity has been cancelled.
+    async fn record_activity_heartbeat_by_id(
```
Sushisource, line 152: Rather than making two separate APIs it may make sense to use an enum that is either a task token or this info

src/activity/activity_heartbeat_manager.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    pub fn shutdown(&self) {
+        self.shutdown_tx.send(true);
+    }
+
+    fn processor(heartbeat_rx: Receiver<ActivityHeartbeatTask>, shutdown_rx: Receiver<bool>) {
+        // TODO this function should:
+        // 1. consume heartbeat tasks
+        // 2. join the futures and for any future that becomes ready:
```
Sushisource, line 80: This operation is called select, not join. Joining them all means they all must finish, whereas you only care about when one finishes. https://docs.rs/futures/0.3.14/futures/future/fn.select_all.html would do what you want.

However, I think there is a simpler design here.

There is no need for the channels and this separate processor thread/task I think. The design can instead be to keep one map of token -> spawn handle, where the spawn handle is the result of a future you have spawned whose job is to issue heartbeats for that task token (like an actor model). Then instead of having this processor thread that manages all hearbeat tasks, you have a task-per-task-token with a simpler responsibility (just managing the timing of sending out heartbeats for an individual activity). Upon shutdown, you just join all the spawn handles, ensuring final heartbeats are sent.

The per-activity spawned task would (mostly pseudocode) look something like:
```rust
let channel_for_activity_tx, channel_for_activity_rx = tokio::sync::watch()
// Put tx side in the map with the spawn handle
async {
  let timeout_or_watch_changed = select!(sleep(no_heartbeats_duration), channel_for_activity_rx.changed());
  while(timeout_or_watch_changed.await != timed out) {
      send_heartbeat_to_server(...).await;
      sleep(interval between sending batches).await;
  }
}
```

Sushisource: The watch channel can contain the heartbeat details

src/lib.rs
vitarb: As discussed, going to ditch this from the core and let lang track heartbeats.

src/pollers/mod.rs
vitarb: That's what I've just ditched, I thought we were going to get rid of this one as well in favor of calling heartbeat by id from the lang?

src/activity/activity_heartbeat_manager.rs
vitarb: Thanks Spencer, let me do it this way.

src/activity/activity_heartbeat_manager.rs
```diff
@@ -1,86 +1,119 @@
...
+            Some(handle) => {
+                handle.heartbeat_tx.send(heartbeat.details);
+            }
+            None => {
+                let heartbeat_timeout: time::Duration = heartbeat
+                    .heartbeat_timeout
+                    .ok_or(ActivityHeartbeatError::HeartbeatTimeoutNotSet)?
+                    .try_into()
+                    .or(Err(ActivityHeartbeatError::InvalidHeartbeatTimeout))?;
+                let delay = heartbeat_timeout.div(2);
```
Sushisource, line 64: This delay factor should probably be configurable

src/activity/activity_heartbeat_manager.rs
```diff
@@ -1,86 +1,119 @@
...
-    task_token: Vec<u8>,
-    next_delay: time::Duration,
-    trigger: Sleep,
+struct ActivityHeartbeatProcessorHandle {
+    heartbeat_tx: Sender<Vec<common::Payload>>,
+    join_handle: JoinHandle<()>,
+}
+
+struct ActivityHeartbeatProcessor {
+    delay: time::Duration,
```
Sushisource, line 33: Slightly more specific name for `delay` might be nice - like `max_delay_between_beats` ?

src/activity/activity_heartbeat_manager.rs
Sushisource: Oh, I see why it isn't at the moment -- it might be nice to do the two different timeouts thing I mentioned which is still pretty simple, where there's one delay for "minimum interval between beats" and another delay for "give up if you haven't seen a heartbeat in this long". Not a huge deal tho

src/activity/activity_heartbeat_manager.rs
```diff
@@ -1,86 +1,119 @@
...
+                }
+                _ = self.heartbeat_rx.changed() => {
+                    // Received new heartbeat details.
+                    let details = self.heartbeat_rx.borrow();
+                    // TODO send details right away since enough time has passed since previous call
+                    false
+                }
+            };
+            if stop {
+                break;
```
Sushisource, line 114: Rather than returning a bool and then `break`ing here you could just do it inside the select. Ultimately a bit less code.

src/lib.rs
```diff
@@ -110,7 +126,7 @@ pub trait Core: Send + Sync {
     /// call [Core::complete_workflow_task] for those workflows until they are done. At that point,
     /// the lang SDK can end the process, or drop the [Core] instance, which will close the
     /// connection.
```
Sushisource, line 128: This should be updated to indicate that now that it's async it returns when shutdown is actually all the way done

tests/integ_tests/simple_wf_tests.rs
```diff
@@ -809,17 +895,17 @@ async fn shutdown_aborts_actively_blocked_poll() {
...
         std::thread::sleep(Duration::from_millis(100));
-        tcore.shutdown();
+        tcore.shutdown().await;
     });
     assert_matches!(
         core.poll_workflow_task().await.unwrap_err(),
         PollWfError::ShutDown
     );
-    handle.join().unwrap();
+    let _ = handle.await;
```
Sushisource, line 906: Should be handle.await.unwrap()

src/lib.rs
Sushisource: And, come to think of it, should return a result now and propagate errors but just make a task for that and we can do it after this. Shouldn't take long.

src/activity/activity_heartbeat_manager.rs
vitarb: Done

src/activity/activity_heartbeat_manager.rs
vitarb: I suggest we keep it simple for now, as both are the same and separate them if we ever need to set different values.

src/activity/activity_heartbeat_manager.rs
vitarb: Done, for some reason I thought that break would break out of the select initially.

tests/integ_tests/simple_wf_tests.rs
vitarb: Done

src/lib.rs
vitarb: Sounds good, slightly updated the comment.

src/activity/activity_heartbeat_manager.rs
vitarb: As mentioned above, let's keep it ambiguous for now as it serves dual purpose, we can choose more specific names if we decide to split up give up timeout and the delay.

Sushisource (approved):

PR #94, opened 2021-04-14
Panic on histories with pagination tokens, for now.

Sushisource:
## What was changed:
Crash if we get a history with a pagination token, until we implement it

## Why?
Better than silently failing to apply the whole history

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -206,6 +206,10 @@ where
 
             match selected_f {
                 Ok(work) => {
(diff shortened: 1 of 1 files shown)


bergundy (approved):
That was a quick one :)

vitarb (approved):

Issue #91, opened 2021-04-12
What happens when a new pending activation interrupts long poll?


Sushisource:
See original discussion here: https://github.com/temporalio/sdk-core/pull/89

We need to test and determine best course of action when a new PA interrupts a poll. Ideally it should cancel the request, but there are many potential weird corner cases here like what to do if things race and we somehow get a new task back but don't acknowledge it.

bergundy:
Does this happen when we receive a workflow task that triggers multiple "sequential" activations and haven't sent them all out to lang while lang polls for the next workflow task?
Perhaps it would make more sense to split polling into 2 methods:
poll_workflow_task_for_specific_workflow
poll_any_workflow_task

Or maybe when a workflow activation completes core responds with next pending activation if one is available.

We should think about it.

bergundy:
Maybe buffering the poll response from the server is a better way to go here.
It keeps the interface simple and the only downside I see is that lang might not pull the buffered task fast enough to avoid WFT timeout.
This might not be such a bad thing considering that lang controls the WFT poll limits.

Sushisource:
I'm worried buffering just opens up too many corner cases. You have to deal with the timeout you mentioned, but also that the buffered poll could be for a *different* task queue than the one you just asked to poll, etc etc.

But.....

Vitaly and I had talked before about potentially responding to completions with new pending activations. We decided against it initially because it felt a little unpleasant to have to have a whole new control flow for that. But, it does solve this problem handily, and the "different task queues" problem actually *already* exists for Pending Activations and it would solve that too. So, that might be a nice way to kill two birds with one stone. It does make things way more clear in that "poll" now always means "yes, actually poll the server" and frankly it'd probably make a bunch of what I am doing in https://github.com/temporalio/sdk-core/pull/93 easier. @vitarb What do you think?

bergundy:
What if we just respond with all pending activation to the poll request? Is there a reason not to do that?

Sushisource:
> What if we just respond with all pending activation to the poll request? Is there a reason not to do that?

Per discussion in call: Not possible. We need lang to reply with appropriate commands during replay to check against history.

mfateev:
Note that synchronous reply can take a long time if the core has to retrieve the next page of a history from the service.

bergundy:
That's not an issue for node and python at least because the api call does
not block a thread.

Sushisource:
[...]

PR #96, opened 2021-04-15
Eliminate pending activations by returning new activations from complete

Sushisource:
## What was changed:
Return new activations from completing them, when required

## Why?
* Eliminates complexity around pending activations
* When you call poll, you always poll the server now
* When you poll, your activation is always for that task queue
* Eliminates question about how to "buffer" server responses if lang is dealing with pending activations - so closes https://github.com/temporalio/sdk-core/issues/91

## Checklist
TODO: 
- [ ] Saw stalling behavior if lang side didn't deal with returned activations from a completion (I think this is because workflow task timeout isn't working right)
- [ ] Cleanup pass

1. Closes issue: 
Closes https://github.com/temporalio/sdk-core/issues/91

2. How was this tested:
Existing & new tests


Sushisource:
This may get closed in favor of another solution Max helped us realize that maintains the existing API by simply not starting any polling behavior until all pending activations are cleared out. Will leave as draft until I've had time to explore that.

vitarb:
As far as I understand, we are not going to land this one, right?

Sushisource:
> As far as I understand, we are not going to land this one, right?

Nope. Not for alpha at least. Want to keep it around if we revisit after.

bergundy:
@Sushisource should we keep this draft?
I think we pretty much realized that this not the direction we're taking.

Sushisource:
:wave: Goodbye, PR. I still think you're a nifty idea but it's been too long.

PR #97, opened 2021-04-16
Buffer poll results rather than cancelling the poll for pending activations

Sushisource:
## What was changed:
Buffer polls rather than cancelling them when a new pending activation comes in that must be handed out first. Also limit the number of concurrent workflow tasks that may be in flight at one time, and use a per-core-instance task queue rather that per-poll.

## Why?
Avoid the PA interrupt problem stated in https://github.com/temporalio/sdk-core/issues/91 without the bigger API changes in https://github.com/temporalio/sdk-core/pull/96 .

However, I'm still of the opinion the different API design is ultimately simpler

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,10 +148,15 @@ struct CoreSDK<WP> {
     server_gateway: Arc<WP>,
     /// Key is run id
     workflow_machines: WorkflowConcurrencyManager,
+    // TODO: Probably move all workflow stuff inside wf manager?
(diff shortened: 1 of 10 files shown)


Sushisource:
This is problematic in it's current form as it limits us to only handling one workflow instance at a time

vitarb (approved):

vitarb (approved):

PR #98, opened 2021-04-20
Allow configuring maximum concurrent activity limits

Sushisource:
## What was changed:
New config option for max concurrent activity tasks

## Why?
Otherwise workers could get immediately swamped

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -125,6 +125,8 @@ pub struct CoreInitOptions {
     /// time. Note that one workflow task may require multiple activations - so the WFT counts as
     /// "outstanding" until all activations it requires have been completed.
     pub max_outstanding_workflow_tasks: usize,
+    /// The maximum allowed number of activity tasks that will ever be given to lang at one time.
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -1579,4 +1600,67 @@ mod test {
...
+                evict_after_pending_cleared: true,
+                max_outstanding_workflow_tasks: 1,
+                max_outstanding_activities: 2,
+            },
+        );
+
+        // We allow two outstanding activities, therefore first two polls should return right away
+        let r1 = core.poll_activity_task().await.unwrap();
+        let _r2 = core.poll_activity_task().await.unwrap();
+        // Third should block until we complete one of the first two
```
bergundy, line 1652: I don't understand how this is tested, how do we know the 3rd poll was blocked?

Sushisource: Good point. Added a way to verify that

vitarb (approved):

PR #99, opened 2021-04-21
Target c5-2xl workers in builds

vitarb:
We have t3-s workers in the mix, alongside c5-s, core builds time out on t3s, this PR sets configuration that should target c5-s for build/test/lint


vitarb:
Followed up with our infrastructure team, and we were able to change defaults to c5a.2xl for everything, so there is no need for this override.

PR #100, opened 2021-04-22
Return evictions to lang and also combine pending activity jobs

Sushisource:
## What was changed:
Evictions are now sent to lang -- new pending activities that get queued will collapse jobs if there is already a pending activity with the same ID.

## Why?
Need to tell lang about cache evictions

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/protos/local/workflow_activation.proto
+++ b/protos/local/workflow_activation.proto
@@ -42,6 +42,13 @@ message WFActivationJob {
         SignalWorkflow signal_workflow = 7;
         /// An activity was resolved with, result could be completed, failed or cancelled
         ResolveActivity resolve_activity = 8;
+
+        /// Remove the workflow identified by the [WFActivation] containing this job from the cache
+        /// after performing the activation.
+        ///
+        /// If other job variant are present in the list, this variant must be the last job in the
+        /// job list. The actual value is irrelevant, and should always be true.
+        bool remove_from_cache = 50;
     }
 }
 
--- a/protos/local/workflow_activation.proto
+++ b/protos/local/workflow_activation.proto
@@ -44,8 +44,9 @@ message WFActivationJob {
         /// Remove the workflow identified by the [WFActivation] containing this job from the cache
         /// after performing the activation.
         ///
-        /// If other job variant are present in the list, this variant must be the last job in the
-        /// job list. The actual value is irrelevant, and should always be true.
+        /// If other job variant are present in the list, this variant will be the last job in the
+        /// job list. The boolean value is irrelevant, since the variant type is what matters. It
+        /// will be set to true if this is the variant.
         bool remove_from_cache = 50;
     }
 }
--- a/protos/local/workflow_activation.proto
+++ b/protos/local/workflow_activation.proto
@@ -22,8 +22,6 @@ message WFActivation {
     string run_id = 3;
     /// The things to do upon activating the workflow
     repeated WFActivationJob jobs = 4;
-    /// True if this activation was internally generated rather than from the server
-    bool from_pending = 5;
 }
 
 message WFActivationJob {
--- a/src/pending_activations.rs
+++ b/src/pending_activations.rs
@@ -1,5 +1,6 @@
 use crate::protos::coresdk::workflow_activation::WfActivation;
 use parking_lot::RwLock;
+use slotmap::SlotMap;
 use std::collections::{HashMap, VecDeque};
 
 /// Tracks pending activations using an internal queue, while also allowing lookup and removal of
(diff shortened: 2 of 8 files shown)


src/pending_activations.rs
```diff
@@ -9,42 +9,74 @@ pub struct PendingActivations {
...
+        if let Some(act) = inner.activations.get_mut(&wf_id) {
+            act.jobs.extend(v.jobs);
+        } else {
+            // Incr run id count
+            *inner
+                .count_by_run_id
+                .entry(v.run_id.clone())
+                .or_insert_with(|| 0) += 1;
+            // TODO: Make better if true
+            assert!(
```
Sushisource, line 48: This does appear to be true - will clean up code a bit.

protos/local/workflow_activation.proto
```diff
@@ -42,6 +42,13 @@ message WFActivationJob {
         SignalWorkflow signal_workflow = 7;
         /// An activity was resolved with, result could be completed, failed or cancelled
         ResolveActivity resolve_activity = 8;
+
+        /// Remove the workflow identified by the [WFActivation] containing this job from the cache
+        /// after performing the activation.
+        ///
+        /// If other job variant are present in the list, this variant must be the last job in the
+        /// job list. The actual value is irrelevant, and should always be true.
+        bool remove_from_cache = 50;
```
vitarb, line 51: Why 50 and not 9?

protos/local/workflow_activation.proto
```diff
@@ -42,6 +42,13 @@ message WFActivationJob {
         SignalWorkflow signal_workflow = 7;
         /// An activity was resolved with, result could be completed, failed or cancelled
         ResolveActivity resolve_activity = 8;
+
+        /// Remove the workflow identified by the [WFActivation] containing this job from the cache
+        /// after performing the activation.
+        ///
+        /// If other job variant are present in the list, this variant must be the last job in the
```
vitarb, line 49: If that's a guarantee that we provide in the core, we should probably be a bit more assertive and say "this variant is guaranteed to be the last job ..."

protos/local/workflow_activation.proto
```diff
@@ -42,6 +42,13 @@ message WFActivationJob {
...
+        /// job list. The actual value is irrelevant, and should always be true.
```
vitarb, line 50: This last sentence is a bit confusing. Do you mean to say that in case if it's set it should always be set to true?

src/pending_activations.rs
vitarb, line 48: Is it considered okay to use assert in non-test code in rust?

vitarb (approved):
LGTM, couple minor comments.

protos/local/workflow_activation.proto
bergundy: It makes sense that it's 50 so the additional commands can all have an incrementing sequence

src/pending_activations.rs
```diff
@@ -9,42 +9,74 @@ pub struct PendingActivations {
...
 
 impl PendingActivations {
     pub fn push(&self, v: WfActivation) {
         let mut inner = self.inner.write();
-        *inner
-            .count_by_id
-            .entry(v.run_id.clone())
-            .or_insert_with(|| 0) += 1;
-        inner.queue.push_back(v);
+        let wf_id = ActivationId {
```
bergundy, line 33: This probably shouldn't be called `wf_id`.

src/pending_activations.rs
bergundy: Does a single WFT alway map to a single activation?

src/protos/mod.rs
```diff
@@ -31,6 +31,17 @@ pub mod coresdk {
...
         include!("coresdk.workflow_activation.rs");
+        pub fn create_evict_activation(task_token: Vec<u8>, run_id: String) -> WfActivation {
+            WfActivation {
+                task_token,
+                timestamp: None,
+                run_id,
+                jobs: vec![WfActivationJob::from(
+                    wf_activation_job::Variant::RemoveFromCache(true),
+                )],
+                from_pending: true,
```
bergundy, line 42: Why do we have `from_pending` in the protocol? It doesn't seem like it should interest lang, am I wrong?

Sushisource: Not anymore it doesn't. I can clean this up now because of this change

src/pending_activations.rs
Sushisource: I think using assertions in production code where you know some invariant must be upheld is a good idea. It makes the code less verbose b/c you can avoid results, while also making it super clear during testing if you violate the invariant. So, this code now actually has a few of them, most of which are totally internal (keeping maps/vec in sync) and one of which is this "no PAs with the same run id but different task tokens" which must be upheld in Core itself

protos/local/workflow_activation.proto
```diff
@@ -42,6 +40,14 @@ message WFActivationJob {
...
         /// An activity was resolved with, result could be completed, failed or cancelled
         ResolveActivity resolve_activity = 8;
+
+        /// Remove the workflow identified by the [WFActivation] containing this job from the cache
+        /// after performing the activation.
+        ///
+        /// If other job variant are present in the list, this variant will be the last job in the
+        /// job list. The boolean value is irrelevant, since the variant type is what matters. It
+        /// will be set to true if this is the variant.
+        bool remove_from_cache = 50;
```
mfateev, line 50: This is a weird place for this flag. Why is this a different type of WFActivationJob? Or at least a flag on WFActivation?

Sushisource: It is a different type of job - it's inside the oneof here for what the job is.

PR #101, opened 2021-04-22
Remove heartbeat timeout from core heartbeat API

vitarb:
This change makes core memorize heartbeat timeout from the poll activity task response, rather than require it from the lang side.

### DIFF
--- a/src/activity/details.rs
+++ b/src/activity/details.rs
@@ -1,13 +1,8 @@
 use prost_types::Duration;
 
 /// Contains minimal set of details that core needs to store, while activity is running.
+#[derive(derive_more::Constructor)]
 pub(crate) struct InflightActivityDetails {
     /// Used to calculate aggregation delay between activity heartbeats.
     pub heartbeat_timeout: Option<Duration>,
-}
-
-impl InflightActivityDetails {
-    pub fn new(heartbeat_timeout: Option<Duration>) -> Self {
-        Self { heartbeat_timeout }
-    }
 }
--- a/src/activity/details.rs
+++ b/src/activity/details.rs
@@ -1,0 +1,13 @@
+use prost_types::Duration;
+
+/// Contains minimal set of details that core needs to store, while activity is running.
+pub(crate) struct InflightActivityDetails {
+    /// Used to calculate aggregation delay between activity heartbeats.
+    pub heartbeat_timeout: Option<Duration>,
+}
+
+impl InflightActivityDetails {
+    pub fn new(heartbeat_timeout: Option<Duration>) -> Self {
+        Self { heartbeat_timeout }
+    }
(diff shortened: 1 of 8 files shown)


src/activity/details.rs
```diff
@@ -0,0 +1,13 @@
...
+pub(crate) struct InflightActivityDetails {
+    /// Used to calculate aggregation delay between activity heartbeats.
+    pub heartbeat_timeout: Option<Duration>,
+}
+
+impl InflightActivityDetails {
+    pub fn new(heartbeat_timeout: Option<Duration>) -> Self {
+        Self { heartbeat_timeout }
+    }
+}
```
Sushisource, line 13: This can be replaced with `#[derive(derive_more::Constructor)]` on the struct

src/activity/details.rs
```diff
@@ -0,0 +1,13 @@
+use prost_types::Duration;
+
+/// Contains minimal set of details that core needs to store, while activity is running.
+pub(crate) struct InflightActivityDetails {
+    /// Used to calculate aggregation delay between activity heartbeats.
+    pub heartbeat_timeout: Option<Duration>,
```
Sushisource, line 6: Why is this optional? We throw errors later on if it was never set. We should reject the activity task as malformed if it is missing a timeout when we first insert it into the map

Sushisource: Unless that can happen because the user never sent one and the server just returns whatever they sent. I'd hope the server just wouldn't allow that.

The situation where you don't set the timeout and then don't learn until you try to beat, at which point it's too late for you to do anything except cancel the activity is kinda annoying

Sushisource: It might make more sense to just assume a default timeout at that point

src/activity/details.rs
vitarb: Nice!

src/activity/details.rs
vitarb: Discussed with Spencer, Max and Alex offline:
- currently server sets `heartbeat_timeout` to `0` when it's not populated (bug), I've followed up with @alexshtin, it will be addressed.
- as per discussion with @mfateev if heartbeat timeout is not set, we shouldn't be allowing heartbeats, Max was opposed to the idea of having defaults
- as to optional vs not, I think it's better to keep it optional here as it would not leak heartbeat specific logic into poll API, especially given that there are legit error cases, and we don't want to see heartbeat related errors when we poll for activity task.

So my suggestion is:
Let's keep it as optional and throw errors to the lang, which can then log them and treat them accordingly. For now let's add a `0` check, if heartbeat timeout is `0` then it should be treated the same way as `None`.

Sushisource (approved):

PR #105, opened 2021-04-23
Handling server errors during heartbeat

vitarb:
This PR adds handling for server errors during activity heartbeat.
Some errors will be retried, while others will be propagated back to the user.
Error dispatch is similar to cancellation dispatch that comes as part of the heartbeat.
I copied current behavior [from java](https://github.com/temporalio/sdk-java/blob/86189b209f18f3d906de8bbd7f7ce2aaadec6414/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java#L221), but I think there is some room for improvement in terms of retry strategy.


Sushisource:
I don't think we want to propagate these back to lang. I think simply logging warnings is adequate, and for things like "workflow already complete" we can just kill the processor.

Issue #110, opened 2021-04-26
[Bug] Completed activity for which cancellation was requested causes duplicate `ResolveActivity` activation job


bergundy:
**Describe the bug**
When a Workflow requests to cancel an Activity, Core uses `TRY_CANCEL` mode and immediately activates the workflow with a `ResolveActivity` activation where the result is `canceled`.
If the activity completes before it heartbeats again core will resolve the activity again, this time with result set to `completed`.
The node SDK will throw an error saying that there's no pending completion for that activity.

**To Reproduce**
Run the `cancel-http-request` workflow in the node sdk and set the activity's `heartbeatTimeout` to something high like `3 minutes`.

**Expected behavior**
Core should not resend an `ActivityResolve` activation job.

**Screenshots/Terminal ouput**
```
Caught error while worker was running Error: No completion for taskSeq 0
```

bergundy:
Run the the test with this command:
```shell
RUN_INTEGRATION_TESTS=1 npx ava packages/test/lib/test-integration.js -m "cancel-http-request don't completeOnActivityCancellation" -T 100s
```

bergundy:
@Sushisource was this fixed by #112?

Sushisource:
Indeed it was, thanks for the reminder.

PR #112, opened 2021-04-26
Fix activity double resolve -- but do we want to?

Sushisource:
## What was changed:
Fixes double-resolves by squelching them lang side, but this introduces a problem where we can get a workflow task from the server that we have nothing to about. We could simply respond immediately with an empty command list, but that feels a little weird and introduces completions inside poll which is nonobvious.

We can instead just not do this, and expect lang to ignore (or do whatever else it wants) with the extra completion. It would then also have to return an empty response if there is no further work to be done.

Currently, what will happen if there are no other events in the history is the workflow task will be dropped and eventually time out. Results in a history like this:
```
11
ActivityTaskCancelRequested

12
TimerStarted

13
ActivityTaskStarted

14
ActivityTaskCompleted <-- we don't care about this because we already cancelled it locally

15
WorkflowTaskScheduled

16
WorkflowTaskStarted

17
WorkflowTaskTimedOut
10s (+10s)
```

## Checklist

1. Closes issue: 
closes https://github.com/temporalio/sdk-core/issues/110
2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/machines/activity_state_machine.rs
+++ b/src/machines/activity_state_machine.rs
@@ -82,7 +82,7 @@ fsm! {
     StartedActivityCancelEventRecorded --(ActivityTaskFailed(ActivityTaskFailedEventAttributes),
         on_activity_task_failed) --> Failed;
     StartedActivityCancelEventRecorded --(ActivityTaskCompleted(ActivityTaskCompletedEventAttributes),
-        on_activity_task_completed) --> Completed;
+        shared on_activity_task_completed) --> Completed;
     StartedActivityCancelEventRecorded --(ActivityTaskTimedOut(ActivityTaskTimedOutEventAttributes),
         shared on_activity_task_timed_out) --> TimedOut;
     StartedActivityCancelEventRecorded --(ActivityTaskCanceled(ActivityTaskCanceledEventAttributes),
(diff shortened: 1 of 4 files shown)


src/machines/activity_state_machine.rs
```diff
@@ -497,12 +497,17 @@ pub(super) struct StartedActivityCancelEventRecorded {}
...
     pub(super) fn on_activity_task_completed(
         self,
+        dat: SharedState,
         attrs: ActivityTaskCompletedEventAttributes,
     ) -> ActivityMachineTransition<Completed> {
-        ActivityMachineTransition::ok(
-            vec![ActivityMachineCommand::Complete(attrs.result)],
-            Completed::default(),
-        )
+        // We don want to tell lang about completions if we already pre-emptively canceled the
```
vitarb, line 503: I would say that in abandon and trycancel, workflow is not waiting for the activity to complete so we don't need to send activation. Also I believe abandon can not get here so it is just try cancel.

src/machines/activity_state_machine.rs
```diff
@@ -497,12 +497,17 @@ pub(super) struct StartedActivityCancelEventRecorded {}
...
+        // activity
+        let cmds = if dat.cancellation_type == ActivityCancellationType::WaitCancellationCompleted {
```
vitarb, line 505: You can make a match with unreachable for abandon i believe.

vitarb (approved):
lgtm, couple small comments

src/machines/activity_state_machine.rs
Sushisource: Yep, that's exactly what this is doing. The empty list means no activation gets sent.

Sushisource:
Just going to merge everything at once in https://github.com/temporalio/sdk-core/pull/113

PR #115, opened 2021-04-28
Add concurrent heartbeat cancellation test

Sushisource:
## What was changed:
Added test to verify core isn't getting hung up on many concurrent heartbeat cancels.
Also undo issuing pointless activity task for unknown activities.

## Why?
Better testing around heartbeat concurrency, bench investigation.

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2171,9 +2171,9 @@ mod test {
                 let calls = match calls_map.entry(tt) {
                     Entry::Occupied(mut e) => {
                         *e.get_mut() += 1;
-                        e.get().clone()
+                        *e.get()
                     }
-                    Entry::Vacant(v) => v.insert(1).clone(),
+                    Entry::Vacant(v) => *v.insert(1),
                 };
                 async move {
                     if calls < 5 {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -310,6 +310,8 @@ where
                         } else {
                             warn!(task_token = ?task_token,
                                   "Unknown activity task when issuing cancel");
+                            // If we can't find the activity here, it's already been completed,
+                            // in which case issuing a cancel again is pointless.
                             continue;
                         }
                     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -301,12 +301,18 @@ where
                         // It's possible that activity has been completed and we no longer have an
                         // outstanding activity task. This is fine because it means that we no
                         // longer need to cancel this activity, so we'll just ignore such orphaned
-                        // cancellation.
-                        if let Some(details) = self.outstanding_activity_tasks.get(&task_token) {
-                            return Ok(ActivityTask::cancel_from_ids(
-                                task_token,
-                                details.activity_id.clone(),
-                            ));
+                        // cancellations.
+                        if let Some(mut details) =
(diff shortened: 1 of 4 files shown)


Sushisource:
This is not right - lang should confirm the cancel with status canceled, but there is still interesting weird behavior to dig into here.

src/lib.rs
```diff
@@ -298,18 +298,27 @@ where
...
+                        if let Some((_, details)) =
+                            self.outstanding_activity_tasks.remove(&task_token) {
+                                return Ok(ActivityTask::cancel_from_ids(
+                                    task_token,
+                                    details.activity_id,
+                                ));
                         } else {
                             // TODO: Unless we return this JS seems to just give up on polling.
                             //   -- we shouldn't need to return it?
+                            // TODO: NEW TODO: (lol) -- This may be fixed now. Confirm.
```
vitarb, line 321: What is this? :)

src/lib.rs
```diff
@@ -298,18 +298,27 @@ where
...
                         // outstanding activity task. This is fine because it means that we no
                         // longer need to cancel this activity, so we'll just ignore such orphaned
                         // cancellation.
-                        if let Some(details) = self.outstanding_activity_tasks.get(&task_token) {
-                            return Ok(ActivityTask::cancel_from_ids(
-                                task_token,
-                                details.activity_id.clone(),
-                            ));
+                        if let Some((_, details)) =
+                            self.outstanding_activity_tasks.remove(&task_token) {
```
vitarb, line 313: I don't understand why is cleanup during completion is insufficient. No matter how cancellation is dispatched activity should at the end be completed by lang, which is when we should remove from this map. Otherwise we lose track of what is actually in progress on the lang side.

Sushisource: Yeah, not doing this.

src/lib.rs
```diff
@@ -308,14 +308,9 @@ where
...
                         } else {
-                            // TODO: Unless we return this JS seems to just give up on polling.
-                            //   -- we shouldn't need to return it?
                             warn!(task_token = ?task_token,
                                   "Unknown activity task when issuing cancel");
-                            return Ok(ActivityTask::cancel_from_ids(
-                                task_token,
-                                "unknown".to_owned(),
-                            ));
+                            continue;
```
vitarb, line 313: Should we leave a comment saying that we just ignore cancels from the heartbeat for activities that already completed?

Sushisource: Yeah, I'll add that. This is actually exactly how the code was before I added this weird fake activity task, which from the old TODO you can see I didn't really want to do anyway.

src/lib.rs
```diff
@@ -308,14 +308,9 @@ where
                                 details.activity_id.clone(),
                             ));
```
vitarb, line 309: Should we record cancel requested flag in the `InflightActivityDetails` and avoid sending extra cancels on every heartbeat? This probably is minor, but I just don't see the reason to deliver cancellation multiple times.

Sushisource: Oh, that's a really good idea... I'll add that.

vitarb (approved):
Approved with a couple small comments.

PR #116, opened 2021-04-29
Task token type, concurrent workflows test, better logs

Sushisource:
## What was changed:
Added a task token type for better formatting and better clarity in the code.
Added big concurrent workflows test & corresponding test upgrades.
Reduced hugely verbose tracing output.

This would've been easier if https://github.com/danburkert/prost/issues/334 was fixed in the protobuf builder. Will be nice when it is.

## Why?
Easier debugging

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,7 @@ version = "0.1"
 
 [dev-dependencies]
 assert_matches = "1.4"
+bimap = "0.6.1"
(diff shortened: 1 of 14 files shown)


vitarb (approved):
Nice!

PR #118, opened 2021-05-04
Log heartbeat errors instead of returning them back to lang

vitarb:
## What was changed:
Errors returned by the heartbeat, such as validation failures and heartbeat during shutdown are logged and swallowed instead of being returned to the lang.

## Why?
It has been agreed that this is an optimal behavior for the user to not interrupt activity flow if heartbeat is not working properly.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -119,11 +119,10 @@ pub trait Core: Send + Sync {
     /// cancellation processing.
     /// For now activity still needs to heartbeat if it wants to receive cancellation requests.
     /// In the future we are going to change this and will dispatch cancellations more proactively.
-    /// Note that this function is not blocking on the server call and will return immediately.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -830,6 +814,34 @@ impl<WP: ServerGatewayApis + Send + Sync + 'static> CoreSDK<WP> {
     }
 }
 
+impl<WP> CoreSDK<WP>
+where
+    WP: ServerGatewayApis + Send + Sync + 'static,
+{
```
Sushisource, line 820: This impl doesn't need to be redeclared. Same as the above one.

Sushisource (approved):
Small bit of redudancy needs fixing otherwise good

src/lib.rs
vitarb: Done

PR #122, opened 2021-05-11
Enable mostly configurable TLS in core gRPC client

Sushisource:
## What was changed:
Enable setting options when initting core that will optionally enable TLS or mTLS

This doesn't nearly expose the full range of things you could possibly configure but covers probably 90% of cases with a simple set of options.

## Why?
Encryption is good

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/pollers/mod.rs
+++ b/src/pollers/mod.rs
@@ -1,15 +1,14 @@
 mod poll_buffer;
 
+#[cfg(test)]
+pub use manual_mock::MockManualGateway;
 pub use poll_buffer::{
     new_activity_task_buffer, new_workflow_task_buffer, PollActivityTaskBuffer,
     PollWorkflowTaskBuffer,
 };
 
-use crate::protos::temporal::api::workflowservice::v1::{
-    RecordActivityTaskHeartbeatRequest, RecordActivityTaskHeartbeatResponse,
-};
-use crate::task_token::TaskToken;
 use crate::{
+    errors::CoreInitError::CertLoadingError,
     machines::ProtoCommand,
     protos::temporal::api::{
         common::v1::{Payloads, WorkflowExecution, WorkflowType},
@@ -19,7 +18,8 @@ use crate::{
         workflowservice::v1::{
             workflow_service_client::WorkflowServiceClient, PollActivityTaskQueueRequest,
             PollActivityTaskQueueResponse, PollWorkflowTaskQueueRequest,
-            PollWorkflowTaskQueueResponse, RespondActivityTaskCanceledRequest,
+            PollWorkflowTaskQueueResponse, RecordActivityTaskHeartbeatRequest,
+            RecordActivityTaskHeartbeatResponse, RespondActivityTaskCanceledRequest,
             RespondActivityTaskCanceledResponse, RespondActivityTaskCompletedRequest,
             RespondActivityTaskCompletedResponse, RespondActivityTaskFailedRequest,
             RespondActivityTaskFailedResponse, RespondWorkflowTaskCompletedRequest,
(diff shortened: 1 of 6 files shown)


src/pollers/mod.rs
```diff
@@ -60,21 +64,84 @@ pub struct ServerGatewayOptions {
...
+
+/// Configuration options for TLS
+#[derive(Clone, Debug)]
+pub struct TlsConfig {
+    /// A location to a file representing the root CA certificate used by the server. If this is
+    /// set, core will attempt to use TLS when connecting to the server. If `client_tls_config` is
+    /// also set, then core will attempt to use mTLS.
+    pub server_root_ca_cert: PathBuf,
+    /// The domain/dns name that the server's certificate uses
+    pub domain: String,
```
bergundy, line 80: AFAIU, you're not supposed to set it a custom domain in production, the domain should be derived from the server address.
In nodejs this is optional: https://gist.github.com/bergundy/32f181f056760928cf934d2d4f695e33#file-client-js-L14

https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#ga218bf55b665134a11baf07ada5980825

src/pollers/mod.rs
```diff
@@ -60,21 +64,84 @@ pub struct ServerGatewayOptions {
...
+    pub tls_cfg: Option<TlsConfig>,
+}
+
+/// Configuration options for TLS
+#[derive(Clone, Debug)]
+pub struct TlsConfig {
+    /// A location to a file representing the root CA certificate used by the server. If this is
+    /// set, core will attempt to use TLS when connecting to the server. If `client_tls_config` is
+    /// also set, then core will attempt to use mTLS.
+    pub server_root_ca_cert: PathBuf,
```
bergundy, line 78: We'll need to support reading certs from buffers instead of paths, we don't know how our users obtain their certificates (see example [here](https://temporalio.slack.com/archives/C01DKSMU94L/p1620409279021300?thread_ts=1620333054.020200&cid=C01DKSMU94L))

src/pollers/mod.rs
```diff
@@ -60,21 +64,84 @@ pub struct ServerGatewayOptions {
...
+    pub domain: String,
+    /// TLS info for the client, if using mTLS
+    pub client_tls_config: Option<ClientTlsConfig>,
+}
+
+/// If using mTLS, both the client cert and private key must be specified, this contains them.
+#[derive(Clone, Debug)]
+pub struct ClientTlsConfig {
+    /// The certificate for this client
+    pub client_cert: PathBuf,
```
bergundy, line 89: Same here, change to `Vec<u8>` instead.

tests/main.rs
```diff
@@ -88,4 +93,39 @@ mod integ_tests {
...
         .unwrap()
     }
+
+    // TODO: Currently ignored because starting up the docker image with TLS requires some hoop
+    //  jumping. We should upgrade CI to be able to do that but this was manually run against
+    //  https://github.com/temporalio/customization-samples/tree/master/tls/tls-simple
+    #[tokio::test]
+    #[ignore]
+    async fn tls_test() {
+        let sgo = ServerGatewayOptions {
```
bergundy, line 103: We should have test for the different TLS setups in that repo, (can wait for another PR though).

bergundy (changes requested):

src/pollers/mod.rs
Sushisource: Ah yeah, reading from a buffer makes perfect sense. What's the scenario where there's no root cert? A cert from a trusted issuer? Is that the case with our SaaS stuff?

Sushisource: Yes, to answer my own question.

vitarb (approved):

vitarb:
Let's land this and iterate on top of it addressing any remaining concerns.

PR #125, opened 2021-05-13
Sticky Queues interface changes

Sushisource:
This PR seeks to iteratively flesh out the design for any interface changes needed to support sticky task queues.

Checklist:
- [x]  Add configuration option to enable or disable sticky mode, with a cache size limit
- [x] Add some way (a new method) to evict specific workflows, or entire cache.
- [x] Decide on if we should change to using run ids rather than task tokens as the primary identifier returned by polls and provided in activations
    - [x] If so, make that change
- [x] Add unit tests for multiple worker scenarios

### DIFF
--- a/protos/local/activity_task.proto
+++ b/protos/local/activity_task.proto
@@ -13,13 +13,16 @@ import "google/protobuf/duration.proto";
 message ActivityTask {
     /// A unique identifier for this task
     bytes task_token = 1;
+    // The task queue that this activity task came from. Corresponds to the worker that should be
+    // tasked with handling it.
+    string task_queue = 2;
     /// The activity's ID
-    string activity_id = 2;
+    string activity_id = 3;
     oneof variant {
         /// Start activity execution.
-        Start start = 3;
+        Start start = 4;
         /// Attempt to cancel activity execution.
-        Cancel cancel = 4;
+        Cancel cancel = 5;
     }
 }
 
--- a/protos/local/activity_task.proto
+++ b/protos/local/activity_task.proto
@@ -13,16 +13,13 @@ import "google/protobuf/duration.proto";
 message ActivityTask {
     /// A unique identifier for this task
     bytes task_token = 1;
-    // The task queue that this activity task came from. Corresponds to the worker that should be
-    // tasked with handling it.
-    string task_queue = 2;
     /// The activity's ID
-    string activity_id = 3;
+    string activity_id = 2;
     oneof variant {
         /// Start activity execution.
-        Start start = 4;
+        Start start = 3;
         /// Attempt to cancel activity execution.
-        Cancel cancel = 5;
+        Cancel cancel = 4;
     }
 }
 
--- a/protos/local/workflow_activation.proto
+++ b/protos/local/workflow_activation.proto
@@ -14,17 +14,16 @@ import "google/protobuf/timestamp.proto";
 /// An instruction to the lang sdk to run some workflow code, whether for the first time or from
 /// a cached state.
 message WFActivation {
-    // A unique identifier for this task
-    bytes task_token = 1;
+    /// The id of the currently active run of the workflow. Also used as a cache key. There may
+    /// only ever be one active workflow task (and hence activation) of a run at one time.
+    string run_id = 1;
     // The task queue that this activation's workflow task came from. Corresponds to the worker
     // that should be tasked with handling it.
     string task_queue = 2;
     /// The current time as understood by the workflow, which is set by workflow task started events
     google.protobuf.Timestamp timestamp = 3;
-    /// The id of the currently active run of the workflow
-    string run_id = 4;
     /// The things to do upon activating the workflow
-    repeated WFActivationJob jobs = 5;
+    repeated WFActivationJob jobs = 4;
 }
 
 message WFActivationJob {
--- a/protos/local/workflow_activation.proto
+++ b/protos/local/workflow_activation.proto
@@ -17,13 +17,10 @@ message WFActivation {
     /// The id of the currently active run of the workflow. Also used as a cache key. There may
     /// only ever be one active workflow task (and hence activation) of a run at one time.
     string run_id = 1;
-    // The task queue that this activation's workflow task came from. Corresponds to the worker
-    // that should be tasked with handling it.
-    string task_queue = 2;
     /// The current time as understood by the workflow, which is set by workflow task started events
-    google.protobuf.Timestamp timestamp = 3;
+    google.protobuf.Timestamp timestamp = 2;
     /// The things to do upon activating the workflow
-    repeated WFActivationJob jobs = 4;
+    repeated WFActivationJob jobs = 3;
 }
 
 message WFActivationJob {
--- a/protos/local/workflow_activation.proto
+++ b/protos/local/workflow_activation.proto
@@ -16,12 +16,15 @@ import "google/protobuf/timestamp.proto";
 message WFActivation {
     // A unique identifier for this task
     bytes task_token = 1;
+    // The task queue that this activation's workflow task came from. Corresponds to the worker
+    // that should be tasked with handling it.
+    string task_queue = 2;
     /// The current time as understood by the workflow, which is set by workflow task started events
-    google.protobuf.Timestamp timestamp = 2;
+    google.protobuf.Timestamp timestamp = 3;
     /// The id of the currently active run of the workflow
-    string run_id = 3;
+    string run_id = 4;
     /// The things to do upon activating the workflow
-    repeated WFActivationJob jobs = 4;
+    repeated WFActivationJob jobs = 5;
 }
 
 message WFActivationJob {
--- a/protos/local/workflow_completion.proto
+++ b/protos/local/workflow_completion.proto
@@ -7,8 +7,8 @@ import "workflow_commands.proto";
 
 /// Result of a single workflow activation, reported from lang to core
 message WFActivationCompletion {
-    // The token from the workflow activation you are completing
-    bytes task_token = 1;
+    // The run id from the workflow activation you are completing
+    string run_id = 1;
     oneof status {
         Success successful = 2;
         Failure failed = 3;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,9 +62,9 @@ use crate::{
     workflow::WorkflowCachingPolicy,
     workflow_tasks::{FailedActivationOutcome, NewWfTaskOutcome, WorkflowTaskManager},
 };
-use dashmap::DashMap;
 use futures::{future::select_all, FutureExt, TryFutureExt};
 use std::{
+    collections::HashMap,
     convert::TryInto,
     future::Future,
     sync::{
@@ -74,7 +74,7 @@ use std::{
 };
 use tokio::sync::{
     mpsc::{unbounded_channel, UnboundedReceiver},
-    Mutex, Notify,
+    Mutex, Notify, RwLock,
 };
 
 /// This trait is the primary way by which language specific SDKs interact with the core SDK. It is
@@ -85,7 +85,7 @@ pub trait Core: Send + Sync {
     /// Register a worker with core. Workers poll on a specific task queue, and when calling core's
     /// poll functions, you must provide a task queue name. Returned activations will include that
     /// task queue name as well, so that lang may route them to the appropriate place.
-    fn register_worker(&self, config: WorkerConfig);
+    async fn register_worker(&self, config: WorkerConfig);
 
     /// Ask the core for some work, returning a [WfActivation]. It is then the language SDK's
     /// responsibility to call the appropriate workflow code with the provided inputs. Blocks
@@ -187,7 +187,7 @@ struct CoreSDK<WP> {
     /// Provides work in the form of responses the server would send from polling task Qs
     server_gateway: Arc<WP>,
     /// Maps task queue names to workers
-    workers: DashMap<String, Worker>,
+    workers: RwLock<HashMap<String, Worker>>,
     /// Workflow task management
     wft_manager: WorkflowTaskManager,
     /// Activity task management
@@ -211,10 +211,10 @@ impl<WP> Core for CoreSDK<WP>
 where
     WP: ServerGatewayApis + Send + Sync + 'static,
 {
-    fn register_worker(&self, config: WorkerConfig) {
+    async fn register_worker(&self, config: WorkerConfig) {
         let tq = config.task_queue.clone();
         let worker = Worker::new(config, self.server_gateway.clone());
-        self.workers.insert(tq, worker);
+        self.workers.write().await.insert(tq, worker);
     }
 
     #[instrument(skip(self))]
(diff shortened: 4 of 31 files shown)


Sushisource:
We had had some discussion in Slack about changing to use run ids for activations rather than task tokens. I think we should decide that first as it will impact the rest of the changes here. The (long) history is in this thread: https://temporaltechnologies.slack.com/archives/C01FG4BRQVB/p1620064199265000 -- at the conclusion it sounds like we were all in agreement but just wanted to confirm.

@mfateev Do you have any opinion here? I think when we talked you were already saying run ids made more sense.

src/workflow/mod.rs
```diff
@@ -175,3 +175,35 @@ impl WorkflowManager {
...
+    NonSticky,
+
+    /// Not a real mode, but good for imitating crashes. Evict workflows after *every* reply,
+    /// even if there are pending activations
+    #[cfg(test)]
+    AfterEveryReply,
+}
+
+/// The default number of workflows sticky mode will cache at maximum
+pub static STICKY_DEFAULT_MAX_WORKFLOWS: usize = 10_000;
```
bergundy, line 199: I wouldn't put a default in core, this is very much language specific.
It's OK if we use this default internally (for core testing purposes) only.

src/workflow/mod.rs
```diff
@@ -175,3 +175,35 @@ impl WorkflowManager {
         Ok(())
     }
 }
+
+/// Determines when workflows are kept in the cache or evicted
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum WorkflowCachingPolicy {
```
bergundy, line 181: Seems like this would be handy for debugging purposes but we probably don't want to expose any of this to the user other than `max_cached_workflows`.

src/workflow/mod.rs
Sushisource: Sure, makes sense.

src/workflow/mod.rs
Sushisource: Yeah, Max mentioned we don't want to expose the non sticky mode. So, I'll have this not be an enum. To be clear, `AfterEveryReply` would never show up publicly because it's compile-time guarded to be test only. It makes this slightly annoying b/c having this be an enum is nice ergonomically inside core, but it ends up only having one variant publicly, but oh well.

Sushisource:
Options for lang sending evictions:

1. Explicit eviction API. Pass run id as argument.
2. Add a "evict" option to the status enum inside `WfActivationCompletion`
3. Add an "evict" `WorkflowCommand`, allowing you to include it as part of a successful activation.

1 is simple, explicit and obvious. 2 I think mixes concerns in a way that is icky and provides little benefit. 3 is more reasonable as it allows you to pretty cleanly actually complete an activation, and then say you want to evict right after that, which could be a desirable thing to do. Returning a failed activation already evicts, so that's covered.

I would be OK with either 1 or 3. 

The Go SDK, for example, also exposes a "blow up the whole cache" function. Arguably that functionality is easier accomplished by just restarting the worker. And, in fact, the docstring `// PurgeStickyWorkflowCache resets the sticky workflow cache. This must be called only when all workers are stopped.` implies it might be better handled by just doing that implicitly once there are no live workers. So, not sure we need this. If we do, it favors solution 1 as we would need some top-level API for it anyway, and at that point might as well allow passing `run_id | entire_cache` as the argument.

@bergundy @mfateev @vitarb Thoughts?

bergundy:
I think we'd need to go with 1 because otherwise eviction can only happen as a part of completion.
With 1 what I imagine that the eviction request is sent to core that will eventually issue an eviction activation.
The shouldn't be an option for lang to just drop a workflow without receiving an instruction from core.
As for going with run id as the identifier, I'm good with that as long as the assumption we made in the slack thread about not starting a new workflow instance if another one exists for the same run id.

Sushisource:
Spoke with Max, other thing we'll want to do as part of these changes is introduce a proper Worker abstraction, with support for registering multiple workers per core instance. Some of the stuff in the gateway options, and some of the stuff in the core options, will move into a new worker options struct, and then we'll need to support registering workers with core. Likely this also means returning which worker some activation is associated with as part of poll responses.

bergundy:
While we're at it are we going to support multiple task queues in a single core instance?

Sushisource:
> While we're at it are we going to support multiple task queues in a single core instance?

That would take the form of registering multiple workers, as any given worker is only supposed to poll from one task queue (with the minor exception of sticky queues, where the worker polls from it's specific queue as well as the shared one)

Sushisource:
@bergundy I've updated the PR to change polling to accept a task queue name, and activations have the associated task queue (/ worker name) included in them.

Pending a discussion with Samar I have set for tomorrow morning, we'll either be able to do sticky queues that are host+queue specific or we'll have to stay with sticky queues that are per-host only. See https://www.notion.so/temporalio/Core-Sticky-Queues-Design-and-docs-c443300a424d4e158e03ab51edb0b668

If we can do the former, then things will just work pretty naturally. Workers poll the queue to which they are assigned, and they get tasks only from that queue, either sticky or not.

If we can't, the sticky queue has to get polled somehow. The options would be to auto-register a "sticky" worker which polls it, or to have it be polled in the background, potentially handing it's tasks to whatever worker happens to be polling next. I'd say the former is probably better if we have can't do the more-specific sticky queues. Either way, lang has to deal with activations for that queue by routing them to a (deterministically chosen) worker than can handle the workflow type, which is the crappy part we can hopefully avoid.

Sushisource:
@bergundy Interface for wf activations is changed to use run id now too

vitarb (approved):

protos/local/workflow_completion.proto
```diff
@@ -7,8 +7,8 @@ import "workflow_commands.proto";
 
 /// Result of a single workflow activation, reported from lang to core
 message WFActivationCompletion {
-    // The token from the workflow activation you are completing
-    bytes task_token = 1;
+    // The run id from the workflow activation you are completing
+    string run_id = 1;
```
bergundy, line 11: We should probably add a unique identifier per activation for debug purposes.

src/lib.rs
```diff
@@ -57,47 +54,53 @@ use crate::{
...
+    /// task queue name as well, so that lang may route them to the appropriate place.
+    fn register_worker(&self, config: WorkerConfig);
+
     /// Ask the core for some work, returning a [WfActivation]. It is then the language SDK's
     /// responsibility to call the appropriate workflow code with the provided inputs. Blocks
     /// indefinitely until such work is available or [Core::shutdown] is called.
     ///
     /// TODO: Examples
-    async fn poll_workflow_task(&self) -> Result<WfActivation, PollWfError>;
+    async fn poll_workflow_task(&self, task_queue: &str) -> Result<WfActivation, PollWfError>;
```
bergundy, line 95: Something still feels a bit off with this API.
We register a Worker with a specific task queue and then later poll for tasks by `task_queue`.
Maybe `register_worker` should return an object with poll methods?

src/lib.rs
```diff
@@ -132,6 +135,12 @@ pub trait Core: Send + Sync {
     /// configured heartbeat options.
     fn record_activity_heartbeat(&self, details: ActivityHeartbeat);
 
+    /// Request that a workflow be evicted by its run id. This will generate a workflow activation
+    /// with the eviction job inside it to be eventually returned by [Core::poll_workflow_task]. If
+    /// the workflow had any existing outstanding activations, such activations are invalidated and
+    /// subsequent completions of them will do nothing and log a warning.
+    fn request_eviction(&self, run_id: &str);
```
bergundy, line 142:
```suggestion
    fn request_workflow_eviction(&self, run_id: &str);
```

bergundy (commented):
I went over the interface part.
Had some comments but nothing too major, this is definitely the right direction.

src/lib.rs
```diff
@@ -57,47 +54,53 @@ use crate::{
...
 
 /// This trait is the primary way by which language specific SDKs interact with the core SDK. It is
 /// expected that only one instance of an implementation will exist for the lifetime of the
 /// worker(s) using it.
 #[async_trait::async_trait]
 pub trait Core: Send + Sync {
+    /// Register a worker with core. Workers poll on a specific task queue, and when calling core's
+    /// poll functions, you must provide a task queue name. Returned activations will include that
+    /// task queue name as well, so that lang may route them to the appropriate place.
+    fn register_worker(&self, config: WorkerConfig);
```
bergundy, line 88: What are the implications of registering a worker?
Can workers be deregistered?

src/lib.rs
Sushisource: Will do

src/lib.rs
Sushisource: So we talked about this before and I thought we decided against it, although I forget exactly why. From an encapsulation perspective I actually really like it. My only concern is that it maybe requires pretty tight integration with Rust that might not work as well in other languages., and it substantially expands the API surface (or, at least shifts things around quite a bit).

src/lib.rs
Sushisource: As it stands they cannot. Not sure I see a big reason to do that unless someone comes up with a use case. I'd just start a new process.

Implications in what sense? Nothing really happens here except allowing you to poll on the task queue for that worker, once you start polling, then pollers are created etc. If you have suggestions on what to add to the docstring lemme know

protos/local/activity_task.proto
```diff
@@ -13,13 +13,16 @@ import "google/protobuf/duration.proto";
 message ActivityTask {
     /// A unique identifier for this task
     bytes task_token = 1;
+    // The task queue that this activity task came from. Corresponds to the worker that should be
+    // tasked with handling it.
+    string task_queue = 2;
```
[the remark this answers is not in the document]
Sushisource: Hmm, I suppose that's fair. There kinda is, which is when it's from the sticky queue, but I realized after I posted this PR that really we need a flag for that (in workflow activations only). Probably activity task can indeed have it removed.

protos/local/workflow_activation.proto
```diff
@@ -14,12 +14,14 @@ import "google/protobuf/timestamp.proto";
...
 /// a cached state.
 message WFActivation {
-    // A unique identifier for this task
-    bytes task_token = 1;
+    /// The id of the currently active run of the workflow. Also used as a cache key. There may
+    /// only ever be one active workflow task (and hence activation) of a run at one time.
+    string run_id = 1;
+    // The task queue that this activation's workflow task came from. Corresponds to the worker
+    // that should be tasked with handling it.
+    string task_queue = 2;
```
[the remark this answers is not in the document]
Sushisource: Will change this just to be a flag like "should_be_cached" to indicate it came from a sticky queue, will remove explicit task queue name.

protos/local/workflow_completion.proto
Sushisource: Let's save this for tracing/logging interface changes as it probably overlaps a lot.

src/lib.rs
Sushisource: Other potential benefit here is it can reduce lock contention in Core, but, that's not exactly a big issue.

protos/local/activity_task.proto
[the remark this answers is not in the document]
Sushisource: OK so I now remember why this is, it's pending activations (and, less frequently, buffered poll responses).

But, they should probably not be returned for queues you didn't poll for, if we're passing queue in the argument. This whole thing goes away if we only return queue in responses and there's just one queue-agnostic poll (see other comment).

src/lib.rs
Sushisource: Talked with @mfateev and he mentioned a preference for poll not taking queue at all and *only* including it in the response. Problem with that is core can "get ahead" of lang (ex: if we're polling queues `a`, `b`, and `c`, if lang polls we might get responses for all three very quickly and need to buffer them, and lang may not poll again for some time for whatever reason). This is pretty desirable to avoid during unit testing where right now a `poll_workflow_task` call always equals exactly one request to the server if nothing is pending, which is good for being able to count expected number of server calls exactly.

Biggest advantages to doing a queue-agnostic poll are that we can always prioritize returning pending activations (instead of not returning them if they don't match the queue you asked for), and no need for a `Worker` type handle / stateful API.

I think I can figure out ways to deal with the unit testing nuisance if only including queue in response sounds good to you @bergundy

Sushisource: So, I've gone ahead and pre-emptively done it this way, which we can reverse if there are strong feelings. It makes a lot of sense to me and my concerns about the unit testing being a problem turned out to not be real at all. This keeps closer to our existing poll-respond loop and overall seems to fit pretty nicely.

protos/local/workflow_activation.proto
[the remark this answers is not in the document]
Sushisource: Consolidate to https://github.com/temporalio/sdk-core/pull/125#discussion_r638163750

protos/local/activity_task.proto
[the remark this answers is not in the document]
Sushisource: Consolidate to https://github.com/temporalio/sdk-core/pull/125#discussion_r638163750

Sushisource:
There's still a little cleanup to do here, but just wanted to get the no-task-queue-name-passing version up. The last open question is if we need an `is_cached` flag for workflow activations. I think actually probably not, considering lang always needs to keep the workflow function suspended while going back and forth with core anyway. Essentially any activation that isn't just a `StartWorkflow` job must be "cached" already.

Does that make sense to you @bergundy ?

src/lib.rs
bergundy: How would this API work on the lang side?
Today a lang worker polls core "directly" which means the core-lang bridge simply passes on poll requests from lang to core.
With this new change a lot more work is needed in the bridge for storing the lang callbacks and buffering poll responses since we need lang to register a new callback and a poll response to be available per queue before a response can be dispatched to lang.

src/lib.rs
bergundy: So there's no way to tell core to stop polling on a specific task queue?
I generally prefer we have a way to clean up every resource that we create.

bergundy:
> There's still a little cleanup to do here, but just wanted to get the no-task-queue-name-passing version up. The last open question is if we need an `is_cached` flag for workflow activations. I think actually probably not, considering lang always needs to keep the workflow function suspended while going back and forth with core anyway. Essentially any activation that isn't just a `StartWorkflow` job must be "cached" already.
> 
> Does that make sense to you @bergundy ?

Yup

src/lib.rs
bergundy: Also take note that this change implies that polling is never initiated by lang which invalidates your assumption that traces always start in lang.

Sushisource: Yeah, the concern about routing makes sense to me. Polling is still initiated by lang, it just fans out to available queues. There are pretty good reasons for either design, but, given the whole point of core is to make lang small, keeping that routing stuff to a minimum does seem a compelling reason. @mfateev ?

src/lib.rs
Sushisource: Sure, makes sense, we can add a `remove_worker` method

Sushisource: Now you can stop polling on the queue by simply not polling on it. In the name of keeping things simple I'm going to hold off on de-registering them for now given it seems super, super unlikely to come up to me. If we get a compelling non-theoretical use case down the line we can always add the feature.

src/lib.rs
Sushisource: We went with passing the queue name

PR #126, opened 2021-06-01
Add is_replaying attribute to WFActivation

bergundy:
## Why?
This is required in order for lang SDKs to be able to dedupe log messages and metrics.
See https://github.com/temporalio/proposals/blob/master/node/logging-and-metrics-for-user-code.md


vitarb (approved):

bergundy:
I'm merging this despite the incorrectness of the value.
Let's prioritize fixing this since a lot of functionality relies on the `is_replaying` flag.
One thing I'm not sure of is if this flag should be set per activation or per activation job.
@Sushisource @mfateev @vitarb WDYT?

PR #128, opened 2021-06-06
Pass history event header fields to StartWorkflow activation job

bergundy:
## Why?
Used by interceptors to pass metadata from the client

## Checklist

2. How was this tested:
Tested with the Node SDK


Sushisource (dismissed):

Sushisource (commented):
Actually, shouldn't we just put this on the whole activation? These headers could change between a start workflow and another activation.

bergundy:
The headers are attached to a specific workflow history event which correlates to an activation job in the core-lang terminology.
As far as ease of use, it's easier for the headers to be attached to a single job and not the whole activation.

Sushisource (approved):

PR #129, opened 2021-06-07
Sticky queues implementation

Sushisource:
## What was changed
Sticky queues are used now when max cached workflows is nonzero

1. How was this tested:
All integ tests were already using a sort of fake sticky mode. Now they use real sticky mode, and some non-sticky tests are added.

### DIFF
--- a/src/core_tests/workflow_tasks.rs
+++ b/src/core_tests/workflow_tasks.rs
@@ -1,3 +1,4 @@
+use crate::machines::test_help::mock_core_with_opts_no_workers;
 use crate::{
     job_assert,
     machines::test_help::{
@@ -872,7 +873,7 @@ async fn max_concurrent_wft_respected() {
         .expect_complete_workflow_task()
         .returning(|_, _, _| Ok(RespondWorkflowTaskCompletedResponse::default()));
 
-    let core = mock_core(mock_gateway);
+    let core = mock_core_with_opts_no_workers(mock_gateway, CoreInitOptionsBuilder::default());
     core.register_worker(
         WorkerConfigBuilder::default()
             .task_queue(TEST_Q)
@@ -880,7 +881,8 @@ async fn max_concurrent_wft_respected() {
             .build()
             .unwrap(),
     )
-    .await;
+    .await
+    .unwrap();
 
     // Poll twice in a row before completing -- we should be at limit
     let r1 = core.poll_workflow_task(TEST_Q).await.unwrap();
@@ -1152,6 +1154,8 @@ async fn complete_after_eviction() {
 
 #[tokio::test]
 async fn sends_appropriate_sticky_task_queue_responses() {
+    // This test verifies that when completions are sent with sticky queues enabled, that they
+    // include the information that tells the server to enqueue the next task on a sticky queue.
     let wfid = "fake_wf_id";
     let t = canned_histories::single_timer("fake_timer");
     let mut mock = MockServerGatewayApis::new();
--- a/src/core_tests/workflow_tasks.rs
+++ b/src/core_tests/workflow_tasks.rs
@@ -1,10 +1,9 @@
-use crate::machines::test_help::{fake_sg_opts, mock_core_with_opts};
 use crate::{
     job_assert,
     machines::test_help::{
         build_fake_core, build_multihist_mock_sg, fake_core_from_mock_sg, gen_assert_and_fail,
-        gen_assert_and_reply, hist_to_poll_resp, mock_core, poll_and_reply, single_hist_mock_sg,
-        FakeCore, FakeWfResponses, TestHistoryBuilder, TEST_Q,
+        gen_assert_and_reply, hist_to_poll_resp, mock_core, mock_core_with_opts, poll_and_reply,
+        single_hist_mock_sg, FakeCore, FakeWfResponses, TestHistoryBuilder, TEST_Q,
     },
     pollers::MockServerGatewayApis,
     protos::{
@@ -27,7 +26,7 @@ use crate::{
     },
     test_help::canned_histories,
     workflow::WorkflowCachingPolicy::{self, AfterEveryReply, NonSticky},
-    Core, CoreInitOptionsBuilder, CoreSDK, WfActivationCompletion, WorkerConfigBuilder,
+    Core, CoreInitOptionsBuilder, WfActivationCompletion, WorkerConfigBuilder,
 };
 use rstest::{fixture, rstest};
 use std::{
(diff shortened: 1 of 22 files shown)


src/worker/mod.rs
```diff
@@ -30,9 +36,15 @@ pub struct WorkerConfig {
     #[builder(default = "100")]
     pub max_outstanding_activities: usize,
     /// Maximum number of concurrent poll workflow task requests we will perform at a time on this
-    /// worker's task queue
+    /// worker's task queue. See also [nonsticky_to_sticky_poll_ratio]. A zero value will be treated
+    /// as if set to `1`.
```
bergundy, line 40: The minimum in case sticky is on is 2 according to the implementation below.
We should just return an error if passed less than 2 concurrent polls with sticky turned on and 1 with sticky off.

bergundy (commented):
Nice!
My biggest concern ATM is the eviction flow as I've pointed out in one of the comments.
One other thing that I think we may have missed is that lang exposes a `Worker.shutdown()` method while core only has `Core.shutdown()`.
We should talk about what it means to shutdown a single worker vs the entire core.

src/worker/mod.rs
Sushisource: Yeah I'm fine with just returning an error. Will do that.

src/lib.rs
```diff
@@ -77,13 +77,21 @@ use tokio::sync::{
...
+
 /// This trait is the primary way by which language specific SDKs interact with the core SDK. It is
 /// expected that only one instance of an implementation will exist for the lifetime of the
 /// worker(s) using it.
 #[async_trait::async_trait]
 pub trait Core: Send + Sync {
     /// Register a worker with core. Workers poll on a specific task queue, and when calling core's
-    /// poll functions, you must provide a task queue name.
+    /// poll functions, you must provide a task queue name. If there was already a worker registered
+    /// with the same task queue name, it will be shut down and a new one will be created.
```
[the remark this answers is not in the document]
Sushisource: I'm happy to just throw an error too. I didn't just to avoid needing to have this return a Result, but, nice to just keep it simple here.

src/core_tests/workflow_tasks.rs
```diff
@@ -1148,5 +1146,35 @@ async fn complete_after_eviction() {
...
+        .unwrap();
+    core.inner.shutdown().await;
+}
+
+#[tokio::test]
+async fn sends_appropriate_sticky_task_queue_responses() {
+    let wfid = "fake_wf_id";
+    let t = canned_histories::single_timer("fake_timer");
+    let mut mock = MockServerGatewayApis::new();
+    mock.expect_complete_workflow_task()
```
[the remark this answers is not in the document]
Sushisource: Added

src/core_tests/workflow_tasks.rs
```diff
@@ -1134,8 +1131,9 @@ async fn wft_timeout_repro(hist_batches: &[usize]) {
 async fn complete_after_eviction() {
     let wfid = "fake_wf_id";
     let t = canned_histories::single_timer("fake_timer");
-    let mut mock = single_hist_mock_sg(wfid, t, &[2]);
-    mock.sg.expect_complete_workflow_task().times(0);
+    let mut mock = MockServerGatewayApis::new();
+    mock.expect_complete_workflow_task().times(0);
```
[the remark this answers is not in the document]
Sushisource: So, per our discussion on the call, we're agreed an ack makes sense here. It's a pretty big change though and I dunno if I want to add it to this PR - it's related but not strictly. We also talked about bringing back the handle-per-workflow concept.

Perhaps we can do that and this together? They interact closely. Thoughts @bergundy ?

Sushisource:
@bergundy Added a worker-specific shutdown method

src/worker/mod.rs
```diff
@@ -30,9 +36,17 @@ pub struct WorkerConfig {
...
+    /// worker's task queue. See also [WorkerConfig::nonsticky_to_sticky_poll_ratio]. Must be at
+    /// least 1.
     #[builder(default = "5")]
     pub max_concurrent_wft_polls: usize,
+    /// [WorkerConfig::max_concurrent_wft_polls] * this number = the number of max pollers that will
+    /// be allowed for the nonsticky queue when sticky tasks are enabled. If both defaults are used,
+    /// the sticky queue will allow 4 max pollers while the nonsticky queue will allow one. The
+    /// minimum for either poller is 1, so if `max_concurrent_wft_polls` is 1 and sticky queues are
+    /// enabled, there will be 2 concurrent polls.
+    #[builder(default = "0.2")]
```
vitarb, line 48: Is there any reason why we have 5:1 ratio here? Since by default all workflow tasks are sticky such default would lead to sub-optimal performance when sticky can not keep up and regular workers are idling. Should we remove this setting entirely and only control sticky cache size assuming 1:1 ratio between sticky and non sticky pollers?

tests/integ_tests/workflow_tests/timers.rs
```diff
@@ -1,30 +1,51 @@
...
+        timer_id: "super_timer_id".to_string(),
+        start_to_fire_timeout: Some(Duration::from_secs(1).into()),
+    };
+    command_sink.timer(timer).await;
+    command_sink.complete_workflow_execution();
+}
+
 #[tokio::test(flavor = "multi_thread")]
-async fn timer_workflow_new_way() {
+async fn timer_workflow_not_sticky() {
```
vitarb, line 20: It would be nice to add a test that polls on sticky, fails or times out and then polls on non-sticky and is able to finish workflow using whatever was in the cache.

vitarb (approved):
LGTM with two nits

src/worker/mod.rs
Sushisource: Talked with Max and we agreed it makes sense to prioritize sticky since that's where most tasks will be for a sticky worker. There is no situation where "regular workers" idle, because task queues sticky per-queue now, the same worker is always servicing both sticky and non sticky tasks. So it's purely a matter of which type of polling to prioritize.

tests/integ_tests/workflow_tests/timers.rs
Sushisource: That's a good idea will add

Sushisource: Turns out this was already covered really, but, added a more specific test for it.

src/core_tests/workflow_tasks.rs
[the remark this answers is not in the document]
bergundy: I don't think we should tie the (potential) redesign to this, that being said, we can wait for the next PR to implement these suggestions.
Let's track and prioritize and figure out how it fits with the greater plans for the pollers.

Issue #127, opened 2021-06-03
Fix WFActivation.is_replaying incorrectly reflecting WF state


bergundy:
I'm merging this despite the incorrectness of the value.
Let's prioritize fixing this since a lot of functionality relies on the `is_replaying` flag.
One thing I'm not sure of is if this flag should be set per activation or per activation job.
@Sushisource @mfateev @vitarb WDYT?

_Originally posted by @bergundy in https://github.com/temporalio/sdk-core/issues/126#issuecomment-853592693_

Sushisource:
For reference: Lang needs this info so it can choose to not emit logs during replay (for example)

PR #130, opened 2021-06-11
Fix workflow activation is_replaying flag

Sushisource:
## What was changed
The replay flag was set to some nonsense copied from Java when I first started working on core. This makes it reflect what you would expect it to be.

## Why?
Needed for lang

## Checklist

- [x] Need to address some TODOs I added, but this should fix the bug. @bergundy Give it a shot.
- [x] Add a few more specific tests

1. Closes https://github.com/temporalio/sdk-core/issues/127

2. How was this tested:
Added tests

### DIFF
--- a/src/core_tests/replay_flag.rs
+++ b/src/core_tests/replay_flag.rs
@@ -39,6 +39,9 @@ fn fire_happy_hist(num_timers: usize) -> WorkflowMachines {
 #[case::one_timer(fire_happy_hist(1), 1)]
 #[case::five_timers(fire_happy_hist(5), 5)]
 fn replay_flag_is_correct(#[case] setup: WorkflowMachines, #[case] num_timers: usize) {
+    // Verify replay flag is correct by constructing a workflow manager that already has a complete
+    // history fed into it. The first (few, depending on test a params) activation(s) will be under
+    // replay while the last should not
     let mut wfm = WorkflowManager::new_from_machines(setup);
 
     for _ in 1..=num_timers {
--- a/src/core_tests/replay_flag.rs
+++ b/src/core_tests/replay_flag.rs
@@ -1,5 +1,5 @@
 use crate::{
-    machines::{test_help::TestHistoryBuilder, WorkflowMachines},
+    machines::WorkflowMachines,
     protos::coresdk::workflow_commands::StartTimer,
     protos::temporal::api::enums::v1::CommandType,
     test_help::canned_histories,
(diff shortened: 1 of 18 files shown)


src/core_tests/replay_flag.rs
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+    let t = canned_histories::single_timer("timer1");
+    let state_machines = WorkflowMachines::new(
+        "wfid".to_string(),
+        "runid".to_string(),
+        Box::new(twd).into(),
+    );
+
+    assert_eq!(2, t.as_history().get_workflow_task_count(None).unwrap());
```
bergundy, line 31: nit: Looks like here you've swapped `actual` and `expected`, is this the right parameter order?

src/core_tests/replay_flag.rs
```diff
@@ -0,0 +1,68 @@
...
+    (t, state_machines)
+}
+
+#[rstest]
+fn replay_flag_is_correct(fire_happy_hist: (TestHistoryBuilder, WorkflowMachines)) {
```
bergundy, line 36: I get what this test does but not why it works.
Maybe add a comment about the history and the state of the state machine in this test?

src/core_tests/replay_flag.rs
```diff
@@ -0,0 +1,68 @@
...
+        commands[0].command_type,
+        CommandType::CompleteWorkflowExecution as i32
+    );
+}
+
+#[rstest]
+fn replay_flag_is_correct_partial_history(fire_happy_hist: (TestHistoryBuilder, WorkflowMachines)) {
+    let (t, state_machines) = fire_happy_hist;
+    let mut wfm =
+        WorkflowManager::new_from_machines(t.as_partial_history(3), state_machines).unwrap();
```
bergundy, line 61: A little context here would really improve readability.

src/core_tests/replay_flag.rs
Sushisource: There is no expected order in Rust asserts, which I think is a great choice :laughing:

src/core_tests/replay_flag.rs
Sushisource: added

src/workflow/history_update.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// Given a workflow task started id, return all events starting at that number (inclusive) to
+    /// the next WFT started event (inclusive). If there is no subsequent WFT started event,
+    /// remaining history is returned.
+    ///
+    /// Events are *consumed* by this process, to keep things efficient in workflow machines
+    pub fn take_next_wft_sequence(&mut self, from_wft_started_id: i64) -> Vec<HistoryEvent> {
+        let mut events_to_next_wft_started = vec![];
+        let mut saw_next_wft = false;
+
+        let mut should_pop = |e: &HistoryEvent| {
```
vitarb, line 53: nit. Could we add a comment explaining what this flag is as well as if branches below?

src/core_tests/replay_flag.rs
```diff
@@ -0,0 +1,84 @@
...
+        "runid".to_string(),
+        t.as_history_update(),
+        Box::new(twd).into(),
+    )
+}
+
+#[rstest]
+#[case::one_timer(fire_happy_hist(1), 1)]
+#[case::five_timers(fire_happy_hist(5), 5)]
+fn replay_flag_is_correct(#[case] setup: WorkflowMachines, #[case] num_timers: usize) {
```
vitarb, line 41: nit. `is_correct` is slightly misleading should we follow given-when-then naming for tests, or at least when-then?

vitarb (approved):
LGTM, couple minor comments.

src/core_tests/replay_flag.rs
Sushisource: Not a huge fan of those naming schemes, because this one would be something like `given_a_history_with_multiple_workflow_tasks_when_polling_and_completing_then_replay_is_true_until_the_last_wft_started`

Which makes test result output pretty awful sometimes. But I'm happy to just add that as a comment at the top of the test. If we want to do a policy of that I'd be happy to. I know I'd like to see them in @bergundy 's node tests and he'd probably like to see them in my tests :)

src/workflow/history_update.rs
Sushisource: Added

PR #132, opened 2021-06-17
Fix worker shutdown

Sushisource:
## What was changed
Worker shutdown would deadlock if you were still polling while trying to shut down. This fixes that, and also fixes a problem where you would not have been able to send completes for pending activations even though that should've worked.

## Why?
Fix

## Checklist

1. Closes 

2. How was this tested:
New tests

3. Any docs updates needed?


src/core_tests/workers.rs
```diff
@@ -138,3 +145,114 @@ async fn pending_activities_only_returned_for_their_queue() {
...
+#[tokio::test]
+async fn worker_shutdown_during_poll_doesnt_deadlock() {
+    let mut mock_gateway = MockManualGateway::new();
+    mock_gateway
+        .expect_poll_workflow_task()
+        .returning(move |_| {
+            // Slow poll response
+            async move {
+                let t = canned_histories::single_timer("fake_timer");
+                sleep(Duration::from_secs(1)).await;
```
bergundy, line 238: I usually would rather avoid sleeping in tests, they tend to break when CI server is under load and make you wait unnecessarily.
Would be much better to use some sort of signaling.

src/lib.rs
```diff
@@ -224,19 +224,37 @@ enum WfActivationUpdate {
...
 }
 
+enum WorkerStatus {
+    Live(Worker),
+    Shutdown,
+}
+
+impl WorkerStatus {
+    #[cfg(test)]
+    pub fn unwrap(&self) -> &Worker {
```
bergundy, line 234: Noice!

src/worker/mod.rs
```diff
@@ -141,7 +141,21 @@ impl Worker {
...
+        if let Some(sq) = self.sticky_queue.as_ref() {
+            sq.poll_buffer.notify_shutdown();
+        }
+        if let Some(b) = self.at_task_poll_buffer.as_ref() {
+            b.notify_shutdown();
+        }
+    }
+
+    /// Resolves when shutdown of the worker is complete
+    pub(crate) async fn await_shutdown(self) {
```
bergundy, line 158: Maybe rename this to `shutdown_complete` since it doesn't actually await?
```rust
worker.shutdown_complete().await
```
as opposed to
```rust
worker.await_shutdown().await
```

bergundy (approved):
Nice, did not have any blocking comments.

bergundy:
@Sushisource confirmed the fix with my branch in the Node SDK.

bergundy:
Managed to create another test in the Node SDK that fails due to Worker not being shutdown in time.
Could be another deadlock.
Test hangs here: https://github.com/temporalio/sdk-node/blob/d21167ecf3377a80f2f3b20f5791b4599bd09e94/packages/test/src/test-core.ts#L22

fsm/rustfsm_procmacro/src/lib.rs
```diff
@@ -191,7 +191,7 @@ struct StateMachineDefinition {
 impl StateMachineDefinition {
     fn is_final_state(&self, state: &Ident) -> bool {
         // If no transitions go from this state, it's a final state.
-        self.transitions.iter().find(|t| t.from == *state).is_none()
+        !self.transitions.iter().any(|t| t.from == *state)
```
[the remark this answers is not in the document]
Sushisource: Huh, there is. Strangely clippy recommends this one (in a new lint)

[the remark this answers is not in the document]
Sushisource: But, actually, I kinda see why. At least in terms of the comment / logic here, `! any` feels more clear than `all` semantically.

vitarb (approved):

PR #133, opened 2021-06-22
Remove an extra map that is not needed in the WFT manager

Sushisource:
## What was changed
Removed an unnecessary redundancy

## Why?
Code cleanup I discovered while getting started on queries

## Checklist

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?


vitarb (approved):
LGTM

PR #135, opened 2021-06-22
Implement Queries

Sushisource:
## What was changed
Implement queries

## Why?
People like queries.

## Checklist

1. Closes 

2. How was this tested:
New tests

3. Any docs updates needed?

### DIFF
--- a/protos/local/workflow_commands.proto
+++ b/protos/local/workflow_commands.proto
@@ -47,7 +47,6 @@ message ScheduleActivity {
     string activity_type = 2;
     string namespace = 3;
     // The name of the task queue to place this activity request in
-    //  TODO: Do we care about sticky/not sticky?
     string task_queue = 4;
     map<string, common.Payload> header_fields = 5;
     /// Arguments/input to the activity. Called "input" upstream.
@@ -62,7 +61,6 @@ message ScheduleActivity {
     google.protobuf.Duration schedule_to_start_timeout = 8;
     /// Maximum time an activity is allowed to execute after a pick up by a worker. This timeout is
     /// always retryable. Either this or schedule_to_close_timeout must be specified.
-    /// TODO: Is this really either or can you do both? Make oneof if mutually exclusive
     google.protobuf.Duration start_to_close_timeout = 9;
     /// Maximum time allowed between successful worker heartbeats.
     google.protobuf.Duration heartbeat_timeout = 10;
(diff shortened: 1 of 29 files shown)


src/lib.rs
```diff
@@ -548,14 +551,14 @@ impl<SG: ServerGatewayApis + Send + Sync + 'static> CoreSDK<SG> {
         )
     }
 
-    /// Apply validated WFTs from the server. Returns an activation if one should be issued to
-    /// lang, or returns `None` in which case the polling loop should be restarted.
+    /// Apply validated poll responses from the server. Returns an activation if one should be
+    /// issued to lang, or returns `None` in which case the polling loop should be restarted.
```
bergundy, line 555: Maybe elaborate on when this returns `None`?

tests/integ_tests/queries_tests.rs
```diff
@@ -0,0 +1,101 @@
...
+                query_type: "myquery".to_string(),
+                query_args: Some(b"hi".into()),
+            },
+        )
+        .await
+        .unwrap()
+    });
+    let workflow_completions_future = async {
+        // Give query a beat to get going
+        tokio::time::sleep(Duration::from_millis(400)).await;
```
bergundy, line 52: I think you know what I have to say about this :)

bergundy (commented):
Awesome that we finally got to queries, go @Sushisource!

tests/integ_tests/queries_tests.rs
Sushisource: Heh, yeah. Unfortunately there is just no reasonable way to hook into "the query request has been issued" because the query call stays open until fulfilled. I would have to somehow instrument all the way down inside of the http handler and it'd just be a monster mess

protos/local/workflow_commands.proto
```diff
@@ -91,9 +89,11 @@ message RequestCancelActivity {
 }
 
 message QueryResult {
+    /// Corresponds to the id provided in the activation job
+    string query_id = 1;
     oneof variant {
-        QuerySuccess succeeded = 1;
-        string failed_with_message = 2;
+        QuerySuccess succeeded = 2;
+        string failed_with_message = 3;
```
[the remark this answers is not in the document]
Sushisource: I don't think so. The typical way these would be failed would actually be by failing the workflow activation. However, in the event that there is an activation with normal jobs and some query jobs, lang may succeed at applying the normal jobs but fail one of the queries. That is where I would expect it to fill this.

But, maybe we should replace w/ `UserCodeFailure` ?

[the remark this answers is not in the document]
bergundy: I think we should preserve the ability to fail individual queries as opposed to the entire activation.
Just to double check though is this what we do in other SDKs?

vitarb (approved):

PR #139, opened 2021-07-01
Mock at the poller level rather than server client

Sushisource:
## What was changed
Simplifies mocking concerns by mocking poller reponses rather than individual client calls. Will allow refactoring to concurrent pollers that are not tied 1:1 to poll calls.

## Why?
Tech debt

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


CODEOWNERS
```diff
@@ -1,3 +1,3 @@
 # Primary owners
 
-* @vitarb @Sushisource 
+* @vitarb @Sushisource @bergundy
```
vitarb, line 3: nit. This should be a separate PR IMO.

vitarb (approved):

PR #140, opened 2021-07-02
Continue as new

Sushisource:
## What was changed
Implement continue as new machine and command

## Why?
Long running workflows

## Checklist
- [x] Integ test
- [x] Unit test

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/protos/local/workflow_commands.proto
+++ b/protos/local/workflow_commands.proto
@@ -20,6 +20,7 @@ message WorkflowCommand {
         CancelTimer cancel_timer = 5;
         CompleteWorkflowExecution complete_workflow_execution = 6;
         FailWorkflowExecution fail_workflow_execution = 7;
+        ContinueAsNewWorkflowExecution continue_as_new_workflow_execution = 8;
 
         // To be added as/if needed:
         //  RequestCancelActivityTask request_cancel_activity_task_command_attributes = 6;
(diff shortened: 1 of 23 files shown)


protos/local/workflow_commands.proto
```diff
@@ -109,4 +110,27 @@ message CompleteWorkflowExecution {
 /// Issued when the workflow errors out
 message FailWorkflowExecution {
     common.UserCodeFailure failure = 1;
+}
+
+// TODO: Maybe combine all execution resolves into one message
```
vitarb, line 115: nit. "Execution" suffix sounds a bit redundant here and in other ones too

src/machines/continue_as_new_workflow_state_machine.rs
```diff
@@ -35,6 +56,110 @@ impl Created {
...
+        &self,
+        _event: &HistoryEvent,
+        _has_next_event: bool,
+        _my_command: Self::Command,
+    ) -> Result<Vec<MachineResponse>, WFMachinesError> {
+        Ok(vec![])
+    }
+}
+
+impl Cancellable for ContinueAsNewWorkflowMachine {}
```
vitarb, line 111: We don't have any use cases for continue as new cancellation, right?

vitarb (approved):

src/machines/continue_as_new_workflow_state_machine.rs
Sushisource: No, hence why the implementation is empty. All machines are required to implement the trait but only the ones that override the method it defines do anything.

PR #141, opened 2021-07-06
Add LRU cache for workflow task eviction

vitarb:
## What was changed
Workflow tasks will be evicted according to the cache policy defined. If non-sticky policy is used, then appropriate number of tasks will be kept in the cache and evictions will only be performed after cache capacity was reached.
This PR doesn't change places where eviction is called, but changes condition for eviction during successful completion.

## Why?
We need sticky cache to be size bound to avoid running out of memory.

## Discussion points
Let's discuss this approach and alternative one, where we touch cache when we receive new tasks.


src/workflow/cache_manager.rs
```diff
@@ -0,0 +1,154 @@
+use crate::workflow::WorkflowCachingPolicy;
+use lru::LruCache;
+use std::collections::HashSet;
+
+pub(crate) struct WorkflowCacheManager {
+    /// LRU cache of run IDs.
+    cache: LruCache<String, bool>,
```
Sushisource, line 7: Value should be `()` rather than a bool, slightly smaller and avoids possible `false` values.

src/workflow/cache_manager.rs
```diff
@@ -0,0 +1,154 @@
...
+        } else {
+            // Cache size is set to 0, we should add run_id to the list of pending evictions.
+            self.pending_evictions.insert(run_id.to_owned());
+        }
+    }
+
+    /// Returns all pending evictions currently recorded.
+    pub fn get_pending_evictions(&self) -> HashSet<String> {
+        self.pending_evictions.clone()
+    }
```
Sushisource, line 56: It would be a good idea to avoid cloning here. This should return by reference. Probably returning a set is not a relevant detail either. It could return `&impl Iterator<Item = String>` (or `&str`, not sure which) and then the method contents are just `self.pending_evictions.iter()`

src/workflow/cache_manager.rs
```diff
@@ -0,0 +1,154 @@
...
+        match self.eviction_policy {
+            WorkflowCachingPolicy::Sticky { .. } => self.pending_evictions.get(run_id).is_some(),
+            _ => true, // Non-sticky modes should always be evicted.
+        }
+    }
+
+    /// Removes run id from the cache and pending evictions.
+    pub fn evict(&mut self, run_id: &str) {
+        // All records in the pending evictions are no longer in the cache, so we need to remove the pending eviction itself.
+        // If there was no pending eviction, item can still be present in the cache and we want to remove it,
```
Sushisource, line 69: This comment implies an else branch that isn't here? It seems like it's saying we would still remove from both places here if the id wasn't in pending evictions

src/workflow/cache_manager.rs
```diff
@@ -0,0 +1,154 @@
...
+use lru::LruCache;
+use std::collections::HashSet;
+
+pub(crate) struct WorkflowCacheManager {
+    /// LRU cache of run IDs.
+    cache: LruCache<String, bool>,
+    /// Defines eviction behavior from the cache when tasks are completed.
+    eviction_policy: WorkflowCachingPolicy,
+    /// List of run ids that should be evicted upon completion.
+    pending_evictions: HashSet<String>,
```
Sushisource, line 11: We have run id in so many places now we might want to use `Arc<String>` to avoid copying them so often but, that's a bigger change for later. Might ultimately not matter much either.

src/workflow_tasks.rs
```diff
@@ -46,7 +48,9 @@ pub struct WorkflowTaskManager {
     ready_buffered_wft: DashMap<String, VecDeque<ValidPollWFTQResponse>>,
     /// Used to wake blocked workflow task polling
     workflow_activations_update: UnboundedSender<WfActivationUpdate>,
-    eviction_policy: WorkflowCachingPolicy,
+    /// Lock guarded cache manager, which is an authority for workflow task eviction from the cache.
```
Sushisource, line 51:
```suggestion
    /// Lock guarded cache manager, which is the authority for limit-based workflow machine eviction from the cache.
```

src/workflow_tasks.rs
```diff
@@ -266,14 +271,28 @@ impl WorkflowTaskManager {
...
 
+        // Evict all run ids that were invalidated by the cache eviction policy and have no pending/outstanding activations or workflow tasks.
+        for run_id in self.workflow_cache_manager.read().get_pending_evictions() {
+            if !self.pending_activations.has_pending(&run_id)
+                && !self.outstanding_activations.contains_key(&run_id)
+                && !self.outstanding_workflow_tasks.contains_key(&run_id)
+            {
+                self.evict_run(&run_id);
+            }
+        }
```
Sushisource, line 282: Need some UTs for this part. One problem I think I see here is if the run for which we are currently applying a poll response is in the pending evictions list we'll issue an eviction and then immediately issue the new activation, which works but is unnecessary. `touch` should probably be at the top of this method.

Also worth extracting this bit to it's own little function for readability

src/workflow/cache_manager.rs
vitarb: This code is no longer needed in the recent revision.

src/workflow/cache_manager.rs
vitarb: Done

src/workflow/cache_manager.rs
vitarb: This function has been changed significantly, comment too.

src/workflow_tasks.rs
vitarb: Updated

src/workflow_tasks.rs
vitarb: This is removed as I've inverted where touch/evict is done.

src/lib.rs
```diff
@@ -337,7 +338,10 @@ where
                 }
                 Ok(Some(work)) => match self.apply_server_work(work).await? {
                     Some(a) => return Ok(a),
-                    None => continue,
+                    None => {
+                        debug!("Skipping to the next poll");
```
Sushisource, line 342: Debug line can get removed

src/machines/workflow_machines.rs
```diff
@@ -139,6 +139,8 @@ pub enum WFMachinesError {
     InvalidTransition(String),
     #[error("Invalid cancelation type: {0}")]
     InvalidCancelationType(i32),
+    #[error("Unable to process partial event history because workflow is no longer cached.")]
+    ResetSticky,
```
Sushisource, line 143: Let's call this `CacheMiss` or something. A little more clear what went wrong rather than the intended outcome.

src/machines/workflow_machines.rs
```diff
@@ -479,6 +481,17 @@ impl WorkflowMachines {
...
         }
 
+        let first_event_id = match events.first() {
+            Some(event) => event.event_id,
+            None => 0,
+        };
+        // Workflow has been evicted, but we've received partial history from the server.
+        // Need to reset sticky and trigger another poll.
+        if self.current_started_event_id == 0 && first_event_id != 1 && !events.is_empty() {
+            debug!("Workflow cache miss. Resetting sticky and triggering another poll.");
```
Sushisource, line 491: I'd just say cache miss here. Technically at this level it's not known what the caller will do with this info so the log line could be slightly misleading

src/workflow_tasks.rs
```diff
@@ -149,6 +155,8 @@ impl WorkflowTaskManager {
             });
         if let Some(act) = maybe_act.as_ref() {
             self.insert_outstanding_activation(&act);
+            // Evict from the workflow cache to ensure that run id is not going to be invalidated before activation is complete.
+            self.cache_manager.write().evict(&act.run_id);
```
Sushisource, line 159: This should be touch, no? Explicitly evicting a run we just queued up an activation for seems wrong. Should have test for this as well

src/workflow_tasks.rs
```diff
@@ -176,10 +184,14 @@ impl WorkflowTaskManager {
     /// Evict a workflow from the cache by its run id and enqueue a pending activation to evict the
     /// workflow. Any existing pending activations will be destroyed, and any outstanding
     /// activations invalidated.
+    /// In case if evict_from_cache flag is passed, run id will also be removed from the LRU cache.
     ///
     /// Returns that workflow's task info if it was present.
-    pub fn evict_run(&self, run_id: &str) -> Option<WorkflowTaskInfo> {
+    pub fn evict_run(&self, run_id: &str, evict_from_cache: bool) -> Option<WorkflowTaskInfo> {
```
Sushisource, line 190: I don't love this extra flag. There's not any reason for it to be false except the line where you mention the lock is already held. We should just... not hold the lock there.

src/workflow_tasks.rs
```diff
@@ -383,14 +403,15 @@ impl WorkflowTaskManager {
             );
             return FailedActivationOutcome::NoReport;
         };
+        self.cache_manager.write().evict(run_id);
```
Sushisource, line 406: Why does this line need to be here? The below call to `evict_run` should be sufficient?

src/workflow_tasks.rs
vitarb: Yeah, I hate it too, removed.

src/workflow_tasks.rs
vitarb: As per our offline discussion, replaced evicts with touches and existing touch with insert.

src/machines/workflow_machines.rs
vitarb: Done

src/machines/workflow_machines.rs
vitarb: Yep

src/lib.rs
vitarb: Done

src/workflow/cache_manager.rs
vitarb: Let's probably defer this.

src/workflow_tasks.rs
vitarb: Yeah, no need to do it here as it's post activation completion.

src/workflow_tasks.rs
```diff
@@ -492,9 +507,11 @@ impl WorkflowTaskManager {
...
             }
 
-            // Blow them up if we're in non-sticky mode as well
-            if self.eviction_policy == WorkflowCachingPolicy::NonSticky {
-                self.evict_run(run_id);
+            // Evict run id if we are in non-sticky mode or if we have too many cached run ids.
+            let maybe_evicted = self.cache_manager.write().insert(run_id);
+            if let Some(evicted_run_id) = maybe_evicted {
+                // Not removing from the cache as record has already been removed and we are still holding cache manager lock.
+                self.evict_run(&evicted_run_id);
```
Sushisource, line 514: We've lost the nonsticky eviction condition. Also the bit in the comment about still holding the lock is irrelevant if insertion removes the one we're about to evict because we're at the limit anyway.

vitarb: It should actually work because we'll initialize cache with 0 capacity, but just in case if we've missed to touch in some situation, let's add a safety condition just in case.

Sushisource (approved):

src/workflow_tasks.rs
```diff
@@ -128,7 +134,7 @@ impl WorkflowTaskManager {
             outstanding_activations: Default::default(),
             ready_buffered_wft: Default::default(),
             workflow_activations_update,
-            eviction_policy,
+            cache_manager: RwLock::new(WorkflowCacheManager::new(eviction_policy)),
```
Sushisource, line 137: I've also realized there's no point in this being an `RwLock` since it's always locked with write. It can be a mutex.

src/workflow_tasks.rs
```diff
@@ -176,10 +184,10 @@ impl WorkflowTaskManager {
     /// Evict a workflow from the cache by its run id and enqueue a pending activation to evict the
     /// workflow. Any existing pending activations will be destroyed, and any outstanding
     /// activations invalidated.
-    ///
     /// Returns that workflow's task info if it was present.
     pub fn evict_run(&self, run_id: &str) -> Option<WorkflowTaskInfo> {
         debug!(run_id=%run_id, "Evicting run");
+        self.cache_manager.write().touch(run_id);
```
bergundy, line 190: Why do we need to touch the LRU here? Can you please document the reason?

bergundy (commented):
Really late to the party here, I wish I had looked at it earlier.
I'm missing the main design principal here.
When should the cache be touched?
Is it every time we access a WF machine? If so, did you consider putting the WF machine as the value in the LRU cache so it doesn't explicitly need to be touched everywhere?

bergundy (commented):
Another thing that comes to mind is when do we create the evict_activation?
Why only if `outstanding_workflow_tasks.remove()` is true?

src/workflow_tasks.rs
vitarb: It got messed up in the process of addressing comments, it should be remove here.

vitarb:
> Another thing that comes to mind is when do we create the evict_activation?
> Why only if `outstanding_workflow_tasks.remove()` is true?

We've agreed that we probably should evict, but we don't have a task queue at that level, just run id and task queue is required to enqueue a pending activation. I'll need to check with Specner on what is the right thing to do there.

vitarb:
Updated PR to store task queue in workflow machines, one side-effect is that it makes eviction async function as access to wf machines needs to be async.

src/workflow/concurrency_manager.rs
```diff
@@ -51,7 +52,7 @@ impl WorkflowConcurrencyManager {
         } else {
             // Create a new workflow machines instance for this workflow, initialize it, and
             // track it.
-            let mut wfm = WorkflowManager::new(history, workflow_execution);
+            let mut wfm = WorkflowManager::new(history, workflow_execution, task_queue);
```
Sushisource, line 55: You can avoid asyncing everything if you just make the value in the machines map a struct with the mutex and also the queue name. No need to pass it into the manager itself. Then you don't need to use `access` and fetching the queue name does not require async.

PR #143, opened 2021-07-07
Implement workflow cancellations

Sushisource:
## What was changed
Implemented workflow cancel machine & sending activation to lang

## Why?
Sometimes users cancel their workflows, ya know.

## Checklist

1. Closes 

2. How was this tested:
New tests

3. Any docs updates needed?

### DIFF
--- a/protos/local/workflow_activation.proto
+++ b/protos/local/workflow_activation.proto
@@ -99,7 +99,7 @@ message QueryWorkflow {
 
 /// Cancel a running workflow
 message CancelWorkflow {
-    // TODO: add attributes here
+    repeated common.Payload details = 1;
 }
 
 /// Send a signal to a workflow
--- a/protos/local/workflow_activation.proto
+++ b/protos/local/workflow_activation.proto
@@ -99,6 +99,7 @@ message QueryWorkflow {
 
 /// Cancel a running workflow
 message CancelWorkflow {
+    /// Information from the cancellation request
     repeated common.Payload details = 1;
 }
 
--- a/src/machines/workflow_machines.rs
+++ b/src/machines/workflow_machines.rs
@@ -339,6 +339,10 @@ impl WorkflowMachines {
         event: &HistoryEvent,
         has_next_event: bool,
     ) -> Result<()> {
+        debug!(
+            event = %event,
+            "handling non-stateful event"
+        );
         match EventType::from_i32(event.event_type) {
             Some(EventType::WorkflowExecutionStarted) => {
                 if let Some(history_event::Attributes::WorkflowExecutionStartedEventAttributes(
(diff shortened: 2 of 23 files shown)


src/machines/workflow_machines.rs
```diff
@@ -591,6 +605,10 @@ impl WorkflowMachines {
                     let canm = self.add_new_machine(continue_as_new(attrs));
                     self.current_wf_task_commands.push_back(canm);
                 }
+                WFCommand::AckCancelled(attrs) => {
```
bergundy, line 608: Didn't we rename this?

src/protos/mod.rs
```diff
@@ -203,6 +226,9 @@ pub mod coresdk {
                         workflow_command::Variant::ContinueAsNewWorkflowExecution(_) => {
                             write!(f, "ContinueAsNewWorkflowExecution")
                         }
+                        workflow_command::Variant::AckExecutionCancelled(_) => {
```
bergundy, line 229: Same here, let's rename

src/protos/mod.rs
```diff
@@ -203,6 +226,9 @@ pub mod coresdk {
...
+                            write!(f, "AckExecutionCancelled")
```
bergundy, line 230: And this string

bergundy (commented):
In the Node SDK I'm going to have these test scenarios:
1. Workflow cancel requested -> workflow is immediately cancelled
2. Workflow cancel requested -> workflow immediately completes
3. Workflow cancel requested -> workflow immediately fails
4. Workflow cancel requested -> workflow completes after cleanup
5. Workflow cancel requested -> workflow is cancelled after cleanup
6. Workflow cancel requested -> workflow fails after cleanup

These are all of the possible scenarios I can think of, would you test these here? Is this overkill?

src/machines/workflow_machines.rs
Sushisource: This is a core-internal name but yeah I'll keep it consistent

protos/local/workflow_activation.proto
```diff
@@ -99,7 +99,7 @@ message QueryWorkflow {
 
 /// Cancel a running workflow
 message CancelWorkflow {
-    // TODO: add attributes here
+    repeated common.Payload details = 1;
```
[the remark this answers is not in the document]
Sushisource: The cancellation request. Will add docstring

Sushisource:
@bergundy Re: testing. I definitely would *like* to, but as I was mentioning to you in slack it's amazing how much time trying to get the baby Rust SDK/worker going takes compared to everything else. So from a practicality standpoint, core getting tested via your tests just makes a lot of sense, and maybe we can defer the closer-to-core testing until we start productionalizing a Rust SDK.

If your tests uncover problems in Core, then adding regression tests for those scenarios has worked well so far.

I am definitely open to investing more time up front, too, just a question of what we feel is the best investment of our time ATM

bergundy (approved):
Awesome, let's merge this

Issue #134, opened 2021-06-22
[Bug] Worker fails with TransportError when trying trying to complete WFT for terminated WF


bergundy:
I've had to [work around](https://github.com/temporalio/sdk-node/pull/132/commits/38196a4adf04738458ee8fdfc680f8c1e1eb2830) it in the Node SDK test suite.
Core should probably ignore this error and evict the WF from the cache.

I've managed to reproduce this by running the test in a loop:
```shell
for x in `seq 1 50`; do RUN_INTEGRATION_TESTS=1 npx ava packages/test/lib/test-interceptors.js; done
```

Sushisource:
```
[TransportError: Unhandled error when calling the temporal server: Status { code: NotFound, message: "workflow execution already completed", details: b"\x08\x05\x12$workflow execution already completed\x1aB\n@type.googleapis.com/temporal.api.errordetails.v1.NotFoundFailure", metadata: MetadataMap { headers: {"content-type": "application/grpc"} } }]
```

PR #144, opened 2021-07-09
Handle completion errors everywhere they can happen

Sushisource:
## What was changed
Handle completion errors everywhere they can be generated

## Why?
Fixes spurious errors on shutdown

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/134
2. How was this tested:
New tests

3. Any docs updates needed?


bergundy (approved):
Nice, thanks!

PR #150, opened 2021-07-18
Fix unnecessary activity not found warnings when we already know

Sushisource:
## What was changed
No longer warn if we have prior knowledge an activity already doesn't exist according to server.

## Why?
No spurious warnings

## Checklist

1. Closes 

2. How was this tested:
Against node js bench

3. Any docs updates needed?

### DIFF
--- a/src/activity/mod.rs
+++ b/src/activity/mod.rs
@@ -1,9 +1,7 @@
 mod activity_heartbeat_manager;
 mod details;
 
-pub(crate) use activity_heartbeat_manager::{
-    ActivityHeartbeatManager, ActivityHeartbeatManagerHandle,
-};
+pub(crate) use activity_heartbeat_manager::ActivityHeartbeatManager;
 pub(crate) use details::InflightActivityDetails;
 
 use crate::{
@@ -17,7 +15,7 @@ use std::{convert::TryInto, ops::Div, sync::Arc, time::Duration};
 
 pub(crate) struct ActivityTaskManager {
     /// Handle to the heartbeat manager
-    activity_heartbeat_manager_handle: ActivityHeartbeatManagerHandle,
+    activity_heartbeat_manager: ActivityHeartbeatManager,
     /// Activities that have been issued to lang but not yet completed
     outstanding_activity_tasks: DashMap<TaskToken, InflightActivityDetails>,
 }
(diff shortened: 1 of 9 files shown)


src/activity/mod.rs
```diff
@@ -125,4 +131,11 @@ impl ActivityTaskManager {
         self.activity_heartbeat_manager_handle
             .record(details, t.div(2))
     }
+
+    pub(crate) fn is_known_not_found(&self, task_token: &TaskToken) -> bool {
+        self.outstanding_activity_tasks
+            .get(&task_token)
+            .map(|d| d.known_not_found)
+            .unwrap_or_default()
```
bergundy, line 139: This means return false if tt is not in the map right?

src/lib.rs
```diff
@@ -108,6 +108,9 @@ pub trait Core: Send + Sync {
     /// The returned activation is guaranteed to be for the same task queue / worker which was
     /// provided as the `task_queue` argument.
     ///
+    /// It is rarely a good idea to call poll concurrently. It handles polling the server
```
bergundy, line 111: This raises the question of when it is a good idea to poll concurrently.

src/lib.rs
```diff
@@ -423,36 +429,47 @@ where
...
-        let (res, should_remove) = match maybe_net_err {
-            Some(e) if e.code() == tonic::Code::NotFound => {
-                warn!(task_token = ?tt, details = ?e, "Activity not found on completion. \
-                 This may happen if the activity has already been cancelled but completed anyway.");
-                (Ok(()), true)
+
+        let is_known_not_found = self.act_manager.is_known_not_found(&task_token);
+
+        // No need to report activities which we already know the server doesn't care about
+        let (res, should_remove) = if !is_known_not_found {
```
bergundy, line 436: I find this hard to read.
Can we maybe split this function and have something like early return on `is_known_not_found`?

bergundy (approved):

src/activity/mod.rs
Sushisource: Yep

src/lib.rs
Sushisource: Good point. Probably never.... but never say never? I can change it to just say don't

src/lib.rs
Sushisource: Not quite, because the call to remove from the manager still has to happen later. I can collapse the first branch into its own function though

Sushisource: Extracting it ends up being harder to read IMO... will leave as is for now since I don't have any great ideas about how to make it look nicer.

PR #152, opened 2021-07-20
Avoid pointless extra calls to the poll buffer

Sushisource:
## What was changed
Previously we always began a poll even if there might be pending notification of a task update. Avoid that.

## Why?
Optimize. Avoid pointless server calls.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


vitarb (approved):

bergundy (commented):
Nice one

PR #154, opened 2021-07-23
Eviction confirmation

Sushisource:
## What was changed
Evictions must now be confirmed with an activation completion. This fixes an obnoxious inconsistency where not all activations were necessarily replied to, and also makes it much easier to keep the cache in sync. Core will not actually evict until lang confirms.

## Why?
Easier to understand and eliminates edge cases.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/workflow_tasks.rs
```diff
@@ -535,10 +580,15 @@ impl WorkflowTaskManager {
...
-                self.evict_run(&evicted_run_id);
+            if !just_evicted {
+                // Evict run id if cache is full. Non-sticky will always evict.
+                let maybe_evicted = self.cache_manager.lock().insert(run_id);
+                if let Some(evicted_run_id) = maybe_evicted {
+                    // Don't try to re-evict something we just evicted
+                    if evicted_run_id != run_id || !just_evicted {
+                        self.request_eviction(&evicted_run_id);
+                    }
+                }
```
bergundy, line 591: This should probably be refactored into a method `self.touch_and_maintain_cache(run_id)` or something like that.

src/workflow_tasks.rs
```diff
@@ -554,11 +604,26 @@ impl WorkflowTaskManager {
         }
     }
 
+    /// Must be called after *every* activation is replied to, regardless of whether or not we
+    /// had some issue reporting it to server or anything else. This upholds the invariant that
+    /// every activation we issue to lang has exactly one reply.
+    ///
+    /// Any subsequent action that needs to be taken will be created as a new activation
+    pub(crate) fn activation_done(&self, run_id: &str) {
+        if self.outstanding_activations.remove(run_id).is_some() {
```
bergundy, line 613: Would it be considered an error if there was nothing removed from `outstanding_activations`?
Also, consider renaming this method to something that clarifies it's a side-effect and doesn't just return a boolean. Maybe `on_activation_done` or `complete_activation`?

src/workflow_tasks.rs
```diff
@@ -554,11 +604,26 @@ impl WorkflowTaskManager {
...
+            let _ = self
+                .workflow_activations_update
+                .send(WfActivationUpdate::NewPendingActivation);
+        }
+    }
+
     fn insert_outstanding_activation(&self, act: &WfActivation) {
```
bergundy, line 620: Will there ever be more than one activation per run_id?
Should this method assert there isn't?

bergundy (approved):
LGTM

src/protos/mod.rs
```diff
@@ -132,14 +132,22 @@ pub mod coresdk {
...
-                matches!(
-                    &self.jobs.as_slice(),
-                    &[WfActivationJob {
-                        variant: Some(wf_activation_job::Variant::RemoveFromCache(_))
-                    }]
-                )
+            /// Returns the index of the eviction job if this activation contains one. If present
+            /// it should always be the last job in the list.
+            pub(crate) fn eviction_index(&self) -> Option<usize> {
+                self.jobs.iter().position(|j| {
```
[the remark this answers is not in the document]
Sushisource: I'd rather just do this which is unlikely to ever get close to a perf problem and enjoy the safety than not

src/pending_activations.rs
```diff
@@ -21,6 +24,10 @@ struct PaInner {
 }
 
 impl PendingActivations {
+    /// Push a pending activation into the given task queue.
+    ///
+    /// Importantly, if there already exist activations for same workflow run in the queue, those
+    /// activations will be merged.
```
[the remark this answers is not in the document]
Sushisource: I believe it's just evictions at this point.

src/protos/mod.rs
```diff
@@ -357,6 +374,45 @@ pub mod coresdk {
             }
         }
 
+        /// Returns true if the activation has either a fail, continue, cancel, or complete workflow
+        /// execution command in it.
+        pub fn has_execution_ending(&self) -> bool {
+            self.has_complete_workflow_execution()
```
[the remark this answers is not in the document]
Sushisource: Yeah, probably worth making that change at some point

src/workflow/cache_manager.rs
```diff
@@ -28,9 +29,13 @@ impl WorkflowCacheManager {
             self.cache.put(run_id.to_owned(), ());
             None
         } else if self.cache.cap() != 0 {
-            let result = self.cache.peek_lru().map(|r| r.0.to_owned());
-            self.cache.put(run_id.to_owned(), ());
-            result
+            let maybe_got_evicted = self.cache.peek_lru().map(|r| r.0.to_owned());
+            let already_existed = self.cache.put(run_id.to_owned(), ()).is_some();
+            if !already_existed {
+                maybe_got_evicted
```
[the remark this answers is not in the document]
Sushisource: `maybe_xxx` is a pretty common term across a few languages for optional types in my experience. If it was just bool it'd be `got_evicted` or `was_evicted`

src/workflow_tasks.rs
Sushisource: Not an error b/c sometimes it's already been removed due to evicting. Added comment.

src/workflow_tasks.rs
Sushisource: Good call. Started doing this and it revealed a bug where we would overwrite an eviction with new work from the server.

src/core_tests/queries.rs
```diff
@@ -74,12 +73,18 @@ async fn legacy_query(#[case] include_history: bool) {
         .await
         .unwrap();
     };
+    let clear_eviction = || async {
+        let t = core.poll_workflow_task(TEST_Q).await.unwrap();
+        core.complete_workflow_task(WfActivationCompletion::empty(t.run_id))
+            .await
+            .unwrap();
+        first_wft().await;
```
bergundy, line 81: I would leave `first_wft` out of this function, it doesn't have anything to do with clearing the eviction.

src/pending_activations.rs
```diff
@@ -98,6 +105,51 @@ impl PendingActivations {
     }
 }
 
+/// TODO: This might break with legacy queries, which cannot be merged with anything. Fix.
```
bergundy, line 108: We should address this before merging.

src/workflow/workflow_tasks/concurrency_manager.rs
```diff
@@ -0,0 +1,284 @@
...
+use tokio::sync::Mutex;
+
+/// Provides a thread-safe way to access workflow machines for specific workflow runs
+pub(crate) struct WorkflowConcurrencyManager {
+    /// Maps run id -> data about and machines for that run
+    machines: RwLock<HashMap<String, ManagedRun>>,
+}
+
+struct ManagedRun {
+    wfm: Mutex<Option<WorkflowManager>>,
```
bergundy, line 28: Why do we need the mutex here and only here?

src/workflow/workflow_tasks/concurrency_manager.rs
```diff
@@ -0,0 +1,284 @@
...
+    collections::HashMap,
+    fmt::Debug,
+    ops::{Deref, DerefMut},
+};
+use tokio::sync::Mutex;
+
+/// Provides a thread-safe way to access workflow machines for specific workflow runs
+pub(crate) struct WorkflowConcurrencyManager {
+    /// Maps run id -> data about and machines for that run
+    machines: RwLock<HashMap<String, ManagedRun>>,
```
bergundy, line 24: This is not `machines`, it's managed runs, should we rename the property?

bergundy (approved):
Looks like it wasn't too hard to refactor all of these maps.
Would the next step be to have a manager like this per worker?

src/workflow/workflow_tasks/concurrency_manager.rs
Sushisource: There's another comment about this, but basically it allows the read lock to be held instead of the write lock on the overall map. Since most work happens in here, it's more efficient (theoretically).

I don't want to bother tweaking this until I have a change to move this all inside of a worker per the other comment. Then will be a good time to simplify.

src/workflow/workflow_tasks/concurrency_manager.rs
Sushisource: yes

PR #156, opened 2021-07-23
Versioning Interface / Machine

Sushisource:
## What was changed
Add `check_version` command

## Why?
Users sometimes version their workflows! :astonished: 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (approved):
LGTM

src/machines/version_state_machine.rs
```diff
@@ -24,10 +27,31 @@ fsm! {
...
+///
+/// `change_id`: identifier of a particular change. All calls to get_version that share a change id
+/// are guaranteed to return the same version number.
+/// `replaying_when_invoked`: If the workflow is replaying when this invocation occurs, this needs
+/// to be set to true.
+pub(super) fn get_version_invoked(
+    change_id: String,
+    replaying_when_invoked: bool,
+) -> NewMachineWithCommand<VersionMachine> {
+    let (activity, add_cmd) = VersionMachine::new_scheduled(attribs);
```
bergundy, line 49:
```suggestion
    let (machine, add_cmd) = VersionMachine::new_scheduled(attribs);
```

bergundy:
I don't see an activation with the response to lang.
I want to support a synchronous method in node but if we use commands and activations we have to make the method async.
One way I thought we could work around this is if we send an activation to lang to turn the marker on ahead of time in case of replay, I'm not sure this can be done reliably.
Another thing to consider with this new flag mechanism is that lang only needs to set it once with the latest value and from that moment on the WF can just assume the flag is on for the duration of the run unless it sees a different marker.
Let's discuss this soon.

Another thing that we have to figure out is how to help users query their WF histories for markers so they know when it's safe to deploy each version of their WF.
Maybe if the marker is registered on WF startup that would give us enough information because then you can easily query for histories without a marker.

src/machines/fail_workflow_state_machine.rs
```diff
@@ -89,13 +90,13 @@ impl TryFrom<HistoryEvent> for FailWorkflowMachineEvents {
...
-        Ok(match EventType::from_i32(e.event_type) {
-            Some(EventType::WorkflowExecutionFailed) => Self::WorkflowExecutionFailed,
+        Ok(match e.event_type() {
+            EventType::WorkflowExecutionFailed => Self::WorkflowExecutionFailed,
             _ => {
-                return Err(WFMachinesError::UnexpectedEvent(
-                    e,
-                    "Fail workflow machine does not handle this event",
-                ))
+                return Err(WFMachinesError::Nondeterminism(format!(
```
vitarb, line 96: Why is it attributed to non-determinism here?

src/machines/workflow_machines.rs
```diff
@@ -516,12 +488,34 @@ impl WorkflowMachines {
...
+                self.encountered_change_markers.insert(
+                    change_id.clone(),
+                    ChangeInfo {
+                        deprecated,
+                        created_command: false,
+                    },
+                );
+                // Found a change marker
+                self.drive_me
+                    .send_job(wf_activation_job::Variant::ResolveHasChange(
```
vitarb, line 510: Is it guaranteed that this job won't be received by lang too early? Or are we expecting lang to cache change statuses?

protos/local/workflow_commands.proto
```diff
@@ -137,4 +138,16 @@ message ContinueAsNewWorkflowExecution {
 
 /// Indicate a workflow has completed as cancelled. Generally sent as a response to an activation
 /// containing a cancellation job.
-message CancelWorkflowExecution {}
\ No newline at end of file
+message CancelWorkflowExecution {}
+
+/// A request to set/check version information for a certain change.
+/// TODO: Add table of outcomes
+message HasChange {
```
vitarb, line 145: nit. I'm not sure about this name as it's a bit too passive, maybe `ApplyChange` would be better name for a job as it would assume that something needs to be done with it?

vitarb (approved):
LGTM with couple small comments.

bergundy (commented):
I know we've almost finished this already but I thought of a way to get rid of the deprecated flag.
Not sure about the API, consider this a draft.

version 0:

```js
async function main() {
  await activity1();
}

export const workflow = { main };
```

version 1:
Make a change, increment backwards-compatible version.
If a worker running version 0 picks up this history it'll fail.

```js
import { Context } from '@temporalio/workflow';

async function main() {
  if (Context.historyVersion >= 1) {
    await activity2();
  } else {
    await activity1();
  }
}

export const workflow = { version: 1, main };
```

version 2:
New version, remove the version check from main.
Mark that we cannot process histories from before version 1.

```js
import { Context } from '@temporalio/workflow';

async function main() {
    await activity2();
}

export const workflow = { version: 2, minHistoryVersion: 1, main };
```

WDYT?
The server can help us out by automatically adding `minHistoryVersion` as a search attribute so we can tell when it's safe to deploy version 2.

bergundy:
Something else that the server can help out with is verifying that the user incremented the Workflow version by comparing the content hash with the previous version's content hash.

src/machines/workflow_machines.rs
Sushisource: It needs to cache them

src/machines/fail_workflow_state_machine.rs
Sushisource: Because if the machine is processing an event it doesn't understand then that means user's WF code failed to produce a machine that *should've* matched it.

src/machines/mod.rs
```diff
@@ -221,6 +234,28 @@ where
...
+    <SM as StateMachine>::Command: Debug + Display,
+    <SM as StateMachine>::State: Display,
+{
+    if !commands.is_empty() {
+        debug!(commands=%commands.display(), state=%machine.state(),
+               machine_name=%StateMachine::name(machine), "Machine produced commands");
+    }
+    let mut machine_responses = vec![];
+    for cmd in commands {
+        machine_responses.extend(machine.adapt_response(cmd, event_info)?);
```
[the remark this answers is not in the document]
Sushisource: Not quite as that'd still generate a new list. I find this pretty readable in any case

src/machines/version_state_machine.rs
```diff
@@ -1,58 +1,131 @@
...
+impl VersionMachine {
+    fn new_scheduled(
+        state: SharedState,
+        replaying_when_invoked: bool,
+        deprecated: bool,
+    ) -> (Self, Command) {
+        let initial_state = if replaying_when_invoked {
+            Replaying {}.into()
+        } else {
+            Executing {}.into()
```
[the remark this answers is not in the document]
Sushisource: It's best to have them different since they immediately transition to different states, also more clear when tracing/debugging

PR #159, opened 2021-07-29
Incorporate proto failures from upstream

bergundy:
Started porting the failures, still needs testing and work on cancellation.


protos/local/failure.proto
```diff
@@ -0,0 +1,77 @@
+syntax = "proto3";
+
+package coresdk.failure;
+
+import "google/protobuf/duration.proto";
+import "common.proto";
+
+/// Represents a failure in user code, workflow or activity, which could've been triggered by
+/// an exception or similar error mechanism like the error half of a Result type.
+message ApplicationFailureInfo {
```
Sushisource, line 10: Can we keep calling this `UserCodeFailure` ? I think that name is way more clear.

Also we can probably drop `Info` off of all these names. Not useful.

protos/local/failure.proto
```diff
@@ -0,0 +1,77 @@
...
+    string namespace = 1;
+    common.WorkflowExecution workflow_execution = 2;
+    string workflow_type = 3;
+    int64 initiated_event_id = 4;
+    int64 started_event_id = 5;
+    common.RetryState retry_state = 6;
+}
+
+/// Structured failure which can be mapped to upstream proto Failure
+message Failure {
```
Sushisource, line 62: As discussed in slack this is nasty and we shouldn't copy it. Way too easy to represent impossible states since this `Failure` includes every possible failure type but only some failures are possible depending on what's failing.

protos/local/failure.proto
bergundy: Let's keep the `ApplicationFailure` name for compatibility with the other SDKs.

bergundy:
@Sushisource I'm wondering if it's worth keeping the core definitions closer to the upstream ones, I would like Core based SDKs to have the same failure classes as the Java SDK, we shouldn't diverge here too much especially considering the documentation / terminology overhead.
I'm not convinced the exact causes are a good enough reason to use different classes.
If you insist we can delay this PR until Maxim has availability later this week but keep in mind a lot is depending on this change.

FTR: There are still some unresolved issues from our conversation with @mfateev:

1. Do we ever instantiate the generic `TemoralFailure` class?
2. Which failures are valid causes of other failures?

There's also the risk of some other SDK or the server chaining errors in a way we didn't expect and breaking Core processing.
The only way to prevent it is to change the Service proto.

Sushisource:
> There's also the risk of some other SDK or the server chaining errors in a way we didn't expect and breaking Core processing.
> The only way to prevent it is to change the Service proto.

I don't think this is a concern. We already do things differently in core (like, change markers are totally different now), and there's no expectation that you should be able to replay history from a workflow that ran with a non-core SDK. I don't know what kind of crazy someone would have to be to want to write the exact same workflow in different languages anyway.

But, it could reasonably happen with a child workflow, so that part makes sense

Sushisource:
If we're explicitly saying we want to be compatible with upstream protos I think we might as well just use them directly at this point.

I do want to edit the upstream repo to actually have docstrings, though.

Sushisource (approved):

PR #163, opened 2021-08-09
Implement child workflow state machine

bergundy:
Implemented the ChildWorkflowMachine, currently we index child WF machines by their workflow ID which is a bug, @Sushisource let's discuss how we'd change this, we'll need to change it for activities as well.
I suggest we make the correlation ID change later and merge this as is.
For complete child workflow functionality we should implement the external WF machines for cancel and signal.

### DIFF
--- a/protos/local/child_workflow.proto
+++ b/protos/local/child_workflow.proto
@@ -38,6 +38,7 @@ message Failure {
  * in case its parent is closed.
  */
 enum ParentClosePolicy {
+    /** Let's the server set the default. */
     PARENT_CLOSE_POLICY_UNSPECIFIED = 0;
     /** Terminate means terminating the child workflow. */
     PARENT_CLOSE_POLICY_TERMINATE = 1;
@@ -57,13 +58,13 @@ enum StartChildWorkflowExecutionFailedCause {
  * Controls at which point to report back to lang when it cancels a child workflow
  */
 enum ChildWorkflowCancellationType {
-    /// Do not request cancellation of the child workflow
+    /** Do not request cancellation of the child workflow if already scheduled */
     ABANDON = 0;
-    /// Initiate a cancellation request and immediately report cancellation to the parent.
+    /** Initiate a cancellation request and immediately report cancellation to the parent. */
     TRY_CANCEL = 1;
-    /// Wait for child cancellation completion.
+    /** Wait for child cancellation completion. */
     WAIT_CANCELLATION_COMPLETED = 2;
-    /// Request cancellation of the child and wait for confirmation that the request was received.
+    /** Request cancellation of the child and wait for confirmation that the request was received. */
     WAIT_CANCELLATION_REQUESTED = 3;
 }
 
--- a/protos/local/child_workflow.proto
+++ b/protos/local/child_workflow.proto
@@ -1,0 +1,69 @@
+syntax = "proto3";
+
+package coresdk.child_workflow;
+
+import "common.proto";
+import "temporal/api/failure/v1/message.proto";
+
+/**
+ * Used by core to resolve child workflow executions.
+ *
+ * When a child workflow is cancelled, `status` is failed and
+ * failure.cause.info is set to CanceledFailure.
+ */
+message ChildWorkflowResult {
+    oneof status {
+        Success completed = 1;
+        Failure failed = 2;
+    }
+}
+
+/**
+ * Used in ChildWorkflowResult to report successful completion.
+ */
+message Success {
+    common.Payload result = 1;
+}
+
+/**
+ * Used in ChildWorkflowResult to report non successful outcomes such as
+ * application failures, timeouts, terminations, and cancellations.
+ */
+message Failure {
+    temporal.api.failure.v1.Failure failure = 1;
+}
+
+/**
+ * Used by the service to determine the fate of a child workflow
+ * in case its parent is closed.
+ */
+enum ParentClosePolicy {
+    PARENT_CLOSE_POLICY_UNSPECIFIED = 0;
+    /** Terminate means terminating the child workflow. */
+    PARENT_CLOSE_POLICY_TERMINATE = 1;
+    /** Abandon means not doing anything on the child workflow. */
+    PARENT_CLOSE_POLICY_ABANDON = 2;
+    /** Cancel means requesting cancellation on the child workflow. */
+    PARENT_CLOSE_POLICY_REQUEST_CANCEL = 3;
+}
+
+/** Possible causes of failure to start a child workflow */
+enum StartChildWorkflowExecutionFailedCause {
+    START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_UNSPECIFIED = 0;
+    START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_WORKFLOW_ALREADY_EXISTS = 1;
+}
+
+/**
+ * Controls at which point to report back to lang when it cancels a child workflow
+ */
+enum ChildWorkflowCancellationType {
+    /// Do not request cancellation of the child workflow
+    ABANDON = 0;
+    /// Initiate a cancellation request and immediately report cancellation to the parent.
+    TRY_CANCEL = 1;
+    /// Wait for child cancellation completion.
+    WAIT_CANCELLATION_COMPLETED = 2;
+    /// Request cancellation of the child and wait for confirmation that the request was received.
+    WAIT_CANCELLATION_REQUESTED = 3;
+}
+
--- a/src/machines/child_workflow_state_machine.rs
+++ b/src/machines/child_workflow_state_machine.rs
@@ -1,31 +1,85 @@
-use rustfsm::{fsm, TransitionResult};
+use crate::machines::activity_state_machine::ActivityMachineCommand::Fail;
+use crate::{
+    machines::{
+        workflow_machines::MachineResponse, Cancellable, EventInfo, MachineKind,
+        NewMachineWithCommand, OnEventWrapper, WFMachinesAdapter, WFMachinesError,
+    },
+    protos::{
+        coresdk::{
+            child_workflow::{
+                self as wfr, child_workflow_result::Status as ChildWorkflowStatus,
+                ChildWorkflowCancellationType, ChildWorkflowResult,
+            },
+            common::Payload,
+            workflow_activation::{
+                resolve_child_workflow_execution_start, ResolveChildWorkflowExecution,
+                ResolveChildWorkflowExecutionStart, ResolveChildWorkflowExecutionStartCancelled,
+                ResolveChildWorkflowExecutionStartFailure,
+                ResolveChildWorkflowExecutionStartSuccess,
+            },
+            workflow_commands::StartChildWorkflowExecution,
+        },
+        temporal::api::{
+            command::v1::{command, Command},
+            common::v1::{Payloads, WorkflowExecution, WorkflowType},
+            enums::v1::{
+                CommandType, EventType, RetryState, StartChildWorkflowExecutionFailedCause,
+                TimeoutType,
+            },
+            failure::v1::{self as failure, failure::FailureInfo, ActivityFailureInfo, Failure},
+            history::v1::{
+                history_event, ChildWorkflowExecutionCanceledEventAttributes,
+                ChildWorkflowExecutionCompletedEventAttributes,
+                ChildWorkflowExecutionFailedEventAttributes,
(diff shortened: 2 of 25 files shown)


protos/local/child_workflow.proto
```diff
@@ -0,0 +1,69 @@
...
+ * Used by the service to determine the fate of a child workflow
+ * in case its parent is closed.
+ */
+enum ParentClosePolicy {
+    PARENT_CLOSE_POLICY_UNSPECIFIED = 0;
+    /** Terminate means terminating the child workflow. */
+    PARENT_CLOSE_POLICY_TERMINATE = 1;
+    /** Abandon means not doing anything on the child workflow. */
+    PARENT_CLOSE_POLICY_ABANDON = 2;
+    /** Cancel means requesting cancellation on the child workflow. */
```
Sushisource, line 46: I enjoy that you really played jazz with the different forms of commenting here `/**` `///` just missing `//` :smile:

protos/local/child_workflow.proto
```diff
@@ -0,0 +1,69 @@
...
+message Failure {
+    temporal.api.failure.v1.Failure failure = 1;
+}
+
+/**
+ * Used by the service to determine the fate of a child workflow
+ * in case its parent is closed.
+ */
+enum ParentClosePolicy {
+    PARENT_CLOSE_POLICY_UNSPECIFIED = 0;
```
Sushisource, line 41: Should probably say what we default to if left unset -- or will we just not allow unset?

src/machines/child_workflow_state_machine.rs
```diff
@@ -62,13 +163,23 @@ pub(super) struct StartEventRecorded {}
 impl StartEventRecorded {
     pub(super) fn on_child_workflow_execution_started(
         self,
+        attrs: ChildWorkflowExecutionStartedEventAttributes,
     ) -> ChildWorkflowMachineTransition<Started> {
-        unimplemented!()
+        ChildWorkflowMachineTransition::ok(vec![
+            ChildWorkflowCommand::Start(attrs.workflow_execution.expect("ChildWorkflowExecutionStartedEventAttributes.workflow_execution should not be empty"))
```
Sushisource, line 169: Rather than `expect`ing here, the conversion from the history event should package exactly what is needed into the machine event. Only some subset of the attributes are required here, in any case.

Sushisource: Same for the other places this happens

src/machines/child_workflow_state_machine.rs
```diff
@@ -111,3 +340,487 @@ pub(super) struct Terminated {}
 
 #[derive(Default, Clone)]
 pub(super) struct TimedOut {}
+
+#[derive(Default, Clone)]
+pub(super) struct SharedState {
+    scheduled_event_id: i64,
+    //     started_event_id: i64,
```
Sushisource, line 347: Comment can be removed?

src/machines/child_workflow_state_machine.rs
```diff
@@ -111,3 +340,487 @@ pub(super) struct Terminated {}
...
+                {
+                    Self::ChildWorkflowExecutionCancelled(attrs)
+                } else {
+                    return Err(WFMachinesError::Fatal(
+                        "ChildWorkflowExecutionCanceled attributes were unset".to_string(),
+                    ));
+                }
+            }
+            _ => {
+                return Err(WFMachinesError::Fatal(
```
Sushisource, line 487: This is a nondeterminism error

src/test_help/canned_histories.rs
```diff
@@ -1362,3 +1368,158 @@ pub fn activity_double_resolve_repro() -> TestHistoryBuilder {
...
+///  3: EVENT_TYPE_WORKFLOW_TASK_STARTED
+///  4: EVENT_TYPE_WORKFLOW_TASK_COMPLETED
+///  5: EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED
+///  6: EVENT_TYPE_CHILD_WORKFLOW_EXECUTION_STARTED
+///  7: EVENT_TYPE_WORKFLOW_TASK_SCHEDULED
+///  8: EVENT_TYPE_WORKFLOW_TASK_STARTED
+///  9: EVENT_TYPE_WORKFLOW_TASK_COMPLETED
+/// 10: EVENT_TYPE_CHILD_WORKFLOW_EXECUTION_COMPLETED
+/// 11: EVENT_TYPE_WORKFLOW_TASK_SCHEDULED
+/// 12: EVENT_TYPE_WORKFLOW_TASK_STARTED
```
Sushisource, line 1383: Usually a good idea to have the history go through workflow execution completed

src/machines/child_workflow_state_machine.rs
```diff
@@ -111,3 +340,487 @@ pub(super) struct Terminated {}
...
+        let commands = wfm.get_server_commands().await.commands;
+        assert_eq!(commands.len(), 1);
+        assert_eq!(
+            commands[0].command_type,
+            CommandType::StartChildWorkflowExecution as i32
+        );
+
+        wfm.get_next_activation().await.unwrap();
+        let commands = wfm.get_server_commands().await.commands;
+        assert_eq!(commands.len(), 0);
```
Sushisource, line 751: What happens during this activation? I would expect this one to not exist - the first starting, and the second resolving, seems like it should be all that is needed?

src/machines/child_workflow_state_machine.rs
```diff
@@ -111,3 +340,487 @@ pub(super) struct Terminated {}
...
+        assert_eq!(2, t.get_full_history_info().unwrap().wf_task_count());
+        ManagedWFFunc::new(t, func, vec![])
+    }
+
+    #[rstest(wfm, case::default(child_workflow_cancel_before_sent()))]
+    #[tokio::test]
+    async fn single_child_workflow_cancel_before_sent(mut wfm: ManagedWFFunc) {
+        wfm.get_next_activation().await.unwrap();
+        let commands = wfm.get_server_commands().await.commands;
+        assert_eq!(commands.len(), 0);
```
Sushisource, line 815: Another curious seemingly unnecessary activation

src/machines/mod.rs
```diff
@@ -100,6 +103,15 @@ impl TryFrom<WorkflowCommand> for WFCommand {
...
             }
             workflow_command::Variant::SetChangeMarker(s) => Ok(WFCommand::SetChangeMarker(s)),
+            workflow_command::Variant::StartChildWorkflowExecution(s) => {
+                Ok(WFCommand::AddChildWorkflow(s))
+            }
+            workflow_command::Variant::RequestCancelExternalWorkflowExecution(s) => {
+                Ok(WFCommand::RequestCancelChildWorkflow(s))
+            }
+            workflow_command::Variant::SignalExternalWorkflowExecution(_) => {
+                Err(EmptyWorkflowCommandErr) // TODO
```
Sushisource, line 113: This can just be a `todo!` macro to be a little more obvious until it's done

src/machines/mod.rs
```diff
@@ -37,7 +37,8 @@ use crate::{
         coresdk::workflow_commands::{
             workflow_command, CancelTimer, CancelWorkflowExecution, CompleteWorkflowExecution,
             ContinueAsNewWorkflowExecution, FailWorkflowExecution, QueryResult,
-            RequestCancelActivity, ScheduleActivity, SetChangeMarker, StartTimer, WorkflowCommand,
+            RequestCancelActivity, RequestCancelExternalWorkflowExecution, ScheduleActivity,
```
Sushisource, line 40: Stupid github won't let me comment on unchanged lines, but, above remove allow unused from child wf machine

src/prototype_rust_sdk/workflow_future.rs
```diff
@@ -255,12 +296,6 @@ impl Future for WorkflowFuture {
                 }
             }
 
-            if activation_cmds.is_empty() {
-                panic!(
-                    "Workflow did not produce any commands or exit, but awaited. This \
-                 means it will deadlock. You probably awaited on a non-WfContext future."
-                );
-            }
```
Sushisource, line 263: Ah, this lines up with the weird non-activations. I'd really like to keep this unless there's a really good reason it needs to go.

tests/integ_tests/workflow_tests/continue_as_new.rs
```diff
@@ -27,9 +27,10 @@ async fn continue_as_new_happy_path() {
     let wf_name = "continue_as_new_happy_path";
     let mut starter = CoreWfStarter::new(wf_name);
     let worker = starter.worker().await;
+    worker.register_wf(wf_name.to_string(), continue_as_new_wf);
 
     worker
-        .submit_wf(vec![[1].into()], wf_name.to_owned(), continue_as_new_wf)
+        .submit_wf(wf_name.to_string(), wf_name.to_string(), vec![[1].into()])
         .await
         .unwrap();
```
Sushisource, line 35: Might be nice to create a combined register & submit convenience function for all these.

protos/local/child_workflow.proto
bergundy: It's copied as is from upstream, I'll add a comment.

protos/local/child_workflow.proto
bergundy: Apparently protobufjs doesn't treat `///` as a docstring, started moving to `/**` for that.

src/machines/child_workflow_state_machine.rs
bergundy: Yup

src/machines/child_workflow_state_machine.rs
bergundy: Got it

src/machines/child_workflow_state_machine.rs
bergundy: no because there's an activation just notifying that Child WF has started.
In the case of the WF above we do not generate any commands between started and complete.

src/machines/child_workflow_state_machine.rs
bergundy: See my comment above.

bergundy: It's totally valid for a WF not to produce any commands during an activation, one obvious case is for waiting for multiple things to resolve.

src/machines/mod.rs
bergundy: Will do

src/machines/mod.rs
bergundy: sgtm

src/prototype_rust_sdk/workflow_future.rs
bergundy: Same response as above

src/test_help/canned_histories.rs
bergundy: I copied this from the history right above it

tests/integ_tests/workflow_tests/continue_as_new.rs
bergundy: I was actually thinking of making `register_wf` return a type-safe function that you can call.
If you don't mind I'll open an issue to unblock this PR.

bergundy (commented):
Thanks!

src/machines/child_workflow_state_machine.rs
Sushisource: Right, yes, that makes sense. OK That sounds good to me

src/machines/child_workflow_state_machine.rs
Sushisource: Add a little comment that explains this please, otherwise makes  sense

src/test_help/canned_histories.rs
Sushisource: Yeah not all of them have been upgraded and for a few it's actually important that they are that way, but generally the histories being complete is nice

tests/integ_tests/workflow_tests/continue_as_new.rs
Sushisource: Sounds good

src/prototype_rust_sdk/workflow_future.rs
Sushisource: Yeah so this can't stay as-is then, clearly. It's still a nice thing to be able to detect, though. Maybe we can upgrade this to also check that there is nothing that needs to be unblocked. *then* it's a deadlock. If that's easy go for it, if not we can fix in another PR

Sushisource (approved):

PR #165, opened 2021-08-10
gRPC Automatic Retries

vitarb:
Added retry gateway that provides default exponential retry strategy for common retryable gRPC errors.


src/pollers/retry.rs
```diff
@@ -0,0 +1,334 @@
...
+
+impl<SG> RetryGateway<SG> {
+    pub fn new(gateway: SG) -> Self {
+        Self { gateway }
+    }
+}
+
+#[async_trait::async_trait]
+impl<SG: ServerGatewayApis + Send + Sync + 'static> ServerGatewayApis for RetryGateway<SG> {
+    async fn start_workflow(
```
Sushisource, line 83: This whole thing should really just be done with a macro, but we can do that in another PR like we talked about

src/pollers/default.rs
```diff
@@ -0,0 +1,13 @@
+/// Number of max retries that will be used in the retry gateway.
```
Sushisource, line 1: Move these into gateway options & possibly use builder as discussed

src/core_tests/retry.rs
```diff
@@ -0,0 +1,64 @@
...
+        Code::DataLoss,
+        Code::DeadlineExceeded,
+        Code::Internal,
+        Code::Unknown,
+        Code::ResourceExhausted,
+        Code::Aborted,
+        Code::OutOfRange,
+        Code::Unimplemented,
+        Code::Unavailable,
+        Code::Unauthenticated,
```
Sushisource, line 43: Probably lets store these in a const array and rather than matching using a bunch of `|` in the non-test code, just check if the code is inside the array. That way we don't need to repeat these lists and possibly have a mismatch in test/real code.

src/pollers/retry.rs
```diff
@@ -0,0 +1,334 @@
...
+impl ErrorHandler<tonic::Status> for TonicErrorHandler {
+    type OutError = tonic::Status;
+
+    fn handle(&mut self, current_attempt: usize, e: tonic::Status) -> RetryPolicy<tonic::Status> {
+        if current_attempt >= self.max_attempts {
+            return RetryPolicy::ForwardError(e);
+        }
+        match e.code() {
+            Code::Cancelled
+            | Code::DataLoss
```
Sushisource, line 55: Here is the part where you would check if code is in the const array

src/core_tests/retry.rs
vitarb: Done

src/pollers/retry.rs
vitarb: Yes, let's do it as a follow up, would be a good exercise for me to learn macros.

src/pollers/default.rs
vitarb: Done

src/pollers/gateway.rs
```diff
@@ -54,6 +59,9 @@ pub struct ServerGatewayOptions {
     /// attempt to use TLS when connecting to the Temporal server. Lang SDK is expected to pass any
     /// certs or keys as bytes, loading them from disk itself if needed.
     pub tls_cfg: Option<TlsConfig>,
+
+    /// Optional retry configuration for the server gateway. If not supplied defaults will be used.
+    pub retry_config: Option<RetryConfig>,
```
Sushisource, line 64: I'd say there's no reason to make this an `Option` -- it's trivial for the caller to use `Default` if they don't want to specify some or all of the specific fields.

src/pollers/gateway.rs
```diff
@@ -79,6 +87,50 @@ pub struct ClientTlsConfig {
     pub client_private_key: Vec<u8>,
 }
 
+#[derive(Clone, Debug)]
+pub struct RetryConfig {
+    /// initial wait time before the first retry.
+    pub initial_interval: Duration,
+    /// randomization jitter that is added or subtracted from the interval length.
```
Sushisource, line 94: Should explicitly state that this is used as a multiplier

src/pollers/gateway.rs
```diff
@@ -79,6 +87,50 @@ pub struct ClientTlsConfig {
...
+    pub randomization_factor: f64,
+    /// rate at which retry time should be increased, until it reaches max_interval.
+    pub multiplier: f64,
+    /// maximum amount of time to wait between retries.
+    pub max_interval: Duration,
+    /// maximum total amount of time requests should be retried for.
+    pub max_elapsed_time: Option<Duration>,
```
Sushisource, line 101: Should say what it means for this to be `None` - presumably infinity

src/pollers/gateway.rs
vitarb: The counter argument would be that it's simpler to set None and then it's as simple for us to default it down the line.

src/pollers/gateway.rs
vitarb: Done

src/pollers/gateway.rs
vitarb: Done

src/pollers/gateway.rs
Sushisource: I'd avoid it. It's not a standard practice, and it results in dumb things happening with more than one layer. EX: if you apply default to *this* struct the field will be `None`. It would make more sense for it to be populated with the defaults.

Sushisource (approved):

PR #173, opened 2021-08-16
Use incremental sequence number as ID for operations between lang and core

### LINKED ISSUE #164: [Bug] Don't use business IDs for correlation in operations between lang-core
We should switch to using an incremental sequence number for correlation IDs between core and lang

bergundy:
Closes #164 

@Sushisource I changed a lot in the prototype rust SDK and tests, I hope you like the new direction.

### DIFF
--- a/src/machines/patch_state_machine.rs
+++ b/src/machines/patch_state_machine.rs
@@ -226,15 +226,20 @@ mod tests {
             coresdk::{
                 common::decode_change_marker_details,
                 workflow_activation::{wf_activation_job, NotifyHasPatch, WfActivationJob},
-                workflow_commands::StartTimer,
             },
             temporal::api::command::v1::{
-                command::Attributes, RecordMarkerCommandAttributes, StartTimerCommandAttributes,
+                command::Attributes, RecordMarkerCommandAttributes,
+                ScheduleActivityTaskCommandAttributes,
             },
+            temporal::api::common::v1::ActivityType,
             temporal::api::enums::v1::{CommandType, EventType},
-            temporal::api::history::v1::{history_event, TimerFiredEventAttributes},
+            temporal::api::history::v1::{
+                history_event, ActivityTaskCompletedEventAttributes,
+                ActivityTaskScheduledEventAttributes, ActivityTaskStartedEventAttributes,
+                TimerFiredEventAttributes,
+            },
         },
-        prototype_rust_sdk::{WfContext, WorkflowFunction},
+        prototype_rust_sdk::{ActivityOptions, WfContext, WorkflowFunction},
         test_help::TestHistoryBuilder,
         workflow::managed_wf::ManagedWFFunc,
     };
@@ -249,18 +254,21 @@ mod tests {
         NoMarker,
     }
 
+    const ONE_SECOND: Duration = Duration::from_secs(1);
+
     /// EVENT_TYPE_WORKFLOW_EXECUTION_STARTED
     /// EVENT_TYPE_WORKFLOW_TASK_SCHEDULED
     /// EVENT_TYPE_WORKFLOW_TASK_STARTED
     /// EVENT_TYPE_WORKFLOW_TASK_COMPLETED
     /// EVENT_TYPE_MARKER_RECORDED (depending on marker_type)
-    /// EVENT_TYPE_TIMER_STARTED
-    /// EVENT_TYPE_TIMER_FIRED
+    /// EVENT_TYPE_ACTIVITY_TASK_SCHEDULED
+    /// EVENT_TYPE_ACTIVITY_TASK_STARTED
+    /// EVENT_TYPE_ACTIVITY_TASK_COMPLETED
     /// EVENT_TYPE_WORKFLOW_TASK_SCHEDULED
     /// EVENT_TYPE_WORKFLOW_TASK_STARTED
     /// EVENT_TYPE_WORKFLOW_TASK_COMPLETED
     /// EVENT_TYPE_WORKFLOW_EXECUTION_COMPLETED
-    fn change_marker_single_timer(marker_type: MarkerType, timer_id: &str) -> TestHistoryBuilder {
+    fn patch_marker_single_activity(marker_type: MarkerType) -> TestHistoryBuilder {
         let mut t = TestHistoryBuilder::default();
         t.add_by_type(EventType::WorkflowExecutionStarted);
         t.add_full_wf_task();
@@ -269,13 +277,41 @@ mod tests {
             MarkerType::NotDeprecated => t.add_has_change_marker(MY_PATCH_ID, false),
             MarkerType::NoMarker => {}
         };
-        let timer_started_event_id = t.add_get_event_id(EventType::TimerStarted, None);
+
+        let scheduled_event_id = t.add_get_event_id(
+            EventType::ActivityTaskScheduled,
+            Some(
+                history_event::Attributes::ActivityTaskScheduledEventAttributes(
+                    ActivityTaskScheduledEventAttributes {
+                        activity_id: "0".to_string(),
+                        activity_type: Some(ActivityType {
+                            name: "".to_string(),
+                        }),
+                        ..Default::default()
+                    },
+                ),
+            ),
+        );
+        let started_event_id = t.add_get_event_id(
+            EventType::ActivityTaskStarted,
+            Some(
+                history_event::Attributes::ActivityTaskStartedEventAttributes(
+                    ActivityTaskStartedEventAttributes {
+                        scheduled_event_id,
+                        ..Default::default()
+                    },
+                ),
+            ),
+        );
         t.add(
-            EventType::TimerFired,
-            history_event::Attributes::TimerFiredEventAttributes(TimerFiredEventAttributes {
-                started_event_id: timer_started_event_id,
-                timer_id: timer_id.to_string(),
-            }),
+            EventType::ActivityTaskCompleted,
+            history_event::Attributes::ActivityTaskCompletedEventAttributes(
+                ActivityTaskCompletedEventAttributes {
+                    scheduled_event_id,
+                    started_event_id,
+                    ..Default::default()
+                },
+            ),
         );
         t.add_full_wf_task();
         t.add_workflow_execution_completed();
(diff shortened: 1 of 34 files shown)


src/prototype_rust_sdk.rs
```diff
@@ -178,21 +180,17 @@ impl TestRustWorker {
...
 #[derive(derive_more::From)]
 #[allow(clippy::large_enum_variant)]
 enum RustWfCmd {
     #[from(ignore)]
-    CancelTimer(String),
-    #[from(ignore)]
-    CancelActivity(String),
-    #[from(ignore)]
-    CancelChildWorkflow(String),
+    Cancel(CommandID),
```
Sushisource, line 193: I don't really love this. It allows the construction of cancel commands that shouldn't be allowed to exist, like later on where child workflow completes can't be cancelled.

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -25,6 +29,7 @@ pub struct WfContext {
     args: Vec<Payload>,
     am_cancelled: watch::Receiver<bool>,
     shared: Arc<RwLock<WfContextSharedData>>,
+    next_sequence_number: AtomicU32,
```
Sushisource, line 32: Why is this atomic? It's only ever accessed from inside the `impl` with `&mut` receivers

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -75,101 +81,97 @@ impl WfContext {
...
 
-    /// Cancel child workflow
-    pub fn cancel_child_workflow(&self, child_workflow_id: &str) {
-        self.send(RustWfCmd::CancelChildWorkflow(
-            child_workflow_id.to_string(),
-        ))
+    /// Cancel any cancellable operation
+    pub fn cancel<T: 'static>(&self, fut: &(impl Future<Output = T> + 'static)) {
+        let fut = fut as &dyn Any;
+        match fut.downcast_ref::<WFCommandFut<T>>() {
```
Sushisource, line 165: This shouldn't be this way. If the `fut` argument should only ever be a `WFCommandFut` then that can be enforced in the type signature.

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -218,21 +220,64 @@ impl WfContext {
...
 }
 
-impl Future for WFCommandFut {
-    type Output = UnblockEvent;
+impl<T> Future for WFCommandFut<T> {
+    type Output = T;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        self.result_rx.poll_unpin(cx).map(|x| x.unwrap())
+        // let mapper = self.mapper;
```
Sushisource, line 253: Leftover comment

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -218,21 +220,64 @@ impl WfContext {
     }
 }
 
-struct WFCommandFut {
+struct WFCommandFut<T> {
+    cancel_command_id: Option<CommandID>,
     result_rx: oneshot::Receiver<UnblockEvent>,
+    mapper: Box<dyn Fn(UnblockEvent) -> T + Send + Sync + 'static>,
```
Sushisource, line 226: Why move the mapping into here at all? This requires a Box now where one was not needed previously, and doesn't really seem to save much.

There's an alternative way to do this though that I think might be worth doing, where `Future` is not implemented for  `WFCommandFut<T>` but rather is implemented for `WFCommandFut<T> where T: From<UnblockEvent>` and then all we need to do is just implement `From` for the different output types. That way the mapping stuff moves out of each command instantiation and is nicely contained, and there's no boxing.

src/prototype_rust_sdk.rs
bergundy: We just make sure never to use that CommandID but I can make a narrower type with `impl From<CancellableID> for CommandID`

src/prototype_rust_sdk/workflow_context.rs
bergundy: I wasn't sure it was bound to one thread, I just realized it's only `&mut` receivers, thanks.

src/prototype_rust_sdk/workflow_context.rs
bergundy: I didn't want the other commands to return a `WFCommandFut`, it's not exported and I don't know if it should be.
WDYT?

src/prototype_rust_sdk/workflow_context.rs
bergundy: 👍

src/machines/patch_state_machine.rs
```diff
@@ -249,18 +254,21 @@ mod tests {
...
-    /// EVENT_TYPE_TIMER_FIRED
+    /// EVENT_TYPE_ACTIVITY_TASK_SCHEDULED
+    /// EVENT_TYPE_ACTIVITY_TASK_STARTED
+    /// EVENT_TYPE_ACTIVITY_TASK_COMPLETED
     /// EVENT_TYPE_WORKFLOW_TASK_SCHEDULED
     /// EVENT_TYPE_WORKFLOW_TASK_STARTED
     /// EVENT_TYPE_WORKFLOW_TASK_COMPLETED
     /// EVENT_TYPE_WORKFLOW_EXECUTION_COMPLETED
-    fn change_marker_single_timer(marker_type: MarkerType, timer_id: &str) -> TestHistoryBuilder {
+    fn patch_marker_single_activity(marker_type: MarkerType) -> TestHistoryBuilder {
```
Sushisource, line 271: Why change this to activity?

bergundy (commented):
Thanks, I'll fix what I can, not sure about exporting `WFCommandFut`.

src/machines/patch_state_machine.rs
```diff
@@ -570,43 +567,110 @@ mod tests {
...
+                    ),
+                ),
+            );
+            t.add(
+                EventType::ActivityTaskCompleted,
+                history_event::Attributes::ActivityTaskCompletedEventAttributes(
+                    ActivityTaskCompletedEventAttributes {
+                        scheduled_event_id,
+                        started_event_id,
+                        // TODO result: Some(Payloads { payloads: vec![Payload{ metadata: Default::default(), data: vec![] }] }),
```
Sushisource, line 601: Leftover todos here and below

src/machines/timer_state_machine.rs
```diff
@@ -362,30 +349,22 @@ mod test {
...
+            cmd_sink.timer(Duration::from_secs(5)).await;
             // Cancel the first timer after having waited on the second
-            cmd_sink.cancel_timer("cancel_timer");
+            cmd_sink.cancel(&cancel_timer_fut);
             cancel_timer_fut.await;
             Ok(().into())
         });
 
-        let t = canned_histories::cancel_timer("wait_timer", "cancel_timer");
+        let t = canned_histories::cancel_timer("1", "0");
```
Sushisource, line 367: I'm not necessarily saying we *need* to change this, but going from readable names to just numbers harms the readability of these tests a decent amount. It would be nice if we could still attach readable IDs to things with sequences somehow, but I understand it's possibly more work than it's worth.

src/prototype_rust_sdk/workflow_context.rs
Sushisource: I also don't love using the same sequence for all IDs. It's probably much easier for users to debug workflows against history if they see timers go 1,2,3,4 instead of 1,5,8 etc.

It's not the biggest deal, but also should be quite easy to do here, and hopefully in node too.

src/prototype_rust_sdk.rs
Sushisource: Yeah what I'm getting at is this line shouldn't need to exist:
https://github.com/temporalio/sdk-core/pull/173/files#diff-0343a18509c554e2669813790fbe84d92fdea82f88d373bdeb5d69576148aac4R225

src/prototype_rust_sdk/workflow_context.rs
Sushisource: I am fine with exposing it. Then we can just make a `.cancel` on the `WFCommandFut` itself too which is possibly more desirable.

Sushisource: Also desirable, combined with the generic changes to `WFCommandFut`, it would be possible to only expose `.cancel` on types that actually make sense to cancel

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -218,21 +186,188 @@ impl WfContext {
...
+    /// Cancel this Future
+    fn cancel(&self, cx: &WfContext);
+}
+
+struct WFCommandFut<T>
+where
+    T: From<UnblockEvent>,
+{
+    /// We need a field that references the template type T, this works
+    unused: PhantomData<T>,
```
Sushisource, line 201: Can start w/ leading underscore - also doc comment not really necessary here, pretty common pattern.

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -218,21 +186,188 @@ impl WfContext {
...
+                unused: PhantomData,
+            },
+            tx,
+        )
     }
 }
 
-impl Future for WFCommandFut {
-    type Output = UnblockEvent;
+impl<T> Unpin for WFCommandFut<T> where T: From<UnblockEvent> {}
```
Sushisource, line 223: I believe this should not be necessary, it should get auto-derived

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -218,21 +186,188 @@ impl WfContext {
...
-        self.result_rx.poll_unpin(cx).map(|x| x.unwrap())
+        self.result_rx.poll_unpin(cx).map(|x| x.unwrap().into())
+    }
+}
+
+impl<T> CancellableFuture<T> for WFCommandFut<T>
+where
+    T: From<UnblockEvent>,
+{
+    fn cancel(&self, cx: &WfContext) {
```
Sushisource, line 240: In principle, the futures could probably store a ref to the `WfContext` and then it doesn't need to be passed here, but We don't really have to bother with that right away

Sushisource (approved):
Thanks this looks good!

src/prototype_rust_sdk/workflow_context.rs
bergundy: let do that in the ... future 😉

src/prototype_rust_sdk/workflow_context.rs
bergundy: It became necessary, complier was complaining, I think it's the phantom data.

src/prototype_rust_sdk/workflow_context.rs
bergundy: Okay, I'll change

PR #177, opened 2021-08-20
Signal external workflow machine implementation

### LINKED ISSUE #174: Implement signal external workflow state machine

Sushisource:
## What was changed
Implement signal external workflow (and child) machine and associated testing

## Why?
Signals!

## Checklist

1. Closes #174

2. How was this tested:
New tests

3. Any docs updates needed?


src/machines/workflow_machines.rs
```diff
@@ -668,14 +674,63 @@ impl WorkflowMachines {
...
+                                    WorkflowExecution {
+                                        namespace: child_machine.namespace().to_string(),
+                                        workflow_id: child_machine.workflow_id().to_string(),
+                                        run_id: "".to_string(),
+                                    },
+                                    true,
+                                )
+                            } else {
+                                todo!("Return signal failed")
+                            }
```
Sushisource, line 715: I sorta hate this downcasting. It can be avoided if lang doesn't do the whole sequence number thing for child workflows and instead just passes the namespace/wfid and says "I think it's my child". That would be more like what the other SDKs do.

This is potentially slightly more correct in that we can prevent things we know won't work ahead of time, but it introduces this whole new downcasting path and can fail in it's own interesting ways...

@bergundy WDYT? Should we keep this or just change the command from lang?

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -335,38 +390,76 @@ impl ChildWorkflowOptions {
...
+    pub status: ChildWorkflowStartStatus,
+}
+
+impl PendingChildWorkflow {
+    /// Returns `None` if the child did not start successfully. The returned [StartedChildWorkflow]
+    /// can be used to wait on, signal, or cancel the child workflow.
+    pub fn as_started(&self) -> Option<StartedChildWorkflow> {
+        match &self.status {
+            ChildWorkflowStartStatus::Succeeded(s) => Some(StartedChildWorkflow {
+                child_wf_cmd_seq_num: self.child_wf_cmd_seq_num,
```
bergundy, line 406: Why does the WF code need to be aware of sequence numbers?

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -335,38 +390,76 @@ impl ChildWorkflowOptions {
...
-    ) -> impl CancellableFuture<ChildWorkflowStartStatus> {
-        let seq = cx.next_child_workflow_sequence_number;
+    pub fn start(&mut self, cx: &mut WfContext) -> impl CancellableFuture<PendingChildWorkflow> {
+        let child_seq = cx.next_child_workflow_sequence_number;
         cx.next_child_workflow_sequence_number += 1;
-        self.sequence_number = Some(seq);
-        let (cmd, unblocker) = WFCommandFut::new(CancellableID::ChildWorkflow(seq));
+        let (cmd, unblocker) = WFCommandFut::new(CancellableID::ChildWorkflow {
+            child_seq,
+            cancel_cmd_seq: None,
```
bergundy, line 426: When would this be `Some(u32)`?

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -335,38 +390,76 @@ impl ChildWorkflowOptions {
...
             .into(),
         );
         cmd
     }
+}
 
-    /// Wait on completion of a child workflow execution started with `start_child_workflow`.
-    /// The returned Future is cancellable.
+impl StartedChildWorkflow {
+    /// Create a future that will wait until completion of this child workflow execution
```
bergundy, line 440: We might want to add an `execute` method too that combines `start` and `result` like we have in the other SDKs.

src/prototype_rust_sdk/workflow_future.rs
```diff
@@ -214,22 +233,30 @@ impl Future for WorkflowFuture {
...
                                 );
                             }
-                            CancellableID::ChildWorkflow(seq) => {
-                                let start_request = self
-                                    .child_workflow_starts
-                                    .get(&seq)
-                                    .expect("Tried to cancel unknown child");
+                            CancellableID::ChildWorkflow {
+                                cancel_cmd_seq,
+                                child_seq,
```
bergundy, line 238: I don't really understand the logic here, we want to cancel the child, why don't we use `child_seq`?

src/test_help/canned_histories.rs
```diff
@@ -1368,21 +1368,7 @@ pub fn activity_double_resolve_repro() -> TestHistoryBuilder {
...
-///  7: EVENT_TYPE_WORKFLOW_TASK_SCHEDULED
-///  8: EVENT_TYPE_WORKFLOW_TASK_STARTED
-///  9: EVENT_TYPE_WORKFLOW_TASK_COMPLETED
-/// 10: EVENT_TYPE_CHILD_WORKFLOW_EXECUTION_COMPLETED
-/// 11: EVENT_TYPE_WORKFLOW_TASK_SCHEDULED
-/// 12: EVENT_TYPE_WORKFLOW_TASK_STARTED
-/// 13: EVENT_TYPE_WORKFLOW_TASK_COMPLETED
-/// 14: EVENT_TYPE_WORKFLOW_EXECUTION_COMPLETED
-pub fn single_child_workflow(child_wf_id: &str) -> TestHistoryBuilder {
+fn start_child_wf_preamble(child_wf_id: &str) -> (TestHistoryBuilder, i64, i64) {
```
bergundy, line 1371: No docstring?

tests/integ_tests/workflow_tests/signals.rs
```diff
@@ -0,0 +1,105 @@
...
+async fn signal_sender(mut ctx: WfContext) -> WorkflowResult<()> {
+    let run_id = std::str::from_utf8(&ctx.get_args()[0].data)
+        .unwrap()
+        .to_owned();
+    let sigres = ctx
+        .signal_workflow(RECEIVER_WFID, run_id, SIGNAME, b"hi!")
+        .await;
+    if ctx.get_args().get(1).is_some() {
+        // We expect failure
+        assert!(sigres.is_err());
```
bergundy, line 18: `unwrap_err` or `expect_err`?

tests/integ_tests/workflow_tests/signals.rs
```diff
@@ -0,0 +1,105 @@
...
+    let wf_name = "sends_signal_to_missing_wf";
+    let mut starter = CoreWfStarter::new(wf_name);
+    let worker = starter.worker().await;
+    worker.register_wf(wf_name.to_owned(), signal_sender);
+
+    worker
+        .submit_wf(
+            wf_name,
+            wf_name,
+            vec![Uuid::new_v4().to_string().into(), [1].into()],
```
bergundy, line 36:
```suggestion
            vec![Uuid::new_v4().to_string().into(), EXPECT_FAILURE_ARGS],
```

tests/integ_tests/workflow_tests/signals.rs
```diff
@@ -0,0 +1,105 @@
...
+
+#[tokio::test]
+async fn sends_signal_to_other_wf() {
+    let mut starter = CoreWfStarter::new("sends_signal_to_other_wf");
+    let worker = starter.worker().await;
+    worker.register_wf("sender", signal_sender);
+    worker.register_wf("receiver", signal_receiver);
+
+    let receiver_run_id = worker
+        .submit_wf(RECEIVER_WFID, "receiver", vec![])
```
bergundy, line 57:
```suggestion
        .submit_wf(RECEIVER_WFID, "receiver", EXPECT_SUCCESS_ARGS)
```

bergundy (commented):
This is mostly there, left some comments.

src/machines/workflow_machines.rs
Sushisource: Yeah but we don't need to downcast in that case because the generic `cancel` method exists

protos/local/common.proto
```diff
@@ -17,8 +17,12 @@ message Payload {
 
 // Identifying information about a particular workflow execution
 message WorkflowExecution {
-    string workflow_id = 1;
-    string run_id = 2;
+    // The namespace the workflow resides in
+    string namespace = 1;
```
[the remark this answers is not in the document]
Sushisource: It just turns out that every single place it gets used always requires a namespace as well... so seemed reasonable to combine things. When you think about it a workflow execution sorta needs a namespace to truly be defined accurately.

That said I can undo this just to keep things consistent

src/machines/workflow_machines.rs
Sushisource: I will undo this as we discussed

src/machines/signal_external_state_machine.rs
```diff
@@ -1,23 +1,95 @@
...
+    workflow_execution: WorkflowExecution,
+    signal_name: String,
+    payloads: impl IntoIterator<Item = Payload>,
+    only_child: bool,
+) -> NewMachineWithCommand<SignalExternalMachine> {
+    let mut s = SignalExternalMachine {
+        state: Created {}.into(),
+        shared_state: SharedState { seq },
+    };
+    OnEventWrapper::on_event_mut(&mut s, SignalExternalMachineEvents::Schedule)
```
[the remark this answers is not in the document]
Sushisource: There's a note somewhere about maybe doing that for *all* the machines b/c a lot are like this. If we do, let's just do it all at once

src/machines/signal_external_state_machine.rs
```diff
@@ -57,14 +129,214 @@ impl SignalExternalCommandRecorded {
...
+            SignalExternalCommand::Failed(f) => {
+                vec![ResolveSignalExternalWorkflow {
+                    seq: self.shared_state.seq,
+                    // TODO: Create new failure type upstream for this
+                    failure: Some(Failure {
+                        message: "Unable to signal external workflow because it was not found"
+                            .to_string(),
+                        failure_info: Some(FailureInfo::ApplicationFailureInfo(
+                            ApplicationFailureInfo {
+                                r#type: f.to_string(),
```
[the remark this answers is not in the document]
Sushisource: This is what Java does, but, this whole thing is kinda silly. As in the TODO we should just make a new failure type for it upstream and then use that

src/machines/signal_external_state_machine.rs
```diff
@@ -57,14 +129,214 @@ impl SignalExternalCommandRecorded {
...
+                    failure: None,
+                }
+                .into()]
+            }
+            SignalExternalCommand::Failed(f) => {
+                vec![ResolveSignalExternalWorkflow {
+                    seq: self.shared_state.seq,
+                    // TODO: Create new failure type upstream for this
+                    failure: Some(Failure {
+                        message: "Unable to signal external workflow because it was not found"
```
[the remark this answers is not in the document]
Sushisource: The only failure cause in `SignalExternalWorkflowExecutionFailedCause` is not found, so for now that's really all we can say

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -176,11 +186,63 @@ impl WfContext {
         res
     }
 
+    /// Send a signal to an external workflow. May resolve as a failure if the signal didn't work
+    /// or was cancelled.
+    pub fn signal_workflow(
+        &mut self,
+        workflow_id: impl Into<String>,
+        run_id: impl Into<String>,
```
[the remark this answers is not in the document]
Sushisource: You just pass an empty string in that case, which is all that ends up happening in the proto anyway. Long run when we have a public Rust SDK I think we'll in general consider what the best way to pass args is

src/prototype_rust_sdk/workflow_context.rs
Sushisource: The WF code itself isn't, this is inside the Rust "sdk"

src/prototype_rust_sdk/workflow_future.rs
Sushisource: I will address all this with the cancel external wf machine, which will end up looking a lot like this one.

tests/integ_tests/workflow_tests/signals.rs
Sushisource: Your two comments are actually for the two different types of workflows -- just the presence of a second argument in the sender indicates failure. Lifting these out just feels more awkward than not to me

src/machines/signal_external_state_machine.rs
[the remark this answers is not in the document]
bergundy: I think I made a similar assumption somewhere in the child WF machine.
Can we match the failure cause reported by the server in case another cause is added?

src/machines/signal_external_state_machine.rs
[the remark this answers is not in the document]
bergundy: I see, okay, let's keep this for now

src/prototype_rust_sdk/workflow_context.rs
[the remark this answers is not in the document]
bergundy: Empty string is not that convenient but it doesn't matter that much for the prototype SDK.

tests/integ_tests/workflow_tests/signals.rs
bergundy: It helps with readability.

bergundy (approved):
I don't think there are any more blockers here, feel free to merge after going over the last comments.

src/prototype_rust_sdk/workflow_future.rs
```diff
@@ -303,16 +352,20 @@ impl Future for WorkflowFuture {
...
+                        // registered
+                        if let Some(SigChanOrBuffer::Buffer(buf)) = self.sig_chans.remove(&signame)
+                        {
+                            for input in buf {
+                                let _ = chan.send(input);
+                            }
+                            // Re-poll wf future since signals may be unblocked
+                            res = self.inner.poll_unpin(cx);
+                        }
+                        self.sig_chans.insert(signame, SigChanOrBuffer::Chan(chan));
```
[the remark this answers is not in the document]
Sushisource: It can't be done at that point unfortunately. It needs to send the send half of the channel to the context where it can be stored, but that can't happen until the underlying workflow code is polled, but we don't want to do that yet. It's a similar problem to what we had in node.

My expectation for the Rust SDK is this will go away by registering signals along with the workflow code statically, by using a struct or something. Dynamically registered signals will maybe still be a reason to have this though.

PR #179, opened 2021-08-23
Revise worker shutdown

bergundy:


src/worker/mod.rs
```diff
@@ -139,39 +140,45 @@ impl Worker {
...
+        self.wf_task_poll_buffer.notify_shutdown();
+        // Notify in case shutdown was requested while there were no more outstanding WFTs.
+        // This is required because the only other place where we notify wfts_drained is on
+        // activation completion and activation polling checks for wfts_drained.
+        self.maybe_notify_wtfs_drained();
+        // wait until all outstanding workflow tasks have been completed before shutting down
+        if !*self.wfts_drained.borrow() {
+            self.wfts_drained
+                .clone()
+                .changed()
```
Sushisource, line 179: I realized while I was looking at some of this that this should probably be moved to `shutdown_complete`, since it waits all this time. That does slightly change some of the existing semantics though where workers will end up gone instead of returning `Shutdown`. 

I don't suuuper care if we move it, or if we maybe just update the docstring for this function, but we should do one of the two.

Sushisource (approved):

Sushisource:
@bergundy Merging this since I want it in for the signal machine work. I'll just do the rename in that branch as well.

PR #182, opened 2021-08-24
Cancel external workflow machines

### LINKED ISSUE #175: Implement cancel external workflow state machine

Sushisource:
## What was changed
Allows cancelling external workflows from within a workflow, including started children.

## Why?
Feature parity

## Checklist

1. Closes #175 
2. How was this tested:
New tests

3. Any docs updates needed?


src/machines/child_workflow_state_machine.rs
```diff
@@ -241,30 +241,25 @@ impl Started {
         self,
         state: SharedState,
     ) -> ChildWorkflowMachineTransition<Cancelled> {
-        match state.cancellation_type {
-            ChildWorkflowCancellationType::Abandon => {
-                ChildWorkflowMachineTransition::ok(vec![], Cancelled::default())
-            }
```
Sushisource, line 247: @bergundy Regarding this change, none of these cancel types can be handled within the child workflow machine.

Essentially they get handled at the lang SDK level (at least, how it works in Java now):
https://github.com/temporalio/sdk-java/blob/374014922d42b792a180ff5c8064e19819db0dc6/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java#L503

However, we could just upgrade the child workflow machine to deal with post-start cancels. It would eliminate lang having to deal with this logic, and it would simplify the cancellation of children from needing to decide if they should send the `CancelUnstartedChildWorkflowExecution` or the `RequestCancelExternalWorkflowExecution` command to just *always* sending a `CancelChildWorkflowExecution` command that always takes the seq number regardless of if the child is started or not.

LMK what you think about this. Maybe we do it another PR too. I think overall it would be a substantial simplification, but it'll take the better part of a day I think.

tests/integ_tests/workflow_tests/cancel_external.rs
```diff
@@ -0,0 +1,64 @@
...
+        assert!(sigres.is_err());
+    } else {
+        sigres.unwrap();
+    }
+    Ok(().into())
+}
+
+async fn cancel_receiver(mut ctx: WfContext) -> WorkflowResult<()> {
+    let cancelled = tokio::select! {
+        _ = ctx.timer(Duration::from_secs(500)) => false,
```
bergundy, line 31: Do we even need this timer?
Can't we just block on `ctx.cancelled`?

bergundy (commented):
Overall LGTM.
I want to try and use it from node and verify we didn't miss anything.

bergundy:
@Sushisource after pulling the branch I see that `coresdk.activity_task.Start.workflow_namespace` was removed in the previous PR.
Please readd it here.

src/machines/signal_external_state_machine.rs
```diff
@@ -68,7 +68,7 @@ pub(super) fn new_external_signal(
         shared_state: SharedState { seq },
     };
     OnEventWrapper::on_event_mut(&mut s, SignalExternalMachineEvents::Schedule)
```
bergundy, line 70: Don't know where to put this comment, when cancelling the signal we might want to resolve the signal with failure.

Issue #185, opened 2021-08-27
Separate protos into own crate & potentially own repo


Sushisource:
At minimum we need a separate crate for the protos.

In addition, it possibly makes sense to have the protos themselves in a separate repo, and possibly the node package there.

PR #186, opened 2021-08-30
Protos moved to own crate

Sushisource:
## What was changed
Generated protobuf code (& associated helpers that are reasonable to publicly expose) moved to own crate

## Why?
Better organization, makes it easier to depend on just interface types for other crates.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/185

2. How was this tested:
Existing tests

3. Any docs updates needed?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,10 +15,10 @@ categories = ["development-tools"]
 # TODO: Feature flag opentelem stuff
 [dependencies]
 anyhow = "1.0"
-async-trait = "0.1"
 arc-swap = "1.3"
-base64 = "0.13"
+async-trait = "0.1"
 backoff = "0.3.0"
+base64 = "0.13"
 crossbeam = "0.8"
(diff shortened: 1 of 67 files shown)


bergundy (approved):
Cool, should we move this crate along with the proto definitions to a new repo?

Sushisource:
> Cool, should we move this crate along with the proto definitions to a new repo?

I think the convenience of keeping them in while we're still changing them every so often is pretty desirable. Even if we moved the protos themselves out, I'd keep the crate in here.

That said, for now I think it's fully reasonable to just use a subtree for the node package and you can bring in just the `protos` dir. Sound good?

bergundy:
@Sushisource from my experience using subtrees isn't really convenient, I'd rather just have 2 submodules (for now).
Anyways, I'm fine with with only moving the proto definitions to a new repo.
Do you have any objections if I do that soon?

Issue #85, opened 2021-04-07
Unifying Logging / Metrics w/ OpenTelemetry between Core & Lang


Sushisource:
We've discussed the desire to unify logging across core/lang.

Right now, core uses https://github.com/tokio-rs/tracing to simultaneously perform traditional logging functions, while also generating OpenTelemetry traces.

When we consider unifying these concerns across the language barrier, we really want to do two things:

1. Have consistent "traditional" logging output - IE: What you'd see in the console / a log file.
2. Pass tracing context across the language barrier

For 1, one option (possibly better option described below) is to expose a `log` function from core which takes a `level` parameter, a `span` parameter, a `message` parameter, and optionally a series of key/value pairs to be logged (all logging is structured). Then, lang side has to translate it's logging interface to the core logging interface in whatever way feels most natural/idiomatic and pass logging through to core. The core implementation of that function then constructs an https://tracing-rs.netlify.app/tracing/struct.event corresponding to the passed in parameters.

If there's something more automatic we could do here I'm all ears but nothing comes to mind.

For 2, is seems best for `poll_workflow_task` and `poll_activity_task` to return a `Span` as defined here https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L57 by OpenTelemetry. Then, lang can convert that into the `Span` type that the tracing library it has chosen to use uses, and proceed from there to pass it around and create new child spans, etc. It would then pass it / such spans back into the logging interface described above as the `span` parameter.

While writing the above, I realized there probably is a more elegant option here. The core side can provide an OpenTelemetry "Collector" https://opentelemetry.io/docs/concepts/data-collection/ which the lang side will need to provide an implementation for. Then, ideally, lang simply uses it's opentelem library, registering the custom collector which is core. On the core side, the collector is just a pass through to tokio tracing. Lang would be expected to disable all output, relying on core to collect the logs/traces and output them as console/file/tracing etc. This is more work up front, and possibly per-lang, but also feels substantially less hacky. Unfortunately, I think the spans from 2 still need to be passed into lang so that what it's doing can be associated with certain workflow/activity tasks.

mfateev:
I like the approach where lang can set a sink for log and any other telemetry. This way lang can fully control how the log output is filtered and reported. If the sink is not set the core can use default output which may be useful to troubleshoot the lang side.

bergundy:
[...]

PR #187, opened 2021-09-01
Export OTel Collector data to Lang

Sushisource:
## What was changed
When otel is enabled, now we turn on the OTel Collector exporter, but *hah*, we don't actually need a collector, because langs will implement the frontend of the collector, passing along spans and metrics to user-registered exporters.

A lot of lifting to avoid running another process, but nice.

## Why?
Enabling unified telemetry across core/lang.

## Checklist
1. Part of #85 

2. How was this tested:
Verified spans are exported to lang.

TODO:
- [ ] Metrics


bergundy (commented):
Overall LGTM,

I'd really rather avoid taking up a port and instead send spans over a channel, can we put more effort into that?

src/telemetry/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+    global, sdk::trace::TracerProvider, trace::TracerProvider as TracerProviderTrait, KeyValue,
+};
+use std::{collections::VecDeque, sync::Once};
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt, EnvFilter};
+
+const TELEM_SERVICE_NAME: &str = "temporal-core-sdk";
+const ENABLE_OPENTELEM_ENV_VAR: &str = "TEMPORAL_ENABLE_OPENTELEMETRY";
+static TRACING_INIT: Once = Once::new();
+
+/// Initialize tracing subscribers and output. Core will not call this itself, it exists here so
```
vitarb, line 18: Why don't we pass all necessary parameters into core init and initialize tracing from there?

Sushisource:
Closing in favor of https://github.com/temporalio/sdk-core/pull/188

PR #188, opened 2021-09-04
Otel collector export

Sushisource:
In contrast to https://github.com/temporalio/sdk-core/pull/187, simply just export spans/metrics to an OTel collector. Users will have to run one, but otherwise this is *dramatically* simpler and more maintainable.

I strongly recommend we go with this approach.


src/telemetry/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+use opentelemetry::sdk::trace::Config;
+use opentelemetry::sdk::Resource;
+use opentelemetry::{global, sdk::metrics::PushController, util::tokio_interval_stream, KeyValue};
+use opentelemetry_otlp::WithExportConfig;
+use std::collections::VecDeque;
+use std::time::Duration;
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt, EnvFilter};
+
+const TELEM_SERVICE_NAME: &str = "temporal-core-sdk";
+const ENABLE_OPENTELEM_ENV_VAR: &str = "TEMPORAL_ENABLE_OPENTELEMETRY";
```
bergundy, line 14: Let's avoid requiring env vars to configure core, maybe only run this with config=test?

src/telemetry/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+            let tracer_cfg = Config::default().with_resource(Resource::new(vec![KeyValue::new(
+                "service.name",
+                TELEM_SERVICE_NAME,
+            )]));
+            let tracer = opentelemetry_otlp::new_pipeline()
+                .tracing()
+                .with_exporter(
+                    opentelemetry_otlp::new_exporter()
+                        .tonic()
+                        .with_endpoint("http://localhost:4317"),
```
bergundy, line 53: This shouldn't be hardcoded

src/telemetry/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+            let opentelemetry = tracing_opentelemetry::layer().with_tracer(tracer);
+
+            let metrics = opentelemetry_otlp::new_pipeline()
+                .metrics(tokio::spawn, |dur| tokio_interval_stream(dur))
+                .with_period(Duration::from_secs(1))
+                .with_exporter(
+                    // No joke exporter builder literally not cloneable for some insane reason
+                    opentelemetry_otlp::new_exporter()
+                        .tonic()
+                        .with_endpoint("http://localhost:4317"),
```
bergundy, line 68: same as above

bergundy (commented):
Looks like this turned out to be so much simpler.
Let's definitely go with this direction.
otel options should be provided by lang and we should avoid using env vars.

src/telemetry/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+/// Initialize tracing subscribers and output. Core [crate::init] calls this, but it may be called
+/// separately so that tests may choose to initialize tracing differently. If this function is
+/// called more than once, subsequent calls do nothing.
+///
+/// See [TelemetryOptions] docs for more on configuration.
+pub(crate) fn telemetry_init(opts: &TelemetryOptions) -> Result<(), anyhow::Error> {
+    TRACING_INIT.get_or_init(|| {
+        let mut globaldat = GlobalTracingDats::default();
+
+        let filter_layer = EnvFilter::try_from_env(LOG_FILTER_ENV_VAR)
```
bergundy, line 57: I'm still not sure I'd put the env var here, it should only be used for tests IMHO.

src/worker/mod.rs
```diff
@@ -277,16 +282,32 @@ impl Worker {
...
-                Some(work) => match self.apply_server_work(work).await? {
-                    NewWfTaskOutcome::IssueActivation(a) => return Ok(a),
-                    NewWfTaskOutcome::TaskBuffered => {
-                        // If the task was buffered, it's not actually outstanding, so we can
-                        // immediately return a permit.
-                        self.return_workflow_task_permit();
+                Some(work) => {
+                    WORKFLOW_TASK_QUEUE_POLL_SUCCEED_COUNTER.add(
+                        1,
+                        &[KeyValue::new(
```
bergundy, line 288: Is the task queue relevant to this metric?

bergundy (approved):
Awesome! I had some non-blocking comments.

src/telemetry/mod.rs
Sushisource: There's no harm and they can be much easier to deal with operationally sometimes when you need to quickly change something. Having the option is nice

src/worker/mod.rs
Sushisource: Yeah, but these are here just as an example. I'm going to flesh them out next.

src/telemetry/mod.rs
bergundy: I'd leave that up to the user to map env vars to config, everybody has their own configuration management systems and I wouldn't enforce a specific pattern.
We don't use env vars anywhere else in the core / node SDKs so this is a bit of a special case.

PR #189, opened 2021-09-09
Ability to export logs to lang

Sushisource:
## What was changed
Adds a facility to export logs to lang so that they can show up with the user's configured loggers. Loses some context from tracing but overall still useful.

## Why?
Easy-to-enable debuggability for users

Relates to https://github.com/temporalio/sdk-core/issues/85

### DIFF
--- a/src/log_export.rs
+++ b/src/log_export.rs
@@ -1,0 +1,40 @@
+use log::{Level, Log, Metadata, Record};
+use std::{collections::VecDeque, sync::Mutex, time::Instant};
+
+/// A log line (which ultimately came from a tracing event) exported from Core->Lang
+#[derive(Debug)]
+pub struct CoreLog {
+    message: String,
+    timestamp: Instant,
+    level: Level,
+    // KV pairs aren't meaningfully exposed yet to the log interface by tracing
+}
+
+#[derive(Default)]
+pub(crate) struct CoreExportLogger {
+    records: Mutex<VecDeque<CoreLog>>,
+}
+
+impl Log for CoreExportLogger {
+    fn enabled(&self, _metadata: &Metadata) -> bool {
+        // TODO: Disable in test only mode where no lang half
+        true
+    }
+
+    fn log(&self, record: &Record) {
+        let clog = CoreLog {
+            message: format!("{}", record.args()),
+            timestamp: Instant::now(),
+            level: record.level(),
+        };
+        dbg!(&clog);
+        self.records
+            .lock()
+            .expect("Logging mutex must be acquired")
+            .push_back(clog)
+    }
+
+    fn flush(&self) {
+        // Does nothing
+    }
+}
--- a/src/log_export.rs
+++ b/src/log_export.rs
@@ -1,6 +1,6 @@
 use log::{Level, LevelFilter, Log, Metadata, Record};
+use ringbuf::{Consumer, Producer, RingBuffer};
 use std::{
-    collections::VecDeque,
     sync::Mutex,
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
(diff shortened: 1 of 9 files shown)


src/telemetry/mod.rs
```diff
@@ -28,21 +31,31 @@ pub struct TelemetryOptions {
...
     /// variable.
     pub tracing_filter: String,
+    /// Core can forward logs to lang for them to be rendered by the user's logging facility.
+    /// The logs are somewhat contextually lacking, but still useful in a local test situation when
+    /// running only one workflow at a time. This sets the level at which they are filtered.
+    /// TRACE level will export span start/stop info as well.
+    ///
+    /// Default is INFO. If set to `Off`, the mechanism is disabled entirely (saves on perf).
+    /// If set to anything besides `Off`, any console output directly from core is disabled.
+    pub log_forwarding_level: LevelFilter,
```
bergundy, line 41: So basically to turn off logging completely we need to set `log_forwarding_level` to `Off` and `tracing_filter` to `"OFF"`?

src/telemetry/mod.rs
```diff
@@ -28,21 +31,31 @@ pub struct TelemetryOptions {
     /// telemetry or console output. May be overridden by the `TEMPORAL_TRACING_FILTER` env
```
bergundy, line 31:
```suggestion
    /// telemetry, forwarding to lang, or console output. May be overridden by the `TEMPORAL_TRACING_FILTER` env
```

src/telemetry/mod.rs
bergundy: I think we might want to have 2 filters instead and 1 boolean:
1. filter for otel collector
2. filter for output / forwarding
3. boolean toggle for forwarding / output directly to console

That'd be the most flexible, is it even possible?

src/log_export.rs
```diff
@@ -0,0 +1,74 @@
...
+
+pub(crate) struct CoreExportLogger {
+    records: Mutex<VecDeque<CoreLog>>,
+    level_filter: LevelFilter,
+}
+
+impl CoreExportLogger {
+    pub(crate) fn new(level: LevelFilter) -> Self {
+        Self {
+            records: Mutex::new(Default::default()),
```
bergundy, line 38: We mentioned limiting the capacity of the buffer, I think it'd be better to do that.

bergundy (approved):
This is almost good to merge, left a few comments, PTAL before merging.

src/telemetry/mod.rs
Sushisource: I don't think there's any reason to forward logs *and* have them go direct to console from core, you'll just get duplicated output. So with that in mind the two filters pretty much do what you need.

bergundy: But there's only 1 filter.
How do I configure ">=error to console and >=info to the collector"?

src/log_export.rs
```diff
@@ -28,24 +28,35 @@ impl CoreLog {
...
 pub(crate) struct CoreExportLogger {
-    records: Mutex<VecDeque<CoreLog>>,
+    logs_in: Mutex<Producer<CoreLog>>,
+    logs_out: Mutex<Consumer<CoreLog>>,
     level_filter: LevelFilter,
 }
 
 impl CoreExportLogger {
     pub(crate) fn new(level: LevelFilter) -> Self {
+        let (lin, lout) = RingBuffer::new(2048).split();
```
bergundy, line 38: This should be configurable

src/log_export.rs
```diff
@@ -64,10 +75,11 @@ impl Log for CoreExportLogger {
             timestamp: SystemTime::now(),
             level: record.level(),
         };
-        self.records
+        let _ = self
+            .logs_in
             .lock()
             .expect("Logging mutex must be acquired")
-            .push_back(clog)
+            .push(clog);
```
bergundy, line 82: There should be a way to know if we are losing messages due to slow consumer.

PR #192, opened 2021-09-15
Metrics parity (mostly) with existing SDKs

Sushisource:
## What was changed
Added metrics collection for most metrics that exist in Java/Go, excepting things that aren't implemented yet.

Some metrics are intentionally omitted:
* `unregistered_activity_invocation` - Can't be done in core unless we add a flag to activity failure result, which is possible and probably a reasonable thing to do.
* `activity_endtoend_latency` - Already redundant since it's essentially `activity_schedule_to_start_latency` + `activity_execution_latency`.
* `activity_task_error` - Doesn't make sense in core. The core/lang contract requires all exceptions to be reported as failures.
* `workflow_task_no_completion` - Not implemented in go. Seemingly redundant with `workflow_task_execution_failed`
* `corrupted_signals` - Cannot be handled in core w/o propagating info back from lang

Other differences:
* `poller_start`  changed to  `num_pollers` to become a gauge, which makes much more sense

TODO:
- [x] Clean up crazy dedicated thread situation

## Why?
Metrics!

## Checklist

1. Relates to https://github.com/temporalio/sdk-core/issues/85

2. How was this tested:
Viewing emitted metrics in prometheus / otel collector

3. Any docs updates needed?


Sushisource:
Draft while wrapping up remaining metrics

Sushisource:
This is missing metrics for gRPC calls, but that can be added in a subsequent PR.

src/pollers/poll_buffer.rs
```diff
@@ -56,6 +81,7 @@ where
                         sp = polls_requested.acquire() => sp.expect("Polls semaphore not dropped"),
                         _ = shutdown.changed() => continue,
                     };
+                    let _active_guard = ActiveCounter::new(ap.as_ref());
```
bergundy, line 84: My favorite Gillette product: "Active Guard"

src/telemetry/mod.rs
```diff
@@ -64,81 +67,126 @@ struct GlobalTelemDat {
...
+    //   env-filter. When it does, allow filtering logs/telemetry separately.
+
+    // This is a bit odd, but functional. It's desirable to create a separate tokio runtime for
+    // metrics handling, since tests typically use a single-threaded runtime and initializing
+    // pipeline requires us to know if the runtime is single or multithreaded, we will crash
+    // in one case or the other. There does not seem to be a way to tell from the current runtime
+    // handle if it is single or multithreaded. Additionally, we can isolate metrics work this
+    // way which is nice.
+    let opts = opts.clone();
+    std::thread::spawn(move || {
```
bergundy, line 80: What's supposed to kill this thread? I couldn't figure out from reading this code.

src/worker/mod.rs
```diff
@@ -75,43 +73,56 @@ pub struct Worker {
...
 }
 
 impl Worker {
     pub(crate) fn new(
         config: WorkerConfig,
         sticky_queue_name: Option<String>,
         sg: Arc<GatewayRef>,
+        metrics: MetricsContext,
     ) -> Self {
+        metrics.worker_registered();
```
bergundy, line 87: Should this include the task queue?

bergundy (approved):
LGTM.

Had 1 question about the telemetry thread.

src/telemetry/mod.rs
Sushisource: It's sorta weird. It immediately gets joined. The different thread context simply needs to exist so I can create a runtime within it, then the runtime manages itself.

src/worker/mod.rs
Sushisource: Already does from higher context. A lot of that is nonobvious though. There is a way I can fix it with clever typing but I didn't want to spend quite that much time initially. There is a note about how it can be done in the future if needed

PR #194, opened 2021-09-20
Update API protos subtree

bergundy:
## What was changed
In title

## Why?
I needed the latest versions of the API protos.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "protos/api_upstream"]
-	path = protos/api_upstream
-	url = git://github.com/temporalio/api.git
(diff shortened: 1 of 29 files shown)


Sushisource:
We should just change the docs to say subtree and include the instructions for updating them - submodules are comparatively a pain in the butt to update properly IMO. Subtrees on the other hand just work when someone pulls the repo. Unless we've got a compelling reason to switch I think I'd prefer to keep it as it

This is the right command for updating I believe:
`git subtree pull --prefix protos/api_upstream/ git://github.com/temporalio/api.git master --squash`

bergundy:
Subtrees are a different type of pain from my experience but I'm fine with using them here since we never edit them directly from this repo.
I was confused by the docs and didn't realize we were using subtree.

Sushisource:
I'll update the readme in https://github.com/temporalio/sdk-core/pull/195

Sushisource:
Whoops, didn't mean to close, we can use this to update the subtree

PR #195, opened 2021-09-20
Allow directly scraping prometheus metrics

Sushisource:
## What was changed
Expose an option to enable direct scraping of prometheus metrics.

## Why?
Many users may not care about exporting metrics to the OTel collector, and only use prometheus for metrics. This way they don't have to run the OTel collector if they don't care about traces.

## Checklist

1. Relates to https://github.com/temporalio/sdk-core/issues/85 

2. How was this tested:
Verified metrics are properly exposed on specified endpoint

3. Any docs updates needed?

### DIFF
--- a/.buildkite/docker/docker-compose.yaml
+++ b/.buildkite/docker/docker-compose.yaml
@@ -5,22 +5,20 @@ services:
     image: cassandra:3.11
     logging:
       driver: none
-    ports:
-      - "9042:9042"
+  #    ports:
+  #      - '9042:9042'
 
   temporal:
     image: temporalio/auto-setup:1.11.0
-    logging:
-      driver: none
     ports:
       - "7233:7233"
       - "7234:7234"
-      - "7235:7235"
-      - "7239:7239"
-      - "6933:6933"
-      - "6934:6934"
-      - "6935:6935"
-      - "6939:6939"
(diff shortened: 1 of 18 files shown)


vitarb (approved):

etc/otel-collector-config.yaml
```diff
@@ -0,0 +1,36 @@
+receivers:
```
bergundy, line 1: I think we should reuse *this* config in `sdk-node` instead of duplicating it in both repos.

bergundy (approved):

etc/otel-collector-config.yaml
Sushisource: Yeah I'll do that after merging

.buildkite/docker/docker-compose.yaml
```diff
@@ -5,22 +5,20 @@ services:
...
     ports:
       - "7233:7233"
       - "7234:7234"
-      - "7235:7235"
-      - "7239:7239"
-      - "6933:6933"
-      - "6934:6934"
-      - "6935:6935"
-      - "6939:6939"
+#      - "7235:7235"
```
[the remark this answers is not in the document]
Sushisource: It's just a nice way to remember what exists

PR #197, opened 2021-09-23
Pass required client name/version headers on rpc calls

### LINKED ISSUE #76: Use lang SDK name+version as client name in GRPC metadata
_Originally posted by @bergundy in https://github.com/temporalio/sdk-core/pull/72#discussion_r601025981_

Sushisource:
* Also use builder pattern for big options structs

## What was changed
Pass needed headers, getting client name/version from lang

## Why?
So server can yell at us when we're too old.

## Checklist

- [x] When https://github.com/temporalio/temporal/pull/1949 is merged do the right thing for that too

1. Closes #76 

2. How was this tested:
Integ tests

3. Any docs updates needed?

### DIFF
--- a/src/pollers/gateway.rs
+++ b/src/pollers/gateway.rs
@@ -43,7 +43,8 @@ use std::str::FromStr;
 use tonic::metadata::{MetadataKey, MetadataValue};
 
 static LONG_POLL_METHOD_NAMES: [&str; 2] = ["PollWorkflowTaskQueue", "PollActivityTaskQueue"];
-const LONG_POLL_TIMEOUT: Duration = Duration::from_secs(60);
+/// The server times out polls after 60 seconds. Set our timeout to be slightly beyond that.
+const LONG_POLL_TIMEOUT: Duration = Duration::from_secs(62);
 const OTHER_CALL_TIMEOUT: Duration = Duration::from_secs(30);
 
 pub struct GatewayRef {
--- a/src/pollers/gateway.rs
+++ b/src/pollers/gateway.rs
@@ -30,7 +30,7 @@ use tonic::{
     codegen::{InterceptedService, Service},
     service::Interceptor,
     transport::{Certificate, Channel, Endpoint, Identity},
-    Status,
+    IntoRequest, Status,
 };
 use tower::ServiceBuilder;
 use url::Url;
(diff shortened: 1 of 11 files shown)


src/pollers/gateway.rs
```diff
@@ -30,16 +30,21 @@ use tonic::{
...
 use uuid::Uuid;
 
 #[cfg(test)]
 use futures::Future;
+use std::collections::HashMap;
+use std::str::FromStr;
+use tonic::metadata::{MetadataKey, MetadataValue};
 
 static LONG_POLL_METHOD_NAMES: [&str; 2] = ["PollWorkflowTaskQueue", "PollActivityTaskQueue"];
+const LONG_POLL_TIMEOUT: Duration = Duration::from_secs(60);
```
mfateev, line 46: I would make it a little big longer than 60 to avoid race condition with the service returning the call after 60 seconds. May be 61 second?

Sushisource: Done. Went with 62 because even numbers are nice :)

Sushisource:
Turns out there's no need for Core to deal with the server change since no one is fetching workflow results using it at present. Will add the supported-features header whenever it affects Core.

Leaving this open since Node needs to pass the header, and I will make that change as part of updating Node for this PR

src/pollers/gateway.rs
```diff
@@ -30,16 +30,22 @@ use tonic::{
...
 #[cfg(test)]
 use futures::Future;
+use std::collections::HashMap;
+use std::str::FromStr;
+use tonic::metadata::{MetadataKey, MetadataValue};
 
 static LONG_POLL_METHOD_NAMES: [&str; 2] = ["PollWorkflowTaskQueue", "PollActivityTaskQueue"];
+/// The server times out polls after 60 seconds. Set our timeout to be slightly beyond that.
+const LONG_POLL_TIMEOUT: Duration = Duration::from_secs(62);
+const OTHER_CALL_TIMEOUT: Duration = Duration::from_secs(30);
```
vitarb, line 48: What does "other" mean in this context? Anything other than a long poll?

Sushisource: Yes, anything that isn't a long poll

vitarb (approved):

src/pollers/gateway.rs
vitarb: I would remove word other then as it's just a "normal" call timeout. Also isn't 30 sec a bit too much for a non long poll?

Issue #170, opened 2021-08-10
Use `WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR` for non-deterministic workflow


wxing1292:
use `WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR` as cause for `RespondWorkflowTaskFailed` if user workflow logic is non-deterministic

Ref: https://github.com/temporalio/temporal/pull/1726

Issue #171, opened 2021-08-12
Stop returning workflow related errors from poll/complete and instead pipe them to evictions & attach reason


Sushisource:
Rather than returning errors that lang can't do anything about except evict the workflow, Core should assume the responsibility of just properly issuing evictions in such situations. Eviction jobs will also need to have reasons attached as a result.

The only errors that should be returned from core's top-level APIs should be fatal:
* Unrecoverable/unretryable network/server error
* Lang sent malformed data
* Worker shutdown / not present

All errors having to do with a specific workflow should instead be piped into an eviction.

PR #200, opened 2021-10-05
Automatically fail WFTs when WF code is nondeterministic w/ right cause

Sushisource:
## What was changed
Now errors on workflow completion result in auto-failure of the workflow task. In principle this should only ever be nondeterministic failures, or other edge-case panic type things.

This code can be cleaned up a bit once we do https://github.com/temporalio/sdk-core/issues/171 which I may just do next for exactly that reason, but it will require a bit more work to verify (and should result in Node changes where some code can be removed too)

## Why?
Seemingly before this change we'd just let WFT time out which I'm slightly surprised by. Now we fail properly, and include the right error code.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/170
2. How was this tested:
Existing and new tests and node integ tests

3. Any docs updates needed?

### DIFF
--- a/src/core_tests/determinism.rs
+++ b/src/core_tests/determinism.rs
@@ -18,13 +18,17 @@ use temporal_sdk_core_protos::temporal::api::enums::v1::WorkflowTaskFailedCause;
 static DID_FAIL: AtomicBool = AtomicBool::new(false);
 pub async fn timer_wf_fails_once(mut ctx: WfContext) -> WorkflowResult<()> {
     ctx.timer(Duration::from_secs(1)).await;
-    if !DID_FAIL.load(Ordering::Relaxed) {
-        DID_FAIL.store(true, Ordering::Relaxed);
+    if DID_FAIL
+        .compare_exchange(false, true, Ordering::Relaxed, Ordering::Relaxed)
+        .is_ok()
+    {
         panic!("Ahh");
     }
     Ok(().into())
 }
 
+/// Verifies that workflow panics (which in this case the Rust SDK turns into workflow activation
+/// failures) are turned into unspecified WFT failures.
 #[tokio::test]
 async fn test_wf_task_rejected_properly() {
     let wf_id = "fakeid";
@@ -49,6 +53,8 @@ async fn test_wf_task_rejected_properly() {
     worker.run_until_done().await.unwrap();
 }
 
+/// Verifies nondeterministic behavior in workflows results in automatic WFT failure with the
+/// appropriate nondeterminism cause.
 #[rstest::rstest]
 #[case::with_cache(true)]
 #[case::without_cache(false)]
@@ -72,16 +78,12 @@ async fn test_wf_task_rejected_properly_due_to_nondeterminism(#[case] use_cache:
     let core = mock_core(mock);
     let mut worker = TestRustWorker::new(Arc::new(core), TEST_Q.to_string(), None);
 
-    worker.register_wf(wf_type.to_owned(), |mut ctx: WfContext| {
-        let did_nondeterminism = AtomicBool::new(false);
-        async move {
-            ctx.timer(Duration::from_secs(1)).await;
-            if !did_nondeterminism.load(Ordering::Relaxed) {
-                did_nondeterminism.store(true, Ordering::Relaxed);
-                ctx.timer(Duration::from_secs(1)).await;
-            }
-            Ok(().into())
-        }
+    // The workflow is replaying all of history, so the when it schedules an extra timer it should
+    // not have, it causes a nondeterminism error.
+    worker.register_wf(wf_type.to_owned(), |mut ctx: WfContext| async move {
+        ctx.timer(Duration::from_secs(1)).await;
+        ctx.timer(Duration::from_secs(1)).await;
+        Ok(().into())
     });
 
     worker
--- a/src/core_tests/determinism.rs
+++ b/src/core_tests/determinism.rs
@@ -1,0 +1,94 @@
+use crate::{
+    pollers::MockServerGatewayApis,
+    prototype_rust_sdk::{TestRustWorker, WfContext, WorkflowResult},
+    test_help::{
+        build_mock_pollers, canned_histories, mock_core, MockPollCfg, ResponseType,
+        DEFAULT_WORKFLOW_TYPE, TEST_Q,
+    },
+};
+use std::{
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+use temporal_sdk_core_protos::temporal::api::enums::v1::WorkflowTaskFailedCause;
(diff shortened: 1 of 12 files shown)


src/core_tests/determinism.rs
```diff
@@ -0,0 +1,94 @@
...
+        Arc,
+    },
+    time::Duration,
+};
+use temporal_sdk_core_protos::temporal::api::enums::v1::WorkflowTaskFailedCause;
+
+static DID_FAIL: AtomicBool = AtomicBool::new(false);
+pub async fn timer_wf_fails_once(mut ctx: WfContext) -> WorkflowResult<()> {
+    ctx.timer(Duration::from_secs(1)).await;
+    if !DID_FAIL.load(Ordering::Relaxed) {
```
bergundy, line 21: Should this be a CAS operation to avoid a race?
I assume it's not critical for this test, just best practice.

src/core_tests/determinism.rs
```diff
@@ -0,0 +1,94 @@
+use crate::{
```
bergundy, line 1: This test could use some more comments all over

src/core_tests/determinism.rs
```diff
@@ -0,0 +1,94 @@
...
+    Ok(().into())
+}
+
+#[tokio::test]
+async fn test_wf_task_rejected_properly() {
+    let wf_id = "fakeid";
+    let wf_type = DEFAULT_WORKFLOW_TYPE;
+    let t = canned_histories::workflow_fails_with_failure_after_timer("1");
+    let mock = MockServerGatewayApis::new();
+    let mut mh = MockPollCfg::from_resp_batches(wf_id, t, [1, 2, 2], mock);
```
bergundy, line 34: Maybe explain what these magic numbers mean?

bergundy (approved):
Nice!
Had a few comments

src/prototype_rust_sdk/workflow_future.rs
```diff
@@ -266,8 +267,28 @@ impl Future for WorkflowFuture {
...
+                        .send(WfActivationCompletion::fail(
+                            &self.task_queue,
+                            run_id,
+                            Failure {
+                                message: format!("Workflow function panicked: {:?}", e),
+                                ..Default::default()
+                            },
+                        ))
+                        .expect("Completion channel intact");
+                    // Loop back up because we're about to get evicted
```
cretz, line 286: Pardon my ignorance to the algorithm, can you help me understand the flow and this comment? Who's listening to `WfActivationCompletion`? Are they then expected to send back a `WfActivation` with an eviction job upon receipt? Is it expected that there is only one `WfActivationCompletion` per workflow and if so, what if this completion is sent _and_ when the incoming eviction is processed another completion is sent?

src/core_tests/determinism.rs
cretz: :+1:  (also makes sense not to use `std::sync::Once` here as it will capture the panic and re-panic everytime)

src/core_tests/determinism.rs
```diff
@@ -0,0 +1,94 @@
...
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+use temporal_sdk_core_protos::temporal::api::enums::v1::WorkflowTaskFailedCause;
+
+static DID_FAIL: AtomicBool = AtomicBool::new(false);
+pub async fn timer_wf_fails_once(mut ctx: WfContext) -> WorkflowResult<()> {
```
cretz, line 19: Why is it important to only fail once? One might expect a deterministic workflow to always panic given the same inputs, so the test might as well too, right?

src/core_tests/determinism.rs
Sushisource: Can't actually race, since WFs aren't called concurrently. Atomics just happen to be a simple way to get a static mutable which is needed for something that will change across invocations

src/core_tests/determinism.rs
```diff
@@ -0,0 +1,94 @@
+use crate::{
+    pollers::MockServerGatewayApis,
+    prototype_rust_sdk::{TestRustWorker, WfContext, WorkflowResult},
```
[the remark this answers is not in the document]
Sushisource: That's what it is - it will be separated into its own crate but it's the API users writing Rust workflows would use. It happens to be the easiest way to write most tests too. Hence Rust SDK rather than Core

src/prototype_rust_sdk.rs
```diff
@@ -175,10 +175,7 @@ impl TestRustWorker {
                     debug!("Workflow {} says it's finishing", &completion.run_id);
                     self.incomplete_workflows.fetch_sub(1, Ordering::SeqCst);
                 }
-                self.core
-                    .complete_workflow_activation(completion)
-                    .await
-                    .unwrap();
```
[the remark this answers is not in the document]
Sushisource: This is only ever used for tests at this point, when we start working on productionizing it that will happen thorougly, though I'm pretty sure there aren't many

src/core_tests/determinism.rs
```diff
@@ -0,0 +1,94 @@
...
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+use temporal_sdk_core_protos::temporal::api::enums::v1::WorkflowTaskFailedCause;
+
+static DID_FAIL: AtomicBool = AtomicBool::new(false);
+pub async fn timer_wf_fails_once(mut ctx: WfContext) -> WorkflowResult<()> {
+    ctx.timer(Duration::from_secs(1)).await;
```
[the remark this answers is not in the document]
Sushisource: It's useful to have more than one workflow task in the history to make sure that the workflow's expected history matches what happens when executing one of the tasks fails and then replay happens again. 

It also matches an existing history I've already built so is easy to use

src/prototype_rust_sdk/workflow_future.rs
Sushisource: This would be easier to explain on a quick call. The actual core library eventually gets the completion. I know it will send an eviction activation for this workflow because the lang SDK (which just happens to be Rust b/c test) just sent a failure.

There can only ever be one outstanding activation for a given workflow run at a time, so the race you describe is not possible.

Feel free to put something on my calendar and I can go over this in more detail

src/core_tests/determinism.rs
Sushisource: Because we're testing nondeterministic behavior here. We want to see that a workflow can fail a WFT for spurious reasons, send the right type of failure, and then recover the next go-round.

src/core_tests/determinism.rs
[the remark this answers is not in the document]
cretz: Does it actually wait the second? Any concerns as the suite grows?

PR #202, opened 2021-10-11
Stop returning tonic errors from polls instead log warnings and retry.

Sushisource:
## What was changed
Polling now never returns network errors

## Why?
There was no reason to. Lang could never do anything sensible about it anyway. Ok, it failed, so what? Now the errors are logged as warnings.

## Checklist

1. Closes https://github.com/temporalio/sdk-node/issues/290

2. How was this tested:
Added UT. Verified manually by shutting server on and off.

3. Any docs updates needed?

### DIFF
--- a/src/pollers/retry.rs
+++ b/src/pollers/retry.rs
@@ -16,7 +16,6 @@ use temporal_sdk_core_protos::{
         query::v1::WorkflowQuery, workflowservice::v1::*,
     },
 };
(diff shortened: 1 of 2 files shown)


Sushisource:
I think I'm actually going to undo half of this. Keep the ability to return tonic errors because some things really are fatal (like bad auth). Instead I need to use a custom retryer for polls.

PR #213, opened 2021-11-02
Allow long polling to retry indefinitely and under more circumstances

Sushisource:
## What was changed
Polling now will retry indefinitely, and ignores cancel/deadline exceeded problems

## Why?
No reason not to. Polling should retry all transient conditions

## Checklist

1. Closes 

2. How was this tested:
Upgraded / added UTs

3. Any docs updates needed?

### DIFF
--- a/src/pollers/gateway.rs
+++ b/src/pollers/gateway.rs
@@ -43,7 +43,7 @@ use tonic::metadata::{MetadataKey, MetadataValue};
 
 static LONG_POLL_METHOD_NAMES: [&str; 2] = ["PollWorkflowTaskQueue", "PollActivityTaskQueue"];
 /// The server times out polls after 60 seconds. Set our timeout to be slightly beyond that.
-const LONG_POLL_TIMEOUT: Duration = Duration::from_secs(62);
+const LONG_POLL_TIMEOUT: Duration = Duration::from_secs(70);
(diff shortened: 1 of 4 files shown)


src/pollers/gateway.rs
```diff
@@ -164,6 +164,19 @@ impl Default for RetryConfig {
     }
 }
 
+impl RetryConfig {
+    pub(crate) fn poll_retry_policy() -> Self {
```
cretz, line 168: Wonder if this value should be static/global and just use references of it when needed? Does it ever get mutated? May have to change the next part to `impl From<&RetryConfig> for ExponentialBackoff` or something.

Sushisource: There's kinda no point since a new owned instance needs to be recreated every time. We'd end up just cloning the static instance.

Sushisource: It's also like ~48 bytes

cretz: > There's kinda no point since a new owned instance needs to be recreated every time

Ah, ok, haven't looked at where all it is sent. Granted I understand that for exponential backoff since it's mutable. I suppose a `Cow` type could be used for this shared type, but if not seen as worth it, no prob. Wasn't too concerned w/ size as much as instantiation over and over, but I understand if that's over-optimization.

cretz (approved):

PR #217, opened 2021-11-10
Stop returning workflow update errors from poll/complete & auto-evict

Sushisource:
## What was changed
We no longer return `WorkflowUpdateError` variants from polling/completing workflow tasks. It put an unnecessary burden on lang to deal with those errors, when the only thing that can really ever happen is to evict them. Now this eviction happens automatically.

## Why?
* Remove work from lang
* Correctness. Failing a WFT before this change could put the workflow in a stuck state.

TODO:
- [x] Attach reasons

Closes https://github.com/temporalio/sdk-core/issues/171

### DIFF
--- a/protos/local/workflow_activation.proto
+++ b/protos/local/workflow_activation.proto
@@ -60,7 +60,7 @@ message WFActivationJob {
         ///
         /// If other job variant are present in the list, this variant will be the last job in the
         /// job list. The string value is a reason for eviction.
-        string remove_from_cache = 50;
+        RemoveFromCache remove_from_cache = 50;
     }
 }
 
@@ -185,4 +185,8 @@ message ResolveRequestCancelExternalWorkflow {
     /// If populated, this signal either failed to be sent or was cancelled depending on failure
     /// type / info.
     temporal.api.failure.v1.Failure failure = 2;
+}
+
+message RemoveFromCache {
+    string reason = 1;
 }
--- a/protos/local/workflow_activation.proto
+++ b/protos/local/workflow_activation.proto
@@ -59,9 +59,8 @@ message WFActivationJob {
         /// after performing the activation.
         ///
         /// If other job variant are present in the list, this variant will be the last job in the
-        /// job list. The boolean value is irrelevant, since the variant type is what matters. It
-        /// will be set to true if this is the variant.
-        bool remove_from_cache = 50;
+        /// job list. The string value is a reason for eviction.
+        string remove_from_cache = 50;
     }
 }
 
--- a/src/core_tests/determinism.rs
+++ b/src/core_tests/determinism.rs
@@ -8,7 +8,7 @@ use crate::{
 };
 use std::{
     sync::{
-        atomic::{AtomicBool, Ordering},
+        atomic::{AtomicBool, AtomicUsize, Ordering},
         Arc,
     },
     time::Duration,
@@ -64,7 +64,13 @@ async fn test_wf_task_rejected_properly_due_to_nondeterminism(#[case] use_cache:
     let wf_type = DEFAULT_WORKFLOW_TYPE;
     let t = canned_histories::single_timer_wf_completes("1");
     let mock = MockServerGatewayApis::new();
-    let mut mh = MockPollCfg::from_resp_batches(wf_id, t, [ResponseType::AllHistory], mock);
+    let mut mh = MockPollCfg::from_resp_batches(
+        wf_id,
+        t,
+        // Two polls are needed, since the first will fail
+        [ResponseType::AllHistory, ResponseType::AllHistory],
+        mock,
+    );
     // We should see one wft failure which has nondeterminism cause
     mh.num_expected_fails = Some(1);
     mh.expect_fail_wft_matcher =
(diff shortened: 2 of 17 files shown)


src/core_tests/workflow_tasks.rs
```diff
@@ -1391,3 +1391,76 @@ async fn buffering_tasks_doesnt_count_toward_outstanding_max() {
...
     );
 }
+
+#[tokio::test]
+async fn fail_wft_then_recover() {
+    let t = canned_histories::long_sequential_timers(1);
+    let mut mh = MockPollCfg::from_resp_batches(
+        "fake_wf_id",
+        t,
+        [ResponseType::AllHistory, ResponseType::AllHistory],
```
bergundy, line 1401: Maybe add a comment here to explain that this will cause the same history to be delivered twice?
I had to scroll back up here to understand why WF gets retried.

src/worker/mod.rs
```diff
@@ -6,7 +6,7 @@ pub use crate::worker::config::{WorkerConfig, WorkerConfigBuilder};
 pub(crate) use dispatcher::WorkerDispatcher;
 
 use crate::{
-    errors::{CompleteWfError, WorkflowUpdateError},
+    errors::CompleteWfError,
```
bergundy, line 9: Looks like this is the only error that can be returned by worker methods.
After skimming the code I'm not convinced we evict in that case, can you verify?
One instance I think that might be missed is here:
https://github.com/temporalio/sdk-core/blob/1e92f51a19d80ba0b3124906d810471e3bad61bd/src/worker/mod.rs#L652

And possibly here:
https://github.com/temporalio/sdk-core/blob/1e92f51a19d80ba0b3124906d810471e3bad61bd/src/worker/mod.rs#L324

This is the TS side of completion:
https://github.com/temporalio/sdk-typescript/blob/53cffb41a1edc7f23a1676c5d1e8cee2421bf4ba/packages/worker/src/worker.ts#L806

bergundy (commented):
Great that we're tackling this now.

protos/local/workflow_activation.proto
```diff
@@ -59,9 +59,8 @@ message WFActivationJob {
         /// after performing the activation.
         ///
         /// If other job variant are present in the list, this variant will be the last job in the
-        /// job list. The boolean value is irrelevant, since the variant type is what matters. It
-        /// will be set to true if this is the variant.
-        bool remove_from_cache = 50;
+        /// job list. The string value is a reason for eviction.
+        string remove_from_cache = 50;
```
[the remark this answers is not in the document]
Sushisource: Yeah can do

src/workflow/workflow_tasks/concurrency_manager.rs
```diff
@@ -89,18 +89,16 @@ impl WorkflowConcurrencyManager {
...
         let writelock = self.runs.write();
         if writelock.contains_key(run_id) {
             Ok(RwLockWriteGuard::map(writelock, |hm| {
                 // Unwrap is safe because we hold the lock and just ensured run is in the map
                 &mut hm.get_mut(run_id).unwrap().wft
             }))
         } else {
-            Err(WorkflowUpdateError {
-                source: WFMachinesError::Fatal("Workflow machines not found".to_string()),
+            Err(WorkflowMissingError {
```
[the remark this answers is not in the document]
Sushisource: Yep

src/core_tests/determinism.rs
```diff
@@ -90,7 +99,6 @@ async fn test_wf_task_rejected_properly_due_to_nondeterminism(#[case] use_cache:
         .submit_wf(wf_id.to_owned(), wf_type.to_owned(), vec![])
         .await
         .unwrap();
-    // TODO: Shouldn't return error, should just queue eviction per
-    //  https://github.com/temporalio/sdk-core/issues/171
-    assert!(worker.run_until_done().await.is_err());
+    worker.run_until_done().await.unwrap();
+    assert_eq!(2, run_ct.load(Ordering::Relaxed));
```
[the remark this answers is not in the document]
Sushisource: Not without some fairly substantial restructuring, but it is inferrable from the fact that we need to fetch all of history twice.

tests/integ_tests/workflow_tests/determinism.rs
```diff
@@ -0,0 +1,47 @@
...
+    let wf_name = "test_determinism_error_then_recovers";
+    let mut starter = CoreWfStarter::new(wf_name);
+    let mut worker = starter.worker().await;
+
+    worker.register_wf(wf_name.to_owned(), timer_wf_nondeterministic);
+    worker
+        .submit_wf(wf_name.to_owned(), wf_name.to_owned(), vec![])
+        .await
+        .unwrap();
+    worker.run_until_done().await.unwrap();
```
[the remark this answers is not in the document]
Sushisource: Run id must be the same or the SDK won't activate the workflow code so that's covered.

protos/local/workflow_activation.proto
```diff
@@ -59,9 +59,8 @@ message WFActivationJob {
         /// after performing the activation.
         ///
         /// If other job variant are present in the list, this variant will be the last job in the
-        /// job list. The boolean value is irrelevant, since the variant type is what matters. It
-        /// will be set to true if this is the variant.
-        bool remove_from_cache = 50;
+        /// job list. The string value is a reason for eviction.
+        RemoveFromCache remove_from_cache = 50;
```
cretz, line 63: I trust we can guarantee that these protos are only used for FFI and never stored or used across versions and therefore are not concerned with compatibility? Or at least while we're in beta we can assume that the typescript SDK version always matches the exact expected core version?

cretz (approved):

protos/local/workflow_activation.proto
Sushisource: Yes exactly

src/worker/mod.rs
Sushisource: Yup, eviction happens there too (and is the most covered path by tests actually)

bergundy (approved):

PR #218, opened 2021-11-11
Fix issue with receiving child wf result before we awaited for it

Sushisource:
## What was changed
Before this fix if the test SDK received an activation resolving a child workflow before the workflow code awaited on the result, it would blow up because there was nothing to "resolve" yet.

This problem is now fixed by creating the future for resolution at the same time we start the child wf. 

## Why?
Bug

### DIFF
--- a/src/core_tests/child_workflows.rs
+++ b/src/core_tests/child_workflows.rs
@@ -44,13 +44,13 @@ async fn signal_child_workflow(#[case] serial: bool) {
             .into_started()
             .expect("Child should get started");
         let (sigres, res) = if serial {
-            let sigfut = start_res.signal(&mut ctx, SIGNAME, b"Hi!");
-            let resfut = start_res.result();
-            join!(sigfut, resfut)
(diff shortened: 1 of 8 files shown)


src/core_tests/child_workflows.rs
```diff
@@ -1,42 +1,68 @@
...
+        let start_res = child
+            .start(&mut ctx)
+            .await
+            .into_started()
+            .expect("Child should get started");
+        let (sigres, res) = if serial {
+            let sigfut = start_res.signal(&mut ctx, SIGNAME, b"Hi!");
+            let resfut = start_res.result();
+            join!(sigfut, resfut)
+        } else {
```
bergundy, line 50: I think this is the serial branch

bergundy (approved):

src/core_tests/child_workflows.rs
Sushisource: derp. Thanks

PR #219, opened 2021-11-12
Added a test to verify we can handle multiple WFT timeouts in a row

Sushisource:
Just adding a test to confirm something I was unsure about. It works.


cretz (approved):

Sushisource:
@bergundy This history was derived from an integration test. I just didn't commit it b/c it takes at least 4 seconds or so by definition

bergundy:
WFT timeout is configurable and server behavior might change and make this test obsolete but this is good enough.
I'll add an integration test to the TS SDK so we know this actually works E2E.

PR #225, opened 2021-11-16
Smattering of bugfixes

Sushisource:
## What was changed
* Fix a situation where we could get stuck polling if WFTs were repeatedly being failed
* Allow lang to send us cancels for already-completed things. This is unavoidable in some situations and we shouldn't blow up. Now they are simply dropped
* Fix for us accidentally still sending commands sometimes for things that were cancelled immediately

## Why?
Bugs!

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/src/core_tests/workflow_tasks.rs
+++ b/src/core_tests/workflow_tasks.rs
@@ -1557,3 +1557,111 @@ async fn lang_slower_than_wft_timeouts() {
     .unwrap();
     core.shutdown().await;
 }
+
+#[tokio::test]
+async fn tries_cancel_of_completed_activity() {
+    let mut t = TestHistoryBuilder::default();
+    t.add_by_type(EventType::WorkflowExecutionStarted);
+    t.add_full_wf_task();
+    let scheduled_event_id = t.add_activity_task_scheduled("1");
+    t.add_we_signaled("sig", vec![]);
+    let started_event_id = t.add_activity_task_started(scheduled_event_id);
+    t.add_activity_task_completed(scheduled_event_id, started_event_id, Default::default());
+    t.add_workflow_task_scheduled_and_started();
+
+    let mock = MockServerGatewayApis::new();
+    let mut mock = single_hist_mock_sg("fake_wf_id", t, &[1, 2], mock, true);
+    mock.worker_cfg(TEST_Q, |cfg| cfg.max_cached_workflows = 1);
+    let core = mock_core(mock);
+
+    let activation = core.poll_workflow_activation(TEST_Q).await.unwrap();
+    core.complete_workflow_activation(WfActivationCompletion::from_cmd(
+        TEST_Q,
+        activation.run_id,
+        ScheduleActivity {
+            seq: 1,
+            activity_id: "1".to_string(),
+            ..Default::default()
+        }
+        .into(),
+    ))
+    .await
+    .unwrap();
+    let activation = core.poll_workflow_activation(TEST_Q).await.unwrap();
+    assert_matches!(
+        activation.jobs.as_slice(),
+        [
+            WfActivationJob {
+                variant: Some(wf_activation_job::Variant::SignalWorkflow(_)),
+            },
+            WfActivationJob {
+                variant: Some(wf_activation_job::Variant::ResolveActivity(_)),
+            }
+        ]
+    );
+    core.complete_workflow_activation(WfActivationCompletion::from_cmds(
+        TEST_Q,
+        activation.run_id,
+        vec![
+            RequestCancelActivity { seq: 1 }.into(),
+            CompleteWorkflowExecution { result: None }.into(),
+        ],
+    ))
+    .await
+    .unwrap();
+
+    core.shutdown().await;
+}
+
+#[tokio::test]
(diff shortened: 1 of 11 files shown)


src/core_tests/workflow_tasks.rs
```diff
@@ -1557,3 +1557,111 @@ async fn lang_slower_than_wft_timeouts() {
...
     core.shutdown().await;
 }
+
+#[tokio::test]
+async fn tries_cancel_of_completed_activity() {
+    let mut t = TestHistoryBuilder::default();
+    t.add_by_type(EventType::WorkflowExecutionStarted);
+    t.add_full_wf_task();
+    let scheduled_event_id = t.add_activity_task_scheduled("1");
+    t.add_we_signaled("sig", vec![]);
```
bergundy, line 1567: signal isn't really relevant to this test case, it incidentally was what unblocked the workflow in the repro.

src/machines/activity_state_machine.rs
```diff
@@ -305,6 +305,20 @@ impl TryFrom<CommandType> for ActivityMachineEvents {
...
 impl Cancellable for ActivityMachine {
     fn cancel(&mut self) -> Result<Vec<MachineResponse>, MachineError<Self::Error>> {
+        if matches!(
+            self.state(),
+            ActivityMachineState::Completed(_)
+                | ActivityMachineState::Canceled(_)
+                | ActivityMachineState::Failed(_)
+                | ActivityMachineState::TimedOut(_)
+        ) {
+            // Ignore attempted cancels in terminal states
```
bergundy, line 315: I'm just wondering if it's worth showing example workflow code that explains why this is a valid scenario.

src/machines/activity_state_machine.rs
```diff
@@ -305,6 +305,20 @@ impl TryFrom<CommandType> for ActivityMachineEvents {
 
 impl Cancellable for ActivityMachine {
     fn cancel(&mut self) -> Result<Vec<MachineResponse>, MachineError<Self::Error>> {
+        if matches!(
+            self.state(),
+            ActivityMachineState::Completed(_)
+                | ActivityMachineState::Canceled(_)
```
bergundy, line 311: Not sure we should allow double cancel.
At least in TS we don't issue cancellations twice.

src/machines/child_workflow_state_machine.rs
```diff
@@ -54,6 +54,12 @@ fsm! {
     Started --(ChildWorkflowExecutionTimedOut(i32), shared on_child_workflow_execution_timed_out) --> TimedOut;
     Started --(ChildWorkflowExecutionCancelled, shared on_child_workflow_execution_cancelled) --> Cancelled;
     Started --(ChildWorkflowExecutionTerminated, shared on_child_workflow_execution_terminated) --> Terminated;
+
+    // Ignore any spurious cancellations after resolution
+    Cancelled --(Cancel) --> Cancelled;
```
bergundy, line 59: Hmm.. just verifying, I think we only use this machine to cancel before sending a command, no?
There's the external WF machine that does cancellation commands.

bergundy: Also, double cancel is probably a lang SDK bug (at least in TS).

src/machines/signal_external_state_machine.rs
```diff
@@ -42,6 +42,10 @@ fsm! {
     SignalExternalCommandRecorded
       --(SignalExternalWorkflowExecutionFailed(SignalExternalWorkflowExecutionFailedCause),
          on_signal_external_workflow_execution_failed) --> Failed;
+
+    // Ignore any spurious cancellations after resolution
+    Cancelled --(Cancel) --> Cancelled;
```
bergundy, line 47: Again double cancel should be a bug

src/machines/timer_state_machine.rs
```diff
@@ -39,6 +39,10 @@ fsm! {
         --(CommandCancelTimer, on_command_cancel_timer) --> CancelTimerCommandSent;
 
     CancelTimerCommandSent --(TimerCanceled) --> Canceled;
+
+    // Ignore any spurious cancellations after resolution
+    Canceled --(Cancel) --> Canceled;
```
bergundy, line 44: If we're disallowing double cancel worth noting this one too.

src/machines/activity_state_machine.rs
Sushisource: I think we might as well. Either we fail a situation that was otherwise recoverable, if misbehaving, or we allow it to proceed. This way if it does happen we keep operating which I think is probably better than not.

Sushisource: The commands sent to core show up in the logs so we still have an avenue to detect that sort of misbehavior.

src/core_tests/workflow_tasks.rs
Sushisource: Yep true, I think it just makes sense to indicate something else happened that would've activated the workflow at the same time

bergundy (approved):
As per off-PR discussion we want to allow double cancellation.

PR #228, opened 2021-11-18
Fix some lints and refactor across crates

tedbyron:
## What was changed

lint fix / code style

Used the following lints on each crate. Didn't fix every lint because of false positives and some require larger changes or are more opinionated. Have not looked at integration tests yet.

```rs
#![warn(clippy::all, clippy::pedantic, clippy::nursery)]
#![allow(
    clippy::wildcard_imports, // no big deal - if changed would require a ton of imports in multiple files
    clippy::module_name_repetitions, // no big deal - if changed it could just make items names shorter, rn sort of redunant like `use workflow::WorkflowBridge`, workflow already in mod name
    clippy::must_use_candidate, // questionable - sort of useful on constructors and functions that return owned values
    clippy::cast_possible_truncation, // questionable
    clippy::cast_sign_loss, // questionable
    clippy::cast_precision_loss, // questionable
    clippy::cast_possible_wrap, // questionable
    clippy::upper_case_acronyms, // ignoreable
    clippy::similar_names, // ignoreable
    clippy::too_many_lines, // ignoreable
    clippy::let_underscore_drop, // idiomatic to let _ = instead of drop()
    clippy::default_trait_access, // `Default::default()` is not very descriptive but most IDE will show types when u use this
    clippy::redundant_pub_crate, // reduntant but explicit if the surrounding module becomes pub in future
    clippy::unused_self, // some WIP methods don't currently use self but may in future
    clippy::unused_async, // common for async wrapper functions to need async fn as parameter, even if the param fn doesn't need to be async
    clippy::doc_markdown, // TODO - i can fix some doc lints in future
    clippy::missing_panics_doc, // TODO - i can fix some doc lints in future
    clippy::missing_errors_doc // TODO - i can fix some doc lints in future
)]
```

Some lints that could be worth taking a deeper look into, but also can't be 100% trusted because they are not in `clippy::all`: 

[`clippy::future_not_send`](https://rust-lang.github.io/rust-clippy/master/index.html#future_not_send) - a lot of futures being produced are not fully `Send`, but I'm not sure which ones are actually being sent across threads

[`clippy::needless_pass_by_value`](https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value) - a lot of fn params move a value but don't consume it, passing by value can sometimes create unnecessary allocations

## Why?

looking over code for now so I can make some actual contribution in future! I'll look at docs and tests more to get a better understanding of the code

## Checklist

1. Closes 
   n/a :-/

2. How was this tested:
   

   tested on arm64 macOS

   Integration tests had some warnings but same result as testing upstream/master

   ```
   cargo t --test integ_tests
   ..
   test result: ok. 42 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 6.67s
   ```

   ```
   cargo t --all-features --all-targets
   ..
   test result: ok. 188 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.76s
   ```

3. Any docs updates needed?
   

   no


tedbyron:
@Sushisource u do not want all these lints in CI or it will never pass...

```sh
cargo clippy -- -W clippy::all -W clippy::pedantic -W clippy::nursery 2>&1 | wc -l
5536
```

src/machines/workflow_machines.rs
```diff
@@ -459,13 +459,7 @@ impl WorkflowMachines {
...
         self.commands
             .iter()
-            .filter_map(|c| {
-                if !self.machine(c.machine).is_final_state() {
-                    Some(c.command.clone())
-                } else {
-                    None
-                }
-            })
+            .filter_map(|c| (!self.machine(c.machine).is_final_state()).then(|| c.command.clone()))
```
Sushisource, line 462: This one I don't totally love. I think the original is easier to parse.

Sushisource (commented):
This is great! I'm wondering if there's some subset of the lint preamble we could put in right now that would pass CI? Otherwise I'm liable to accumulate more of these :)

src/machines/workflow_machines.rs
tedbyron: https://github.com/temporalio/sdk-core/pull/228/commits/5eb12623e7f0ead32968a9469deedcb227c397c7

tedbyron:
Here are a bunch of lints that would pass now but I really don't know if it's a good idea to add, it will break stuff in future. If you look at the lints on https://rust-lang.github.io/rust-clippy/ almost every pedantic/nursery lint has issues with false positives. 

```rust
#![warn(
    clippy::all,
    clippy::use_self, // use Self when applicable
    clippy::explicit_iter_loop, // loop over &T instead of explicitly calling iter()
    clippy::semicolon_if_nothing_returned, // semicolon!
    clippy::if_not_else, // for readability, just reverse if/else when conditional has !
    clippy::explicit_deref_methods, // no explicit deref() or deref_mut()
    clippy::unseparated_literal_suffix, // 123832_i32 instead of 123832i32
    clippy::match_same_arms // same match arm bodies
)]
```

I looked at tokio and hyper which are big libs and they don't add lints except documentation lints and then once in a while people fix lints manually.

hyper only has at top level
```rust
#![deny(missing_docs)]
#![deny(missing_debug_implementations)]
```

tokio has
```rust
#![allow(
    clippy::cognitive_complexity,
    clippy::large_enum_variant,
    clippy::needless_doctest_main
)]
#![warn(
    missing_debug_implementations,
    missing_docs,
    rust_2018_idioms,
    unreachable_pub
)]
#![deny(unused_must_use)]
```

Sushisource:
Interesting, that makes sense to me then. Approved! Thanks so much for the contribution

Sushisource (approved):

PR #232, opened 2021-11-25
Fix activity heartbeat behavior

bergundy:
Behavior was incorrect, heartbeats need to be throttled using a timer and not count on Lang to keep sending heartbeats in order to be flushed.

Added 2 new Worker options to control throttling:
* `max_heartbeat_throttle_interval`
* `default_heartbeat_throttle_interval`

### DIFF
--- a/src/worker/activities/activity_heartbeat_manager.rs
+++ b/src/worker/activities/activity_heartbeat_manager.rs
@@ -341,7 +341,7 @@ mod test {
     };
     use tokio::time::sleep;
 
-    /// Ensure that heartbeats that are sent with a small throttleInterval are aggregated and sent roughly once
+    /// Ensure that heartbeats that are sent with a small `throttle_interval` are aggregated and sent roughly once
     /// every 1/2 of the heartbeat timeout.
     #[tokio::test]
     async fn process_heartbeats_and_shutdown() {
--- a/src/worker/activities/activity_heartbeat_manager.rs
+++ b/src/worker/activities/activity_heartbeat_manager.rs
@@ -15,10 +15,11 @@ use temporal_sdk_core_protos::{
 use tokio::{
     sync::{
         mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
-        oneshot, watch, Mutex,
+        Mutex,
     },
     task::JoinHandle,
 };
+use tokio_util::sync::CancellationToken;
 
 /// Used to supply new heartbeat events to the activity heartbeat manager, or to send a shutdown
 /// request.
@@ -26,15 +27,15 @@ pub(crate) struct ActivityHeartbeatManager {
     /// Cancellations that have been received when heartbeating are queued here and can be consumed
     /// by [fetch_cancellations]
     incoming_cancels: Mutex<UnboundedReceiver<PendingActivityCancel>>,
-    shutting_down: watch::Sender<bool>,
+    cancellation_token: CancellationToken,
     /// Used during `shutdown` to await until all inflight requests are sent.
     join_handle: Mutex<Option<JoinHandle<()>>>,
-    heartbeat_tx: UnboundedSender<HBAction>,
+    heartbeat_tx: UnboundedSender<HeartbeatAction>,
 }
 
 #[derive(Debug)]
-enum HBAction {
-    HB(ValidActivityHeartbeat),
+enum HeartbeatAction {
+    SendHeartbeat(ValidActivityHeartbeat),
     Evict(TaskToken),
     CompleteReport(TaskToken),
     CompleteThrottle(TaskToken),
@@ -48,9 +49,9 @@ pub struct ValidActivityHeartbeat {
 }
 
 #[derive(Debug)]
-enum HBExecutorAction {
+enum HeartbeatExecutorAction {
     /// Heartbeats are throttled for this task token, sleep until duration or wait to be cancelled
-    Sleep(TaskToken, Duration, oneshot::Receiver<()>),
+    Sleep(TaskToken, Duration, CancellationToken),
     /// Report heartbeat to the server
     Report(TaskToken, Vec<common::Payload>),
 }
@@ -71,11 +72,11 @@ impl ActivityHeartbeatManager {
         hb: ActivityHeartbeat,
         throttle_interval: Duration,
     ) -> Result<(), ActivityHeartbeatError> {
-        if *self.shutting_down.borrow() {
+        if self.cancellation_token.is_cancelled() {
             return Err(ActivityHeartbeatError::ShuttingDown);
         }
         self.heartbeat_tx
-            .send(HBAction::HB(ValidActivityHeartbeat {
+            .send(HeartbeatAction::SendHeartbeat(ValidActivityHeartbeat {
                 task_token: TaskToken(hb.task_token),
                 details: hb.details,
                 throttle_interval,
@@ -88,7 +89,7 @@ impl ActivityHeartbeatManager {
     /// Tell the heartbeat manager we are done forever with a certain task, so it may be forgotten.
     /// Record *should* not be called with the same TaskToken after calling this.
     pub(super) fn evict(&self, task_token: TaskToken) {
-        let _ = self.heartbeat_tx.send(HBAction::Evict(task_token));
+        let _ = self.heartbeat_tx.send(HeartbeatAction::Evict(task_token));
     }
 
     /// Returns a future that resolves any time there is a new activity cancel that must be
@@ -98,10 +99,10 @@ impl ActivityHeartbeatManager {
     }
 
     // TODO: Can own self now!
-    /// Initiates shutdown procedure by stopping lifecycle loop and awaiting for all heartbeat
-    /// processors to terminate gracefully.
+    /// Initiates shutdown procedure by stopping lifecycle loop and awaiting for all in-flight
+    /// heartbeat requests to be flushed to the server.
     pub(super) async fn shutdown(&self) {
-        let _ = self.shutting_down.send(true);
+        let _ = self.cancellation_token.cancel();
         let mut handle = self.join_handle.lock().await;
         if let Some(h) = handle.take() {
             h.await.expect("shutdown should exit cleanly");
@@ -110,15 +111,15 @@ impl ActivityHeartbeatManager {
 }
 
 #[derive(Debug)]
-struct ActivityHbState {
+struct ActivityHeartbeatState {
     task_token: TaskToken,
     last_recorded_details: Option<Vec<common::Payload>>,
     last_send_requested: Instant,
     throttle_interval: Duration,
-    sleep_cancel_tx: Option<oneshot::Sender<()>>,
+    sleep_cancel_tx: Option<CancellationToken>,
 }
 
-impl ActivityHbState {
+impl ActivityHeartbeatState {
     /// Get duration to sleep by subtracting `throttle_interval` by elapsed time since
     /// `last_send_requested`
     fn get_throttle_sleep_duration(&self) -> Duration {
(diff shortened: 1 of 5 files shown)


src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -105,143 +99,173 @@ impl ActivityHeartbeatManager {
...
-    fn new(incoming_hbs: UnboundedReceiver<HBAction>) -> Self {
-        Self {
-            last_sent: Default::default(),
-            incoming_hbs,
+async fn send_heartbeat(
+    sg: Arc<impl ServerGatewayApis + Send + Sync + 'static + ?Sized>,
+    cancels_tx: UnboundedSender<PendingActivityCancel>,
+    task_token: TaskToken,
+    details: Vec<common::Payload>,
+) {
```
Sushisource, line 112: Why make such dramatic changes to the entire structure here? The existing implementation had the nice property that there aren't bunches of `Arc`s and locks.

I think the streaming approach could pretty easily be kept while still accomplishing the goal

src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -105,143 +99,173 @@ impl ActivityHeartbeatManager {
...
+type JoinHandleMap = Arc<Mutex<Option<HashMap<TaskToken, JoinHandle<()>>>>>;
+
+async fn activity_heartbeat_loop(
+    tt: TaskToken,
+    throttle_interval: Duration,
+    heartbeat_states: HBStateMap,
+    join_handles: JoinHandleMap,
+    mut shutdown_rx: UnboundedReceiver<()>,
+    sg: Arc<impl ServerGatewayApis + Send + Sync + 'static + ?Sized>,
+    cancels_tx: UnboundedSender<PendingActivityCancel>,
```
Sushisource, line 154: I think this big 'ol parameter list is an example of that I mean. At minimum this could be wrapped up in a struct, but I still suspect an overall simpler take is totally doable

src/worker/config.rs
```diff
@@ -51,6 +51,17 @@ pub struct WorkerConfig {
...
+    /// Longest interval for throttling activity heartbeats
+    #[builder(default = "Duration::from_secs(60)")]
+    pub max_heartbeat_throttle_interval: Duration,
+
+    /// Default interval for throttling activity heartbeats in case
+    /// `ActivityOptions.heartbeat_timeout` is unset.
+    /// When the timeout *is* set in the `ActivityOptions`, throttling is set to
+    /// `heartbeat_timeout * 0.8`.
+    #[builder(default = "Duration::from_secs(30)")]
+    pub default_heartbeat_throttle_interval: Duration,
```
Sushisource, line 64: Do these really need to be configurable? Are we going to actually expose them to users? Seems like it'd cause more confusion than anything.

Sushisource:
I'm not clear on what the incorrect behavior was since I couldn't be in the meeting. Can we spell that out more here? Seems like from the PR we could fail to send some details if they were sent during throttle period but nothing was sent after?

Also from what I read in slack, it sounded like we might keep issuing heartbeats repeatedly with the last details even if user code/lang is no longer sending anything. Doesn't look like that's happening in this PR though which I think is good.

Overall the fix makes sense to me but I think the structure could be quite a bit simpler. Seems like we could keep the streaming approach and just use a task to push new items into it periodically that would flush out anything that needed to be

bergundy:
Sorry for the lack of details, I submitted it after midnight last night.
The incorrect behavior is that throttling worked differently in the other sdks.
There may be situations where it’s time to flush heartbeats but heartbeats get stuck in the buffer.
We discussed adding the default and max throttle intervals and exposing them to users.

I agree the implementation is not the prettiest and I think I can clean it up make it work with streams again.

bergundy:
@Sushisource changed to use streams instead.
I still will probably want to clean this up a bit but I think this is better.

src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -88,50 +97,145 @@ impl ActivityHeartbeatManager {
...
+                state.last_recorded_details = Some(hb.details);
+                None
+            }
+        }
+    }
+
+    /// Heartbeat report to server completed
+    fn complete_report(&mut self, tt: TaskToken) -> Option<HBExecutorAction> {
+        if let Some(st) = self.tt_to_state.get_mut(&tt) {
+            let (sleep_cancel_tx, sleep_cancel_rx) = unbounded_channel();
```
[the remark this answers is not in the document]
bergundy: Just figured that out myself and fixed :)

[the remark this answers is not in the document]
bergundy: Thanks for the link to `CancellationToken`, sounds like exactly what we need here, even better than `oneshot`.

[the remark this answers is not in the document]
bergundy: We should probably be using `CancellationToken`s and `DropGuard`s all over the codebase.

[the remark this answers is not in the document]
Sushisource: `CancellationToken` is nice. It's a lot like `watch` or `Notify` but the child thing is very useful.

src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -105,143 +111,221 @@ impl ActivityHeartbeatManager {
...
 #[derive(Debug)]
-struct ActivityHbState {
-    delay: Duration,
-    last_sent: Instant,
+struct HeartbeatStreamState {
+    tt_to_state: HashMap<TaskToken, ActivityHeartbeatState>,
+    incoming_hbs: UnboundedReceiver<HeartbeatAction>,
+    /// Token that can be used to cancel the entire stream.
+    /// Requests to the server are not cancelled with this token.
+    pub cancellation_token: CancellationToken,
```
Sushisource, line 144: Shoudn't / doesn't need to be be pub

src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -105,143 +111,221 @@ impl ActivityHeartbeatManager {
...
-            last_sent: Default::default(),
-            incoming_hbs,
-        }
+impl ActivityHeartbeatState {
+    /// Get duration to sleep by subtracting `throttle_interval` by elapsed time since
+    /// `last_send_requested`
+    fn get_throttle_sleep_duration(&self) -> Duration {
+        let time_since_last_sent = self.last_send_requested.elapsed();
+
+        return if time_since_last_sent > Duration::ZERO
```
Sushisource, line 128: Pretty sure this is in the CI errors but this return is unneeded / noisy

src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -105,143 +111,221 @@ impl ActivityHeartbeatManager {
...
+
+    /// Record a heartbeat received from lang
+    fn record(&mut self, hb: ValidActivityHeartbeat) -> Option<HeartbeatExecutorAction> {
+        match self.tt_to_state.entry(hb.task_token.clone()) {
+            Entry::Vacant(e) => {
+                let state = ActivityHeartbeatState {
+                    task_token: hb.task_token.clone(),
+                    throttle_interval: hb.throttle_interval,
+                    last_send_requested: Instant::now(),
+                    last_recorded_details: None,
```
Sushisource, line 168: Shouldn't this be set here since it is in the other branch? If not seemingly it should be set in `complete_report` rather than the other branch

src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -105,143 +111,221 @@ impl ActivityHeartbeatManager {
 }
 
 #[derive(Debug)]
-struct HbStreamState {
-    last_sent: HashMap<TaskToken, ActivityHbState>,
-    incoming_hbs: UnboundedReceiver<HBAction>,
+struct ActivityHeartbeatState {
+    task_token: TaskToken,
+    last_recorded_details: Option<Vec<common::Payload>>,
```
Sushisource, line 116: This probably could do without the `Option` since an empty `Vec` isn't worth reporting either.

src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -105,143 +111,221 @@ impl ActivityHeartbeatManager {
...
+    /// Throttling completed, report or stop tracking task token
+    fn complete_throttle(&mut self, tt: TaskToken) -> Option<HeartbeatExecutorAction> {
+        match self.tt_to_state.entry(tt.clone()) {
+            Entry::Occupied(mut e) => {
+                let state = e.get_mut();
+                if state.last_recorded_details.is_none() {
+                    // Nothing to report, forget this task token
+                    e.remove();
+                    None
+                } else {
```
Sushisource, line 207: These branches could be simplified by having the first one be

`if let Some(details) = state.last_recorded_details.take() {`

src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -105,143 +111,221 @@ impl ActivityHeartbeatManager {
...
+                }
+            }
+            Entry::Vacant(_) => None,
+        }
+    }
+
+    /// Activity should not be tracked anymore, cancel throttle timer if running
+    fn evict(&mut self, tt: TaskToken) -> Option<HeartbeatExecutorAction> {
+        if let Some(state) = self.tt_to_state.remove(&tt) {
+            if let Some(tx) = state.sleep_cancel_tx {
```
Sushisource, line 224: Could collapse branches by using `.map(|state| state.sleep_cancel_tx)`

src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -257,7 +341,7 @@ mod test {
     };
     use tokio::time::sleep;
 
-    /// Ensure that heartbeats that are sent with a small delay are aggregated and sent roughly once
+    /// Ensure that heartbeats that are sent with a small throttleInterval are aggregated and sent roughly once
```
Sushisource, line 344:
```suggestion
    /// Ensure that heartbeats that are sent with a small `throttle_interval` are aggregated and sent roughly once
```

src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -105,143 +111,221 @@ impl ActivityHeartbeatManager {
...
+            Entry::Occupied(mut o) => {
+                let state = o.get_mut();
+                state.last_recorded_details = Some(hb.details);
+                None
+            }
+        }
+    }
+
+    /// Heartbeat report to server completed
+    fn complete_report(&mut self, tt: TaskToken) -> Option<HeartbeatExecutorAction> {
```
Sushisource, line 183: I think this name would be better as `report_completed` since it is indicating it happened rather than doing it. Same with throttle.

Sushisource (commented):
This looks good to me this way :+1:  just a few things to fix and the CI lints

src/worker/activities/activity_heartbeat_manager.rs
bergundy: We don't record it here because we send the details out.
We use `None` to denote that after throttling we can stop tracking this task token.
Adding a comment.

src/worker/activities/activity_heartbeat_manager.rs
bergundy: But then `report` sounds like it's a verb, I'll change to `handle_report_completed`

src/worker/activities/activity_heartbeat_manager.rs
bergundy: You're right.

src/worker/activities/activity_heartbeat_manager.rs
bergundy: Tried it but it made the code less readable imho especially because `sleep_cancel_tx` is an `Option`.

bergundy (commented):
Thanks, fixing

src/worker/config.rs
cretz: Sorry, I _do_ think it's fine to open to users. I forgot discussion. Doesn't matter to much on sdk-core repo because it should at least be exposed to languages.

Sushisource (approved):

PR #233, opened 2021-11-30
Add the license field in all Cargo.tomls

bergundy:
I'm hoping this will help FOSSA find the licenses.


Sushisource (changes requested):
This makes the compiler warn constantly:

`warning: /sdk-core/Cargo.toml: only one of `license` or `license-file` is necessary`

I would prefer to avoid that unless we know for sure this helps with FOSSA

bergundy:
We've managed to get around this for now, closing

PR #235, opened 2021-12-01
Fix cache miss not returning WFT permits

Sushisource:
## What was changed
Title

## Why?
Bugfix

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Sushisource:
This branch will be merged into master. The https://github.com/temporalio/sdk-core/tree/ts-release branch should hang out for TS to use until we finish LA work and go back to master

src/worker/mod.rs
```diff
@@ -528,6 +528,7 @@ impl Worker {
                         }),
                     )
                     .await?;
+                self.return_workflow_task_permit();
```
Sushisource, line 531: This is the fix

src/core_tests/workflow_tasks.rs
```diff
@@ -1665,3 +1665,72 @@ async fn failing_wft_doesnt_eat_permit_forever() {
 
     core.shutdown().await;
 }
+
+#[tokio::test]
+async fn cache_miss_doesnt_eat_permit_forever() {
```
bergundy, line 1670: Wouldn't it have been more useful to have an integration test here so we could verify the SDK behaves correctly with an actual server implementation?

src/test_help/history_info.rs
```diff
@@ -137,4 +149,11 @@ mod tests {
         let history_info = t.get_history_info(2).unwrap();
         assert_eq!(8, history_info.events.len());
     }
+
+    #[test]
+    fn incremental_works() {
+        let t = canned_histories::single_timer("timer1");
+        let hi = t.get_one_wft(2).unwrap();
+        dbg!(hi.events);
```
bergundy, line 157: Shouldn't there be an assertion here?

src/worker/mod.rs
bergundy: lol

bergundy (approved):
I would have made an integration test with a worker with 1 concurrent wft execution and called evict to reproduce the exact scenario but we can merge as is if you don't want to invest in it.

cretz (approved):

src/test_help/history_info.rs
Sushisource: Yeah, just forgot to do that. Done.

src/core_tests/workflow_tasks.rs
Sushisource: Added one as well

PR #238, opened 2021-12-06
Local Activity Retries

Sushisource:
## What was changed
Adds retry capability to local activities

## Why?
Sometimes they fail don'tcha know

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Sushisource:
Need to add server-timer based backoff for long backoffs

src/worker/activities/local_activities.rs
```diff
@@ -135,17 +152,61 @@ impl LocalActivityManager {
...
-        self.complete_notify.notify_one();
-        Some(info)
+            .remove(task_token)
+        {
+            self.semaphore.add_permits(1);
+
+            match status {
+                LocalActivityExecutionResult::Completed(_) => {
+                    self.complete_notify.notify_one();
+                    LACompleteAction::Report(info)
```
bergundy, line 172: Not for this PR but as part of the "refactor core to use more streams and reducers" I think we would just push the result into the notification channel.

src/worker/activities/local_activities.rs
```diff
@@ -135,17 +152,61 @@ impl LocalActivityManager {
...
+                }
+                LocalActivityExecutionResult::Failed(_) => {
+                    match &info.la_info.schedule_cmd.retry_policy {
+                        None => {
+                            // If there's no retry policy... don't retry
+                            LACompleteAction::Report(info)
+                        }
+                        Some(rp) => {
+                            if let Some(backoff_dur) =
+                                rp.should_retry(info.attempt as usize, "TODO")
```
bergundy, line 182: We'd use the failure type here, right?

src/worker/activities/local_activities.rs
```diff
@@ -135,17 +152,61 @@ impl LocalActivityManager {
...
+                            {
+                                // Send the retry request after waiting the backoff duration
+                                let send_chan = self.act_req_tx.clone();
+                                tokio::spawn(async move {
```
bergundy, line 186: There's nothing to cancel this task for example on shutdown, I don't know if I'd leave it dangling like this.

src/worker/mod.rs
```diff
@@ -275,35 +278,44 @@ impl Worker {
...
+                            &la_info.workflow_exec_info.run_id,
+                            "Issue while completing local activity",
+                        );
+                    }
+                }
+                LACompleteAction::WillBeRetried => {
+                    // Nothing to do here
+                }
+                LACompleteAction::Untracked => {
+                    error!("Tried to complete untracked local activity {}", task_token);
```
bergundy, line 317: Should this error just be logged?
Is this "error" severity? It would probably happen if user ignores cancellation in their activity implementation.

src/workflow/mod.rs
```diff
@@ -314,12 +316,18 @@ pub mod managed_wf {
...
-            result: ActivityResult,
+            result: ActivityExecutionResult,
         ) -> Result<()> {
             self.mgr.notify_of_local_result(
                 seq_num,
                 LocalResolution::LocalActivity {
-                    result,
+                    // We accept normal execution results and do this conversion because there
+                    // are more helpers for constructing them.
+                    result: result
```
bergundy, line 326: Should this not return a meaningful error back instead of panicking?

protos/local/common.proto
```diff
@@ -38,10 +38,11 @@ message NamespacedWorkflowExecution {
 // Defines how an activity or workflow should be retried in the event of failure, timeout, etc.
 message RetryPolicy {
     // Interval of the first retry. If backoff_coefficient is 1.0 then it is used for all
-    // retries.
+    // retries. If this is not specified, there will be *no retrying*. Lang SDKs should require
```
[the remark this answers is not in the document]
Sushisource: Because the right way to say that is max attempts 1.

Lang sdks should just default this to some reasonable value always, rather than allowing the odd situation of max attempts being > 1 but this being unset.

protos/local/activity_result.proto
```diff
@@ -17,30 +18,52 @@ message ActivityResult {
...
  * Core will forget about this Activity and free up resources used to track this Activity.
  */
 message WillCompleteAsync {
 }
 
+/**
+ * Issued when a local activity needs to retry but also wants to back off more than would be
+ * reasonable to WFT heartbeat for. Lang is expected to schedule a timer for the duration
+ * and then start a local activity of the same type & same inputs with the provided attempt number
+ * after the timer has elapsed.
```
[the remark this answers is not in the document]
Sushisource: Sure will do

protos/local/activity_result.proto
```diff
@@ -17,30 +18,52 @@ message ActivityResult {
...
     temporal.api.failure.v1.Failure failure = 1;
 }
 
 /**
- * Used in ActivityResult to report cancellation from both Core and Lang.
- * When Lang reports a cancelled ActivityResult, it must put a CancelledFailure in the failure field.
- * When Core reports a cancelled ActivityResult, it must put an ActivityFailure with CancelledFailure
+ * Used to report cancellation from both Core and Lang.
+ * When Lang reports a cancelled activity, it must put a CancelledFailure in the failure field.
+ * When Core reports a cancelled activity, it must put an ActivityFailure with CancelledFailure
```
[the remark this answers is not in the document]
Sushisource: You wrote it! :stuck_out_tongue: 

But, no I think the way it is makes sense. More info is needed in the core->lang direction than the reverse.

protos/local/common.proto
```diff
@@ -38,10 +38,11 @@ message NamespacedWorkflowExecution {
...
+    // this to be set.
     google.protobuf.Duration initial_interval = 1;
     // Coefficient used to calculate the next retry interval. The next retry interval is previous
-    // interval multiplied by the coefficient. Must be 1 or larger.
+    // interval multiplied by the coefficient. Must be 1 or larger. Defaults to 2.
```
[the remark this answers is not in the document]
Sushisource: Yeah I just realized that this is unclear. I will remove the defaults from here because they may actually be different depending on what the policy is used for.

src/retry_logic.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+impl RetryPolicyExt for RetryPolicy {
+    fn should_retry(&self, attempt_number: usize, err_str: &str) -> Option<Duration> {
+        if attempt_number >= self.maximum_attempts.max(0) as usize {
+            return None;
+        }
+
+        for pat in &self.non_retryable_error_types {
+            if err_str.contains(pat.as_str()) {
```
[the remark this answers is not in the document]
Sushisource: So I think that approach kinda sucks and we should ideally change it on server at some point too. It's deeply inflexible.

It's even worse in this context because I have no idea what type name you're talking about and the user likely doesn't either. The type names they see in lang I won't even see in Core. It just doesn't make sense. We would have to add an explicit field in `Failure` (which is upstream, not owned by us) along the lines of `failure_type_string` just to be able to check this reliably.

I'd much rather make it clear that for local activities it works this way, which most users I think will prefer in any case.

src/retry_logic.rs
```diff
@@ -0,0 +1,110 @@
...
+        if let Some(interval) = converted_interval {
+            let max_iv = self
+                .maximum_interval
+                .clone()
+                .map(TryInto::try_into)
+                .transpose()
+                .ok()
+                .flatten()
+                .unwrap_or_else(|| interval * 100);
+            Some((interval.mul_f64(coeff.powi(attempt_number as i32 - 1))).min(max_iv))
```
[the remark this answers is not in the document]
Sushisource: I don't think we care. It doesn't exist for normal activities. Users who really need it can fairly trivially implement it inside the activity itself.

src/worker/activities/local_activities.rs
```diff
@@ -38,13 +41,22 @@ impl Debug for NewLocalAct {
     }
 }
 
+#[derive(Debug)]
+enum NewOrRetry {
```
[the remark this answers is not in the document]
Sushisource: Better living through type systems. This is more explicit (and, uninterestingly, saves a word).

src/worker/activities/local_activities.rs
Sushisource: Yeah, fixed in next PR

src/worker/activities/local_activities.rs
Sushisource: I don't think it's relevant at all here. If shutdown is ongoing it cannot proceed until the LAs are complete anyway. If somehow we've decided that's true then this just gets dropped on the floor which seemingly is what you'd want at that point anyway.

src/worker/mod.rs
Sushisource: You mean should it do more than just log? All it does is log now.

We can change it to warn when working on cancels.

src/workflow/mod.rs
Sushisource: This is test code

Sushisource: `ManagedWFFunc` is test only

protos/local/common.proto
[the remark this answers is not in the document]
bergundy: Looks like that's what java does: https://github.com/temporalio/sdk-java/blob/820627bc6050dfba13c252f46c3518715004a02e/temporal-sdk/src/main/java/io/temporal/common/RetryOptions.java#L112

src/retry_logic.rs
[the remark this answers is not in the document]
cretz: > We would have to add an explicit field in Failure (which is upstream, not owned by us) along the lines of failure_type_string just to be able to check this reliably.

:+1: If we don't want to be limited to a specific set of types but also don't want to do gross error message string-contains, we'd need some kind of "tagging".

> I'd much rather make it clear that for local activities it works this way, which most users I think will prefer in any case.

I do think `non_retryable_error_types` is specified by https://github.com/temporalio/api/blob/master/temporal/api/common/v1/message.proto#L109-L111 to mean failure type for now and I worry that someone may reuse a retry policy for normal activities and local activities expecting this field to work the same across both. Can a language not provide the failure types of their failures to core for purposes of this kind of matching?

src/retry_logic.rs
```diff
@@ -0,0 +1,110 @@
...
+            .ok()
+            .flatten();
+        if attempt_number == 1 {
+            return converted_interval;
+        }
+        let coeff = if self.backoff_coefficient != 0. {
+            self.backoff_coefficient
+        } else {
+            2.0
+        };
```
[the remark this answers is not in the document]
cretz: I kinda understand wanting the other single statement to become a single-statement method since it has reuse potential (even then single statements with hardly any logic rarely deserve to be broken out, otherwise one could only imagine the extent which that could be applied). But for this simple if statement I, subjectively, prefer it remain inline.

src/retry_logic.rs
```diff
@@ -0,0 +1,110 @@
...
+            }
+        }
+
+        let converted_interval = self
+            .initial_interval
+            .clone()
+            .map(TryInto::try_into)
+            .transpose()
+            .ok()
+            .flatten();
```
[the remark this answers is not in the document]
cretz: FWIW, I'm have no hard preference here since it's reused, but am totally fine leaving inline

src/retry_logic.rs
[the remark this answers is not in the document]
cretz: :+1: As a dev I would expect to be able calculate my retry time based on the inputs I gave.

src/worker/activities/local_activities.rs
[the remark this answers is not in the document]
cretz: Seeing how it just translates later, I like the single struct for both cases myself, but meh.

src/worker/activities/local_activities.rs
```diff
@@ -135,17 +152,61 @@ impl LocalActivityManager {
...
+
+            match status {
+                LocalActivityExecutionResult::Completed(_) => {
+                    self.complete_notify.notify_one();
+                    LACompleteAction::Report(info)
+                }
+                LocalActivityExecutionResult::Failed(_) => {
+                    match &info.la_info.schedule_cmd.retry_policy {
+                        None => {
+                            // If there's no retry policy... don't retry
```
cretz, line 177: In practice, do we ever expect lang to not provide a retry policy?

Sushisource: A default should be provided

protos/local/activity_result.proto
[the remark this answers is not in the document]
bergundy: 👍

src/worker/activities/local_activities.rs
bergundy: In the near future we should move shutdown grace period from TS into core and then you'll need to forcefully shut these down.

src/worker/mod.rs
bergundy: I'd maybe just warn because the case I stated above

src/worker/activities/local_activities.rs
bergundy: Hmm.. lang doesn't provide a default right now.
What do the other SDKs do? What is the default policy for LAs?

src/retry_logic.rs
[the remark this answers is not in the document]
Sushisource: @bergundy I didn't see the `type` field on `ApplicationFailure`. I will only use/look for that which is seemingly closest to normal activities.

I still think it's really difficult for users to make use of this well, but better to be consistent I suppose

src/retry_logic.rs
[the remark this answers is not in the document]
Sushisource: Yeah, not gonna change this one. This would be like making a fn for a ternary op

src/worker/activities/local_activities.rs
Sushisource: It's to retry forever I believe. We can check the defaults when impling lang side

src/worker/mod.rs
Sushisource: I'll do it when working on cancels

src/worker/activities/local_activities.rs
Sushisource: It still won't matter? Nothing will happen. They'll just be dropped. They won't hang shutdown like a thread would.

PR #239, opened 2021-12-09
Local activity timer-based backoff

Sushisource:
## What was changed
Long backoff periods while retrying LAs are implemented in other SDKs (or should be) by scheduling a timer rather than WFT heartbeating for a huge period of time.

This is more complex to do totally internally to core than would be ideal because Core never expected to create most types of commands internally. To solve this, we send a special type of activity resolution to lang that says "You should schedule a timer and *then* try to run this activity again".

Much blood is also spilled creating a new `ScheduleLocalActivity` type since `ScheduleActivity` was getting too cluttered with local-specific stuff and the local version had a bunch of things that don't apply.

## Why?
Parity

## Checklist
Need to add some more tests

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/core_tests/local_activities.rs
+++ b/src/core_tests/local_activities.rs
@@ -17,8 +17,8 @@ use std::{
     time::Duration,
 };
 use temporal_sdk_core_protos::{
-    coresdk::{common::RetryPolicy, AsJsonPayloadExt},
-    temporal::api::enums::v1::EventType,
+    coresdk::{activity_result::activity_resolution, common::RetryPolicy, AsJsonPayloadExt},
+    temporal::api::{enums::v1::EventType, failure::v1::Failure},
 };
 use tokio::sync::Barrier;
 
@@ -244,7 +244,7 @@ async fn local_act_fail_and_retry(#[case] eventually_pass: bool) {
     let attempts: &'static _ = Box::leak(Box::new(AtomicUsize::new(0)));
     worker.register_activity("echo", move |_: String| async move {
         // Succeed on 3rd attempt
-        if 2 == attempts.fetch_add(1, Ordering::Relaxed) && eventually_pass {
+        if 3 == attempts.fetch_add(1, Ordering::Relaxed) && eventually_pass {
             Ok(())
         } else {
             Err(anyhow!("Oh no I failed!"))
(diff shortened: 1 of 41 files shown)


src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -162,11 +195,38 @@ impl WfContext {
...
+                opts.attempt = Some(b.attempt);
+                last_resolution = self.local_activity_no_timer_retry(opts.clone()).await;
+            }
+            last_resolution
+        })
+    }
+
+    /// Request to run a local activity with no implementation of timer-backoff based retrying.
+    /// TODO: Should not be public in non-prototype SDK
+    pub(crate) fn local_activity_no_timer_retry(
```
bergundy, line 225: This could probably just be private, I don't see a reason to use it anywhere.
Even the test that used it should probably just use `local_activity` directly.

src/prototype_rust_sdk/workflow_context.rs
```diff
@@ -162,11 +195,38 @@ impl WfContext {
...
     /// Request to run a local activity
     pub fn local_activity(
-        &mut self,
+        &self,
+        opts: LocalActivityOptions,
+    ) -> impl CancellableFuture<ActivityResolution> + '_ {
+        // We need to immediately create the first LA future to be consistent with the behavior
+        // of other functions where the command is scheduled before being awaited on.
+        let first_la_fut = self.local_activity_no_timer_retry(opts.clone());
+        ManualCancellableFuture::new(|_ct| async move {
```
bergundy, line 204: Can you please explain what this cancellable future is for?
Are we using this type of cancellation in workflow context?
Cancelling a LA should be done via sending a cancellation request AFAIK.

src/prototype_rust_sdk/workflow_context/options.rs
```diff
@@ -91,21 +94,27 @@ pub struct LocalActivityOptions {
     pub input: Payload,
     /// Retry policy
     pub retry_policy: RetryPolicy,
+    /// Override attempt number rather than using 1.
+    /// Ideally we would not expose this in a released Rust SDK, but it's needed for test.
```
bergundy, line 98: Yes ideally we wouldn't expose this.

src/worker/activities/local_activities.rs
```diff
@@ -171,16 +179,42 @@ impl LocalActivityManager {
...
-                                rp.should_retry(info.attempt as usize, "TODO")
-                            {
+                            if let Some(backoff_dur) = rp.should_retry(
+                                info.attempt as usize,
+                                &f.failure
+                                    .as_ref()
+                                    .map(|f| format!("{:?}", f))
+                                    .unwrap_or_else(|| "".to_string()),
+                            ) {
+                                warn!(
```
bergundy, line 196: Sounds like debug to me

src/worker/activities/local_activities.rs
```diff
@@ -171,16 +179,42 @@ impl LocalActivityManager {
...
+                                if backoff_dur
+                                    > info
+                                        .la_info
+                                        .schedule_cmd
+                                        .local_retry_threshold
+                                        .clone()
+                                        .try_into_or_none()
+                                        .unwrap_or_else(|| Duration::from_secs(60))
+                                {
+                                    warn!("Backoff is past local retry threshold");
```
bergundy, line 210: Also debug or maybe even trace

tests/integ_tests/workflow_tests/local_activities.rs
```diff
@@ -144,3 +145,38 @@ async fn local_act_fanout() {
...
     starter.shutdown().await;
 }
+
+#[tokio::test]
+async fn local_act_retry_timer_backoff() {
+    let wf_name = "local_act_retry_timer_backoff";
+    let mut starter = CoreWfStarter::new(wf_name);
+    let mut worker = starter.worker().await;
+    worker.register_wf(wf_name.to_owned(), |ctx: WfContext| async move {
+        ctx.local_activity(LocalActivityOptions {
```
bergundy, line 155: No checking that the activity eventually fails?

src/prototype_rust_sdk/workflow_context.rs
cretz: I think I understand the concept here, but yeah seems not to make sense if the cancellation token parameter is never used.

cretz, line 204: What is the purpose of your custom `ManualCancellableFuture` here if the cancellation token is unused?

cretz (approved):

src/prototype_rust_sdk/workflow_context.rs
Sushisource: Still have to add that bit :)

src/core_tests/local_activities.rs
```diff
@@ -255,6 +259,95 @@ async fn local_act_fail_and_retry(#[case] eventually_pass: bool) {
         .await
         .unwrap();
     worker.run_until_done().await.unwrap();
-    let expected_attempts = if eventually_pass { 3 } else { 4 };
+    let expected_attempts = if eventually_pass { 3 } else { 5 };
     assert_eq!(expected_attempts, attempts.load(Ordering::Relaxed));
 }
+
+#[tokio::test]
+async fn local_act_retry_long_backoff_uses_timer() {
```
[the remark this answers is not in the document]
Sushisource: There is one as well. Both have value. It's much easier to fix bugs with the UTs since they run faster, can use a debugger, etc.

src/prototype_rust_sdk/workflow_context.rs
Sushisource: The unit test is able to verify directly that the right response is returned for telling lang to back off, but I can probably do that lower level in the machine test so I can change this one.

src/worker/activities/local_activities.rs
Sushisource: Yeah forgot to change this one

src/prototype_rust_sdk/workflow_context.rs
bergundy: This is just a prototype SDK but if it turns into a real SDK we'll need to put more effort into the cancellation design.

Sushisource: Everything will need a once over but I don't think there's anything problematic about the current state

Sushisource: (Which I will do with next PR adding LA cancels since it'll just blow up right now trying to cancel them since it's unimplemented)

PR #240, opened 2021-12-14
Fix WFT failures sometimes getting stuck in a spam loop

Sushisource:
## What was changed
While working on LA implementation, discovered that all-internal state machine errors would not avoid spamming WFT failures properly. Fixed that.

## Why?
:bug: 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,6 +1,6 @@
 //! Error types exposed by public APIs
 
-use crate::{machines::WFMachinesError, task_token::TaskToken, WorkerLookupErr};
+use crate::{machines::WFMachinesError, WorkerLookupErr};
 use temporal_sdk_core_protos::coresdk::{
     activity_result::ActivityExecutionResult, workflow_completion::WfActivationCompletion,
 };
@@ -13,8 +13,6 @@ pub(crate) struct WorkflowUpdateError {
     /// The run id of the erring workflow
     #[allow(dead_code)] // Useful in debug output
     pub run_id: String,
-    /// The task token associated with this update, if one existed yet.
-    pub task_token: Option<TaskToken>,
 }
 
 impl From<WorkflowMissingError> for WorkflowUpdateError {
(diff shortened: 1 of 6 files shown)


src/worker/mod.rs
```diff
@@ -538,7 +546,7 @@ impl Worker {
         &self,
         run_id: &str,
         success: workflow_completion::Success,
-    ) -> Result<bool, CompleteWfError> {
+    ) -> Result<WFTReportOutcome, CompleteWfError> {
```
cretz, line 549: Pedantic - Cannot tell with `Wf` vs `WFT` whether we should capitalize abbreviations or not

cretz (approved):

src/worker/mod.rs
Sushisource: Yeah I've been kinda inconsistent here. Clippy by default says don't, but I kinda like it. :man_shrugging:

src/errors.rs
```diff
@@ -13,16 +13,13 @@ pub(crate) struct WorkflowUpdateError {
...
-    /// The task token associated with this update, if one existed yet.
-    pub task_token: Option<TaskToken>,
 }
 
 impl From<WorkflowMissingError> for WorkflowUpdateError {
     fn from(wme: WorkflowMissingError) -> Self {
         Self {
             source: WFMachinesError::Fatal("Workflow machines missing".to_string()),
             run_id: wme.run_id,
-            task_token: None,
```
bergundy, line 25: Why did you remove this?

Sushisource: It's no longer needed for anything.

src/worker/mod.rs
```diff
@@ -648,35 +648,47 @@ impl Worker {
     async fn wf_activation_failed(
         &self,
         run_id: &str,
+        cause: WorkflowTaskFailedCause,
         failure: workflow_completion::Failure,
-    ) -> Result<bool, CompleteWfError> {
+    ) -> Result<WFTReportOutcome, CompleteWfError> {
+        warn!(run_id, failure=?failure, "Failing workflow activation");
```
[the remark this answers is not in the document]
Sushisource: I suppose if something is going wrong it could be... but Shawn did specifically mention users wanting to see this in the logs. I could shift it to be only when we *report* to server which would make sense. I'll do that.

PR #241, opened 2021-12-15
Local Activity Cancellation

Sushisource:
Implement cancellation of local activities

Draft until:

- [ ] Test timeouts ( NEXT PR )
- [ ] Ensure workflows with outstanding LAs can't be evicted ( NEXT PR )
- [x] Implement different cancel types

### DIFF
--- a/sdk-core-protos/src/lib.rs
+++ b/sdk-core-protos/src/lib.rs
@@ -79,18 +79,7 @@ pub mod coresdk {
 
             pub fn cancel_from_details(payload: Option<Payload>) -> Self {
                 Self {
-                    status: Some(aer::Status::Cancelled(Cancellation {
-                        failure: Some(APIFailure {
-                            // CanceledFailure
-                            message: "Activity cancelled".to_string(),
-                            failure_info: Some(failure::FailureInfo::CanceledFailureInfo(
-                                CanceledFailureInfo {
-                                    details: payload.map(Into::into),
-                                },
-                            )),
-                            ..Default::default()
-                        }),
-                    })),
+                    status: Some(aer::Status::Cancelled(Cancellation::from_details(payload))),
                 }
             }
 
@@ -136,6 +125,22 @@ pub mod coresdk {
 
             pub fn cancelled(&self) -> bool {
                 matches!(self.status, Some(activity_resolution::Status::Cancelled(_)))
+            }
+        }
+
+        impl Cancellation {
+            pub fn from_details(payload: Option<Payload>) -> Self {
+                Cancellation {
+                    failure: Some(APIFailure {
+                        message: "Activity cancelled".to_string(),
+                        failure_info: Some(failure::FailureInfo::CanceledFailureInfo(
+                            CanceledFailureInfo {
+                                details: payload.map(Into::into),
+                            },
+                        )),
+                        ..Default::default()
+                    }),
+                }
             }
         }
     }
--- a/sdk-core-protos/src/lib.rs
+++ b/sdk-core-protos/src/lib.rs
@@ -99,6 +99,10 @@ pub mod coresdk {
                     status: Some(aer::Status::WillCompleteAsync(WillCompleteAsync {})),
                 }
             }
+
+            pub fn is_cancelled(&self) -> bool {
+                matches!(self.status, Some(aer::Status::Cancelled(_)))
+            }
         }
 
         impl From<Result<APIPayload, APIFailure>> for ActivityExecutionResult {
@@ -128,6 +132,10 @@ pub mod coresdk {
 
             pub fn failed(&self) -> bool {
                 matches!(self.status, Some(activity_resolution::Status::Failed(_)))
+            }
+
+            pub fn cancelled(&self) -> bool {
+                matches!(self.status, Some(activity_resolution::Status::Cancelled(_)))
             }
         }
     }
--- a/sdk-core-protos/src/lib.rs
+++ b/sdk-core-protos/src/lib.rs
@@ -523,6 +523,12 @@ pub mod coresdk {
             }
         }
 
+        impl Display for RequestCancelLocalActivity {
+            fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+                write!(f, "RequestCancelLocalActivity({})", self.seq)
+            }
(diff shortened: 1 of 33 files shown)


sdk-core-protos/src/lib.rs
```diff
@@ -129,6 +122,26 @@ pub mod coresdk {
...
                 matches!(self.status, Some(activity_resolution::Status::Failed(_)))
             }
+
+            pub fn cancelled(&self) -> bool {
+                matches!(self.status, Some(activity_resolution::Status::Cancelled(_)))
+            }
+        }
+
+        impl Cancellation {
+            pub fn from_details(payload: Option<Payload>) -> Self {
```
cretz, line 132: Pedantic, but maybe "activity_cancelled_with_details" or something if `"Activity cancelled"` is hardcoded in this utility.

src/machines/local_activity_state_machine.rs
```diff
@@ -78,19 +100,32 @@ pub(super) struct ResolveDat {
 pub(crate) enum LocalActivityExecutionResult {
     Completed(Success),
     Failed(ActFail),
+    /// The boolean must be set true if we should not record the marker. EX: If we cancelled before
+    /// dispatching.
+    Cancelled(ActCancel, bool),
```
cretz, line 105: Use named fields instead of simple bool or break into `Cancelled` and `CancelledWithMarker` or something? Doesn't matter to me, just a thought.

src/machines/workflow_machines.rs
```diff
@@ -990,6 +973,48 @@ impl WorkflowMachines {
...
+                    // we sometimes pre-resolve activities when first requested.
+                    if let Some(preres) = self.local_activity_data.take_preresolution(seq) {
+                        if let Machines::LocalActivityMachine(lam) = self.machine_mut(m_key) {
+                            let more_responses = lam.try_resolve_with_dat(preres)?;
+                            self.process_machine_responses(m_key, more_responses)?;
+                        } else {
+                            panic!("A non local-activity machine returned a request cancel LA response");
+                        }
+                    }
+                    // If it's in the request queue, just rip it out.
```
cretz, line 989: Is this what happens when a local activity is started and then cancelled before any commands are sent to the server? What does the history look like when the cancel is slow enough to not be in this queue vs when it is? (just curious how it appears)

src/machines/workflow_machines/local_acts.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// next WFT
+    resolutions: HashMap<u32, ResolveDat>,
+}
+
+impl LocalActivityData {
+    pub(super) fn enqueue(&mut self, act: ScheduleLocalActivity) {
+        self.new_requests.push(act);
+    }
+
+    pub(super) fn enqueue_cancel(&mut self, cancel: ExecutingLAId) {
```
cretz, line 33: I would personally like to see a cleanup PR that goes back through and makes all acronyms consistent (I like pascal case instead of all caps), decide whether to use Dat or Data, Act or Activity, LA or LocalActivity, etc.

src/retry_logic.rs
```diff
@@ -38,15 +39,29 @@ impl RetryPolicyExt for RetryPolicy {
...
             // No retries if initial interval is not specified
             None
         }
     }
 }
 
+const NANOS_PER_SEC: u32 = 1_000_000_000;
+/// modified from rust stdlib since this feature is currently nightly only
+fn try_from_secs_f64(secs: f64) -> Option<Duration> {
+    const MAX_NANOS_F64: f64 = ((u64::MAX as u128 + 1) * (NANOS_PER_SEC as u128)) as f64;
```
cretz, line 56: Heh, I just saw that `9223372036854775807000000000` can't be represented as f64 but `9223372036854775808000000000` can.

General question, why can't we always operate in nanos and just do integer math? Go uses i64 as a duration and says "+/- 290 years is good enough".

cretz (approved):
Initial review complete. Be advised that I don't know enough from a logic POV to determine right from wrong.

sdk-core-protos/src/lib.rs
Sushisource: This struct can only go inside of an activity result, so that's implied by the types - the `from_` prefix is idiomatic rust here when consuming one value to produce another.

src/retry_logic.rs
Sushisource: Just because then it'd be harder to use `Duration`s everywhere which are extremely convenient

src/machines/workflow_machines/local_acts.rs
Sushisource: Yep, we can do that at some point.

src/machines/workflow_machines.rs
Sushisource: Yes. When the cancel happens before lang knows about the activity, no history entry is created.

src/machines/local_activity_state_machine.rs
Sushisource: :+1:

sdk-core-protos/src/lib.rs
cretz: Ah! I was using that contextual info at the top of the GH bar and thought this was in `coresdk`, I didn't see it was in `activity_result` mod.

src/retry_logic.rs
cretz: I think you could use `Duration::from_nanos` and `Duration::as_nanos` when you need to go from integer scalar to/from `Duration` (unless I'm misunderstanding).

Sushisource: Oh, you can - but at least for this part the multiplication factor is in f64 and all this hoop jumping is to ensure the nanos portion *isn't* screwed up - it's actually just copied from the nightly stdlib for the most part (with the real error type stripped out), so I trust this method makes sense.

cretz: K. I just wince at floats as a reflect. I prefer int math, but obv doesn't matter much here as whatever precision may be lost is minimal regardless of primitive chosen.

Sushisource:
Something stupid happened with git and I had to recreate the PR here https://github.com/temporalio/sdk-core/pull/243

Issue #246, opened 2021-12-29
[Bug] Worker should accept activity completions and heartbeats after it has been shutdown


bergundy:
It seems like we don't wait for that when we shutdown the worker: https://github.com/temporalio/sdk-core/blob/a0e40add5382b0b59ab5f8f0a5151890a41a3867/src/worker/mod.rs#L213

See repro here: https://github.com/temporalio/sdk-typescript/pull/439/files#diff-7f668d27625b25e6cee9b071034059939dde93a8b1fb41f6bc06091bf5c8db91R1

PR #250, opened 2022-01-05
Make shutdown wait on outstanding activity tasks

Sushisource:
## Why?
Shutdown would finish while activities are still running, missing giving the worker a chance to complete any that are outstanding. This can easily hang forever now, so a timeout must be imposed on shutdowns. This can easily be added to core but is currently done in TS.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/246

2. How was this tested:
Tests added / updated

3. Any docs updates needed?

### DIFF
--- a/src/core_tests/activity_tasks.rs
+++ b/src/core_tests/activity_tasks.rs
@@ -31,11 +31,11 @@ use temporal_sdk_core_protos::{
     },
     temporal::api::workflowservice::v1::{
         PollActivityTaskQueueResponse, RecordActivityTaskHeartbeatResponse,
-        RespondActivityTaskCompletedResponse,
+        RespondActivityTaskCanceledResponse, RespondActivityTaskCompletedResponse,
     },
 };
 use test_utils::{fanout_tasks, start_timer_cmd};
-use tokio::{sync::Notify, time::sleep};
+use tokio::{join, sync::Notify, time::sleep};
 
 #[tokio::test]
 async fn max_activities_respected() {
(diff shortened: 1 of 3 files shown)


src/worker/activities.rs
```diff
@@ -119,6 +122,13 @@ impl WorkerActivityTasks {
         self.poller.notify_shutdown();
     }
 
+    /// Wait for all outstanding activity tasks to finish
+    pub(crate) async fn wait_all_finished(&self) {
+        while !self.outstanding_activity_tasks.is_empty() {
+            self.complete_notify.notified().await
+        }
```
bergundy, line 129: How about this along with the next suggestion?
```suggestion
        self.complete_notify.notified().await
```

bergundy: Or maybe even with a oneshot channel

src/worker/activities.rs
```diff
@@ -197,6 +207,7 @@ impl WorkerActivityTasks {
             self.heartbeat_manager.evict(task_token.clone());
             let known_not_found = act_info.known_not_found;
             drop(act_info); // TODO: Get rid of dashmap. If we hold ref across await, bad stuff.
+            self.complete_notify.notify_waiters();
```
bergundy, line 210: (or with a oneshot channel)
```suggestion
            if self.outstanding_activity_tasks.is_empty() {
              self.complete_notify.notify_one();
            }
```

src/core_tests/activity_tasks.rs
```diff
@@ -511,12 +552,66 @@ async fn only_returns_cancels_for_desired_queue() {
...
+            activity_id: "act1".to_string(),
+            heartbeat_timeout: Some(Duration::from_millis(1).into()),
+            ..Default::default()
+        }],
+    ));
+
+    let act = core.poll_activity_task(TEST_Q).await.unwrap();
+    // Start shutdown before completing the activity
+    let shutdown_fut = core.shutdown();
+    join!(shutdown_fut, async {
```
bergundy, line 603: It would be nice to have the test prove that completing the activity is what unblocks core completion

bergundy (approved):
Had some comments, overall this fix is good

src/worker/activities.rs
Sushisource: In principle you could use this function anywhere to wait for all current oustanding tasks to complete (not just while shutting down) which is nice. Oneshot would prevent that.

PR #251, opened 2022-01-06
Fix ActivityTaskManager shutdown sequence

bergundy:
Use a watch channel in order to guarantee shutdown of ActivityTaskManager does not get stuck and wait for poll buffer to be emptied before waiting for outstanding activities to be drained.

### DIFF
--- a/src/worker/activities.rs
+++ b/src/worker/activities.rs
@@ -30,7 +30,8 @@ use temporal_sdk_core_protos::{
         workflowservice::v1::PollActivityTaskQueueResponse,
     },
 };
-use tokio::sync::{Notify, Semaphore};
+use tokio::sync::{watch, Semaphore};
+use tokio_util::sync::CancellationToken;
 
 #[derive(Debug, derive_more::Constructor)]
 struct PendingActivityCancel {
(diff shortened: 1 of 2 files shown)


src/worker/activities.rs
```diff
@@ -178,11 +188,13 @@ impl WorkerActivityTasks {
                                 work.heartbeat_timeout.clone()
                             ),
                         );
+                        self.outstanding_watch_tx.send(self.outstanding_activity_tasks.len()).expect("watch channel should not be dropped");
                         // Only permanently take a permit in the event the poll finished properly
                         sem.forget();
                         Ok(Some(ActivityTask::start_from_poll_resp(work)))
                     }
                     None => {
+                        self.poller_finished_token.cancel();
```
Sushisource, line 197: Pretty sure you're hanging because this won't be called if poll isn't. I don't think there's really any reasonable way to deal with that either. Lang might just not call poll and then, well, too bad. I think you could probably just eliminate this token.

Notifying the pollers that they should stop first is the important part. In principle something maybe could be in the buffer I think... but we can only fix that by requiring lang to poll until empty. That's not necessarily unreasonable, but I don't know how important it is.

Sushisource: One workaround is maybe only wait on the token if there *is* anything in the buffer. I think the poll buffer exposes a fn to check if that's true.

bergundy: I think it's reasonable to expect lang to poll until it gets a shutdown error, that's what we do today anyways.
I'll work on fixing the tests so that they poll until shutdown is complete.

Sushisource: I would still do the check if the buffer is empty. That way the tests don't have to poll just to get an empty response.

bergundy:
I'll reopen this later.

Issue #220, opened 2021-11-12
[Feature Request] Split into more crates to allow shared workflows between integ/replay tests


Sushisource:
Right now there's no good way to define shared workflows in the `test_utils` crate because it depends on core in non-dev mode, but core depends on it in dev-mode, and thus if you try to use shared workflows in unit tests types mismatch and everything breaks.

Seemingly the only real way to fix this is to split out all the code that deals with mocking server responses and building fake histories into yet another crate (as well as possibly another crate for all the gateway stuff, possibly the pollers abstraction) so that we can define tests which can use mock responses and the history builder etc and also depend on the shared workflows.

Pretty big and annoying chunk of work but ultimately will result in a nicer structure

Sushisource:
Almost certainly makes sense to just do https://github.com/temporalio/sdk-core/issues/176 at the same time too

Issue #176, opened 2021-08-19
Move prototype rust sdk to separate crate


Sushisource:
We need to move this to its own crate, since it'll need to be one long run and to prevent accruing bad dependencies.

tedbyron:
@Sushisource tried separating crates, i'm getting one error in `core_tests`:

```rs
error[E0277]: the trait bound `CoreSDK: temporal_sdk_core::Core` is not satisfied
  --> src\core_tests\child_workflows.rs:32:42
   |
32 |     let mut worker = TestRustWorker::new(Arc::new(core), TEST_Q.to_string(), None);
   |                                          ^^^^^^^^^^^^^^ the trait `temporal_sdk_core::Core` is not implemented for `CoreSDK`
   |
   = note: required for the cast to the object type `dyn temporal_sdk_core::Core`
```

Is this because CoreSDK isn't `pub`?

There is a cyclical dependency between the prototype sdk and the core sdk (prototype sdk a dev-dep in core, required by `core_tests`), but it still compiles and can run `cargo integ-test` just fine. IDE won't be happy about the cyclical deps though. I think this can be fixed in the future by not using path dependencies, once api is more stable.

[tedbyron/sdk-core@05a1b70](https://github.com/tedbyron/sdk-core/tree/issue176)

Sushisource:
@tedbyron Thanks for giving this a shot! The reason you ran into that error is because of the dependency "cycle". The problem is  described here: https://github.com/temporalio/sdk-core/issues/220

I have no problem making `CoreSDK` pub, but as you found that's not the root issue here.

To really resolve this, it may need to be the case that all the tests which currently use the prototype SDK need to be lifted up and out of the core crate into the integ tests. The reason is that when you run unit tests, the crate is complied in dev mode, but when the prototype-sdk crate you just separated out is compiled, it is compiled in release mode. This is why the "cycle" is permitted, because technically they are totally different crates which is why you get weird type errors where the types look the same but aren't.

And then of course, those tests use a bunch of stuff like these mock pollers and history builders etc, and that'll have to be accessible, etc etc. It's a pretty big amount of stuff to shuffle around. Probably results in splitting core into more sub crates.

If you'd like to continue on the path of trying to get it going please let me know, I'm happy to answer questions etc. It's a pretty big chunk of work though for sure.

tedbyron:
[...]

PR #252, opened 2022-01-06
Great & Glorious Crate Refactoring

Sushisource:
## What was changed
Split things up into way more crates
* Rust sdk now its own crate
* Client now its own crate
* API surface now its own crate

## Why?
Necessary to split out Rust SDK, so this was inevitable, and puts us in a good position for starting FFI work. Also faster compile times in some cases.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/220 & closes https://github.com/temporalio/sdk-core/issues/176

2. How was this tested:
Existing tests

3. Any docs updates needed?


cretz (approved):
Since this is just restructuring, I am marking approved without reading the diff.

One suggestion I would make is to not have a top-level module at all, and take `src` and put it under another dir too.

PR #253, opened 2022-01-07
Fix possible too-early shutdown case

Sushisource:
## What was changed
Fix scenario where workflow activation polling could return shutdown earlier than it should, when there are still outstanding workflow tasks.

## Why?
Need to complete outstanding tasks

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -14,7 +14,7 @@ pub use crate::retry::RetryGateway;
 
 use crate::metrics::{svc_operation, MetricsContext};
 use backoff::{ExponentialBackoff, SystemClock};
-use futures::{future::BoxFuture, task::Context, Future, FutureExt};
+use futures::{future::BoxFuture, task::Context, FutureExt};
 use http::uri::InvalidUri;
 use opentelemetry::metrics::Meter;
 use std::{
@@ -48,6 +48,9 @@ use tonic::{
 use tower::ServiceBuilder;
 use url::Url;
 use uuid::Uuid;
+
+#[cfg(any(feature = "mocks", test))]
+use futures::Future;
 
 static LONG_POLL_METHOD_NAMES: [&str; 2] = ["PollWorkflowTaskQueue", "PollActivityTaskQueue"];
 /// The server times out polls after 60 seconds. Set our timeout to be slightly beyond that.
--- a/core/src/core_tests/workers.rs
+++ b/core/src/core_tests/workers.rs
@@ -408,7 +408,6 @@ async fn complete_with_task_not_found_during_shutdwn() {
         complete_order.borrow_mut().push(2);
     };
     let complete_fut = async {
-        tokio::time::sleep(Duration::from_millis(50)).await;
         core.complete_workflow_activation(WfActivationCompletion::from_cmds(
             TEST_Q,
             res.run_id,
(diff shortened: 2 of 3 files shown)


core/src/worker/mod.rs
```diff
@@ -422,7 +420,15 @@ impl Worker {
             tokio::select! {
                 biased;
 
-                r = self.workflow_poll().map_err(Into::into) => return r,
+                r = self.workflow_poll().map_err(Into::into) => {
+                    if matches!(r, Err(PollWfError::ShutDown)) {
+                        // Don't actually return shutdown until workflow tasks are drained.
+                        // Outstanding tasks being completed will generate new pending activations
+                        // which will cause us to abort this function.
+                        self.all_wfts_drained().await;
```
bergundy, line 428: Oh, nice approach, maybe we can do the same for activities

core/src/worker/mod.rs
```diff
@@ -791,9 +797,11 @@ impl Worker {
             })
     }
 
-    /// Returns true if shutdown has been requested and there are no more outstanding WFTs
-    fn all_wfts_drained(&self) -> bool {
-        *self.shutdown_requested.borrow() && self.outstanding_workflow_tasks() == 0
+    /// Resolves when shutdown has been requested and there are no more outstanding WFTs
+    async fn all_wfts_drained(&self) {
+        while !(*self.shutdown_requested.borrow() && self.outstanding_workflow_tasks() == 0) {
```
bergundy, line 802: You probably don't need `shutdown_requested` here if you're only calling this after a shutdown error.

If you feel like we do need it, it might be better to first wait for shutdown and then wait for no outstanding tasks.

core/src/core_tests/workers.rs
```diff
@@ -373,3 +374,50 @@ async fn can_shutdown_local_act_only_worker_when_act_polling(
...
+            res.jobs[0].variant,
+            Some(wf_activation_job::Variant::RemoveFromCache(_))
+        );
+        core.complete_workflow_activation(WfActivationCompletion::empty(TEST_Q, res.run_id))
+            .await
+            .unwrap();
+        complete_order.borrow_mut().push(2);
+    };
+    let complete_fut = async {
+        tokio::time::sleep(Duration::from_millis(50)).await;
```
bergundy, line 411: Is this going to be racy? What is this sleep supposed to achieve?

client/src/lib.rs
```diff
@@ -49,6 +49,9 @@ use tower::ServiceBuilder;
 use url::Url;
 use uuid::Uuid;
 
+#[cfg(any(feature = "mocks", test))]
```
bergundy, line 52: Where are we using this feature?

bergundy (commented):
The more I think about it the more I feel like the way shutdown is implemented today is error prone.

It'd make more sense IMHO to have 2 separate calls for shutdown:
- `shutdown` - initiate shutdown sequence
- `dispose` - dispose the core instance

That's exactly the way lang interacts with the bridge and they approach also will give us graceful shutdown OOTB.

https://github.com/temporalio/sdk-typescript/blob/2232465a4f9b0cade28f0c21c2d7856053678728/packages/worker/src/worker.ts#L225

client/src/lib.rs
Sushisource: It's needed by the unit tests since the client is a different crate now and needs to export the mock

bergundy (approved):

PR #255, opened 2022-01-10
Proto refactor

### LINKED ISSUE #247: [Feature Request] Make protos consistent and in same package

Protos in `protos/local/` are inconsistently named and are difficult to work with in some languages do to sharing a directory but being unnecessarily put into separate packages.

Put protos of the same directory into the same package (or move each proto to its own directory more reflective of its package). Not only does this help some code generation (it's nice when your protos are in the same dir that your gen'd code is too, but conflicts abound in languages that leverage proto packages for language packages), but it also makes sense from a discoverability perspective (most protos in the wild have the import dir reflect part of the package, not the import file).

There is also no reason to have two of the same `Failure` types for example.

Also disambiguate the names and properly qualify them. It makes little sense to have `coresdk.activity_task.Start` and `coresdk.child_workflow.ChildWorkflowResult`. While changing the latter to `Result` would technically fix this inconsistency, changing the former to `ActivityStart` would be clearer.

IMO we should change the proto package to `temporal.sdk.core` (can suffix `.v1` if you _really_ want).

Fix all the other little inconsistencies (e.g. change `WFActivation` to `WorkflowActivation`).

### Additional context

I can take this on myself soon.

I am planning on making a single gRPC service like:

```
service Core {
  rpc Stream(stream StreamRequest) returns (stream StreamResponse);
}
```

But we need the protos cleaned up first.

cretz:
## What was changed

* Moved protos with separate packages into separate dirs
* Changed inconsistent `wf_*` and `Wf*` to `workflow_` and `Workflow*` respectively

One thing I didn't do but I want to do is change the `coresdk` package base to `temporal.sdk.core`. Some discussions need to be had here.

## Why?

* It is common practice to not have different proto packages in the same dir
  * Often code generators may choose to use the proto dir/import path instead of the proto package name to choose where to place files
  * It is confusing for an importer to not know based on `import` what proto package is being used

## Checklist

1. Closes #247


sdk-core-protos/src/lib.rs
```diff
@@ -30,9 +30,9 @@ pub mod coresdk {
...
         iter::FromIterator,
     };
-    use workflow_activation::{wf_activation_job, WfActivationJob};
+    use workflow_activation::{workflow_activation_job, WorkflowActivationJob};
     use workflow_commands::{workflow_command, workflow_command::Variant, WorkflowCommand};
-    use workflow_completion::{wf_activation_completion, WfActivationCompletion};
+    use workflow_completion::{workflow_activation_completion, WorkflowActivationCompletion};
 
     #[allow(clippy::module_inception)]
     pub mod activity_task {
```
cretz, line 38: For the most part I don't think it's best to add hand-written code to the same module as the generated code for extension purposes.

https://github.com/temporalio/sdk-core/blob/2fe8665cbcf0048c91440589a6276728bbbc8113/sdk-core-protos/src/lib.rs#L1148-L1153

That is a reasonable proto package name and mimics a name I'd like to do for the core protos (not the `.v1` suffix since internal, but `temporal.sdk.core` package prefix). But I have not yet because of the concern around this ugliness. Thoughts?

bergundy (approved):
I'd use this opportunity and see if there are other proto changes we'd like to make.
I think we have some other inconsistencies.

cretz:
@bergundy - I haven't dug, but not sure we have to do it in this PR.

cretz:
Looks like my build is failing, will investigate shortly

sdk-core-protos/src/lib.rs
Sushisource: The implementations can be moved to different modules in the same crate if needed, though I'm not really sure it buys us much since all the actual generated code is still in separate files.

The package renaming I'd only bother with if it really makes a real improvement in another lang or something. I get that it'd look more cannonical but I'm not sure we care much given this isn't really a public API

Sushisource (approved):

bergundy:
> @bergundy - I haven't dug, but not sure we have to do it in this PR. But if you can think of any specific simple ones, let me know and I can add them here.

I took a quick look and could find anything.
@lorensr pointed something out a while back but I can't remember what now.

lorensr:
I noticed variation, both internally and compared to Temporal API, in naming fields of repeated Payloads. Don't know if matching Temporal API is desired (eg `QueryWorkflow.arguments` matching [`WorkflowQuery.query_args`](https://github.com/temporalio/api/blob/master/temporal/api/query/v1/message.proto#L43)), or if we want internal consistency, like `QueryWorkflow.arguments` and `SignalWorkflow.input`:

image

lorensr:
Things like:

```
startWorkflow, 'arguments'),
queryWorkflow, 'arguments'),
signalWorkflow, 'input'),
scheduleActivity, 'arguments'),
continueAsNewWorkflowExecution, 'arguments'),
startChildWorkflowExecution, 'input'),
signalExternalWorkflowExecution, 'args'),
```

cretz:
@lorensr - I support either approach (match API or be internally consistent). @bergundy - I don't think this relates to this PR nor can I see a benefit in hashing it out then adding it to this PR as opposed to a successive one.

Issue #254, opened 2022-01-07
[Feature Request] Add file-based replay facility


Sushisource:
We need the ability to do replay from history files grabbed from server.

This will also make life easier for including rust-only tests in the feature tests repo.

PR #256, opened 2022-01-11
Add history replay from proto binaries

Sushisource:
## What was changed
Added capability to replay history from file

## Why?
Easier compat testing in the sdk features repo, and easy replay verification for integ tests.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/254

2. How was this tested:
Added one integ test. Will look into upgrading integ tests which use workflow functions (instead of manual activation responses) so they download & replay their own history after finishing.

3. Any docs updates needed?


Sushisource:
Note the serde stuff added to the protos isn't actually necessary here, since we're using the binary for this, but it's convenient to have regardless

README.md
```diff
@@ -65,12 +65,19 @@ sure you are running the collector via docker, which can be done like so:
...
 
 This repo uses a subtree for upstream protobuf files. The path `protos/api_upstream` is a
 subtree. To update it, use:
 `git subtree pull --prefix protos/api_upstream/ git://github.com/temporalio/api.git master --squash`
 
+## Fetching Histories
+Tests which would like to replay stored histories rely on that history being made available in
+binary format. You can fetch histories in that format like so (from a local docker server):
+
+`cargo run --bin histfetch {workflow_id}`
```
bergundy, line 78: We should support run_id too for good measure.

README.md
```diff
@@ -65,12 +65,19 @@ sure you are running the collector via docker, which can be done like so:
...
+
+You can change the `TEMPORAL_SERVICE_ADDRESS` env var to fetch from a different address.
```
bergundy, line 80: I'm fine with env vars since this is an internal tool but in general a CLI option would be better.

sdk-core-protos/src/lib.rs
```diff
@@ -258,75 +259,7 @@ pub mod coresdk {
     }
 
     pub mod external_data {
-        use prost_types::{Duration, Timestamp};
-        use serde::{Deserialize, Deserializer, Serialize, Serializer};
         tonic::include_proto!("coresdk.external_data");
-
-        // Buncha hullaballoo because prost types aren't serde compat.
-        // See https://github.com/tokio-rs/prost/issues/75 which hilariously Chad opened ages ago
```
bergundy, line 266: Nice! I'm assuming histories are still compatible, right?

test_utils/src/history_replay.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    let wf_clone = wf.clone();
+    mg.expect_start_workflow().returning(move |_, _, _, _, _| {
+        Ok(StartWorkflowExecutionResponse {
+            run_id: wf_clone.run_id.clone(),
+        })
+    });
+
+    mg.expect_poll_workflow_task().returning(move |_, _| {
+        let mut resp = hist_info.as_poll_wft_response(TEST_Q);
```
bergundy, line 33: Does this return the history only once? Does it matter?

bergundy (commented):
I'm wondering if there should be an easier way to replay workflows without requiring instantiating an entire core.

Also, are we sure about prost-wkt? Looks like it has 10 stars and the docs build fails.

sdk-core-protos/src/lib.rs
Sushisource: This is what using prost-wkt gives us. I would prefer pbjson but it has this problem https://github.com/influxdata/pbjson/issues/23

Perhaps I'll just revert this for now and keep my manual solution rather than adding all these deps.

test_utils/src/history_replay.rs
Sushisource: It will return it for every poll call.

sdk-core-protos/src/lib.rs
Sushisource: Really I just want this upstreamed into prost

bergundy (approved):

Issue #248, opened 2022-01-04
[Feature Request] Create single bidirectional stream w/ gRPC and FFI interfaces


cretz:
### Is your feature request related to a problem? Please describe.

Right now communication w/ core is over the primary 12 trait functions, mostly using protos to communicate. This adds overhead to those using core directly, remotely, etc.

### Describe the solution you'd like

* Complete #247 
* Add a single, thread-safe bidirectional stream of proto in/out representing all communication with core (can use gRPC messages from  next bullet point as the in/out)
* Expose the single bidirectional stream over a gRPC service like `service Core { rpc Stream(stream StreamRequest) returns (stream StreamResponse) {} }`
* Expose the single bidirectional stream over C FFI in a separate crate

Benefits:

* Makes things very easy for lang impls
* Enables hosted/remote core

Sushisource:
@cretz Why just one method? We could define it as a service with the existing set of methods and I think that'd be a bit more well defined / readable. Additionally it'll require some pretty big restructuring to fit everything into that pattern. Things like shutdown, for example, will I think require some pretty substantial change.

I'm not sure what overhead exists right now vs. this approach either.

I'm thinking it might make more sense to just create the crate that exposes whatever we want for a C FFI on top of the existing APIs. I don't think it's much effort, and it'll accomplish the same goal (except remote Core, but it's not clear at all that we need that, and it doesn't prevent us from doing it later)

cretz:
@Sushisource - maybe not one method, maybe two. The primary reason for a single point is to make FFI easier. It's much easier to have one proto request type than 12. This also affects error handling and other things. The other benefit is that it might be easier to put gRPC and FFI on top of a single in/out engine.

But I don't hold too strong of an opinion here. I was literally just thinking about taking single requests w/ `oneof`s and calling the related `Core` trait. The core side doesn't need to do anything special, it's just a very thin layer multiplexing 12 methods into 1. But yes, the FFI and gRPC layers can leave the calls uncombined, it's not that big of an improvement.

> I'm thinking it might make more sense to just create the crate that exposes whatever we want for a C FFI on top of the existing APIs

That is fine, this is just a suggestion to be simpler. This affects error handling and typing and such. Maybe we can meet in the middle. One of my goals was to only have to create core requests one way on the lang side, but having 2 ways to call core (FFI and gRPC) means I have two request builders, response handlers, error handlers, etc.

[...]

PR #259, opened 2022-01-12
Initial FFI Draft

cretz:
## What was changed

Initial impl of bridge FFI. This is still in development so lacks a lot of docs.

## Why?

Go and others need a C API to communicate with core.

## Notes

1. See #248
2. This adds a `bridge` set of protos which is essentially just the `Core` trait wrapped in proto form. It also provides a gRPC stream service proto.
3. This adds a `bridge-ffi` crate which exposes core over C API
4. This may add a `bridge-grpc` crate which exposes core over a gRPC stream
5. We have discussed and decided that protobuf is the encoding language of choice. The serialization/deserialization on both sides of the boundary comes with a known performance cost.
    * This performance cost already exists in core as protobuf is already being used as the encoding between core and lang
    * If we want to improve this, I suggest either https://google.github.io/flatbuffers/ or https://capnproto.org/, both of which operate on byte arrays without explicit serialization and have decent cross-language support
6. Since core calls are async, we have chosen C function pointer callbacks to report results. There are upsides and downsides to this approach:
    * Anything can be built on callbacks, including event loops or whatever
    * Callbacks do require the lang to respect speed inside the callback. The Rust thread is held up while a callback is executed, so lang must do what it needs to do inside the callback to get the data on its other thread for working (resolve a JS promise, put on a Go channel, etc).
    * As most C APIs do, we allow user data in the callback. This helps callers access data inside the callback and is a very common approach.
    * We could have done an event loop, but you essentially are using function pointers anyways, the only difference is you've now made the response handling single threaded which doesn't benefit languages that can use more than one thread.
7. This includes a Go example and may include examples of other languages. The examples may not ever be in the final product, we just need to prove things work


bridge-ffi/include/sdk-core-bridge.h
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * A set of bytes owned by Core. This must always be passed to
+ * tmprl_bytes_free when no longer in use.
+ */
+typedef struct tmprl_bytes_t {
+  const uint8_t *bytes;
+  size_t len;
+  size_t cap;
+  bool disable_free;
```
Sushisource, line 29: What's this param for?

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+
+    fn into_raw(self) -> *mut tmprl_bytes_t {
+        Box::into_raw(Box::new(self))
+    }
+}
+
+// We only impl this because they are required by lazy_static and raw pointers
+// are not usually safe for send
+unsafe impl Send for tmprl_bytes_t {}
+unsafe impl Sync for tmprl_bytes_t {}
```
Sushisource, line 45: Do we need to even use lazy static for these defaults? Could just construct them every time and avoid this. If we do this, we have to ensure all C side implementations never use thread-local locking etc, or they could end up violating safety

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+
+/// A core instance owned by Core. This must be passed to tmprl_core_shutdown
+/// when no longer in use which will free the resources.
+pub struct tmprl_core_t {
+    tokio_runtime: std::sync::Arc<tokio::runtime::Runtime>,
+    // TODO(cretz): Any concerns with the dynamic dispatch overhead here?
+    // TODO(cretz): Should we ask Rust SDK to give us back a struct instead?
+    // TODO(cretz): We could use generics, but impl traits don't cross the FFI
+    // boundary properly
+    core: std::sync::Arc<dyn temporal_sdk_core_api::Core>,
```
Sushisource, line 143: In my latest PR the real struct may be used directly so we can do that, but I'm not concerned about dispatch overhead

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+        unsafe {
+            Box::from_raw(runtime);
+        }
+    }
+}
+
+/// A core instance owned by Core. This must be passed to tmprl_core_shutdown
+/// when no longer in use which will free the resources.
+pub struct tmprl_core_t {
+    tokio_runtime: std::sync::Arc<tokio::runtime::Runtime>,
```
Sushisource, line 138: Why does this guy have it's own runtime instead of the `tmprl_runtime_t`?

Additionally it probably makes sense to have this type be an alias for an `Arc` around a struct with these fields rather than using two `Arc`s, if both fields are needed.

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+/// present, the core instance must be freed via tmprl_core_shutdown when no
+/// longer in use. The third parameter of the callback is a byte array for a
+/// InitResponse protobuf message which must be freed via tmprl_bytes_free.
+type tmprl_core_init_callback = unsafe extern "C" fn(
+    user_data: *mut libc::c_void,
+    core: *mut tmprl_core_t,
+    resp: *const tmprl_bytes_t,
+);
+
+// TODO(cretz): Catch unwind for badly behaved callbacks?
```
Sushisource, line 158: I don't think we need to -- lang should never panic in its callback handlers. If we prevent it we are likely in a broken state anyway

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+                callback(
+                    user_data,
+                    std::ptr::null_mut(),
+                    tmprl_bytes_t::from_vec(resp.encode_to_vec()).into_raw(),
+                );
+            }
+            return;
+        }
+    };
+    let user_data = Box::new(UserDataHandle(user_data));
```
Sushisource, line 195: Is this box for deallocating the memory after the task runs? An explicit `drop` inside the task at the end might make that more clear

Sushisource: Looks like this pattern is in a few places

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+
+impl tmprl_core_t {
+    async fn new(
+        tokio_runtime: std::sync::Arc<tokio::runtime::Runtime>,
+        req: bridge::InitRequest,
+    ) -> Result<tmprl_core_t, String> {
+        let mut core_opts = temporal_sdk_core::CoreInitOptionsBuilder::default();
+        if let Some(req_gateway_opts) = req.gateway_options {
+            let mut gateway_opts = temporal_sdk_core::ServerGatewayOptionsBuilder::default();
+            if !req_gateway_opts.target_url.is_empty() {
```
Sushisource, line 385: All this conversion goop for all these structs should be moved into `TryInto` implementations.

I would also create a dedicated module/file for it.

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+            .poll_activity_task(&req.task_queue)
+            .await
+            .map_err(|err| bridge::poll_activity_task_response::Error {
+                message: format!("{}", err),
+            })
+    }
+
+    fn borrow_buf(&mut self) -> Vec<u8> {
+        // TODO(cretz): Implement thread-safe byte pool?
+        Vec::new()
```
Sushisource, line 615: Even if we don't use a pool, makes sense to pass capacity here so the vec can be created with it immediately rather than calling `reserve` subsequently and performing another allocation

Sushisource (commented):
Cool, this overall makes sense to me.

I think we want to double make sure we're getting the semantics right around the raw pointer handling, but I don't think that's too bad and we really only need to do it once. We should try to encapsulate that stuff as much as we reasonably can so the tricky bits are well contained. Perhaps some very simple macros for some of the common patterns? Just thinking out loud.

The gRPC service definition looks like it isn't used yet - and my understanding is it won't be until (if) we actually expose a gRPC endpoint, correct?

bridge-ffi/include/sdk-core-bridge.h
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * A core instance owned by Core. This must be passed to tmprl_core_shutdown
+ * when no longer in use which will free the resources.
+ */
+typedef struct tmprl_core_t tmprl_core_t;
+
+/**
+ * A runtime owned by Core. This must be passed to tmprl_runtime_free when no
+ * longer in use. This must not be freed until every call to every tmprl_core_t
+ * instance created with this runtime has been shutdown.
```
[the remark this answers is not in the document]
cretz: I abstract Tokio runtime creation, so they just create a generic runtime. They don't even know that this is backed by Rust much less Tokio. They just get a header and a static lib.

bridge-ffi/include/sdk-core-bridge.h
cretz: I figured it wasn't worth creating this struct for empty responses since they happen so often, so I have them as lazy_statics which means I don't want them freed.

bridge-ffi/src/lib.rs
cretz: Yeah, it may be an over-optimization.

> If we do this, we have to ensure all C side implementations never use thread-local locking etc, or they could end up violating safety

This shouldn't be affected by any of that because from a caller/lang perspective, `tmprl_bytes_t` are immutable.

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+impl Drop for tmprl_bytes_t {
+    fn drop(&mut self) {
+        // In cases where freeing is disabled (or technically some other
+        // drop-but-not-freed situation though we don't expect any), the bytes
+        // remain non-null so we re-own them here
+        if !self.bytes.is_null() {
+            unsafe { Vec::from_raw_parts(self.bytes as *mut u8, self.len, self.cap) };
+        }
+    }
+}
```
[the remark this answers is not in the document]
cretz: Nah, when we do `Box::from_raw`, you disable dropping because you remove ownership. The only reason I added this was for static ones, and they'd only be dropped on entire library unload. It's just a memory safety mechanism for those cases since `tmprl_bytes_free` is disabled for them.

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+
+    static ref DEFAULT_REGISTER_WORKER_RESPONSE_BYTES: tmprl_bytes_t = {
+        tmprl_bytes_t::from_vec_disable_free(bridge::RegisterWorkerResponse::default().encode_to_vec())
+    };
+}
+
+/// A runtime owned by Core. This must be passed to tmprl_runtime_free when no
+/// longer in use. This must not be freed until every call to every tmprl_core_t
+/// instance created with this runtime has been shutdown.
+pub struct tmprl_runtime_t {
```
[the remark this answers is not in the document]
cretz: This is not a `repr(C)` but it is a public C-visible struct, it just appears opaque. I think we should shoot for C header prettiness/ergonomics over all else.

bridge-ffi/src/lib.rs
cretz: > Why does this guy have it's own runtime instead of the tmprl_runtime_t?

They're shared instances of the same runtime. I just felt this read simpler than embedding the runtime struct in here. Technically this is a bit safer because on `tmprl_runtime_free`, that will just decrement a reference count instead of segfault here.

bridge-ffi/src/lib.rs
Sushisource: Ah, OK good point. We should explicitly mention that in the docstring too, C/lang should never modify the bytes.

bridge-ffi/include/sdk-core-bridge.h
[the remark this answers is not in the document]
Sushisource: Sure, that's reasonable

bridge-ffi/src/lib.rs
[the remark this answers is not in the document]
Sushisource: K, shouldn't this check that field then?

bridge-ffi/src/lib.rs
cretz: The box and the `UserDataHandle` is only for raw pointer access across threads. See https://doc.rust-lang.org/nomicon/send-and-sync.html.

I'd have to recheck the compiler error on why I needed to box this struct, but for whatever reason, just marking it as unsafe send/sync was not enough..

bridge-ffi/src/lib.rs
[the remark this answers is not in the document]
Sushisource: Right, got it

bridge-ffi/src/lib.rs
cretz: I didn't originally do `TryFrom` because I would have to either edit the proto project (I don't like adding functions to generated code if I can help it) or or the core one. But I agree I can move the conversions to their own module/file.

bridge-ffi/src/lib.rs
cretz: Per the docs, it doesn't perform two allocations. From the docs:

> if you construct a Vec with capacity 0 via Vec::new, vec![], Vec::with_capacity(0), or by calling shrink_to_fit on an empty Vec, it will not allocate memory.

So `Vec::new()` + `reserve` is effective the same as `Vec::with_capacity`

bridge-ffi/src/lib.rs
Sushisource: Segfault could be prevented w/ custom drop impl, but yeah I get it. 

I would make a field docstring indicating that it's always going to be the same runtime as whatever was in the runtime struct.

It also sorta makes me wonder if it makes sense to expect the runtime to be passed by ownership into `tmprl_core_init`, and then free it on core shutdown finishing. Do we have a reason for C/lang to want to keep that runtime around independently of core? If it wants to do other async-y things, perhaps it makes sense to just use a different runtime for that anyway, and Core always operates on its own dedicated runtime?

bridge-ffi/src/lib.rs
cretz: > We should explicitly mention that in the docstring too, C/lang should never modify the bytes.

Ha, just did this pending push anyways (also mentioned `cap` and `disable_free` are for internal use only)

bridge-ffi/src/lib.rs
Sushisource: I would add it to the proto project. It's not adding to generated code - the generated code is _included_ by that project, so I really don't think it's a problem. It's very reasonable to augment generated code with hand-written code, and if I hadn't done that for the existing protos things would be a huge mess. Being able to use the traits is very desirable as they compose well with other stuff.

bridge-ffi/src/lib.rs
Sushisource: Cool, makes sense (though you'll probably need to add it as a param anyway if you use a pool later, and I think it encapsulates the better)

cretz:
> The gRPC service definition looks like it isn't used yet - and my understanding is it won't be until (if) we actually expose a gRPC endpoint, correct?

@Sushisource - Correct. I included the service definition because the encapsulation of request/response as proto is partially driven by that use case, and I wanted to explain my thoughts on using a stream there instead of unary RPC calls. I also want SDK langs to consider, when being designed, that they may be asked to communicate w/ core over in addition to C.

But it is not used today though I expect a `bridge-grpc` crate to come about soon. There will have to be some more thoughts around server connection reuse across streams so each stream doesn't represent a separate gRPC instance, but that may be affected by your new potential work separating the client from the core instance.

Or maybe a `bridge-web` crate which includes gRPC capabilities but also a web socket + proto json equivalent (using gRPC from a browser is annoying).

bridge-ffi/src/lib.rs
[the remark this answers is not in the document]
cretz: That is the purpose of one part of the comment there: 

> or technically some other drop-but-not-freed situation though we don't expect any

So I check whether the pointer is null instead of whether freeing was disabled (during explicit free, we set this pointer as null for this reason)

bridge-ffi/src/lib.rs
cretz: > I would make a field docstring indicating that it's always going to be the same runtime as whatever was in the runtime struct.

:+1:

> Do we have a reason for C/lang to want to keep that runtime around independently of core? If it wants to do other async-y things, perhaps it makes sense to just use a different runtime for that anyway, and Core always operates on its own dedicated runtime?

I think Tokio's task management and scheduling and thread pool works better if reused. Say you had 10 core instances because you were running on 10 namespaces. A Tokio thread pool shared across all of them in theory may be better performing than separate ones.

By default Tokio will make a thread pool of core count for each runtime. I could also see this being configurable (or other things like thread name) in the future and the separation of runtime and core to the caller is logical since it's the separation that exists.

bridge-ffi/src/lib.rs
cretz: Ok, will do

bridge-ffi/src/lib.rs
cretz: Agreed, though I very likely may become convinced that a pool just isn't worth it assuming the allocator is high performing. I'll leave this borrow/return abstraction here, but remove the `TODO`s and just say that a pool impl can be added here if some microbenchmark shows it has value.

bridge-ffi/example/goffi/ffi.go
```diff
@@ -0,0 +1,169 @@
+package goffi
```
bergundy, line 1: This is pretty much what I'd imagined this to look like for go.

bridge-ffi/example/goffi/ffi.go
```diff
@@ -0,0 +1,169 @@
...
+	defer C.tmprl_bytes_free(req.core.core, bytes)
+	if req.resp != nil {
+		if err := protoFromBytes(bytes, req.resp); err != nil {
+			// TODO(cretz): Warn?
+			req.resp = nil
+		}
+	}
+
+	// Non-blocking channel send
+	select {
```
bergundy, line 165: Is this required with channel buffer size of 1?

bridge-ffi/include/sdk-core-bridge.h
bergundy: I'm wondering if we should have a different name for this like `is_static` or at least document what this flag does and when to use it.

protos/local/temporal/sdk/core/bridge/bridge.proto
```diff
@@ -0,0 +1,188 @@
...
+  google.protobuf.Duration sticky_queue_schedule_to_start_timeout = 10;
+  google.protobuf.Duration max_heartbeat_throttle_interval = 11;
+  google.protobuf.Duration default_heartbeat_throttle_interval = 12;
+}
+
+message RegisterWorkerResponse {
+  Error error = 1;
+
+  message Error {
+    string message = 1;
```
bergundy, line 90: All of these errors should be enums (unions) IMHO so they can be properly handled.

bridge-ffi/src/lib.rs
cretz: After research, I can remove the box alloc and just reference the handle

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+            return;
+        }
+    }
+    let bytes = bytes as *mut tmprl_bytes_t;
+    // Return vec back to core before dropping bytes
+    let vec = unsafe { Vec::from_raw_parts((*bytes).bytes as *mut u8, (*bytes).len, (*bytes).cap) };
+    // Set to null so the byte dropper doesn't try to free it
+    unsafe { (*bytes).bytes = std::ptr::null_mut() };
+    // Return only if core is non-null
+    if !core.is_null() {
```
bergundy, line 75: I'd maybe even panic if core is null here, it never should be

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,650 @@
...
+    static ref DEFAULT_SHUTDOWN_RESPONSE_BYTES: tmprl_bytes_t = {
+        tmprl_bytes_t::from_vec_disable_free(bridge::ShutdownResponse::default().encode_to_vec())
+    };
+
+    static ref DEFAULT_REGISTER_WORKER_RESPONSE_BYTES: tmprl_bytes_t = {
+        tmprl_bytes_t::from_vec_disable_free(bridge::RegisterWorkerResponse::default().encode_to_vec())
+    };
+}
+
+/// A runtime owned by Core. This must be passed to tmprl_runtime_free when no
```
bergundy, line 103: Technically not owned by `Core`, `Core` is spawned in a runtime. We could be more accurate in the terminology here.

bergundy (commented):
Overall lgtm.

We've talked about trying to reuse the protos for the TS bridge without the FFI layer.
This might be the only thing we do share if any.
The request / response layer you added if very similar to the TS bridge but I'm not convinced it makes sense to reuse this code.

I also think that ideally the core trait should just use all of these protos directly instead of having the translation layer, @Sushisource would probably have something to say about that though.

bridge-ffi/example/goffi/ffi.go
cretz: Not technically required, but it's safer to do a non-blocking send when you know you should never block a send

bridge-ffi/include/sdk-core-bridge.h
cretz: My next push is saying this is for internal use only

bridge-ffi/src/lib.rs
cretz: Not true. For example, what if core init fails? The error is returned in a proto but core is nil, yet they should still free the proto. I even specifically call this out on. And really, I could remove this parameter altogether right now as we don't even have a byte pool (that's its only purpose).

bridge-ffi/src/lib.rs
cretz: To clarify, I use the capitalized `Core` to refer to this whole static library (so it's owned by the library), and I use the lowercased `core` to refer to an instance. This statement is saying the caller does not own the runtime, this library does.

protos/local/temporal/sdk/core/bridge/bridge.proto
cretz: By "enums (unions)" I assume you mean "oneof" in protobuf parlance? Proto enums can't carry data.

I am taking the add-as-needed approach, because mapping the entirety of the error space to protos is brittle. Only some need to be properly handled.

bridge-ffi/include/sdk-core-bridge.h
cretz: Done

bridge-ffi/src/lib.rs
cretz: Done

bridge-ffi/src/lib.rs
cretz: Cool. Documented that we accept the perf atm.

bridge-ffi/src/lib.rs
cretz: Removed this TODO. I shouldn't have put it here btw, it is the Rust code where I was wondering if we should catch unwind, but per discussion we want it to crash.

bridge-ffi/src/lib.rs
cretz: Done

bridge-ffi/src/lib.rs
cretz: Could not add to proto project due to needing the core project to translate into. Added to bottom of this file.

bridge-ffi/src/lib.rs
```diff
@@ -81,11 +94,18 @@ pub extern "C" fn tmprl_bytes_free(core: *mut tmprl_core_t, bytes: *const tmprl_
     }
 }
 
-// Used for maintaining pointer to user data across threads
+// Used for maintaining pointer to user data across threads. See
+// https://doc.rust-lang.org/nomicon/send-and-sync.html
```
Sushisource, line 98: Can be a docstring

bridge-ffi/src/lib.rs
```diff
@@ -224,7 +262,7 @@ pub extern "C" fn tmprl_core_init(
 /// protobuf message. The callback is invoked on completion with a
 /// ShutdownResponse protobuf message.
 #[no_mangle]
-#[allow(unused_variables)]
+#[allow(unused_variables)] // We intentionally ignore the request
```
Sushisource, line 265: Don't do this - just use a leading underscore for unused params

bridge-ffi/src/lib.rs
```diff
@@ -274,16 +312,54 @@ pub extern "C" fn tmprl_register_worker(
...
+    });
+}
+
+/// Shutdown registered worker.
+///
+/// The req_proto and req_proto_len represent a byte array for a
+/// ShutdownWorkerRequest protobuf message. The callback is invoked on
+/// completion with a ShutdownWorkerResponse protobuf message.
+#[no_mangle]
+#[allow(unused_variables)]
```
Sushisource, line 335: Same here, just use leading underscores.

bridge-ffi/src/lib.rs
```diff
@@ -362,23 +444,393 @@ pub extern "C" fn tmprl_poll_activity_task(
...
+    });
+}
+
+/// Fetch buffered logs.
+///
+/// The req_proto and req_proto_len represent a byte array for a
+/// FetchBufferedLogsRequest protobuf message. The callback is invoked
+/// on completion with a FetchBufferedLogsResponse protobuf message.
+#[no_mangle]
+#[allow(unused_variables)] // We intentionally ignore the request
```
Sushisource, line 643: And this guy

bridge-ffi/src/lib.rs
```diff
@@ -362,23 +444,393 @@ pub extern "C" fn tmprl_poll_activity_task(
...
+    });
+}
+
+/// Request workflow eviction.
+///
+/// The req_proto and req_proto_len represent a byte array for a
+/// RequestWorkflowEvictionRequest protobuf message. The callback is invoked
+/// on completion with a RequestWorkflowEvictionResponse protobuf message.
+#[no_mangle]
+pub extern "C" fn tmprl_request_workflow_eviction(
```
Sushisource, line 600: Almost all these functions have very similar structure. I think it's worth seeing if we can de-dupe that structure with macro or a higher-order function.

bridge-ffi/src/lib.rs
```diff
@@ -500,27 +952,19 @@ impl tmprl_core_t {
...
-        })
+            .map_err(|err| format!("invalid options: {}", err))
     }
+}
 
-    async fn shutdown(&self) {
-        self.core.shutdown().await;
-    }
+// Present for try-from only
+struct WorkerConfig(bridge::RegisterWorkerRequest);
```
Sushisource, line 962: Let's move the wrapper dudes into their own module

Sushisource:
The Go code might make sense to have in another repo, but I can kinda go either way on that one.

bridge-ffi/src/lib.rs
```diff
@@ -1,18 +1,31 @@
-#![allow(non_camel_case_types)]
+#![allow(
+    // Non-camel-case types needed since this is exported as a C header and we
+    // want C-like underscores in our type names
+    non_camel_case_types,
+
+    // We choose to have narrow "unsafe" blocks instead of marking entire
+    // functions as unsafe. Even the example in clippy's docs at
```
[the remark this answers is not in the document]
cretz: K, will add

bridge-ffi/src/lib.rs
cretz: That affects the header output, I'd like to keep that the way it is (these are Rust rules, they shouldn't affect C ergonomics)

bridge-ffi/src/lib.rs
cretz: Same as other comment

bridge-ffi/src/lib.rs
cretz: Same as other comment

cretz:
> The Go code might make sense to have in another repo, but I can kinda go either way on that one.

@Sushisource - I'd rather just delete it instead of letting it grow stale in a separate repo. I would also be willing to show how to use this with, say, .Net here (it'd only take a couple hours probably). But once someone actually uses this crate, it definitely makes sense to have that serve as the canonical example/usage instead of this. This was more for vetting.

Sushisource:
@cretz Ya, sounds good to me. Let's just delete it before merging then. It might be nice to keep for reference, maybe just publish it to another branch that we can leave around for a bit?

bridge-ffi/src/lib.rs
Sushisource: Riiight. K, if this annotation works being moved to apply to just the arg, then I'd do that, otherwise I'm fine with it.

Alternatively, maybe we just don't have the parameter at all? It's not like it's a huge lift to go refactor users of this package if we ever do add meaningful content to the proto. Not a huge deal though if we don't want to.

cretz: I like the consistency. There's also a case of an empty return type too (shutdown). I kinda followed gRPC philosophy of "always have request and response even if empty" even though you're right, this is internal API and therefore easily refactored when we change.

cretz:
I am closing this draft PR in favor of a real PR.

bridge-ffi/src/lib.rs
cretz: Done

bridge-ffi/src/lib.rs
cretz: Rust does support this per param, so I set them there.

bridge-ffi/src/lib.rs
cretz: Done

PR #260, opened 2022-01-13
Bridge C API

cretz:
## What was changed

Implementation of C API for FFI usage.

## Why?

Go and others need a C API to communicate with core.

## Notes

1. See #248
2. See the draft PR at #259
3. I removed the Go example using this SDK, but I'm gonna try to leave my forked branch around, so it can be seen at https://github.com/cretz/temporal-sdk-core/tree/e22bd2759de6a860489134cf9491e3aec51aa9c5/bridge-ffi/example/goffi
4. This adds a `bridge` set of protos which is essentially just the `Core` trait wrapped in proto form. It also provides a gRPC stream service proto but no gRPC implementation as of yet (may be a `bridge-grpc` crate in the future).
5. This adds a `bridge-ffi` crate which exposes core over C API
6. We have discussed and decided that protobuf is the encoding language of choice. The serialization/deserialization on both sides of the boundary comes with a known performance cost.
    * This performance cost already exists in core as protobuf is already being used as the encoding between core and lang
    * If we want to improve this, I suggest either https://google.github.io/flatbuffers/ or https://capnproto.org/, both of which operate on byte arrays without explicit serialization and have decent cross-language support
7. Since core calls are async, we have chosen C function pointer callbacks to report results. There are upsides and downsides to this approach:
    * Anything can be built on callbacks, including event loops or whatever
    * Callbacks do require the lang to respect speed inside the callback. The Rust thread is held up while a callback is executed, so lang must do what it needs to do inside the callback to get the data on its other thread for working (resolve a JS promise, put on a Go channel, etc).
    * As most C APIs do, we allow user data in the callback. This helps callers access data inside the callback and is a very common approach.
    * We could have done an event loop, but you essentially are using function pointers anyways, the only difference is you've now made the response handling single threaded which doesn't benefit languages that can use more than one thread.


bridge-ffi/build.rs
```diff
@@ -0,0 +1,16 @@
+extern crate cbindgen;
+
+use std::env;
+
+fn main() {
+    let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+
+    cbindgen::Builder::new()
+        .with_crate(crate_dir)
+        // .with_no_includes()
```
Sushisource, line 10: Leftover comment line

bridge-ffi/include/sdk-core-bridge.h
```diff
@@ -0,0 +1,216 @@
+#pragma once
```
Sushisource, line 1: Just double checking we want to check this in?

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+    // API.
+    clippy::not_unsafe_ptr_arg_deref,
+)]
+
+mod wrappers;
+
+use prost::Message;
+use temporal_sdk_core_protos::coresdk::bridge;
+
+extern crate libc;
```
Sushisource, line 20: I don't think this extern declaration is necessary -- if it is style-wise I generally prefer these at the very top, but they should only be necessary with `#[macro_use]`

bridge-ffi/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+                callback(user_data, core.encode_proto(&resp).into_raw());
+            }
+            return;
+        }
+    };
+    let user_data = UserDataHandle(user_data);
+    core.tokio_runtime.clone().spawn(async move {
+        let resp = bridge::PollWorkflowActivationResponse {
+            response: Some(match core.poll_workflow_activation(req).await {
+                Ok(act) => bridge::poll_workflow_activation_response::Response::Activation(act),
```
Sushisource, line 413: Nit: Dunno if we need to prefix everything with `bridge::` rather than just importing directly

Sushisource (approved):
Looking good! Just a few minor things

bridge-ffi/build.rs
cretz: I was struggling with this because I couldn't decide whether I wanted the includes. I think I am more sure I do now, so I can remove this.

bridge-ffi/include/sdk-core-bridge.h
cretz: This is a tough question. Usually in Rust, unlike Go, we don't want generated code in. However, when people are developing, they often want a header file already there. I think there's enough value to leave here.

We could have a CI check make sure it doesn't change on build from what is already in repo, but not sure this PR should wait on that.

bridge-ffi/src/lib.rs
cretz: I can change if you want. This reads better to me, but I don't mind adding a bunch of `use` statements for all the proto types if you want. Of course with this specific case, `Response` is ambiguous, so I'd still need `poll_workflow_activation_response::Response` and import that module, but for things like `bridge::PollWorkflowActivationResponse` I can import the type instead of the module.

bridge-ffi/include/sdk-core-bridge.h
cretz: > We could have a CI check make sure it doesn't change on build from what is already in repo, but not sure this PR should wait on that.

Well, I may have found an easy way to do this, I will confirm and get back

bridge-ffi/build.rs
cretz: Removed line

bridge-ffi/include/sdk-core-bridge.h
cretz: I have added this check. The BuildKite lint step will fail if the cbindgen generates something different than what's in repo with a:

> bridge-ffi's header file changed unexpectedly from what's on disk

Example at https://buildkite.com/temporal/core-sdk/builds/1150#a06e1d7b-d281-47bf-b24b-fbafa143945e

bridge-ffi/src/lib.rs
cretz: Removed

bridge-ffi/include/sdk-core-bridge.h
Sushisource: :+1:

PR #261, opened 2022-01-15
History Replay pt.2

Sushisource:
## What was changed
A bunch of upgrades to make using history replay nice. Developed with TS changes.


cretz:
@Sushisource - looks like my bridge merge caused conflict in root Cargo.toml.

test-utils/src/lib.rs
```diff
@@ -47,23 +49,32 @@ pub async fn init_core_and_create_wf(test_name: &str) -> (Arc<dyn Core>, String)
...
+pub fn init_core_replay_preloaded(test_name: &str, history: &History) -> (Arc<dyn Core>, String) {
+    let replay_core = init_core_replay(get_integ_telem_options());
+    let worker_cfg = WorkerConfigBuilder::default()
+        .task_queue(test_name)
+        .build()
+        .expect("Configuration options construct properly");
+    replay_core
+        .make_replay_worker(worker_cfg, history)
+        .expect("Worker registration works");
+    (Arc::new(replay_core), test_name.to_string())
```
cretz, line 77: I don't think it's worth returning the string they passed in just to match the signature of `init_core_and_create_wf`, but not a big deal

test-utils/src/lib.rs
```diff
@@ -47,23 +49,32 @@ pub async fn init_core_and_create_wf(test_name: &str) -> (Arc<dyn Core>, String)
...
-        .register_worker(starter.worker_config.clone())
-        .await
-        .unwrap();
-    starter.initted_core = Some(Arc::new(core_fakehist));
-    let core = starter.get_core().await;
-    starter.start_wf().await;
-    core
+/// Create a core instance that can be used for replay. See the [ReplayCore] trait for adding
+/// canned history.
+pub fn init_core_replay(opts: TelemetryOptions) -> ReplayCoreImpl {
```
cretz, line 54: This is a pretty core feature, I wonder if it should be in the primary crate. Is the concern having to enable mocks or some other dependency? I expect every lang to expose this feature.

cretz (approved):

test-utils/src/lib.rs
Sushisource: Yeah, enabling mocks adds a good amount of bloat. That said it is true probably every SDK is going to turn it on, so your point is valid for sure. The layout just happens to make this easier for now... but we could potentially move this all inside the normal core package. Different PR for that tho I think

core-api/src/lib.rs
```diff
@@ -27,7 +27,7 @@ pub trait Core: Send + Sync {
     /// Register a worker with core. Workers poll on a specific task queue, and when calling core's
     /// poll functions, you must provide a task queue name. If there was already a worker registered
     /// with the same task queue name, it will be shut down and a new one will be created.
-    async fn register_worker(&self, config: WorkerConfig) -> Result<(), WorkerRegistrationError>;
+    fn register_worker(&self, config: WorkerConfig) -> Result<(), WorkerRegistrationError>;
```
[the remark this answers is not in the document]
Sushisource: It will. I'll fix.

core/src/lib.rs
```diff
@@ -231,20 +231,55 @@ impl CoreSDK {
...
+        let tq = config.task_queue.clone();
+        let mut worker = Worker::new(config, None, gateway, self.metrics.clone());
+        worker.set_post_activate_hook(move |worker| {
+            if worker
+                .wft_manager
+                .most_recently_processed_event(&run_id)
+                .unwrap_or_default()
+                >= last_event
+            {
+                worker.initiate_shutdown();
```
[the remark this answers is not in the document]
Sushisource: Good point. I should make a test that intentionally has nondeterminism and make sure we report that properly

bridge-ffi/src/lib.rs
```diff
@@ -319,17 +319,15 @@ pub extern "C" fn tmprl_register_worker(
...
-            Ok(()) => unsafe {
-                callback(user_data.into(), &*DEFAULT_REGISTER_WORKER_RESPONSE_BYTES);
-            },
-            Err(err) => {
-                let resp = bridge::RegisterWorkerResponse { error: Some(err) };
-                unsafe { callback(user_data.into(), core.encode_proto(&resp).into_raw()) };
-            }
+    match core.register_worker(wrappers::WorkerConfig(req)) {
+        Ok(()) => unsafe {
+            callback(user_data.into(), &*DEFAULT_REGISTER_WORKER_RESPONSE_BYTES);
```
bergundy, line 324: Should this be turned into a sync api and drop the callback completely? Do we even have any of those for the FFI layer (don't remember)?

core/src/lib.rs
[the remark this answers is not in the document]
bergundy: How would lang be notified of nondeterminism right now? Do we need to intercept the completion and propagate that to the user?

sdk/src/lib.rs
```diff
@@ -239,21 +249,33 @@ impl TestRustWorker {
...
     /// fetch work for them.
-    pub async fn run_until_done(self) -> Result<(), anyhow::Error> {
+    pub async fn run_until_done(&mut self) -> Result<(), anyhow::Error> {
         self.run_until_done_shutdown_hook(|| {}).await
     }
 
+    /// Temporarily swap the core implementation used by this worker. When the returned value is
+    /// dropped, the old core implementation will be restored. This can be used to run against
+    /// mocked-out core instances.
+    pub fn swap_core(&mut self, other_core: Arc<dyn Core>) -> impl DerefMut<Target = Self> + '_ {
```
bergundy, line 271: Why would we want to swap the core if we have `register_replay_worker`?

test-utils/src/history_replay.rs
```diff
@@ -0,0 +1,179 @@
+mod history_builder;
```
bergundy, line 1: I agree with @cretz, all of this is part of core and should be in the main crate.

test-utils/src/lib.rs
```diff
@@ -47,23 +49,32 @@ pub async fn init_core_and_create_wf(test_name: &str) -> (Arc<dyn Core>, String)
...
+    let shared_mock_gateway = mock_gateway();
+    let init_opts = CoreInitOptionsBuilder::default()
+        .gateway_opts(shared_mock_gateway.get_options().clone())
+        .telemetry_opts(opts)
```
bergundy, line 58: Should the replay core even emit telemetry?

tests/integ_tests/workflow_tests/replay.rs
```diff
@@ -0,0 +1,86 @@
...
+        );
+    };
+    let poll_fut = async {
+        assert_matches!(
+            core.poll_workflow_activation(&task_q).await,
+            Err(PollWfError::ShutDown)
+        );
+    };
+    let complete_fut = async {
+        core.complete_execution(&task_q, &task.run_id).await;
```
bergundy, line 50: I didn't even know core has this API, is it for test purposes only?

tests/integ_tests/workflow_tests/replay.rs
```diff
@@ -0,0 +1,86 @@
...
+    history_from_proto_binary, history_replay::mock_gateway_from_history,
+    init_core_replay_preloaded, CoreTestHelpers,
+};
+use tokio::join;
+
+#[tokio::test]
+async fn timer_workflow_replay() {
+    let (core, task_q) = init_core_replay_preloaded(
+        "fail_wf_task",
+        &history_from_proto_binary("histories/fail_wf_task.bin")
```
bergundy, line 19: Should we use a history with a more appropriate name?
This is a bit misleading.

Sushisource: Heh, I had a history file but wasn't using it since both histories worked with what this test did.

tests/integ_tests/workflow_tests/replay.rs
Sushisource: It's an extension for testing

test-utils/src/lib.rs
Sushisource: Yes, why not? It's configurable in the same way and it's necessary for console output. It doesn't automatically mean exporting is on, and langs can default it to off when constructing

sdk/src/lib.rs
Sushisource: Because there is a shared client Core uses for some operations that isn't tied to a worker, that client needs to be changed out so it's a mock one and not a real one.

core/src/lib.rs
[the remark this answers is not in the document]
Sushisource: You'll get an eviction notification with a nondeterminism reason. It's a bit.. roundabout but it fits our established way of doing things.

bridge-ffi/src/lib.rs
Sushisource: I dunno, good question but we can change in another PR if we want. Keeping it callback based is sorta nice in that it's consistent, but I could go either way. @cretz ?

sdk/src/lib.rs
bergundy: I saw where it's used, I guess my question is why not make a new core with the new worker?

test-utils/src/lib.rs
bergundy: Sure for logging that makes sense.
We could limit the supported options on the rust side to be extra safe but I'm fine with limiting this on the lang side too.
For the TS PR we'd need to provide a way to expose `ReplayCore` in order to configure logging which I was hoping we could avoid.

sdk/src/lib.rs
Sushisource: Could do. This just ends up being easier for tests where you want to do it "real" and then download the replay. In either case this only gets used internally in integ tests. Now that the test_utils crate isn't needed for replay no one else sees this.

PR #263, opened 2022-01-27
Core -> Worker refactor

Sushisource:
Will fill in details here once we agree on top level design

NOTE: Docstrings are not all properly updated yet


core/src/lib.rs
```diff
@@ -93,6 +94,17 @@ pub struct CoreInitOptions {
     pub telemetry_opts: TelemetryOptions,
 }
 
+/// TODO:  This holds an actual worker now. Lang keeps track of taskq->worker mapping
+pub struct WorkerImpl {}
```
Sushisource, line 98: Good chance this just ends up being the actual worker struct

client/src/lib.rs
```diff
@@ -205,6 +214,16 @@ pub enum GatewayInitError {
...
+    /// Attempt to establish a connection to the Temporal server
+    pub async fn connect(
+        &self,
+        _metrics_meter: Option<&Meter>,
+    ) -> Result<RetryGateway<NamespacedServerGateway>, GatewayInitError> {
+        todo!()
+    }
+}
+
 impl ServerGatewayOptions {
```
bergundy, line 227: Is namespace going to be removed from this struct?
What about the metrics? Will they be namespaceless?

client/src/lib.rs
```diff
@@ -518,8 +550,17 @@ pub trait ServerGatewayApis {
     fn get_options(&self) -> &ServerGatewayOptions;
 }
 
+impl WorkflowServiceClientTrait for ServerGateway {
+    // TODO: Will implement 1:1 generated client methods. Boilerplate bonanza
+}
+
+impl WorkflowServiceClientTrait for NamespacedServerGateway {
+    // TODO: Will implement 1:1 generated client methods. Boilerplate bonanza
+}
```
bergundy, line 559: probably not the same trait because the namespaced one doesn't (or at least shouldn't) accept a namespace.

client/src/lib.rs
```diff
@@ -63,6 +63,18 @@ const OTHER_CALL_TIMEOUT: Duration = Duration::from_secs(30);
 
 type Result<T, E = tonic::Status> = std::result::Result<T, E>;
 
+/// Options for a namespace-bound connection to the temporal server
+pub struct NamespacedServerGatewayOptions {
+    pub namespace: String,
+    pub opts: ServerGatewayOptions,
+}
```
bergundy, line 70: I'm not sure we want this at all.
It might make more sense to just have a single `ServerGateway` or `WorkflowService` that's not bound to a namespace and attach the namespace to worker instances or the higher level rust client for test purposes.

client/src/lib.rs
Sushisource: This is the raw trait, they both must implement it. The methods will be exactly as in the generated client.

client/src/lib.rs
Sushisource: I need it to avoid remaking the world for testing. I am totally fine with non-rust langs not using it, or migrating away from it long term.

client/src/lib.rs
Sushisource: Ya, I just somehow missed that. It won't have it.

Metrics will always have namespaces (assuming the call involves a namespace). The only difference is at what point that label is bound. Namespaced clients would bind it when constructed, non-namespaced ones would bind it at call time.

Sushisource: Wait, I didn't miss it. It is removed. https://github.com/temporalio/sdk-core/pull/263/files/3af8ae06394d089141f600f99ee09e42199b56a9#diff-3f7e738903cc8204a4e3b1c8ab4db5812898e6bb5ad8b6c7b36df42e29481a8fL75

client/src/lib.rs
bergundy: The namespaced client wouldn't have the same API though..

core-api/src/lib.rs
```diff
@@ -19,6 +20,119 @@ use temporal_sdk_core_protos::coresdk::{
...
+
+    /// Request that a workflow be evicted by its run id. This will generate a workflow activation
+    /// with the eviction job inside it to be eventually returned by
+    /// [Core::poll_workflow_activation]. If the workflow had any existing outstanding activations,
+    /// such activations are invalidated and subsequent completions of them will do nothing and log
+    /// a warning.
+    fn request_workflow_eviction(&self, run_id: &str);
+
+    /// Returns this worker's instance of the [ServerGatewayApis] implementor it is using.
+    fn server_gateway(&self) -> Arc<dyn ServerGatewayApis + Send + Sync>;
```
bergundy, line 105: Should get call this `get_service_client`?

core/src/lib.rs
```diff
@@ -93,6 +94,28 @@ pub struct CoreInitOptions {
...
+    todo!()
+}
+
+/// Create a worker for replaying a specific history. It will auto-shutdown as soon as the history
+/// has finished being replayed. The provided gateway should be a mock, and this should only be used
+/// for workflow testing purposes.
+pub fn init_replay_worker(
+    config: WorkerConfig,
+    gateway: Arc<dyn ServerGatewayApis + Send + Sync>,
+    history: &History,
```
bergundy, line 114: Would we need the gateway here?
I think we can just create the gateway based on history.

client/src/lib.rs
Sushisource: It won't. It will also implement the higher level trait (which is mentioned in the notes). This is just Rust trait stuff. It needs to implement the lower level one because the higher level one depends on it.

core/src/lib.rs
Sushisource: I can certainly add a convenience function for it. This level of control is useful for my integration tests (hence why it needs to be public, but can be doc-hidden), because they might want to ensure certain rpc calls are or are not made.

core-api/src/lib.rs
Sushisource: This is returning the higher-level one, which lets you use the lower level one. I need this public for integ tests. Having both feels a bit overkill

client/src/lib.rs
cretz: Arguably the `NamespacedServerGateway` could just have a reference to the underlying `ServerGateway` in its struct instead of impl'ing the same trait.

core/src/lib.rs
cretz: :+1: Since it's all internal no need to prematurely abstract to trait

client/src/lib.rs
Sushisource: It does, and the traits could just be totally separate and if you want the raw thing you just have to call the accessor first. Perhaps that's a better direction.

cretz (approved):

Sushisource:
Closing in favor of impl PR https://github.com/temporalio/sdk-core/pull/265

PR #265, opened 2022-01-31
Core to worker refactor implementation

Sushisource:
## What was changed
Change the top level APIs to operate in terms of clients/workers/telemetry rather than "core".

## Why?
Makes using different clients for different purposes much more reasonable, and eliminates the confusing concept of a "core" type that we have to expose in the lang APIs.

## Checklist
TODO:

- [x] Make client API changes as described in https://github.com/temporalio/sdk-core/pull/263
- [x] Go over docstrings
- [x] Attach namespace to metrics on raw client as appropriate

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?
Ya bro


client/src/lib.rs
```diff
@@ -518,6 +523,11 @@ pub trait ServerGatewayApis {
     fn get_options(&self) -> &ServerGatewayOptions;
 }
 
+impl WorkflowServiceClientTrait for ServerGateway {
+    // TODO: Will implement 1:1 generated client methods. Boilerplate bonanza
```
cretz, line 527: I'd be ok if you wanted to just get rid of this trait and ask everyone to use the actual type.

core/src/lib.rs
```diff
@@ -71,97 +60,79 @@ use temporal_sdk_core_protos::{
...
-pub fn init_mock_gateway<SG: ServerGatewayApis + Send + Sync + 'static>(
-    opts: CoreInitOptions,
-    server_gateway: SG,
-) -> Result<CoreSDK, CoreInitError> {
-    telemetry_init(&opts.telemetry_opts).map_err(CoreInitError::TelemetryInitError)?;
-    Ok(CoreSDK::new(server_gateway, opts))
+/// Initialize a worker bound to a task queue, if the client is already behind an `Arc`.
+// Ideally this would be done with `impl Into<Arc<SG>>` but that imposes type clarifications
+// at the callsite sometimes.
+pub fn init_worker_pre_arcd<SG: ServerGatewayApis + Send + Sync + 'static>(
```
cretz, line 81: I'd be ok if you wanted to always require an Arc and just get rid of the non-Arc overload

core-api/src/lib.rs
```diff
@@ -107,24 +93,30 @@ pub trait Core: Send + Sync {
...
     /// connection.
     async fn shutdown(&self);
 
-    /// Shut down a specific worker. Will cease all polling on the task queue and future attempts
-    /// to poll that queue will return [PollWfError::NoWorkerForQueue].
-    async fn shutdown_worker(&self, task_queue: &str);
+    /// Completes shutdown and frees all resources. You should avoid simply dropping workers, as
+    /// this does not allow async tasks to report any panics that may have occurred cleanly.
+    ///
+    /// This should be called only after [Worker::shutdown] has resolved.
```
cretz, line 113: Does `shutdown` return before shutdown complete (i.e. before it receives `complete_workflow_activation`s for outstanding activations)? If not, is there a good use case to separate `shutdown` and `finalize_shutdown`? If `shutdown` does return before workflow shutdown is technically all done, ignore me and leave them separate.

cretz (approved):
At a glance, feels like a large refactoring and I admit to not being able to view it well for bugs or anything.

Feel fairly confident it's mostly naming and rescoping a couple of things, but otherwise doesn't really affect logic?

core-api/src/lib.rs
Sushisource: It waits until everything is genuinely complete. The use case for `finalize_shutdown` is it actually consumes the worker struct and is able to free the inner resources like pollers in an ordered way. EX: Checking that spawned tasks did not panic.

Basically this is because you cannot do `async` things in `impl Drop`

client/src/lib.rs
Sushisource: Problem is the retryer needs to be able to wrap the raw thing too (Otherwise the only benefit we get is connection sharing, and not also retry logic, at which point... why bother). 

So, since the retryer should be transparently usable in place of a non-retryer thing, we need some trait for it, and hence we need that 1:1 trait.

So, no way around it. I'll just apply a little macro magic.

Sushisource: So I reconsidered here after back and forth and decided to go with a little helper macro for calling the raw functions. See here: https://github.com/temporalio/sdk-core/pull/265/files#diff-3d28ff5c2d21a40cf191650a5a2043817ee022f49bbd012fb3c6526add336ce7R159

protos/local/temporal/sdk/core/bridge/service.proto
```diff
@@ -31,7 +31,7 @@ message StreamRequest {
 
   oneof request {
     InitRequest init = 2;
```
cretz, line 33: This has changed too. We can just remove this entire file if it's not worth keeping updated since nothing implements or uses it.

bergundy (approved):
I'm approving this.

As discussed on slack, we should make it easy to create a raw grpc client that isn't bound to a namespace and has a configurable retry policy and metrics should be tagged properly based on client context (use the namespace tag when possible).

bridge-ffi/include/sdk-core-bridge.h
```diff
@@ -37,29 +43,39 @@ typedef struct tmprl_bytes_t {
...
- * parameter is a core instance if the call is successful or null if not. If
- * present, the core instance must be freed via tmprl_core_shutdown when no
+ * parameter is a worker instance if the call is successful or null if not. If
+ * present, the worker instance must be freed via [tmprl_worker_shutdown] when no
  * longer in use. The third parameter of the callback is a byte array for a
- * InitResponse protobuf message which must be freed via tmprl_bytes_free.
+ * [InitResponse] protobuf message which must be freed via [tmprl_bytes_free].
  */
-typedef void (*tmprl_core_init_callback)(void *user_data, struct tmprl_core_t *core, const struct tmprl_bytes_t *resp);
+typedef void (*tmprl_worker_init_callback)(void *user_data, struct tmprl_worker_t *core, const struct tmprl_bytes_t *resp);
```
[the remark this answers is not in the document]
Sushisource: Already fixed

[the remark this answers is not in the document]
Sushisource: Wait, this is still already updated/fixed

protos/local/temporal/sdk/core/bridge/service.proto
Sushisource: Done

client/src/raw.rs
```diff
@@ -0,0 +1,419 @@
...
+            r.extensions_mut().insert(labels);
+            r.set_timeout(LONG_POLL_TIMEOUT);
+        }
+    );
+    proxy!(
+        respond_workflow_task_completed,
+        RespondWorkflowTaskCompletedRequest,
+        RespondWorkflowTaskCompletedResponse,
+        |r| {
+            let labels = AttachMetricLabels::new(vec![namespace_kv(r.get_ref().namespace.clone())]);
```
bergundy, line 160: Should the task queue be attached here (and other calls)?

bergundy (commented):
As mentioned on slack I feel like it would be cleaner to have a way to make a worker from a raw client since that's what the SDKs will essentially be doing. Server gateway is internal to core.

client/src/raw.rs
Sushisource: Ya, may as well

Issue #264, opened 2022-01-31
[Bug] Ensure that heartbeat-just-before-fail sends latest details before completion


Sushisource:
See https://temporalio.slack.com/archives/C01DKSMU94L/p1643647337531829 for details

Summary: Users would like to record HB details to save progress right before failing an activity. If we do not delay the activity completion until after flushing those details, they will be lost since the activity task is now gone. 

Test that we handle that properly

Also try to avoid "Unknown activity task when issuing cancel" extra warning logging when possible, although doing this properly should probably mean it won't happen at least in this case.

PR #266, opened 2022-02-01
Fix heartbeats not flushing on activity completion

Sushisource:
## What was changed
Make completion of activities wait on the eviction/flushing of heartbeats before sending a complete.

## Why?
See https://github.com/temporalio/sdk-core/issues/264

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/264

2. How was this tested:
Added test

3. Any docs updates needed?

### DIFF
--- a/core/src/core_tests/activity_tasks.rs
+++ b/core/src/core_tests/activity_tasks.rs
@@ -653,7 +653,7 @@ async fn complete_act_with_fail_flushes_heartbeat() {
         [PollActivityTaskQueueResponse {
             task_token: vec![1],
             activity_id: "act1".to_string(),
-            heartbeat_timeout: Some(Duration::from_millis(1).into()),
+            heartbeat_timeout: Some(Duration::from_secs(10).into()),
             ..Default::default()
         }],
     ));
--- a/core/src/core_tests/activity_tasks.rs
+++ b/core/src/core_tests/activity_tasks.rs
@@ -1,4 +1,3 @@
-use crate::telemetry::test_telem_console;
 use crate::{
     job_assert,
     test_help::{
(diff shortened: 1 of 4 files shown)


core/src/core_tests/activity_tasks.rs
```diff
@@ -621,3 +623,60 @@ async fn can_heartbeat_acts_during_shutdown() {
...
+        .times(1)
+        .returning(|_, _| Ok(RespondActivityTaskFailedResponse::default()));
+
+    let core = mock_core(MocksHolder::from_gateway_with_responses(
+        mock_gateway,
+        [],
+        [PollActivityTaskQueueResponse {
+            task_token: vec![1],
+            activity_id: "act1".to_string(),
+            heartbeat_timeout: Some(Duration::from_millis(1).into()),
```
bergundy, line 656: Maybe make this an extremely long value to verify we get throttled before failing the activity.

bergundy (approved):

core/src/core_tests/activity_tasks.rs
Sushisource: Kinda already happening by asserting the heartbeat RPC is only called twice. Bumped it up anyway in case the test runs slowly

core/src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -50,7 +53,11 @@ enum HeartbeatExecutorAction {
     /// Heartbeats are throttled for this task token, sleep until duration or wait to be cancelled
     Sleep(TaskToken, Duration, CancellationToken),
     /// Report heartbeat to the server
-    Report(TaskToken, Vec<common::Payload>),
+    Report {
+        task_token: TaskToken,
+        details: Vec<common::Payload>,
+        on_reported: Option<Arc<Notify>>,
```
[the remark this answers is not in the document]
Sushisource: Ehh, I tried it and it's the same LOC and not obviously cleaner to me. You have to maintain another mapping like `tts_to_final_completes` because evict removes from `tt_to_state`, which is a little more action-at-a-distance-y than everything going in the messages.

PR #269, opened 2022-02-05
Replay mock client wasn't allowing completes

Sushisource:
Which can happen with partial history

### DIFF
--- a/client/src/raw.rs
+++ b/client/src/raw.rs
@@ -2,11 +2,15 @@
 //! information we need via interceptors. This module contains the necessary stuff to make that
 //! happen.
 
-use crate::metrics::task_queue_kv;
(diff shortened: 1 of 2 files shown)


client/src/raw.rs
```diff
@@ -2,11 +2,15 @@
 //! information we need via interceptors. This module contains the necessary stuff to make that
 //! happen.
 
-use crate::metrics::task_queue_kv;
-use crate::{metrics::namespace_kv, raw::sealed::RawClientLike, LONG_POLL_TIMEOUT};
+use crate::{
```
bergundy, line 5: Is this relevant to the PR?

bergundy (approved):

cretz (approved):

PR #278, opened 2022-02-24
Add upsert search2

mmcshane:
## What was changed

Adds support for sending search attribute upsert commands to the server through the UpsertSearchAttributesMachine state machine implementation and related types/transforms.

## Why?

Feature was missing from sdk-core

## Checklist

1. Closes 

2. How was this tested:

New unit tests

3. Any docs updates needed?

### DIFF
--- a/core/src/workflow/machines/mod.rs
+++ b/core/src/workflow/machines/mod.rs
@@ -15,7 +15,6 @@ mod patch_state_machine;
 mod side_effect_state_machine;
 mod signal_external_state_machine;
 mod timer_state_machine;
-#[allow(unused)]
 mod upsert_search_attributes_state_machine;
 mod workflow_task_state_machine;
 
--- a/core/src/workflow/machines/mod.rs
+++ b/core/src/workflow/machines/mod.rs
@@ -45,6 +45,7 @@ use temporal_sdk_core_protos::temporal::api::{
     command::v1::Command as ProtoCommand, enums::v1::CommandType, history::v1::HistoryEvent,
 };
 use timer_state_machine::TimerMachine;
+use upsert_search_attributes_state_machine::UpsertSearchAttributesMachine;
 use workflow_machines::MachineResponse;
 use workflow_task_state_machine::WorkflowTaskMachine;
 
@@ -65,6 +66,7 @@ enum MachineKind {
     SignalExternalWorkflow,
     CancelExternalWorkflow,
     LocalActivity,
+    UpsertSearchAttributes,
 }
 
 #[enum_dispatch::enum_dispatch]
@@ -82,6 +84,7 @@ enum Machines {
     SignalExternalMachine,
     TimerMachine,
     WorkflowTaskMachine,
+    UpsertSearchAttributesMachine,
 }
 
 /// Extends [rustfsm::StateMachine] with some functionality specific to the temporal SDK.
--- a/core/src/workflow/machines/upsert_search_attributes_state_machine.rs
+++ b/core/src/workflow/machines/upsert_search_attributes_state_machine.rs
@@ -1,5 +1,3 @@
-use std::convert::TryFrom;
-
 use rustfsm::{fsm, TransitionResult};
 use temporal_sdk_core_protos::{
     coresdk::workflow_commands::UpsertWorkflowSearchAttributes,
--- a/core/src/workflow/machines/upsert_search_attributes_state_machine.rs
+++ b/core/src/workflow/machines/upsert_search_attributes_state_machine.rs
@@ -1,7 +1,26 @@
 use rustfsm::{fsm, TransitionResult};
+use super::{
+    NewMachineWithCommand
+};
+use temporal_sdk_core_protos::{
+    coresdk::{
+        // workflow_activation::FireTimer,
+        // workflow_commands::{UpsertCommandCreated, UpsertCommandRecorded, UpsertSearchAttributesCommand},
+        workflow_commands::{UpsertSearchAttributes},
+        HistoryEventId,
+    },
+    temporal::api::{
+        command::v1::Command,
+        enums::v1::{CommandType, EventType},
+        history::v1::{history_event, HistoryEvent, TimerFiredEventAttributes},
+    },
+};
 
 fsm! {
-    pub(super) name UpsertSearchAttributesMachine; command UpsertSearchAttributesCommand; error UpsertSearchAttributesMachineError;
+    pub(super) name UpsertSearchAttributesMachine;
+    command UpsertSearchAttributesCommand;
+    error UpsertSearchAttributesMachineError;
+    shared_state SharedState;
 
     Created --(Schedule, on_schedule) --> UpsertCommandCreated;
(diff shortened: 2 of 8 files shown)


core/src/workflow/machines/mod.rs
```diff
@@ -37,6 +37,7 @@ use patch_state_machine::PatchMachine;
 use prost::alloc::fmt::Formatter;
 use rustfsm::{MachineError, StateMachine};
 use signal_external_state_machine::SignalExternalMachine;
+use upsert_search_attributes_state_machine::UpsertSearchAttributesMachine;
```
Sushisource: I can't comment on unchanged lines, but, you'll want to remove the `#[allow(unused)]` attribute at the top of the file for the upsert module

core/src/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -1,40 +1,216 @@
...
+    // upon observing a history event indicating that the command has been recorded. Note that this
+    // does not imply that the command has been _executed_, only that it _will be_ executed at some
+    // point in the future.
+    CommandIssued --(CommandRecorded) --> Done;
 }
 
-#[derive(thiserror::Error, Debug)]
-pub(super) enum UpsertSearchAttributesMachineError {}
+// Instantiates an UpsertSearchAttributesMachine and packs it together with an initial command
+// to apply the provided search attribute update.
```
Sushisource, line 44: This can be a docstring `///` -- applies to a bunch of other places too

core/src/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -1,40 +1,216 @@
...
+// Unused but must exist
+type SharedState = ();
 
-#[derive(Default, Clone)]
+// The state-machine-specific set of commands that are the results of state transition in the
+// UpsertSearchAttributesMachine.
+#[derive(Debug, derive_more::Display)]
+pub(super) enum UpsertSearchAttributesMachineCommand {
+    #[display(fmt = "SendUpsertCommand")]
+    SendUpsertCommand(Attributes),
```
Sushisource, line 67: This could potentially contain any attribute, where it should only contain the inner type which is just `UpsertSearchAttributes` (or something like that)

core/src/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -1,40 +1,239 @@
...
+    fn new() -> Self {
+        Self {
+            state: Created {}.into(),
+            shared_state: (),
+        }
+    }
+
+    fn state(&self) -> String {
+        self.state.to_string()
     }
```
Sushisource, line 94: This is auto-implemented as `.state().name()` I believe

core/src/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -1,40 +1,239 @@
...
+        Self {}
+    }
+}
+
+// There is no Command/Reponse associated with this transition
+impl From<Created> for CommandIssued {
+    fn from(_: Created) -> Self {
+        Self {}
+    }
+}
```
Sushisource, line 175: I think you can remove these and the state machine impl automatically uses `Default`

core/src/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -1,40 +1,239 @@
...
+        });
+        let mut wfm = ManagedWFFunc::new(t, wff, vec![]);
+
+        wfm.get_next_activation().await.unwrap();
+        let commands = wfm.get_server_commands().commands;
+        assert!(!commands.is_empty());
+        assert_eq!(
+            commands[0].command_type,
+            CommandType::UpsertWorkflowSearchAttributes as i32
+        );
```
Sushisource, line 212: Probably worth asserting the name/payload come out as expected too. I didn't do a great job of this everywhere in existing tests so blame me for the bad example. A lot of them are handled at the integ test layer where it's a bit easier, but in this case the workflow never gets any "result" from the command, so it's useful to ensure it's sent with the right details.

core/src/workflow/machines/mod.rs
mmcshane: Not sure I follow - what is unused?

core/src/workflow/machines/upsert_search_attributes_state_machine.rs
mmcshane: there's no `.name()` but I have `.to_string()` which works. Thanks

core/src/workflow/machines/upsert_search_attributes_state_machine.rs
mmcshane: Doesn't compile without them - maybe I'm missing something somewhere else?

core/src/workflow/machines/mod.rs
Sushisource: Line 18 has that, you can remove it now

mmcshane: Totally misread your initial comment, sorry. I see what you're saying.

core/src/workflow/machines/upsert_search_attributes_state_machine.rs
mmcshane: You're right but it turns out this is dead code so deleting instead

core/src/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -1,40 +1,250 @@
...
+    coresdk::workflow_commands::UpsertWorkflowSearchAttributes,
+    temporal::api::{
+        command::v1::Command,
+        enums::v1::{CommandType, EventType},
+        history::v1::HistoryEvent,
+    },
+};
+
+use crate::workflow::machines::{Cancellable, EventInfo, MachineKind, WFMachinesAdapter};
+
```
Sushisource, line 12: Nit: Collapse these spaces everywhere and use `cargo +nightly fmt` (need to install nightly rust) which will apply the nesting formatting (annoyingly not stabilized yet.

The only place I use spaces between imports is when they are `pub use`d to differentiate re-exports.

core/src/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -1,40 +1,250 @@
...
+            _ => Err(Self::Error::Nondeterminism(format!(
+                "UpsertWorkflowSearchAttributesMachine does not handle command type {c:?}"
+            ))),
+        }
     }
 }
 
-#[derive(Default, Clone)]
-pub(super) struct UpsertCommandRecorded {}
+/// There is no Command/Response associated with this transition
```
Sushisource, line 150: Nit/FYI These docstrings can't act as docstrings since `From`  already has trait level docs. They could be normal comments.

Sushisource (approved):
Looking good! Thanks @mmcshane

PR #281, opened 2022-03-02
Upsert search attrs integration test

mmcshane:
## What was changed

Adds an integration test for search attribute upsert

As a side effect, adds the WorkflowOptions parameter to calls that start workflows.

## Why?

Tests are lawful good.

## Checklist

1. Closes 

2. How was this tested:

It _is_ a test

3. Any docs updates needed?

Update to dependency graph visualization is included


sdk/src/lib.rs
```diff
@@ -130,7 +136,9 @@ impl TestRustWorker {
         workflow_id: impl Into<String>,
         workflow_type: impl Into<String>,
         input: Vec<Payload>,
+        mut options: WorkflowOptions,
     ) -> Result<String, tonic::Status> {
+        options.task_timeout = options.task_timeout.or(self.task_timeout);
```
mmcshane, line 141: This seems ... ok-ish? We want to keep the value present in `options` if there is one, otherwise use the value in `self.task_timeout`, which itself may be `None`.

Sushisource: This makes sense. Honestly `task_timeout` shouldn't even be on the whole worker probably, at least not in the way it is currently - but all this is going to get the treatment when I give this stuff a polish for real release. So this makes perfect sense for now

Sushisource (approved):
_chef's kiss_

PR #288, opened 2022-03-15
Ensure `service.name` is set for metrics as well

Sushisource:
## What was changed
This attribute was only being set for traces

## Why?
Helps users filter for temporal metrics

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (approved):
LGTM. How was this tested?

Sushisource:
> LGTM. How was this tested?

There aren't any tests that verify metric output right now since it's extremely boilerplate-y to do comprehensively. We could add a few sanity checks for certain really important things, like some labels, but I find this sort of test is generally pretty low ROI.

bergundy:
It'd be enough just to verify this manually once, no need for automated tests.

Sushisource:
> It'd be enough just to verify this manually once, no need for automated tests.

Oh, sure, I've done that.

PR #289, opened 2022-03-15
Fix possibly unflushed heartbeats if they are the first heartbeat

Sushisource:
## What was changed
It was possible for the very first heartbeat of a new activity task to not be flushed if all the activity does is record the heartbeat and then exit extremely fast, while the server also is slow and takes longer to record the heartbeat than it does to complete the activity.

This happens:

```
| record hb RPC started
|
|    complete activity RPC 
|    |
|    |
|    done
|
| activity not found ?!??
```
Which arguably is also a bit of a server-side bug, but regardless now we make sure the record RPC _completes_ before allowing activity completion.

## Why?
Bugfix.
Chad noticed while working on Python https://github.com/temporalio/sdk-typescript/issues/538

## Checklist

1. Closes 

2. How was this tested:
New UT/IT

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/activities/activity_heartbeat_manager.rs
+++ b/core/src/worker/activities/activity_heartbeat_manager.rs
@@ -56,7 +56,6 @@ enum HeartbeatExecutorAction {
     Report {
         task_token: TaskToken,
         details: Vec<common::Payload>,
-        on_reported: Option<Arc<Notify>>,
     },
 }
 
@@ -139,6 +138,8 @@ impl ActivityHeartbeatManager {
 struct ActivityHeartbeatState {
     /// If None and throttle interval is over, untrack this task token
     last_recorded_details: Option<Vec<common::Payload>>,
+    /// True if we've queued up a request to record against server, but it hasn't yet completed
+    is_record_in_flight: bool,
     last_send_requested: Instant,
     throttle_interval: Duration,
     throttled_cancellation_token: Option<CancellationToken>,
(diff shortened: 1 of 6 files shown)


core/src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -332,7 +342,8 @@ impl ActivityHeartbeatManager {
                                 },
                             };
                         }
-                        HeartbeatExecutorAction::Report { task_token: tt, details, on_reported } => {
+                        HeartbeatExecutorAction::Report { task_token: tt, details } => {
+                            debug!("Recording hb {details:?}");
```
bergundy: This should probably be less verbose than `debug` and have a more meaningful message

core/src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -353,6 +364,7 @@ impl ActivityHeartbeatManager {
                                 // finished (which is one thing not found implies - other reasons
                                 // would seem equally valid).
                                 Err(s) if s.code() == tonic::Code::NotFound => {
+                                    warn!("Act not found when recording heartbeat");
```
bergundy: Don't we warn about this somewhere else?
Also, if this is user visible, let's not shorten the word `Activity`.

tests/integ_tests/heartbeat_tests.rs
```diff
@@ -83,3 +84,80 @@ async fn activity_heartbeat() {
...
+        .into(),
+    ))
+    .await
+    .unwrap();
+
+    // Multiple times, poll for the activity, heartbeat, and then immediately fail
+    // Poll activity and verify that it's been scheduled with correct parameters
+    for i in 0u8..=3 {
+        let task = core.poll_activity_task().await.unwrap();
+        let hb_details = assert_matches!(task.variant, Some(activity_task::Variant::Start(s)) => s);
```
bergundy, line 120: The should be a better name than `hb_details` for this, it's just the activity start message..

bergundy (approved):

core/src/worker/activities/activity_heartbeat_manager.rs
Sushisource: Both these I already fixed

Issue #258, opened 2022-01-12
[Feature Request] Reduce required surface area of mock client


Sushisource:
Right now the `ServerGatewayApis` trait is too broad. Core only needs a subset of the functionality internally, and what it requires there is the actual surface area that should be mocked.

The trait should be split up into sub-traits, and only the sub-trait that Core makes use of needs to be mocked. This would allow for more easily extending the client without needing to go pointlessly update mocks.

A long time ago we made the decision that we wanted to accept simpler parameters rather than the raw protos, too. This may be a good time to revisit that. The client meant for end-user consumption can wrap the traits which can operate in terms of the raw protos.

cretz:
@Sushisource - There are a couple use cases of that gRPC client - for core's internal and for caller use. I think it makes plenty of sense to have a core-specific idiomatic client API trait w/ your own params and limited surface (though often the proto params are good enough here like they are for other core calls).

But, where you instantiate this higher-level client backed by gRPC, I think we need a way to still get the raw gRPC trait from there (or maybe the raw gRPC client is what is provided, but it's better to have the high-level client "close"/"drop" handle closing the gRPC conn). So yeah, a `Client` trait with what core needs and a runtime gRPC-based impl in a struct which provides access to the pure gRPC client iface would be nice.

I'm half-tempted to tell new SDK clients not to communicate with gRPC directly but always use the core client just to keep config and connection count centralized (becomes a big deal on gRPC load balancing, cert mgmt, etc). But that's a different topic for a different day.

PR #291, opened 2022-03-22
More client refactoring for release prep

Sushisource:
## What was changed
Renamed a number of things (see commits) to have more straightforward names
Made initting a worker accept different client types more ergonomically
No longer exporting mocks from the client library

## Why?
Ergo, cleanliness, production prep

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/258
2. How was this tested:
Existing tests

3. Any docs updates needed?


cretz (approved):
Approving under the assumption that logic didn't really change

PR #296, opened 2022-03-31
Eager activity dispatch

Sushisource:
## What was changed
Implement newly added optimization for eagerly dispatched activities on WFT completion response

## Why?
Optimization

## Checklist

1. Closes 

2. How was this tested:
Unit tests. Integ tests will be added once a server release is cut with this functionality.

3. Any docs updates needed?


cretz (approved):

core/src/worker/mod.rs
```diff
@@ -937,6 +970,33 @@ impl Worker {
...
     }
+
+    /// Attempt to reserve activity slots for activities we could eagerly execute on
+    /// this worker.
+    ///
+    /// Returns the number of activity slots that were reserved
+    fn reserve_activity_slots_for_outgoing_commands(&self, commands: &mut [Command]) -> usize {
+        let mut reserved = 0;
+        for cmd in commands {
+            if let Some(Attributes::ScheduleActivityTaskCommandAttributes(attrs)) =
```
cretz, line 981: @bergundy's discussion off-PR made me realize we need to probably fail if `ScheduleActivityTaskCommandAttributes` has `request_eager_execution` but has a different task queue than this one.

protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto
```diff
@@ -77,6 +77,10 @@ message ScheduleActivity {
     common.RetryPolicy retry_policy = 12;
     /// Defines how the workflow will wait (or not) for cancellation of the activity to be confirmed
     ActivityCancellationType cancellation_type = 13;
+    /// If set, the worker will not tell the service that it can immediately start executing this
+    /// activity. When unset/default, workers will always attempt to do so if activity execution
+    /// slots are available.
+    bool do_not_eagerly_execute = 14;
```
bergundy, line 83: Can we make this `eagerly_execute` instead of using the negative form?
Lang can put `true` in here by default.

bergundy (commented):
We could really use some integration tests but that might be difficult before 1.16 is released.
We should also discuss cross task queue eager execution.

None of these are blocking the PR, only thing that needs to be fixed is making sure not to eagerly execute if the activity task queue is different from the workflow task queue.

There might also be a case where Lang doesn't register any activities which should mean that it doesn't poll for activity tasks, we should figure out how we want to pass this information to core so it can disable eager execution.

protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto
Sushisource: I think it's a better to have it work this way. Lang shouldn't need to explicitly pass something that is the default, it makes it easier for any given lang implementation to make a mistake. More importantly, lang _doesn't have_ the information needed to know if it actually can eagerly execute.

However, it very much can explicitly state when it would _not_ like eager execution to happen.

sdk-core-protos/src/lib.rs
```diff
@@ -1513,6 +1513,7 @@ pub mod temporal {
                                 start_to_close_timeout: s.start_to_close_timeout,
                                 heartbeat_timeout: s.heartbeat_timeout,
                                 retry_policy: s.retry_policy.map(Into::into),
+                                request_eager_execution: !s.do_not_eagerly_execute,
```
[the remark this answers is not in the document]
Sushisource: IMO none of that should be a concern. If we happen to set it to `true` for an activity sched. cmd that isn't for the same TQ the workflow is on, server should just ignore our request.

It feels off for the WFT completed response for a workflow on one queue to return activities from another. 

@yycptt What will happen in this case?

bergundy (approved):
Nice! CC @yiminc

PR #297, opened 2022-04-11
Fix possible deadlock in machine concurrency manager

Sushisource:
## What was changed
It turns out that even though the *inner* lock for a particular run is fine to hold across an await point, the *outer* lock on the whole map was also being held (which seems obvious in retrospect). If a write to the overall map came in while the inner await was pending, you'd get an ABA lock situation and deadlock everything.

Now the lock is behind an Arc to prevent holding the outer lock while accessing.

## Why?
Deadlocks bad

## Checklist

1. Closes 

2. How was this tested:
Added unit test. Repro load test provided by Roey in typescript also fixed.

3. Any docs updates needed?

### DIFF
--- a/core/src/workflow/workflow_tasks/concurrency_manager.rs
+++ b/core/src/workflow/workflow_tasks/concurrency_manager.rs
@@ -22,7 +22,7 @@ pub(crate) struct WorkflowConcurrencyManager {
 }
 
 struct ManagedRun {
-    wfm: Mutex<WorkflowManager>,
+    wfm: Mutex<Option<WorkflowManager>>,
     wft: Option<OutstandingTask>,
     activation: Option<OutstandingActivation>,
     metrics: MetricsContext,
@@ -36,7 +36,7 @@ struct ManagedRun {
 impl ManagedRun {
     fn new(wfm: WorkflowManager, metrics: MetricsContext) -> Self {
         Self {
-            wfm: Mutex::new(wfm),
+            wfm: Mutex::new(Some(wfm)),
             wft: None,
             activation: None,
             metrics,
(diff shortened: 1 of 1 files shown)


core/src/workflow/workflow_tasks/concurrency_manager.rs
```diff
@@ -266,15 +266,35 @@ impl WorkflowConcurrencyManager {
...
+        let res = mutator(&mut wfm).await;
+
+        {
+            let readlock = self.runs.read();
+            // SAFETY: Since runs are accessed serially the machines cannot go missing while
+            // we are modifying them for a given run
+            let m = readlock
+                .get(run_id)
+                .expect("Machine cannot be missing as each run is accessed serially");
+            *m.wfm.lock() = Some(wfm);
```
cretz, line 296: Can you hold the option obtained from `lock` above and just use this setter or `replace` or whatever here instead of another map lookup?

core/src/workflow/workflow_tasks/concurrency_manager.rs
```diff
@@ -266,15 +266,35 @@ impl WorkflowConcurrencyManager {
...
+
+        let res = mutator(&mut wfm).await;
+
+        {
+            let readlock = self.runs.read();
+            // SAFETY: Since runs are accessed serially the machines cannot go missing while
+            // we are modifying them for a given run
+            let m = readlock
+                .get(run_id)
+                .expect("Machine cannot be missing as each run is accessed serially");
```
cretz, line 295: If this is the case, can you hold the `wfm.lock()` across the life of this call even though you don't hold the map lock?

cretz (approved):

core/src/workflow/workflow_tasks/concurrency_manager.rs
Sushisource: I cannot - and this was in fact the non-obvious part. The problem is that I have a _reference_ to the mutex through the map-level readlock. So if I keep the inner lock guard around, I am by definition _also_ keeping the outer map lock guard around since I don't own that mutex, I am only borrowing it.

Sushisource: Adding the scoped blocks clarifies this since now you get lifetime problems if you try to only keep the inner mutex around but not the outer one.

Sushisource: @cretz You made me realize this can look much nicer with an Arc though instead of this option nonsense, so thanks for mentioning.

cretz: Makes sense you need the map in scope to access its value as a reference :+1:

I do think a future refactor that moves mutability/locking down to a lower level than this higher level "access" abstraction could make sense.

PR #300, opened 2022-04-14
Fix evicts for missing runs eating a permit

Sushisource:


core/src/core_tests/workflow_tasks.rs
```diff
@@ -1694,3 +1695,48 @@ async fn tasks_from_completion_are_delivered() {
...
+
+    let tasks = [hist_to_poll_resp(
+        &t,
+        wfid.to_owned(),
+        // Use a partial task so that we'll fetch history
+        ResponseType::OneTask(2),
+        TEST_Q.to_string(),
+    )];
+    let mut mock = mock_workflow_client();
+    mock.expect_get_workflow_execution_history()
```
bergundy, line 1719: This is for paginating when workflow run is not in cache, right?

bergundy (approved):

core/src/core_tests/workflow_tasks.rs
Sushisource: Yup

cretz (approved):

Issue #301, opened 2022-04-16
[Feature Request] Rust SDK Activity Context


dt665m:
As per the discussion here https://temporalio.slack.com/archives/CTRCR8RBP/p1649650077749749, exposing some context to activity functions is still missing in the Rust SDK.  

### Describe the solution you'd like

A context struct with helper functions for activities.

PR #302, opened 2022-04-16
add activity context for rust sdk

dt665m:
## What was changed
An activity context that tries to be as close to the go sdk as possible.  Implementation added in sdk/src/activity_context.rs.  sdk/src/lib.rs was also modified to support the new context for each activity worker function.

1. Closes https://github.com/temporalio/sdk-core/issues/301

2. How was this tested:
- Ran Unit Tests
- Ran Integration tests
- Sanity Tested with External Project on Docker-compose temporal environment

3. Any docs updates needed?
Not that I know of.  Would be happy to contribute with some direction.

### DIFF
--- a/sdk/src/activity_context.rs
+++ b/sdk/src/activity_context.rs
@@ -1,27 +1,29 @@
-use std::collections::HashMap;
-use std::sync::Arc;
-use std::time::{Duration as StdDuration, SystemTime};
-
 use prost_types::{Duration, Timestamp};
+use std::{
+    collections::HashMap,
+    sync::Arc,
+    time::{Duration as StdDuration, SystemTime},
+};
 use temporal_sdk_core_api::Worker;
-use temporal_sdk_core_protos::coresdk::{
-    activity_task,
-    common::{Payload, RetryPolicy, WorkflowExecution},
-    ActivityHeartbeat,
+use temporal_sdk_core_protos::{
+    coresdk::{
+        activity_task,
+        common::{Payload, RetryPolicy, WorkflowExecution},
+        ActivityHeartbeat,
+    },
+    utilities::TryIntoOrNone,
 };
 use tokio_util::sync::CancellationToken;
 
 /// Used within activities to get info, heartbeat management etc.
 #[derive(Clone)]
-#[allow(dead_code)]
 pub struct ActContext {
-    pub(crate) worker: Arc<dyn Worker>,
-    pub(crate) cancellation_token: CancellationToken,
-    pub(crate) input: Vec<Payload>,
-    pub(crate) heartbeat_details: Vec<Payload>,
-    /// #NOTE future usage?
-    pub(crate) header_fields: HashMap<String, Payload>,
-    pub(crate) info: ActivityInfo,
+    worker: Arc<dyn Worker>,
+    cancellation_token: CancellationToken,
+    input: Vec<Payload>,
+    heartbeat_details: Vec<Payload>,
+    header_fields: HashMap<String, Payload>,
+    info: ActivityInfo,
 }
 
 #[derive(Clone)]
@@ -33,31 +35,31 @@ pub struct ActivityInfo {
     pub activity_id: String,
     pub activity_type: String,
     pub task_queue: String,
-    pub heartbeat_timeout: Option<Duration>,
-    /// Time of activity scheduled by a workflow
-    pub scheduled_time: Option<Timestamp>,
+    pub heartbeat_timeout: Option<StdDuration>,
+    /// Time activity was scheduled by a workflow
+    pub scheduled_time: Option<SystemTime>,
     /// Time of activity start
-    pub started_time: Option<Timestamp>,
+    pub started_time: Option<SystemTime>,
     /// Time of activity timeout
-    pub deadline: Option<Timestamp>,
-    /// Attempt starts from 1, and increased by 1 for every retry if retry policy is specified.
+    pub deadline: Option<SystemTime>,
+    /// Attempt starts from 1, and increase by 1 for every retry, if retry policy is specified.
     pub attempt: u32,
-
-    /// #NOTE Not in sdk-go
-    pub current_attempt_scheduled_time: Option<Timestamp>,
+    /// Time this attempt at the activity was scheduled
+    pub current_attempt_scheduled_time: Option<SystemTime>,
     pub retry_policy: Option<RetryPolicy>,
     pub is_local: bool,
 }
 
 impl ActContext {
-    /// Construct new Activity Context
-    pub fn new(
+    /// Construct new Activity Context, returning the context and the first argument to the activity
+    /// (which may be a default [Payload]).
+    pub(crate) fn new(
         worker: Arc<dyn Worker>,
         cancellation_token: CancellationToken,
         task_queue: String,
         task_token: Vec<u8>,
         task: activity_task::Start,
-    ) -> Self {
+    ) -> (Self, Payload) {
         let activity_task::Start {
             workflow_namespace,
             workflow_type,
@@ -65,7 +67,7 @@ impl ActContext {
             activity_id,
             activity_type,
             header_fields,
-            input,
+            mut input,
             heartbeat_details,
             scheduled_time,
             current_attempt_scheduled_time,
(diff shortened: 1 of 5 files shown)


dt665m:
@Sushisource Hope this could serve as a starting point for activities in the rust SDK.

Sushisource:
@dt665m Cool! Thanks for the PR.

One thing I haven't decided yet is if we should expose the context as an explicit parameter as is done here, or if we want to make it implicit and accessed via static functions (which is typical of all the non-Go SDKs). I would do the same thing with the workflow context if so.

Any thoughts on that?

That aside I think this is a good starting point.

sdk/src/activity_context.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    /// Extract heartbeat details from last failed attempt. This is used in combination with retry policy.
+    pub fn get_heartbeat_details(&self) -> &[Payload] {
+        &self.heartbeat_details
+    }
+
+    /// Helper function to check if details exist from previous execution of this activity
+    pub fn has_heartbeat_details(&self) -> bool {
+        !self.heartbeat_details.is_empty()
+    }
```
Sushisource, line 125: Given  `get_heartbeat_details` returns a ref, I don't know that we need this extra function

sdk/src/activity_context.rs
```diff
@@ -0,0 +1,202 @@
...
+                current_attempt_scheduled_time,
+                retry_policy,
+                is_local,
+            },
+        }
+    }
+
+    /// Retrieve the rest of the parameters
+    /// #NOTE the first parameter is popped for convenience, the rest is still stored in the
+    /// activity state.  There are no variadic parameters in Rust.
```
Sushisource, line 112: This is probably a bit confusing. I think I want to continue enforcing that Rust-defined activities/workflows only take 1 parameter.

In that case, we should maybe give it a more specific name like `extra_inputs` . The docstring can indicate the first parameter is passed to the activity function explicitly, and there should not be any extra unless the caller is doing something a bit odd by passing more than one, and those can be retrieved here.

dt665m:
> @dt665m Cool! Thanks for the PR.
> 
> One thing I haven't decided yet is if we should expose the context as an explicit parameter as is done here, or if we want to make it implicit and accessed via static functions (which is typical of all the non-Go SDKs). I would do the same thing with the workflow context if so.
> 
> Any thoughts on that?
> 
> That aside I think this is a good starting point.

I don't have strong feelings either way.  Intuitively, it seems like the static function style would make the default workflow/activities easier to pick up and learn.  However, I imagine as soon as you actually start doing interesting workflows, you'll inevitably need the context.

Are you thinking of implementing this through tokio's task_local LocalKey?

sdk/src/activity_context.rs
dt665m: Reasonable.  I was trying to keep the API's in line but yes, with the reference, callers can check themselves without the helper.  Fixed

sdk/src/activity_context.rs
dt665m: Fixed by renaming + docs.  I think it's healthy to keep the function even if a little confusing because it is possible to have polyglot environments where workflows are deployed from Go, but certain activities are handled in Rust specific task_queues.  In that case, it could be surprising if the Go SDK naturally allows many parameters but Rust's activities can never access them.

Sushisource:
>  I imagine as soon as you actually start doing interesting workflows, you'll inevitably need the context.
Can you elaborate a bit on any examples you're thinking of? Would be curious to hear what that is. Java and TS sdks don't have explicit context parameters for example.

> Are you thinking of implementing this through tokio's task_local LocalKey?
Yeah. There actually is already one here: https://github.com/temporalio/sdk-core/blob/c87202812b91d5987f34e6a4338701aaef49c973/sdk/src/lib.rs#L355 which we should eliminate if we got with the explicit parameter.

dt665m:
> > I imagine as soon as you actually start doing interesting workflows, you'll inevitably need the context.
> 
> Can you elaborate a bit on any examples you're thinking of? Would be curious to hear what that is. Java and TS sdks don't have explicit context parameters for example.
> 
> > Are you thinking of implementing this through tokio's task_local LocalKey?
> 
> Yeah. There actually is already one here:
> 
> https://github.com/temporalio/sdk-core/blob/c87202812b91d5987f34e6a4338701aaef49c973/sdk/src/lib.rs#L355
> 
> which we should eliminate if we got with the explicit parameter.

If you decide to go with the task_local,  we can fold the cancellation token into the struct.  I think it really depends on the 90/10 use case.  

To elaborate a bit, I want to be able to use snapshots and heartbeat details to handle idempotent requests (ie. commit some key/id first before making some API calls).  I haven't done this yet... but I plan on trying long living workflows with asynchronous activity completion as well.  Additionally, I want to tag the workflow id in activities as a tracing ray.  These things all require the context.  Doing task_local or explicit parameter is just semantically different.  Both ways achieve the same things.  As my use cases mostly require the context, it seems natural to pass one to the closures.  Again, I don't feel strongly either way and if you decide to go with the static approach for API parity, I can make that change.

I can't see what is erroring in the buildkite though, so I'll need some help/direction on what is failing.  I've run the buildkite integration test pipeline locally using the docker-compose and only got warnings that I'm unsure on how to fix.

Sushisource:
I think let's go with explicit for now then. I think it's probably more idiomatic for Rust, and it's certainly more obvious what's going on.

Re: Buildkite failures - we'll probably move this repo to GHA soon, but for now it's just failing on lints and formatting. `cargo fmt` will fix the formatting and `cargo clippy --test integ_tests --test load_tests  --all && cargo clippy --all --all-features --all-targets` will cover all lints.

dt665m:
> I think let's go with explicit for now then. I think it's probably more idiomatic for Rust, and it's certainly more obvious what's going on.
> 
> Re: Buildkite failures - we'll probably move this repo to GHA soon, but for now it's just failing on lints and formatting. `cargo fmt` will fix the formatting and `cargo clippy --test integ_tests --test load_tests --all && cargo clippy --all --all-features --all-targets` will cover all lints.

Alright, explicit param it is.  I'll try to run the clippy stuff to find things to fix.  Should we also move the cancellation token into the Activity Context?

Sushisource:
Yeah, I think it should be moved in. Thanks!

dt665m:
> Yeah, I think it should be moved in. Thanks!

I think I got most of the stuff we discussed covered.  Buildkite is passing now as well.  Hope this is good enough for a merge back to upstream.  Thanks for the guidance!  If I get some free time, I'm gonna need more for implementing workflow checkpoints.  If that works out, the Rust SDK will have enough for me to not have to maintain another go project for workflows.

sdk/src/activity_context.rs
```diff
@@ -0,0 +1,212 @@
...
+use temporal_sdk_core_protos::coresdk::{
+    activity_task,
+    common::{Payload, RetryPolicy, WorkflowExecution},
+    ActivityHeartbeat,
+};
+use tokio_util::sync::CancellationToken;
+
+/// Used within activities to get info, heartbeat management etc.
+#[derive(Clone)]
+#[allow(dead_code)]
```
Sushisource, line 16: What was the dead code here? I'd prefer to avoid the annotation here and just dropped whatever was unused, can always add it back.

sdk/src/activity_context.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Time of activity scheduled by a workflow
+    pub scheduled_time: Option<Timestamp>,
+    /// Time of activity start
+    pub started_time: Option<Timestamp>,
+    /// Time of activity timeout
+    pub deadline: Option<Timestamp>,
+    /// Attempt starts from 1, and increased by 1 for every retry if retry policy is specified.
+    pub attempt: u32,
+
+    /// #NOTE Not in sdk-go
```
Sushisource, line 46: Not super important as a docstring. Feel free to either drop it and I'll add one later, or give it a description.

sdk/src/activity_context.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub activity_id: String,
+    pub activity_type: String,
+    pub task_queue: String,
+    pub heartbeat_timeout: Option<Duration>,
+    /// Time of activity scheduled by a workflow
+    pub scheduled_time: Option<Timestamp>,
+    /// Time of activity start
+    pub started_time: Option<Timestamp>,
+    /// Time of activity timeout
+    pub deadline: Option<Timestamp>,
```
Sushisource, line 42: Sorry I didn't notice this earlier - but lets change these to stdlib types. They're much easier for users to work with than the prost types.

If they cannot be converted the context constructor can bubble up an error.

sdk/src/activity_context.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    /// #NOTE Not in sdk-go
+    pub current_attempt_scheduled_time: Option<Timestamp>,
+    pub retry_policy: Option<RetryPolicy>,
+    pub is_local: bool,
+}
+
+impl ActContext {
+    /// Construct new Activity Context
+    pub fn new(
```
Sushisource, line 54: Probably should be `pub(crate)`

sdk/src/lib.rs
```diff
@@ -347,30 +350,15 @@ impl WorkflowHalf {
...
 impl ActivityHalf {
     /// Spawns off a task to handle the provided activity task
     fn activity_task_handler(
         &mut self,
         worker: Arc<dyn CoreWorker>,
+        task_queue: String,
         activity: ActivityTask,
     ) -> Result<(), anyhow::Error> {
-        match activity.variant {
+        match activity.variant.clone() {
```
Sushisource, line 361: We should avoid this clone - cloning the payloads can be potentially quite expensive.

It loooks like all we really need to to is take out the task token earlier (and it's fine to clone that).

Sushisource (approved):
A few more things I'd like to fix but hitting approve so you don't need to wait on another round for me. If you want fix them, awesome, if not I'll address them. (Though I'm realizing I probably have to hit merge anyway, so just let me know and if you'd prefer me to make them I'll merge this first)

Thanks _so much_ for this awesome contribution!

Issue #298, opened 2022-04-12
[Bug] Workflow cache limit might be exceeded


Sushisource:
We saw during the load test which repro'd the recently fixed deadlock that we may sometimes be allowing more workflows to be cached than the max. Investigate / fix.

PR #304, opened 2022-04-19
Fix problems with exceeding cache limit

Sushisource:
## What was changed
* If a new task would overflow the cache, buffer it an instead return an eviction from polling
* Do not allow configuring worker with more concurrent WFTs than the cache limit
* Also fix a bug where we could potentially fail to send a WFT complete to server if the run had a pending eviction

## Why?
If lang polls faster than it completes, it was possible for it to exceed the cache limit. See added unit test for an example.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/298
2. How was this tested:
Added unit test, new load tests

3. Any docs updates needed?

### DIFF
--- a/core-api/src/worker.rs
+++ b/core-api/src/worker.rs
@@ -20,6 +20,8 @@ pub struct WorkerConfig {
     /// The maximum allowed number of workflow tasks that will ever be given to this worker at one
     /// time. Note that one workflow task may require multiple activations - so the WFT counts as
     /// "outstanding" until all activations it requires have been completed.
+    ///
+    /// Cannot be larger than `max_cached_workflows`.
     #[builder(default = "100")]
     pub max_outstanding_workflow_tasks: usize,
     /// The maximum number of activity tasks that will ever be given to this worker concurrently
@@ -89,6 +91,11 @@ impl WorkerConfigBuilder {
     fn validate(&self) -> Result<(), String> {
         if self.max_concurrent_wft_polls == Some(0) {
             return Err("`max_concurrent_wft_polls` must be at least 1".to_owned());
+        }
+        if self.max_outstanding_workflow_tasks > self.max_cached_workflows {
+            return Err(
+                "`max_outstanding_workflow_tasks` cannot exceed `max_cached_workflows`".to_owned(),
+            );
         }
         Ok(())
     }
--- a/core-api/src/worker.rs
+++ b/core-api/src/worker.rs
@@ -94,7 +94,9 @@ impl WorkerConfigBuilder {
         }
         if self.max_outstanding_workflow_tasks > self.max_cached_workflows {
             return Err(
-                "`max_outstanding_workflow_tasks` cannot exceed `max_cached_workflows`".to_owned(),
+                "Maximum concurrent workflow tasks cannot exceed the maximum number of cached \
+                 workflows"
+                    .to_owned(),
             );
         }
         Ok(())
--- a/core/src/core_tests/queries.rs
+++ b/core/src/core_tests/queries.rs
@@ -80,6 +80,10 @@ async fn legacy_query(#[case] include_history: bool) {
     };
     let clear_eviction = || async {
         let t = worker.poll_workflow_activation().await.unwrap();
+        assert_matches!(
+            t.jobs[0].variant,
+            Some(workflow_activation_job::Variant::RemoveFromCache(_))
+        );
         worker
             .complete_workflow_activation(WorkflowActivationCompletion::empty(t.run_id))
             .await
@@ -127,10 +131,10 @@ async fn legacy_query(#[case] include_history: bool) {
 
     let task = worker.poll_workflow_activation().await.unwrap();
     assert_matches!(
-        task.jobs.as_slice(),
-        [WorkflowActivationJob {
+        task.jobs.as_slice()[0],
+        WorkflowActivationJob {
             variant: Some(workflow_activation_job::Variant::FireTimer(_)),
-        }]
+        }
     );
     worker
         .complete_workflow_activation(WorkflowActivationCompletion::from_cmds(
--- a/core/src/core_tests/queries.rs
+++ b/core/src/core_tests/queries.rs
@@ -131,10 +131,10 @@ async fn legacy_query(#[case] include_history: bool) {
 
     let task = worker.poll_workflow_activation().await.unwrap();
     assert_matches!(
-        task.jobs.as_slice()[0],
-        WorkflowActivationJob {
+        task.jobs.as_slice(),
+        [WorkflowActivationJob {
             variant: Some(workflow_activation_job::Variant::FireTimer(_)),
-        }
+        }]
     );
     worker
         .complete_workflow_activation(WorkflowActivationCompletion::from_cmds(
--- a/core/src/core_tests/workflow_tasks.rs
+++ b/core/src/core_tests/workflow_tasks.rs
@@ -71,7 +71,7 @@ fn single_activity_failure_setup(hist_batches: &'static [usize]) -> Worker {
 #[case::incremental(single_timer_setup(&[1, 2]), NonSticky)]
 #[case::replay(single_timer_setup(&[2]), NonSticky)]
 #[case::incremental_evict(single_timer_setup(&[1, 2]), AfterEveryReply)]
-#[case::replay_evict(single_timer_setup(&[2, 2]), AfterEveryReply)]
+#[case::replay_evict(single_timer_setup(&[2]), AfterEveryReply)]
 #[tokio::test]
 async fn single_timer(#[case] worker: Worker, #[case] evict: WorkflowCachingPolicy) {
     poll_and_reply(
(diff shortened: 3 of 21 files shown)


core/src/workflow/workflow_tasks/cache_manager.rs
```diff
@@ -55,8 +105,19 @@ impl WorkflowCacheManager {
     }
 
     pub fn remove(&mut self, run_id: &str) {
+        let outstanding = self.outstanding_eviction.take();
+        if matches!(outstanding.as_ref(), Some(r) if r != run_id) {
+            // TODO: This isn't an error since eviction could be completed out-of-order?
+            error!("Evicted not-outstanding run");
```
Sushisource, line 111: This likely won't be an error, but I need to add some tests around this. It's unclear to me exactly how this can happen, but it can.

core-api/src/worker.rs
```diff
@@ -90,6 +92,11 @@ impl WorkerConfigBuilder {
         if self.max_concurrent_wft_polls == Some(0) {
             return Err("`max_concurrent_wft_polls` must be at least 1".to_owned());
         }
+        if self.max_outstanding_workflow_tasks > self.max_cached_workflows {
+            return Err(
+                "`max_outstanding_workflow_tasks` cannot exceed `max_cached_workflows`".to_owned(),
+            );
+        }
```
bergundy, line 99: The only issue with this error string is that it uses the rust config property names which are different than the lang ones so it might confuse some users.

Sushisource: I don't think there's any good resolution to that though, besides lang intercepting the error.

core/src/core_tests/queries.rs
```diff
@@ -127,10 +131,10 @@ async fn legacy_query(#[case] include_history: bool) {
 
     let task = worker.poll_workflow_activation().await.unwrap();
     assert_matches!(
-        task.jobs.as_slice(),
-        [WorkflowActivationJob {
+        task.jobs.as_slice()[0],
```
bergundy, line 134: Why only check the first element?

core/src/core_tests/workflow_tasks.rs
```diff
@@ -1737,3 +1737,102 @@ async fn evict_missing_wf_during_poll_doesnt_eat_permit() {
 
     core.shutdown().await;
 }
+
+#[tokio::test]
+async fn poll_faster_than_complete_wont_overflow_cache() {
```
bergundy, line 1742: Love the comments in this test

core/src/core_tests/queries.rs
Sushisource: This can be reverted. I had an earlier change that more eagerly merged evictions but it just caused way too many test failures for me to be comfortable with it.

core/src/core_tests/workflow_tasks.rs
```diff
@@ -1737,3 +1737,102 @@ async fn evict_missing_wf_during_poll_doesnt_eat_permit() {
...
+    // Now we're at cache limit, but there's an open WFT slot. Next poll should return a new
+    // WFT, but also generate an eviction since we're now over the cache size
+    let p4 = core.poll_workflow_activation().await.unwrap();
+    assert_matches!(
+        &p4.jobs[0].variant,
+        Some(workflow_activation_job::Variant::StartWorkflow(sw))
+        if sw.workflow_id == format!("wf-{}", 4)
+    );
+    // Free another WFT slot
+    core.complete_workflow_activation(WorkflowActivationCompletion::from_cmd(
```
bergundy, line 1791: I mentioned this f2f but worth noting here too.
We could send an eviction to lang before accepting the new workflow task and never go over the limit.

core-api/src/worker.rs
bergundy: Lang would need to intercept this specific error string.
We could just rename this from "outstanding" to "concurrent" in Core and it would be good enough.

cretz: FWIW, in Python we also changed `outstanding` to `concurrent`. As for exact var naming, using English words instead of var names can help alleviate those concerns (e.g. just remove the backticks and underscores).

Sushisource: Just used plain English.

core/src/core_tests/workflow_tasks.rs
```diff
@@ -1705,3 +1705,133 @@ async fn evict_missing_wf_during_poll_doesnt_eat_permit() {
...
+            )
+        })
+        .collect();
+    let mut mock = mock_workflow_client();
+    mock.expect_complete_workflow_task()
+        .returning(|_| Ok(Default::default()));
+    let mut mock = MocksHolder::from_client_with_responses(mock, tasks, []);
+    mock.worker_cfg(|wc| {
+        wc.max_cached_workflows = 3;
+        wc.max_outstanding_workflow_tasks = 3;
```
bergundy, line 1729: Worth testing different values here

bergundy: In addition to this test

core/src/core_tests/workflow_tasks.rs
```diff
@@ -1705,3 +1705,133 @@ async fn evict_missing_wf_during_poll_doesnt_eat_permit() {
...
+        start_timer_cmd(1, Duration::from_secs(1)),
+    ))
+    .await
+    .unwrap();
+    // Now we're at cache limit. We will poll for a task, discover it is for a new run, issue
+    // an eviction, and buffer the new run task. However, the run we're trying to evict has pending
+    // activations! Thus, we must complete them first before this poll will unblock, and then it
+    // will unblock with the eviciton.
+    let p4 = async {
+        let p4 = core.poll_workflow_activation().await.unwrap();
```
bergundy, line 1755: Should we verify this poll blocks until the 1st completion or is this covered in other tests?

core/src/core_tests/workflow_tasks.rs
```diff
@@ -1705,3 +1705,133 @@ async fn evict_missing_wf_during_poll_doesnt_eat_permit() {
...
+        assert_matches!(
+            &p4.jobs.as_slice(),
+            [WorkflowActivationJob {
+                variant: Some(workflow_activation_job::Variant::RemoveFromCache(_)),
+            }]
+        );
+        p4
+    };
+    let p2_pending_completer = async {
```
bergundy, line 1764: Not sure I get this part. Aren't we supposed to have enough permits for `p4` to resolve without this `p2_pending_completer`?

core/src/core_tests/workflow_tasks.rs
Sushisource: Unfortunately I can't come up with a good way to do this other than introducing some very intrusive test-specific interceptors or something. I think the best way to deal with this is a bigger restructuring like we've talked about that would make the whole polling process testable at a more granular level.

core/src/core_tests/workflow_tasks.rs
Sushisource: The blocker isn't WFT slots, it's cache limit. The comment above explains this -- because we *have* WFT slots, we get a new task, but we can't issue it, because it's for a new run, so we're at the limit. But we also can't immediately issue an eviction because the run we want to evict has a pending activation.

core/src/core_tests/workflow_tasks.rs
bergundy: You can "race" the two calls and verify the right one completes first.

Sushisource: Ah, sure. Yeah I have ordering checks like that in other places. I can add them here.

core/src/core_tests/workflow_tasks.rs
bergundy: So technically we should be able to issue the eviction above since it doesn't cause the cache to grow

core/src/core_tests/workflow_tasks.rs
```diff
@@ -1705,3 +1705,134 @@ async fn evict_missing_wf_during_poll_doesnt_eat_permit() {
...
+        // Sleep needed because otherwise the complete unblocks waiting for the cache to free a slot
+        // before we have a chance to actually... wait for it.
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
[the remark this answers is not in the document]
Sushisource: Not without putting the worker behind an Arc, and it'd still be timing sensitive. In fact, my guess is the poll would actually happen _later_ in that case because of the small overhead involved in spawning. Right now with it coming first in the `join` it is at least polled once first.

In any case, I can actually remove this now.

cretz (approved):

core/src/core_tests/workflow_tasks.rs
Sushisource: No, we cannot, because the run we would be evicting has an already-outstanding activation.

core/src/worker/mod.rs
```diff
@@ -878,6 +907,7 @@ impl Worker {
                         Ok(())
                     }
                     tonic::Code::NotFound => {
+                        // TODO: Remove outstanding workflow task from workflow task manager
```
bergundy, line 910: Leftover TODO?

core/src/workflow/workflow_tasks/cache_manager.rs
```diff
@@ -28,6 +38,29 @@ impl WorkflowCacheManager {
         Self::new(policy, Default::default())
     }
 
+    /// Resolves once there is an open slot in the cache. The passed in closure can be used to
+    /// exit the wait loop if it returns true even if the cache is not below the limit. It will be
+    /// re-evaluated every time there is an insert or remove to the cache, even if it did not change
+    /// the siz.e
```
bergundy, line 44:
```suggestion
    /// the size.
```

core/src/workflow/workflow_tasks/mod.rs
```diff
@@ -186,6 +198,67 @@ impl WorkflowTaskManager {
...
+    /// Returns number of currently cached workflows
+    pub fn cached_workflows(&self) -> usize {
+        self.workflow_machines.cached_workflows()
+    }
+
+    /// Resolves once there is either capacity in the cache, or there are no pending evictions.
+    /// Inversely: Waits while there are pending evictions and the cache is full.
+    /// Waiting while there are no pending evictions must be avoided because it would block forever,
+    /// since there is no way for the cache size to be reduced.
+    pub async fn wait_for_cache_capacity(&self) {
```
bergundy, line 210: Does this ever get called concurrently?
Is there a scenario where we have 2 waiters on cache capacity and we go over cap?

bergundy (approved):
I wish I could reason about this code better but I haven't touched this in a while and some of it I don't know well at all.
I still wish we would simplify some of the codebase to centralize the logic but of course it doesn't affect this PR.

core/src/workflow/workflow_tasks/mod.rs
Sushisource: That's why, inside the cache manager, there's a lock held while waiting for things to change. All waiters on cache capacity are effectively serialized (though I just realized I can get rid of the watch channel since it can just be a normal channel inside that mutex)

Sushisource:
> I wish I could reason about this code better but I haven't touched this in a while and some of it I don't know well at all. I still wish we would simplify some of the codebase to centralize the logic but of course it doesn't affect this PR.

Agreed. This is too complex right now. I have some good ideas about how to simplify.

In fact while working on the integ test / SDK changes I realized there is another edge case which doesn't break anything but can cause unnecessary slowdown - if we get a workflow task, start replaying, and then evict that workflow while it's replaying, we will fail to then retry replay and time out the task.

Probably, the fix there is to only evict after replay is totally complete (instead of after any outstanding activation is complete), which in theory could cause starvation if long-history workflows dominate the cache, but replay should be fast enough that isn't a concern.

PR #306, opened 2022-04-20
Fix bug where partial-task queries could cause us to empty-reply to WFT

Sushisource:
## What was changed
Queries arriving with a partial workflow task which also missed the cache would trigger a page fetch and then we would erroneously attach the query job to the activation right then. Instead, replay must complete as normal before we finish the query. Fix that.

## Why?
Bugs bad!

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/core_tests/queries.rs
+++ b/core/src/core_tests/queries.rs
@@ -1,4 +1,3 @@
-use crate::telemetry::test_telem_console;
 use crate::{
     test_help::{
         canned_histories, hist_to_poll_resp, mock_worker, MocksHolder, ResponseType, TEST_Q,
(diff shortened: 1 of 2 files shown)


core/src/core_tests/queries.rs
```diff
@@ -381,3 +382,110 @@ async fn legacy_query_after_complete(#[values(false, true)] full_history: bool)
...
+        }]
+    );
+    core.complete_workflow_activation(WorkflowActivationCompletion::from_cmd(
+        task.run_id,
+        start_timer_cmd(1, Duration::from_secs(1)),
+    ))
+    .await
+    .unwrap();
+
+    warn!("Done first compl");
```
bergundy, line 450: Are you leaving this here?

Sushisource: Oops, can get rid of it

bergundy (approved):

PR #307, opened 2022-04-20
Fix extra WFT permits being added in some situations

Sushisource:
## What was changed
* Fixed a few situations where extra permits could get added when they shouldn't have been.
* Further mitigate this by not ever going over permit cap
* Noticed and fixed a situation where local activities that were abandoned completing after cancel would enqueue a pending activation that did nothing.
* Also mitigate that by ignoring pending activations that do nothing.

## Why?
Bugfixing

## Checklist

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?

### DIFF
--- a/core/src/abstractions.rs
+++ b/core/src/abstractions.rs
@@ -7,6 +7,7 @@ use tokio::sync::{AcquireError, Semaphore, SemaphorePermit};
 /// acquired or restored through the provided methods
 pub(crate) struct MeteredSemaphore {
     pub sem: Semaphore,
+    max_permits: usize,
     metrics_ctx: MetricsContext,
     record_fn: fn(&MetricsContext, usize),
 }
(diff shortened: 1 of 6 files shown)


core/src/abstractions.rs
```diff
@@ -30,9 +32,14 @@ impl MeteredSemaphore {
         res
     }
 
-    /// Adds just one permit
+    /// Adds just one permit. Will not add if already at the initial/max capacity.
     pub fn add_permit(&self) {
-        self.sem.add_permits(1);
-        (self.record_fn)(&self.metrics_ctx, self.sem.available_permits());
+        if self.sem.available_permits() < self.max_permits {
```
bergundy, line 37: Isn't this racy?

Sushisource: Shouldn't be, since acquire is async and this is happening in sync code. This is effectively atomic from the viewpoint of someone trying to acquire.

bergundy (approved):

PR #311, opened 2022-04-25
More cache and query fixes

Sushisource:
This needs a follow up PR to fix a problem that doesn't impact correctness, but could cause bad slowdowns:

Right now we'll evict potentially in the middle of replay. In some cases, that will also cause us to drop the workflow task on the floor, timing it out. This can be fixed by never issuing an eviction until replay is complete and WFT is responded to, but that will require a lot of changes to test code which assumes this all over the place.

## What was changed
* Fixed bug with local activity shutdown
* Fixed another possible empty-commands response while querying if there is also outstanding eviction
* Fixed problem where workflow time is unset if queried before first WFT is started
* Refactored Rust SDK to poll independently of completions

## Why?
* Bugfixing

## Checklist

1. Closes 

2. How was this tested:
New and existing tests

3. Any docs updates needed?

### DIFF
--- a/core/src/core_tests/queries.rs
+++ b/core/src/core_tests/queries.rs
@@ -11,8 +11,13 @@ use std::{
 use temporal_sdk_core_api::Worker as WorkerTrait;
 use temporal_sdk_core_protos::{
     coresdk::{
-        workflow_activation::{workflow_activation_job, WorkflowActivationJob},
-        workflow_commands::{CompleteWorkflowExecution, QueryResult, QuerySuccess},
+        workflow_activation::{
+            remove_from_cache::EvictionReason, workflow_activation_job, WorkflowActivationJob,
+        },
+        workflow_commands::{
+            ActivityCancellationType, CompleteWorkflowExecution, ContinueAsNewWorkflowExecution,
+            QueryResult, QuerySuccess, RequestCancelActivity,
+        },
         workflow_completion::WorkflowActivationCompletion,
     },
     temporal::api::{
@@ -26,7 +31,7 @@ use temporal_sdk_core_protos::{
         },
     },
 };
-use temporal_sdk_core_test_utils::start_timer_cmd;
+use temporal_sdk_core_test_utils::{schedule_activity_cmd, start_timer_cmd};
 
 #[rstest::rstest]
 #[case::with_history(true)]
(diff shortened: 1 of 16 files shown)


core/src/core_tests/queries.rs
```diff
@@ -528,3 +533,118 @@ async fn query_cache_miss_causes_page_fetch_dont_reply_wft_too_early(
...
+                ..Default::default()
+            }
+            .into(),
+        ],
+    ))
+    .await
+    .unwrap();
+
+    // Activity unblocked
+    let task = core.poll_workflow_activation().await.unwrap();
```
bergundy, line 600: This shouldn't happen.
The workflow is already completed at this point, what's the use of resolving this activity?

core/src/test_help/mod.rs
```diff
@@ -264,6 +281,10 @@ pub(crate) struct MockPollCfg {
     /// All calls to fail WFTs must match this predicate
     pub expect_fail_wft_matcher:
         Box<dyn Fn(&TaskToken, &WorkflowTaskFailedCause, &Option<Failure>) -> bool + Send>,
+    /// If being used with the Rust SDK, this is set true. It ensures pollers will not error out
+    /// early with no work, since we cannot know the exact number of times polling will happen.
+    /// Instead, they will just block forever.
+    pub for_sdk: bool,
```
bergundy, line 287: Could use a better name for this variable

core/src/workflow/machines/workflow_machines.rs
```diff
@@ -477,6 +477,10 @@ impl WorkflowMachines {
                     if let Some(st) = event.event_time {
                         let as_systime: SystemTime = st.try_into()?;
                         self.workflow_start_time = Some(as_systime);
+                        // Set the workflow time to be the event time of the first event, so that
+                        // if there is a query issued before first WFT started event, there is some
+                        // workflow time set.
+                        self.set_current_time(as_systime);
```
bergundy, line 483: Didn't go in depth here but we should never set the time unless there's no WFT started event.
Can we ensure that?

Sushisource: Why? This will be overridden as soon as there is.

core/src/core_tests/queries.rs
Sushisource: That might be true but it's a whole separate set of changes I'd rather not include in this PR

bergundy: In any case this is fine for TS because we do not activate the workflow after it has completed: https://github.com/temporalio/sdk-typescript/blob/cb6151fcaa29e9cda4a1ae90dda31ae007a6dd10/packages/workflow/src/worker-interface.ts#L218

core/src/workflow/machines/workflow_machines.rs
bergundy: Just that it's a potential source for bugs

Sushisource: I suppose what I mean to say is this is already doing that. This time can never be used unless there is no WFT started event.

core/src/test_help/mod.rs
```diff
@@ -264,6 +281,10 @@ pub(crate) struct MockPollCfg {
     /// All calls to fail WFTs must match this predicate
     pub expect_fail_wft_matcher:
         Box<dyn Fn(&TaskToken, &WorkflowTaskFailedCause, &Option<Failure>) -> bool + Send>,
+    /// If being used with the Rust SDK, this is set true. It ensures pollers will not error out
+    /// early with no work, since we cannot know the exact number of times polling will happen.
+    /// Instead, they will just block forever.
+    pub using_rust_sdk: bool,
```
bergundy, line 287: nit: I'd just call this `poller_error_if_no_work` but this is fine.

bergundy (approved):
My 2 comments about no default time except for query and not issuing activations after workflow transitions to a terminal state still stand but not blocking this PR.
Wanna open issues for those?

core/src/workflow/machines/workflow_machines.rs
bergundy: Why not only use system time when you know we're in a scenario where there's no WFT start event?

Sushisource:
@bergundy Added https://github.com/temporalio/sdk-core/issues/312 for the activations. The time thing is already that way.

PR #313, opened 2022-04-26
Stop evict issue during replay & use `return_new_workflow_task` by default

Sushisource:
…fault

## What was changed
* Do not issue evictions in the middle of replay. This really only serves to cause thrashing if a worker is getting overloaded.
* Use `return_new_workflow_task` when completing tasks by default. Noticeable performance improvement under high load.

## Why?
Better behavior for high-load workers

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (approved):

bergundy:
Not sure what other sdks do here, is this flag on by default in go and java?

cretz:
@bergundy - Via a quick walk of the Go code, this field seems to always be set to true in Go

PR #315, opened 2022-04-29
Fix hang waiting on cache slot

Sushisource:
This line is wrong either way. Trying to verify if this fixes a deadlock.

### DIFF
--- a/core/src/workflow/workflow_tasks/mod.rs
+++ b/core/src/workflow/workflow_tasks/mod.rs
@@ -23,6 +23,7 @@ use futures::FutureExt;
 use parking_lot::Mutex;
 use std::{
     fmt::Debug,
+    future::Future,
     ops::Add,
     sync::Arc,
     time::{Duration, Instant},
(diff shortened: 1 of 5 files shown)


core/src/workflow/workflow_tasks/mod.rs
```diff
@@ -840,10 +840,11 @@ impl WorkflowTaskManager {
...
     fn on_activation_done(&self, run_id: &str) {
-        if self.workflow_machines.delete_activation(run_id).is_some() {
-            self.pending_activations_notifier.notify_waiters();
-        }
-        // It's possible the activation is already removed due to completing an eviction
+        self.workflow_machines.delete_activation(run_id);
+        // It's important to use `notify_one` here to avoid possible races where we're waiting
+        // on a cache slot and fail to realize pending activations must be issued before a slot
+        // will free up.
+        self.pending_activations_notifier.notify_one();
```
Sushisource, line 847: This is the actual fix. When this was protected by the if / not storing the notification, if timing was juuuust right we could be sitting waiting on the cache when we should've realized we instead need to issue a pending activation in order to keep making progress.

tests/load_tests.rs
```diff
@@ -110,7 +110,7 @@ async fn activity_load() {
     dbg!(running.elapsed());
 }
 
-#[tokio::test]
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
```
cretz, line 113: How reliably can you replicate this?

cretz (approved):

bergundy (approved):

PR #319, opened 2022-05-03
Don't retry non retryable local activity application failure

bergundy:


core/src/protosext/mod.rs
```diff
@@ -223,6 +223,25 @@ impl HistoryEventExt for HistoryEvent {
...
+            ..
+        } = self
+        {
+            Some(f)
+        } else {
+            None
+        }
+    }
+}
+
```
Sushisource, line 244: This can go as a direct implementation in the protos crate

Sushisource (approved):

core/src/protosext/mod.rs
bergundy: Missed that, thanks

Issue #321, opened 2022-05-05
[Feature Request] Custom Application Data Propagation (rust sdk)


dt665m:
In the Go sdk, the custom application data is propagated using the standard library context object with a custom key.  The Rust SDK accepts Activities and Workflows as anything that implements the trait `IntoActivityFunc` or `Into<WorkflowFunction>`.  There are helper functions for closures, which are nice.  However, implementing custom structs that convert to these traits takes a lot of boilerplate code.

In real usage, the closures are cumbersome because it is common for an activity to make some external calls (rpc or other).  This usually requires a client object of some sort.   This makes the activity registration look a bit like this if using closures:

```
let client = client.clone();
worker.register_activity(
  "some_activity",
   move |ctx: ActContext, params: Params| {
     let client = client.clone();
     async move { activity_impl(client, ctx, params).await }
    },
);
```

It would be nice to have an API that looks a bit like this:
```
let client = client.clone();
worker.register_activity(
  "some_activity",
  activity_impl,
);

fn activity_impl(ctx: ActCtx, params: Params) -> anyhow::Result<()> {
  let client = ctx.app_data::<Client>();
  ....
}
```

PR #322, opened 2022-05-05
implement appdata propagation

dt665m:
## What was changed
Added an AppData struct that contains a map for any custom data.  The API is heavily inspired by Actix-web.  This AppData is propagated to workflows and activities after being initially setup when creating the workflow client.

## Why?
Rationale in the open issues here: https://github.com/temporalio/sdk-core/issues/321

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/321

2. How was this tested:
Ran Integration Tests and added new Integration Tests

3. Any docs updates needed?
Could use some guidance

### DIFF
--- a/sdk/src/app_data.rs
+++ b/sdk/src/app_data.rs
@@ -1,14 +1,13 @@
 use std::{
     any::{Any, TypeId},
+    collections::HashMap,
     fmt,
 };
-
-use ahash::AHashMap;
 
 /// A Wrapper Type for workflow and activity app data
 #[derive(Default)]
 pub struct AppData {
-    map: AHashMap<TypeId, Box<dyn Any + Send + Sync>>,
+    map: HashMap<TypeId, Box<dyn Any + Send + Sync>>,
 }
 
 impl AppData {
--- a/sdk/src/app_data.rs
+++ b/sdk/src/app_data.rs
@@ -1,0 +1,38 @@
+use std::{
+    any::{Any, TypeId},
+    fmt,
+};
+
+use ahash::AHashMap;
+
+/// A Wrapper Type for workflow and activity app data
+#[derive(Default)]
+pub struct AppData {
+    map: AHashMap<TypeId, Box<dyn Any + Send + Sync>>,
+}
+
+impl AppData {
+    /// Insert an item, overwritting duplicates
+    pub fn insert<T: Send + Sync + 'static>(&mut self, val: T) -> Option<T> {
+        self.map
+            .insert(TypeId::of::<T>(), Box::new(val))
+            .and_then(downcast_owned)
+    }
+
+    /// Get a reference to a type in the map
+    pub fn get<T: 'static>(&self) -> Option<&T> {
+        self.map
+            .get(&TypeId::of::<T>())
+            .and_then(|boxed| boxed.downcast_ref())
+    }
+}
+
+impl fmt::Debug for AppData {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("AppData").finish()
+    }
+}
+
+fn downcast_owned<T: Send + Sync + 'static>(boxed: Box<dyn Any + Send + Sync>) -> Option<T> {
+    boxed.downcast().ok().map(|boxed| *boxed)
+}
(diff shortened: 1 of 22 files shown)


tests/integ_tests/workflow_tests/appdata_propagation.rs
```diff
@@ -0,0 +1,63 @@
...
+use temporal_sdk_core_test_utils::CoreWfStarter;
+
+const TEST_APPDATA_MESSAGE: &str = "custom app data, yay";
+
+struct Data {
+    message: String,
+}
+
+pub async fn appdata_activity_wf(ctx: WfContext) -> WorkflowResult<()> {
+    let data = ctx.app_data::<Data>().expect("appdata exists. qed");
```
bergundy, line 15: I'm worried that users will use this to access environment specific data resulting in non-deterministic behavior.
For activities this API is totally fine.

dt665m: I mulled over this quite a bit too.  It is a little like providing nulls in a programming language and trusting everyone will check it all the time (or be cognizant).  However, I ultimately added it in because I imagine someone will ask for this kind of context propagation for things like tracing (which is in one of the go sdk samples).  I'm happy to take it out for now.  Just having it for activities is enough for my own use case.

Sushisource: I think we should leave it out for now. Keep in mind the Workflow APIs especially are nowhere close to what I expect them to be later.

Ultimately I think workflows will probably look more like implementing a trait in the long run, in which case you could just include this stuff in the fields of your struct which implements that trait.

sdk/src/app_data.rs
```diff
@@ -0,0 +1,38 @@
+use std::{
+    any::{Any, TypeId},
+    fmt,
+};
+
+use ahash::AHashMap;
```
Sushisource, line 6: I think we can just use a normal hash map here

Sushisource: Alternatively, https://crates.io/crates/typemap

sdk/src/lib.rs
```diff
@@ -206,11 +210,18 @@ impl Worker {
         );
     }
 
+    /// Insert Custom App Context for Workflows and Activities
+    pub fn insert_app_data<T: Send + Sync + 'static>(&mut self, data: T) {
+        (*Arc::get_mut(&mut self.app_data)
+            .expect("app_data can only be inserted during initial setup"))
+        .insert(data);
```
Sushisource, line 217: What would probably make sense to do here is not `Arc` this data in `Worker`, but only once `run` has started, that way you can avoid this expectation. It could be behind an `Option` and is `take`n when run starts, and replaced when it's done.

tests/integ_tests/workflow_tests/appdata_propagation.rs
cretz: :+1: For data on `self`

sdk/src/app_data.rs
dt665m: Typemap is pretty cool, didn't know about it.  It requires a "Key" trait implementation in order to insert and retrieve though.  Do you want to expose this trait as a "pub use" from the rust SDK or ask people to import the crate themselves?  I'm going to remove the ahash crate first and use a normal hashmap first while we decide about this.

sdk/src/lib.rs
dt665m: done.

sdk/src/lib.rs
```diff
@@ -115,6 +117,7 @@ pub struct Worker {
     common: CommonWorker,
     workflow_half: WorkflowHalf,
     activity_half: ActivityHalf,
+    app_data: Option<AppData>,
```
Sushisource, line 120: Docstring stating why this is optional and how it gets used that way would be a useful reminder

Sushisource (approved):
Sorry for the delayed approval, lost this on my radar for a bit.

Thanks again for contributing!

Sushisource:
@dt665m I keep forgetting I need to merge these for you! Merged now, sorry about that :)

dt665m:
> @dt665m I keep forgetting I need to merge these for you! Merged now, sorry about that :)

No worries.  Just glad it got merged.  Thanks! 🚀

Issue #303, opened 2022-04-18
[Feature Request] Dynamic HTTP headers


bergundy:
Currently it's only possible to configure static HTTP headers on the Client, we need to add a way to specify dynamic ones.
A typical use case for this would be to refresh an auth token and users should not be required to recreate their workers for that.

cretz:
I think an `Option<Fn(&mut HashMap<String, String>)>` is reasonable at the client level (or `tonic::metadata::MetadataMap` or `http::header::HeaderMap` or whatever). But I wonder if we want to discuss general purpose gRPC interception lang callbacks at this time or save that discussion for later.

bergundy:
We will general interceptors at some point but I don't think it has to be tied to dynamic headers.
At some point soon we should also consider exposing the deadline for individual calls too.

bergundy:
We already have dynamic headers, keeping this open for per request headers and deadline

jrpedrianes:
Is there any news on this? we need to change a header per request to include a security token. In java we can do it, but I'm not pretty sure that we can do it in JS SDK

bergundy:
@jrpedrianes for JS clients, you can set per request headers and deadline. This issue should be closed.
See the api reference here: https://typescript.temporal.io/api/classes/client.Connection#withmetadata

jrpedrianes:
Thanks, @bergundy.

At worker level I see that I can use https://typescript.temporal.io/api/classes/worker.NativeConnection#setmetadata, the method signature is not the same, but it seems that I can call it as many times as I need, in my case to update a security token that a worker needs to send to the temporal server, this token expires, so I need to update it frequently.

PR #326, opened 2022-05-10
Support modifying Client headers

bergundy:
Does not address need to set per-call headers, make the Client headers mutable so they can be updated from Lang
Related: #303

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -31,7 +31,7 @@ use std::{
     fmt::{Debug, Formatter},
     ops::{Deref, DerefMut},
     str::FromStr,
-    sync::{Arc},
+    sync::Arc,
     time::{Duration, Instant},
 };
 use temporal_sdk_core_protos::{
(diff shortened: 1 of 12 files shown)


client/src/lib.rs
```diff
@@ -80,9 +80,9 @@ pub struct ClientOptions {
     /// in all RPC calls. The server decides if the client is supported based on this.
     pub client_version: String,
 
-    /// Other non-required static headers which should be attached to every request
+    /// Other non-required headers which should be attached to every request
     #[builder(default)]
-    pub static_headers: HashMap<String, String>,
+    pub headers: Arc<Mutex<HashMap<String, String>>>,
```
Sushisource, line 85: Use a parking_lot mutex to avoid fallible locking

Sushisource: I also don't really love the idea of this being what gets mutated. IMO the options should remain static.

We could change this to be `default_headers` and then put this as a field on the interceptor specifically named `header_overrides` and that can be changed with a function call.

bergundy: Can you explain why that's better?

client/src/lib.rs
```diff
@@ -394,6 +408,7 @@ pub struct WorkflowTaskCompletion {
 #[derive(Clone)]
 pub struct ServiceCallInterceptor {
     opts: ClientOptions,
+    headers: Arc<Mutex<HashMap<String, String>>>,
```
Sushisource, line 411: Should probably doc here that they are only ever meant to be _read_ from here, not written.

Additionally, probably makes a lot of sense for this to be a `RwLock`

Sushisource (approved):

Issue #180, opened 2021-08-24
[Refactor] Separate polling & completing "halves" of WFT/activity management


Sushisource:
The last round of refactoring that moved WFT management inside of workers still leaves room for improvement. Hopefully the last round of work we need to do here for likely a rather long time.

There are a bunch of synchronization concerns that would be made simpler by following the below pattern for both WFTs and Activities:

- Polling happens in the background (this is largely already true, just need to remove limiting semaphore)
    - Locks workflow run state to make needed updates
    - Poll results are fed into a stream
- Poller selects from stream(s) (for wft: pending, real poll results)
- Completing locks workflow run state to make needed updates

By keeping the polling half separate from the completing half, locking (or, whatever synchronization mechanism, could be message passing) permits polling and completing to happen simultaneously per-worker, but there cannot be multiple concurrent polls. Multiple concurrent completions for different workflow runs is permitted (as it is now).

Before doing this I think it's probably worth diagramming it out, to land on a really nice final iteration

PR #328, opened 2022-05-13
Workflow state refactor

Sushisource:
Still draft while I clean up a few things

## What was changed
Big refactoring of how all workflow state and task processing is managed

## Why?
Avoid locks, which caused a few production issues w/ deadlocks
Generally attempt to make code-flow simpler to understand. Honestly, I don't think this is an unqualified success here, it's still very complex, but that's the nature of this beast. I think it is at least _better_, and more amenable to future encapsulation, but this has already taken much longer than desired

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/180

2. How was this tested:
Existing and new tests

3. Any docs updates needed?


core-api/src/errors.rs
```diff
@@ -4,7 +4,6 @@ use prost_types::TimestampOutOfSystemRangeError;
 use temporal_sdk_core_protos::coresdk::{
```
bergundy, line 4: Looks like this file still has `NoWorkerForQueue` errors defined but AFAIK that's not returned anymore.

core/src/abstractions.rs
```diff
@@ -32,6 +39,12 @@ impl MeteredSemaphore {
         res
     }
 
+    pub fn try_acquire_owned(&self) -> Result<OwnedSemaphorePermit, TryAcquireError> {
+        let res = self.sem.clone().try_acquire_owned();
+        (self.record_fn)(&self.metrics_ctx, self.sem.available_permits());
```
bergundy, line 44: Does it make sense to record if res is error?

core/src/abstractions.rs
```diff
@@ -43,3 +56,84 @@ impl MeteredSemaphore {
         }
     }
 }
+
+#[derive(Clone)]
+pub(crate) struct StreamAllowHandle {
+    acceptable_to_get_next: Arc<AtomicBool>,
```
bergundy, line 62: I'm pretty sure you can implement this without the atomic bool if you call `notify_one` on construction.

core/src/test_help/mod.rs
```diff
@@ -285,17 +341,31 @@ pub(crate) struct MockPollCfg {
     /// Instead, they will just block forever.
     pub using_rust_sdk: bool,
 }
+#[derive(Copy, Clone)]
+pub(crate) enum ExpectationAmount {
+    Zero,
+    Specific(usize),
+}
```
bergundy, line 348: Why not just use `usize`?

core/src/core_tests/workflow_tasks.rs
```diff
@@ -1694,16 +1711,28 @@ async fn evict_missing_wf_during_poll_doesnt_eat_permit() {
...
         wc.max_outstanding_workflow_tasks = 1;
     });
     let core = mock_worker(mock);
 
-    // Should error because mock is out of work
-    assert_matches!(core.poll_workflow_activation().await, Err(_));
-    assert_eq!(core.available_wft_permits(), 1);
+    // We will get an eviction for the workflow (which lang never saw in the first place, but,
+    // that's fine) since it had an error while being created, and then we will evict it and
+    // the permit shall be restored.
```
bergundy, line 1723: I verified, this should be okay in TS but we should probably test this flow in all Core based SDKs

core/src/core_tests/workflow_tasks.rs
```diff
@@ -1893,3 +1918,69 @@ async fn eviction_waits_until_replay_finished() {
...
+#[tokio::test]
+async fn autocompletes_wft_no_work() {
+    let wfid = "fake_wf_id";
+    let activity_id = "1";
+
+    let mut t = TestHistoryBuilder::default();
+    t.add_by_type(EventType::WorkflowExecutionStarted);
+    t.add_full_wf_task();
+    let scheduled_event_id = t.add_activity_task_scheduled(activity_id);
+    let started_event_id = t.add_activity_task_started(scheduled_event_id);
```
bergundy, line 1931: Started would only come right before completed

core/src/core_tests/workflow_tasks.rs
```diff
@@ -1893,3 +1918,69 @@ async fn eviction_waits_until_replay_finished() {
 
     core.shutdown().await;
 }
+
+#[tokio::test]
+async fn autocompletes_wft_no_work() {
```
bergundy, line 1923: Curious why this test was added, what's special about this flow?

core/src/abstractions.rs
Sushisource: It's cheap. Branching just adds extra lines here.

core/src/worker/workflow/managed_run.rs
```diff
@@ -0,0 +1,606 @@
...
+
+        if activation.jobs.is_empty() {
+            if self.wfm.machines.outstanding_local_activity_count() > 0 {
+                // If the activation has no jobs but there are outstanding LAs, we need to restart the
+                // WFT heartbeat.
+                if let Some(ref mut lawait) = self.waiting_on_la {
+                    if lawait.completion_dat.is_some() {
+                        panic!("Should not have completion dat when getting new wft & empty jobs")
+                    }
+                    lawait.heartbeat_timeout_task.abort();
```
bergundy, line 200: What happens to the result of this task when aborted? Do we care?

core/src/test_help/mod.rs
Sushisource: This went through a few evolutions, but yeah it can just be a usize now

core/src/worker/workflow/managed_run.rs
Sushisource: The task has no result. We don't care

core/src/abstractions.rs
Sushisource: Yeah good point

core/src/abstractions.rs
```diff
@@ -43,3 +53,74 @@ impl MeteredSemaphore {
...
+/// From the input stream, create a new stream which only pulls from the input stream when allowed.
+/// The stream will start allowed, and sets itself to unallowed each time it delivers the next item.
+/// The returned [StreamAllowHandle::allow_one] function can be used to indicate it is OK to deliver
+/// the next item.
+pub(crate) fn stream_when_allowed<S>(input: S) -> (StreamAllowHandle, impl Stream<Item = S::Item>)
+where
+    S: Stream + Send + 'static,
+{
+    let acceptable_notify = Arc::new(Notify::new());
+    acceptable_notify.notify_one();
```
bergundy, line 76: After thinking about it some more you can probably even remove this if you change the waiting order below to:

```
input.next()....
notified()...
```

core/src/worker/workflow/managed_run.rs
```diff
@@ -0,0 +1,602 @@
...
+            .take()
+            .expect("Completion response channel must be populated");
+
+        let outcome = async move {
+            // Send commands from lang into the machines then check if the workflow run
+            // needs another activation and mark it if so
+            self.wfm.push_commands(completion.commands).await?;
+            // Don't bother applying the next task if we're evicting at the end of
+            // this activation
+            if !completion.activation_was_eviction {
```
bergundy, line 240: What happens if there's an unprocessed WFT while we're evicting? Is this possible?

core/src/worker/workflow/managed_run.rs
```diff
@@ -0,0 +1,602 @@
...
+        &mut self,
+        want_to_evict: Option<RequestEvictMsg>,
+        has_pending_queries: bool,
+    ) -> Result<Option<ActivationOrAuto>, RunUpdateErr> {
+        if self.wfm.machines.has_pending_jobs() && !self.am_broken {
+            Ok(Some(ActivationOrAuto::LangActivation(
+                self.wfm.get_next_activation().await?,
+            )))
+        } else {
+            if has_pending_queries && !self.am_broken {
```
bergundy, line 319: Does this mean we can't query failed workflows? (not 100% sure what it means to be broken)

core/src/worker/workflow/managed_run.rs
```diff
@@ -0,0 +1,602 @@
...
+                }
+            }
+        }
+        Ok(())
+    }
+
+    /// Returns true if autocompletion should be issued, which will actually cause us to end up
+    /// in [completion] again, at which point we'll start a new heartbeat timeout, which will
+    /// immediately trigger and thus finish the completion, forcing a new task as it should.
+    fn heartbeat_timeout(&mut self) -> Result<bool, RunUpdateErr> {
```
bergundy, line 408: Can just return bool

core/src/worker/workflow/managed_run.rs
```diff
@@ -0,0 +1,602 @@
...
+            // Cancel the heartbeat timeout
+            wait_dat.heartbeat_timeout_task.abort();
+            if let Some((completion_dat, resp_chan)) = wait_dat.completion_dat.take() {
+                self.finish_completion(resp_chan, completion_dat, true);
+            } else {
+                // Auto-reply WFT complete
+                return Ok(true);
+            }
+        }
+        Ok(false)
```
bergundy, line 419: Add some comment saying this shouldn't happen

core/src/worker/workflow/wft_poller.rs
```diff
@@ -0,0 +1,86 @@
...
+                        Err(e) => {
+                            error!(error=?e, "Server returned an unparseable workflow task");
+                            continue;
+                        }
+                    };
+                    metrics.wf_tq_poll_ok();
+                    return Some((work, (poller, metrics)));
+                }
+                Some(Err(e)) => {
+                    warn!(error=?e, "Error while polling for workflow tasks");
```
bergundy, line 33: Metrics?

core/src/worker/workflow/wft_poller.rs
```diff
@@ -0,0 +1,86 @@
...
+                }
+                None => return None,
```
bergundy, line 35: Could use a comment explaining what's the meaning of `None` here

core/src/worker/workflow/workflow_stream.rs
```diff
@@ -0,0 +1,927 @@
...
+    fn run_metrics(&mut self, run_id: &str) -> Option<&MetricsContext> {
+        self.runs.get(run_id).map(|r| &r.metrics)
+    }
+
+    fn activation_has_only_eviction(&mut self, run_id: &str) -> bool {
+        self.runs
+            .get(run_id)
+            .and_then(|rh| rh.activation)
+            .map(OutstandingActivation::has_only_eviction)
+            .unwrap_or_default()
```
bergundy, line 885: Should we be returning default or an error if run isn't found?
Looks like we might be silently ignoring errors.

(Applies below too)

core/src/worker/workflow/workflow_stream.rs
```diff
@@ -0,0 +1,927 @@
...
+            let about_to_issue_evict = run.trying_to_evict.is_some() && !run.last_action_acked;
+            let has_wft = run.wft.is_some();
+            let has_activation = run.activation.is_some();
+            if has_wft
+                || has_activation
+                || about_to_issue_evict
+                || run.more_pending_work
+                || !run.last_action_acked
+            {
+                debug!(run_id = %run_id, run = ?run,
```
bergundy, line 778: There are a few debug statements in this file which I would consider making trace level, this one included.

core/src/worker/workflow/workflow_stream.rs
```diff
@@ -0,0 +1,927 @@
...
+                Some(work)
+            }
+        } else {
+            Some(work)
+        }
+    }
+
+    fn buffer_resp_on_full_cache(&mut self, work: ValidPollWFTQResponse) {
+        debug!(run_id=%work.workflow_execution.run_id, "Buffering WFT because cache is full");
+        // If there's already a buffered poll for the run, replace it.
```
bergundy, line 792: When would this happen?

bergundy: And is it okay to replace the already buffered task?

core/src/worker/workflow/workflow_stream.rs
```diff
@@ -0,0 +1,927 @@
...
+            if !activation_has_eviction && rh.trying_to_evict.is_none() {
+                debug!(run_id=%info.run_id, reason=%info.message, "Eviction requested");
+                rh.trying_to_evict = Some(info);
+                rh.check_more_activations();
+                EvictionRequestResult::EvictionRequested(attempts)
+            } else {
+                EvictionRequestResult::EvictionAlreadyRequested(attempts)
+            }
+        } else {
+            warn!(run_id=%info.run_id, "Eviction requested for unknown run");
```
bergundy, line 672: Is this for lang requesting eviction from core?

bergundy: Technically this should never be possible otherwise, right?

core/src/worker/workflow/workflow_stream.rs
```diff
@@ -0,0 +1,927 @@
...
+                (
+                    entry.info.task_token.clone(),
+                    !entry.pending_queries.is_empty(),
+                    entry.start_time,
+                )
+            } else {
+                if !activation_was_only_eviction {
+                    // Don't bother warning if this was an eviction, since it's normal to issue
+                    // eviction activations without an associated workflow task in that case.
+                    warn!(
```
bergundy, line 509: Should this be an `error!`? When can it happen?

core/src/worker/workflow/managed_run.rs
Sushisource: Can't query workflows who've had some internal error. Failing them the normal way is fine.

core/src/worker/workflow/managed_run.rs
Sushisource: WFT has to be finished and replied to before eviction will happen

core/src/worker/workflow/wft_poller.rs
Sushisource: There is no specific metric for task queue poll failure. There is one for all grpc errors, which is handled inside the client.

core/src/worker/workflow/workflow_stream.rs
Sushisource: For the purposes of these methods we only care if the thing we're looking for is true or not. The run's existence will need to be ensured elsewhere anyway - so it just introduces extra verbosity to have these return results

core/src/worker/workflow/workflow_stream.rs
Sushisource: Timing out repeatedly while the workflow is getting new events (like signals). It's rare, but possible.

core/src/worker/workflow/workflow_stream.rs
Sushisource: It could be either lang or internally

core/src/worker/workflow/workflow_stream.rs
```diff
@@ -0,0 +1,927 @@
...
+        run_id: String,
+        cause: WorkflowTaskFailedCause,
+        reason: EvictionReason,
+        failure: Failure,
+        resp_chan: oneshot::Sender<ActivationCompleteResult>,
+    ) {
+        let tt = if let Some(tt) = self.get_task(&run_id).map(|t| t.info.task_token.clone()) {
+            tt
+        } else {
+            warn!(
```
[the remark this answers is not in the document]
Sushisource: Shouldn't really be possible I think. Can probably also make this a debug-only panic

core/src/worker/workflow/workflow_stream.rs
Sushisource: Same, this shouldn't happen in the course of normal operations. I can also make it a debug panic.

core/src/worker/workflow/managed_run.rs
bergundy: So there's no way there could be a new task to apply if `completion.activation_was_eviction`?

core/src/worker/workflow/managed_run/managed_wf_test.rs
```diff
@@ -0,0 +1,194 @@
+use super::*;
```
bergundy, line 1: This file is named `..._test.rs` but does not appear to contain any tests.

core/src/worker/workflow/run_cache.rs
```diff
@@ -0,0 +1,152 @@
...
+        &mut self,
+        run_id: &str,
+        workflow_id: &str,
+        wf_type: &str,
+        history_update: HistoryUpdate,
+        start_time: Instant,
+    ) -> &mut ManagedRunHandle {
+        let cur_num_cached_runs = self.runs.len();
+
+        if self.runs.contains(run_id) {
```
bergundy, line 59: Is there not a sticky_cache_miss metric?

core/src/worker/workflow/managed_run/managed_wf_test.rs
Sushisource: The code is only in support of tests. The whole module is behind a `cfg(test)`

bergundy (approved):
Didn't manage to finish the review but I'm going to approve to unblock further work.
I plan on reviewing later and will make sure to summarize any late comments.

core/src/abstractions.rs
bergundy: nit @Sushisource

PR #329, opened 2022-05-17
Removed typo in workflow state machines.

debuggerpk:
## What was changed

When terminating a workflow, the error message read "The event is non a non-stateful event, but we tried to handle it as one.". Should be "The event is not a non-stateful event, but we tried to handle it as one"

## Why?

For verbosity.

## Checklist

1. Closes 

2. How was this tested:

No.

3. Any docs updates needed?

No

### DIFF
--- a/core/src/workflow/machines/workflow_machines.rs
+++ b/core/src/workflow/machines/workflow_machines.rs
@@ -525,7 +525,7 @@ impl WorkflowMachines {
             }
             _ => {
                 return Err(WFMachinesError::Fatal(format!(
-                    "The event is non a non-stateful event, but we tried to handle it as one: {}",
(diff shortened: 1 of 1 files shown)


Sushisource:
@debuggerpk Thanks for the fix! Out of curiosity - did you encounter this error in a real situation?

debuggerpk:
> @debuggerpk Thanks for the fix! Out of curiosity - did you encounter this error in a real situation?

_Real_?

Not Real, but during development with Typescript SDK. I am watching both the `workflow` files and `worker.ts` to restart and the workflow IDs that are being generated are currently random. The activities currently return `void`. 

I think it has got something to do with auto restart.

PR #330, opened 2022-05-18
Fix documentation for the sdk worker example

jakejscott:
## What was changed
The sdk worker example doesn't run out of the box, had to make a couple tweaks

## Why?
I wanted to test out the rust sdk

## Checklist

- [ ] Create an issue

2. How was this tested:
Create a test app locally, got it running and then pasted the code back into the docs.

3. Any docs updates needed?
N/A

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -35,7 +35,6 @@
 //!         |_ctx: ActContext, echo_me: String| async move { Ok(echo_me) },
 //!     );
 //! 
-//!     println!("Running worker");
 //!     worker.run().await?;
 //! 
 //!     Ok(())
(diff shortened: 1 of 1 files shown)


jakejscott:
NOTE: If you don't initialize the telemetry it panics

```
thread 'main' panicked at 'Tried to use a metric but telemetry has not been initialized', C:\Users\jaken\.cargo\git\checkouts\sdk-core-3d5ba50f52602518\1f4a9ac\core\src\telemetry\metrics.rs:157:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\debug\temporal-rust.exe` (exit code: 101)
```

Sushisource (approved):
Thanks for this! I'll make a note to fix telemetry as well so it just no-ops rather than panicking if not initted.

PR #334, opened 2022-05-31
Add per-worker activity ratelimiting

Sushisource:
## What was changed
Add local ratelimiting which exists in other SDKs

## Why?
Parity / user ask

## Checklist

1. Enables https://github.com/temporalio/sdk-typescript/issues/48

2. How was this tested:
Added UT. 90% is implemented by library.

3. Any docs updates needed?


core-api/src/worker.rs
```diff
@@ -73,6 +73,13 @@ pub struct WorkerConfig {
     /// winning.
     #[builder(setter(strip_option), default)]
     pub max_task_queue_activities_per_second: Option<f64>,
+
+    /// Limits the number of activities per second that this worker will process. The worker will
+    /// not poll for new activities if by doing so it might receive and execute an activity which
+    /// would cause it to exceed this limit. Negative, zero, or NaN values will be ignored, and is
+    /// equivalent to not setting any value (ie: no rate limit).
+    #[builder(setter(strip_option), default)]
+    pub max_worker_activities_per_second: Option<f64>,
```
bergundy, line 82: Are we ignoring for the sake of langs that don't have a built-in option type? AFAIC, we can error in this case and let each lang wrap appropriately.
I'd rather that then silently ignoring invalid values.

bergundy (approved):

bergundy:
That was fast :)

core-api/src/worker.rs
Sushisource: Option had nothing to do with it. Sure, we can error on bad values instead.

PR #339, opened 2022-06-01
Don't panic if an empty evict reply comes in after shutdown is done

Sushisource:
All in title

### DIFF
--- a/core/src/worker/workflow/mod.rs
+++ b/core/src/worker/workflow/mod.rs
@@ -210,13 +210,25 @@ impl Workflows {
         &self,
         completion: WorkflowActivationCompletion,
     ) -> Result<usize, CompleteWfError> {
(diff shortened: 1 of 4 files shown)


core/src/worker/workflow/mod.rs
```diff
@@ -210,13 +210,25 @@ impl Workflows {
         &self,
         completion: WorkflowActivationCompletion,
     ) -> Result<usize, CompleteWfError> {
+        let is_empty_completion = completion.is_empty();
```
bergundy, line 213: Empty completion doesn't necessarily imply eviction response (as you obviously know).

bergundy (approved):
I'm fine having this workaround for now but let's open an issue to properly fix shutdown.
As discussed on slack, there's not much to fix in the actual code, the pain is just in test refactoring.

Issue #305, opened 2022-04-19
[Bug] Responding to legacy query can fail with not found


Sushisource:
```
"log":" error: [TransportError: Unhandled grpc error when completing workflow task: Status { code: NotFound, message: \"query task not found, or already expired\", details: b\"\\x08\\x05\\x12(query task not found, or already expired\\x1aB\\n@type.googleapis.com/temporal.api.errordetails.v1.NotFoundFailure\", metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\"} }, source: None }]"
```

Looks like we just need to trap not found errors specifically for legacy query responses like we do for normal WFT responses.

bergundy:
Happened again in TS nightly, we consider this a fatal error and it crashes the worker, let's fix this ASAP.

PR #342, opened 2022-06-06
Fix panic on not-found when responding to legacy queries

Sushisource:
## Why?
Not a fatal error

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/305

2. How was this tested:
UT

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/mod.rs
+++ b/core/src/worker/workflow/mod.rs
@@ -281,7 +281,7 @@ impl Workflows {
                     task_token,
                     action: ActivationAction::RespondLegacyQuery { result },
                 } => {
-                    self.client.respond_legacy_query(task_token, result).await?;
+                    self.respond_legacy_query(task_token, result).await?;
                     true
                 }
             },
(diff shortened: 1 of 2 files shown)


core/src/worker/workflow/mod.rs
```diff
@@ -489,6 +488,22 @@ impl Workflows {
...
+    /// Wraps responding to legacy queries. Handles ignore-able failures.
+    async fn respond_legacy_query(
+        &self,
+        tt: TaskToken,
+        res: QueryResult,
+    ) -> Result<(), tonic::Status> {
+        match self.client.respond_legacy_query(tt, res).await {
+            Ok(_) => Ok(()),
+            Err(e) if e.code() == tonic::Code::NotFound => {
+                warn!(error=?e,"Query not found when attempting to respond to it");
```
bergundy, line 501: I'm assuming you want to warn to let users know that query handling was too slow.. if that's the rationale LGTM

bergundy (approved):

core/src/worker/workflow/mod.rs
Sushisource: It is :)

PR #343, opened 2022-06-06
Fix possible panic when racing WFT permits

Sushisource:
## What was changed
Fixed possible race in logic around whether poller should poll or not.
Also make this no longer panic in production code. Temporarily exceeding WFT capacity is better than crashing.

## Why?
Bug

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/core_tests/workflow_tasks.rs
+++ b/core/src/core_tests/workflow_tasks.rs
@@ -40,7 +40,7 @@ use temporal_sdk_core_protos::{
     temporal::api::{
         enums::v1::{EventType, WorkflowTaskFailedCause},
         failure::v1::Failure,
-        history::v1::{history_event, History, TimerFiredEventAttributes},
+        history::v1::{history_event, TimerFiredEventAttributes},
         workflowservice::v1::{
             GetWorkflowExecutionHistoryResponse, RespondWorkflowTaskCompletedResponse,
         },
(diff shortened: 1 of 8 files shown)


core/src/core_tests/workflow_tasks.rs
```diff
@@ -1687,54 +1694,6 @@ async fn tasks_from_completion_are_delivered() {
     core.shutdown().await;
 }
 
-#[tokio::test]
-async fn evict_missing_wf_during_poll_doesnt_eat_permit() {
```
Sushisource, line 1691: This test serves little purpose now since all the permits are owned. It's very high-effort to make work since the poll acquires the permit internally, thus a permit "used" immediately again before we get a new task. In this test, there is no new task, which means we shut down (which releases the permit, but also renders the check pointless).

core/src/worker/workflow/workflow_stream.rs
```diff
@@ -631,8 +633,16 @@ impl WFStream {
...
         };
 
-        // If we reported to server, we always want to mark it complete.
-        self.complete_wft(run_id, report.reported_wft_to_server);
+        if let Some(wft) = report.wft_from_complete {
+            debug!(run_id=%wft.workflow_execution.run_id, "New WFT from completion");
+            if let Some(t) = maybe_t {
+                self.instantiate_or_update(PermittedWFT {
+                    wft,
+                    permit: t.permit,
```
bergundy, line 641: Nice way to solve this

bergundy (approved):
This is better IMHO, nice!

core/src/worker/workflow/workflow_stream.rs
```diff
@@ -113,29 +110,37 @@ impl WFStream {
...
+            basics.max_outstanding_wfts,
+            basics.metrics.with_new_attrs([workflow_worker_type()]),
+            MetricsContext::available_task_slots,
         );
+        let wft_sem_clone = wft_semaphore.clone();
+        let proceeder = move || {
+            let wft_sem_clone = wft_sem_clone.clone();
+            async move { wft_sem_clone.acquire_owned().await.unwrap() }
+        };
+        let (external_wft_allow_handle, poller_wfts) =
```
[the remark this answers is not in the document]
Sushisource: The handle? You very much do, otherwise we'll poll when the cache is full.

Issue #335, opened 2022-05-31
[Feature Request] WorkflowService missing get_workflow_execution_history_reverse


cretz:
### Is your feature request related to a problem? Please describe.

I have a CI process in Python that tells me when any `WorkflowService` call wrapper is missing compared to the protos included, and it flagged `get_workflow_execution_history_reverse` as missing, but unable to add to my wrapper because I assume it's not proxied in `WorkflowService` in `client/src/raw.rs`.

### Describe the solution you'd like

* Add `get_workflow_execution_history_reverse` to `WorkflowService` trait
* If possible, add test or some other CI check that makes sure any new API in the `protos/api_upstream` `WorkflowService` is also proxied in source

Issue #333, opened 2022-05-31
[Bug] Cannot set `WorkflowOptions.max_cached_workflows` to 0


bergundy:
Since we have a check that verifies that `max_outstanding_workflow_tasks` is `<=` `max_cached_workflows`

Issue #325, opened 2022-05-06
[Feature Request] Account for new `NAMESPACE_NOT_FOUND` enums.


cretz:
### Is your feature request related to a problem? Please describe.

https://github.com/temporalio/api/pull/173 introduced namespace not found failure cause enums.

### Describe the solution you'd like

* Update upstream API
* Need `coresdk.child_workflow.StartChildWorkflowExecutionFailedCause.START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_NAMESPACE_NOT_FOUND`
* Fix match for `CancelExternalWorkflowExecutionFailedCause`
* Fix match for `SignalExternalWorkflowExecutionFailedCause`

cretz:
Also confirm that we fail the poll when this is returned (which I believe we do for NotFound)

Issue #318, opened 2022-05-03
[Bug] ReplayWorker errors if history contains WorkflowExecutionTerminated event


bergundy:
### What are you really trying to do?

Replay a previously terminated workflow

### Describe the bug

Don't remember the exact details but should be easily reproducible

### Minimal Reproduction

Start a workflow that blocks, terminate it and use history in replayer

PR #345, opened 2022-06-10
Fix a handful of minor bugs / feature reqs

Sushisource:
## What was changed
* Remove not-necessary allow call on WFT stream
* Fixed https://github.com/temporalio/sdk-core/issues/333
* Fixed https://github.com/temporalio/sdk-core/issues/318
* Resolves https://github.com/temporalio/sdk-core/issues/325
* Resolves https://github.com/temporalio/sdk-core/issues/335


client/src/raw.rs
```diff
@@ -262,6 +262,15 @@ where
...
     );
+    proxy!(
+        get_workflow_execution_history_reverse,
+        GetWorkflowExecutionHistoryReverseRequest,
+        GetWorkflowExecutionHistoryReverseResponse,
+        |r| {
+            let labels = AttachMetricLabels::namespace(r.get_ref().namespace.clone());
+            r.extensions_mut().insert(labels);
+        }
+    );
```
cretz, line 273: Is there an easy way to make a test or compile fail if one of these is missing from the generated `WorkflowService`?

Sushisource:
@cretz #1 lets do in another issue/pr. #2 I will confirm here. I will see if I can find a simple way to double-check that all methods are exposed, too.

core/src/test_help/mod.rs
```diff
@@ -171,6 +171,10 @@ impl MocksHolder {
     pub fn set_act_poller(&mut self, poller: BoxedActPoller) {
         self.inputs.act_poller = Some(poller);
     }
+    pub fn make_wft_stream_interminable(&mut self) {
```
bergundy, line 174: Could use a comment explaining why this is needed

core/src/worker/workflow/machines/signal_external_state_machine.rs
```diff
@@ -234,7 +234,9 @@ impl WFMachinesAdapter for SignalExternalMachine {
             SignalExternalCommand::Failed(f) => {
                 let reason = match f {
                     SignalExternalWorkflowExecutionFailedCause::Unspecified => "unknown",
-                    SignalExternalWorkflowExecutionFailedCause::ExternalWorkflowExecutionNotFound => "it was not found"
+                    SignalExternalWorkflowExecutionFailedCause::ExternalWorkflowExecutionNotFound
+                    | SignalExternalWorkflowExecutionFailedCause::NamespaceNotFound =>
```
bergundy, line 238: What will happen in older SDKs if they get this variant?

bergundy: I've started creating `UNKOWN` variants for user facing enums to avoid unexpected behavior or exceptions.

core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -318,6 +318,17 @@ impl WorkflowMachines {
         if event.is_final_wf_execution_event() {
             self.have_seen_terminal_event = true;
         }
+        if event.event_type() == EventType::WorkflowExecutionTerminated {
```
bergundy, line 321: Would there be other possible events here?

bergundy (approved):
Without diving too deeply into the code, does the removal of the allow stream thingy now effectively allow issuing more than one concurrent poll?

cretz:
@Sushisource - I opened #346

Sushisource:
> Without diving too deeply into the code, does the removal of the allow stream thingy now effectively allow issuing more than one concurrent poll?

Yeah, I can see them spun up in metrics, but, I that already worked anyway. Even after this change, it's still a stream that's pulled from serially. Whether or not more pollers poll concurrently is controlled by the relative rates of consumption. IE: If the stream is being pulled from faster than it's being filled, the concurrent pollers start working. After this change, it's easier to get into that state though I think (which is good)

core/src/worker/workflow/machines/signal_external_state_machine.rs
Sushisource: It'll be unspecified since that's the default proto value.

core/src/worker/workflow/machines/workflow_machines.rs
Sushisource: No - the other terminal events are already handled via the normal mechanisms - since completing the workflow any other way is like issuing a command and has the normal determinism constraints.

Sushisource: Although there is timed out which isn't like that. Thanks for mentioning

PR #347, opened 2022-06-14
Move worker build id to worker options, allow client id override there

Sushisource:
## What was changed
Moved (and renamed) worker build id to be on worker options, fairly obvious there.
Add ability to override client id from worker options too.

## Why?
Support https://github.com/temporalio/sdk-typescript/issues/697

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -85,9 +85,6 @@ pub struct ClientOptions {
     #[builder(default)]
     pub identity: String,
 
-    /// A string that should be unique to this process' binary
-    pub worker_binary_id: String,
-
     /// If specified, use TLS as configured by the [TlsConfig] struct. If this is set core will
     /// attempt to use TLS when connecting to the Temporal server. Lang SDK is expected to pass any
     /// certs or keys as bytes, loading them from disk itself if needed.
@@ -456,6 +453,8 @@ pub struct Client {
     inner: ConfiguredClient<WorkflowServiceClientWithMetrics>,
     /// The namespace this client interacts with
     namespace: String,
+    /// If set, attach as the worker build id to relevant calls
+    bound_worker_build_id: Option<String>,
 }
 
 impl Client {
@@ -467,6 +466,7 @@ impl Client {
         Client {
             inner: client,
             namespace,
+            bound_worker_build_id: None,
         }
     }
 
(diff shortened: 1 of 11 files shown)


cretz (approved):

client/src/lib.rs
```diff
@@ -494,6 +494,17 @@ impl Client {
...
+    /// Return the options this client was initialized with mutably
+    pub fn options_mut(&mut self) -> &mut ClientOptions {
+        &mut self.inner.options
+    }
+
+    /// Set a worker build id to be attached to relevant requests. Unlikely to be useful outside
+    /// of core.
+    pub fn set_worker_build_id(&mut self, id: String) {
+        self.bound_worker_build_id = Some(id)
+    }
```
bergundy, line 507: Do you not plan on sharing this Client between different workers?

core/src/lib.rs
```diff
@@ -68,11 +68,14 @@ where
...
     let as_enum = client.into();
-    // TODO: Assert namespaces match
     let client = match as_enum {
         AnyClient::HighLevel(ac) => ac,
         AnyClient::LowLevel(ll) => {
-            let client = Client::new(*ll, worker_config.namespace.clone());
+            let mut client = Client::new(*ll, worker_config.namespace.clone());
+            client.set_worker_build_id(worker_config.worker_build_id.clone());
+            if let Some(ref id_override) = worker_config.client_identity_override {
+                client.options_mut().identity = id_override.clone();
```
bergundy, line 77: Are the options here a clone of the original options that created the client?

bergundy (approved):

client/src/lib.rs
Sushisource: All the important stuff is shared. Cloning them cones the underlying channel.

core/src/lib.rs
Sushisource: Yes

core-api/src/worker.rs
```diff
@@ -10,6 +10,14 @@ pub struct WorkerConfig {
     /// What task queue will this worker poll from? This task queue name will be used for both
     /// workflow and activity polling.
     pub task_queue: String,
+    /// A string that should be unique to the set of code this worker uses. IE: All the workflow,
+    /// activity, interceptor, and data converter code.
+    pub worker_build_id: String,
```
bergundy, line 15: We should have just called this `build_id`, I overlooked this in the review

Issue #349, opened 2022-06-17
[Bug] Core thread panics on SIGINT


lorensr:
In TS SDK, when I run a worker and send SIGINT (`^C` below, after 'RUNNING'), I get this log:

```
$ node packages/test/lib/run-a-worker.js
...
2022-06-17T16:03:14.131Z [INFO] webpack 5.72.0 compiled with 12 warnings in 467 ms
2022-06-17T16:03:14.133Z [INFO] Workflow bundle created { size: '0.39MB' }
2022-06-17T16:03:14.308Z [INFO] Worker state changed { state: 'RUNNING' }
^C2022-06-17T16:03:16.487Z [INFO] Worker state changed { state: 'STOPPING' }
2022-06-17T16:03:16.489Z [INFO] Worker state changed { state: 'DRAINING' }
  2022-06-17T16:03:16.491213Z  WARN temporal_sdk_core::worker::workflow::workflow_stream: WFT poller died, shutting down

2022-06-17T16:03:16.492Z [INFO] Worker state changed { state: 'DRAINED' }
2022-06-17T16:03:16.493Z [INFO] Worker state changed { state: 'STOPPED' }
thread 'core' panicked at 'shutdown should exit cleanly: JoinError::Cancelled', sdk-core/core/src/worker/activities/activity_heartbeat_manager.rs:137:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Worker gracefully shutdown
```

PR #350, opened 2022-06-17
Don't panic on legit cancel errors inside heartbeat tasks on shutdown

Sushisource:
## What was changed
Fix possible panic during shutdown handling

## Why?
Panic bad

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/349
2. How was this tested:

3. Any docs updates needed?


bergundy (approved):
Do we want to only log or panic on any other error?

Sushisource:
> Do we want to only log or panic on any other error?

Honestly, no real reason to panic here. We're shutting down anyway, no action to be taken.

PR #353, opened 2022-07-07
Make sure workers do not propagate retryable errors as fatal

Sushisource:
## What was changed
Workers should not bubble up retryable errors as fatal, rather they should prefer to just drop the task and let it expire naturally.

## Why?
Robustness

## Checklist

1. Closes 

2. How was this tested:
New UT. Can't test specific scenario user reported b/c of Tonic internals, but it's not exactly complex.

3. Any docs updates needed?


Sushisource:
Will take a different approach - makes more sense to just drop these rather than retrying forever.

bergundy (approved):

PR #357, opened 2022-07-19
Fix bug where LA resolutions could trigger activations with no associated WFT

Sushisource:
## What was changed
This fixes two related problems:
* It does not make sense to activate lang with a local activity resolution if there is no associated WFT with the run at the moment
* Replying to that activation would then cause things to break, because there is nowhere to send the commands

This is fixed in general terms by just not doing work until we have the next WFT

## Why?
It doesn't make sense to activate lang with LA resolutions if there is a missing WFT because, when we _do_ get a WFT, there might be information in there we need to send in the same activation (ex: signal). If that information changes the commands we should send in response, then we've done something wrong.

## Checklist

1. Closes 

2. How was this tested:
Added unit test

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/workflow_stream.rs
+++ b/core/src/worker/workflow/workflow_stream.rs
@@ -45,7 +45,7 @@ pub(crate) struct WFStream {
     metrics: MetricsContext,
 }
 /// All possible inputs to the [WFStream]
-#[derive(derive_more::From)]
+#[derive(derive_more::From, Debug)]
 enum WFStreamInput {
     NewWft(PermittedWFT),
     Local(LocalInput),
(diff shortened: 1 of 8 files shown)


core/src/worker/workflow/workflow_stream.rs
```diff
@@ -167,7 +168,8 @@ impl WFStream {
         };
         all_inputs
             .map(move |action| {
-                let span = span!(Level::DEBUG, "new_stream_input");
+                // TODO: readable version of action
+                let span = span!(Level::DEBUG, "new_stream_input", action=?action);
```
Sushisource, line 172: I want to do this before merge, but it won't affect anything interesting.

bergundy (approved):
LGTM, didn't follow the test but the fix makes total sense.

cretz (approved):

PR #360, opened 2022-07-26
implement SignalWithStartWorkflow for client + tests

dt665m:
## What was changed
```WorkflowClientTrait``` has a new ```signal_with_start_workflow_execution``` method.  ```RetryClient``` and ```Client``` implements this new method.  Added ```ActivityResultPendingError``` to support asynchronous activity completions.

## Why?
Nice to get closer to parity with other lang sdks.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/359

2. How was this tested:
Integration tests added for this feature in ```tests/integ_tests/workflow_tests/signals.rs```

3. Any docs updates needed?
Not that I am aware of.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -711,7 +711,7 @@ impl WorkflowClientTrait for Client {
                 }),
                 task_queue: Some(TaskQueue {
                     name: task_queue,
-                    kind: 0,
+                    kind: TaskQueueKind::Unspecified as i32,
                 }),
                 request_id,
                 workflow_task_timeout: options.task_timeout.map(Into::into),
@@ -963,9 +963,7 @@ impl WorkflowClientTrait for Client {
                 }),
                 task_queue: Some(TaskQueue {
                     name: task_queue,
-                    // #FIXME the go sdk gets this from a proto enum.  Not sure where
-                    // this is available in the rust prost generated stuff is at the moment.
-                    kind: 1,
+                    kind: TaskQueueKind::Normal as i32,
                 }),
                 request_id,
                 input,
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -609,6 +609,20 @@ pub trait WorkflowClientTrait {
         payloads: Option<Payloads>,
     ) -> Result<SignalWorkflowExecutionResponse>;
 
+    /// Send signal and start workflow transcationally
+    //#TODO maybe lift the Signal type from sdk::workflow_context::options
+    #[allow(clippy::too_many_arguments)]
+    async fn signal_with_start_workflow_execution(
+        &self,
+        input: Option<Payloads>,
+        task_queue: String,
+        workflow_id: String,
+        workflow_type: String,
+        options: WorkflowOptions,
+        signal_name: String,
+        signal_input: Option<Payloads>,
+    ) -> Result<SignalWithStartWorkflowExecutionResponse>;
+
     /// Request a query of a certain workflow instance
     async fn query_workflow_execution(
         &self,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by Sushisource — APPROVED** [review]
Thanks for the contribution! I'll take a look at how we might best expose the client in activities.

I'll merge this after you change the task queue kind thing, assuming that works.

File: client/src/lib.rs [hunk 1]
```diff
@@ -928,6 +942,44 @@ impl WorkflowClientTrait for Client {
...
+            .signal_with_start_workflow_execution(SignalWithStartWorkflowExecutionRequest {
+                namespace: self.namespace.clone(),
+                workflow_id,
+                workflow_type: Some(WorkflowType {
+                    name: workflow_type,
+                }),
+                task_queue: Some(TaskQueue {
+                    name: task_queue,
+                    // #FIXME the go sdk gets this from a proto enum.  Not sure where
+                    // this is available in the rust prost generated stuff is at the moment.
```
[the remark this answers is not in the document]
dt665m: Great.  Found it.  I was looking at the ```start_workflow``` function as a reference and it was using 0 instead of the enum.  Didn't notice the other functions were though.  Fixed both.

sdk/src/lib.rs
```diff
@@ -488,6 +488,9 @@ impl ActivityHalf {
                         Ok(res) => ActivityExecutionResult::ok(res),
                         Err(err) => match err.downcast::<ActivityCancelledError>() {
                             Ok(ce) => ActivityExecutionResult::cancel_from_details(ce.details),
+                            Err(other_err) if other_err.is::<ActivityResultPendingError>() => {
```
dt665m, line 491: btw, I also snuck this in for activities to begin async completion.

Sushisource: I saw :) -- I'm thinking I might actually want this to be a success value, like continue-as-new is for workflows. Since this really isn't an "error" state, I think it's semantically better to have it be a non-error value as long as the ergonomics remain good.

WDYT? The workflow result seems pretty reasonable right now, you just call `into` before returning in the normal path. I think that can work here too.

dt665m: I'm assuming you're thinking of making an explicit enum struct like ```WfExitValue<T>``` for activities right?  I know it uses derive_more for the happy path convenient ```into()```.  Nice trick ignoring all the other enum variants. 

Just so we're clear, you are suggesting we introduce the ```anyhow::Result<ActExitValue<T>>``` as the ActivityFunc return type, allowing the SDK to handle this better and also make the implementer of the ActivityFunc return this new enum explicitly instead of using a magical ```Err(ActivityCancelledError)``` ?  If so, this is the right way to go, but also causes a big change.  I implemented it this way because the go sdk does this (which is also odd), it's quick and dirty, and also doesn't affect/break the API.

I'm for it though.  This should suffice: 
```
#[derive(Debug, derive_more::From)]
pub enum ActExitValue<T: Debug> {
    /// Completion requires an asynchronous callback
    WillCompleteAsync,
    /// Finish with a result
    Normal(T),
}
```
and the rest is just cleanup in the tests.  Give me the green light and I can take a crack at it.

Sushisource: Yep, that's exactly what I mean. I would love it if you made that change. I think it's potentially quite confusing to new users to see this as an "error" value.

The blast radius is nonzero though indeed.

One thing that would be worth playing with, which I haven't been able to try yet, is if we could possibly avoid needing the explicit `into` call as well, by doing something like this (also reduces the blast radius, which is great):

Change:
```rust
impl<A, Rf, R, F> IntoActivityFunc<A, Rf> for F
where
    F: (Fn(ActContext, A) -> Rf) + Sync + Send + 'static,
    A: FromJsonPayloadExt + Send,
    Rf: Future<Output = Result<R, anyhow::Error>> + Send + 'static,
    R: AsJsonPayloadExt,
```

to

```rust
impl<A, Rf, R, F> IntoActivityFunc<A, Rf> for F
where
    F: (Fn(ActContext, A) -> Rf) + Sync + Send + 'static,
    A: FromJsonPayloadExt + Send,
    Rf: Future<Output = Result<RR, anyhow::Error>> + Send + 'static,
    R: AsJsonPayloadExt,
    RR: Into<ActExitValue<R>>
```

If that works, that'd be real nice

dt665m: Implemented this in the latest commit and removed the pending error.  I had to introduce the type parameter into ```IntoActivityFunc``` trait though, but otherwise your new trait gymnastics works well.  Take a look and let me know what you think.  Everything else remains unchanged because of this impl.

Sushisource (approved):
Awesome! Thanks so much for this!

Issue #364, opened 2022-07-31
[Feature Request] More complete Rust SDK Workflow Options


dt665m:
### Is your feature request related to a problem? Please describe.

Many options for starting workflows are missing.  

### Describe the solution you'd like

More of the options made available through the WorkflowOption struct.

dt665m:
closing for now after #365

PR #365, opened 2022-07-31
Add more missing workflow options and add request_id as parameter for some calls

dt665m:
Add more missing workflow options and add request_id as parameter for some calls

### Rust SDK
- Request_id is now a user level optional parameter with a UUID_v4 default if none is given.
- Workflow Options are now more complete (extra timeout options, Workflow ID Reuse Policy, etc)
- Implemented Async Activity Completion
- Added `describe_namespace` for client
- Added `list_open_workflow_executions` for client
- Added `list_closed_workflow_executions` for client
- Added `list_workflow_executions` for client

## Why?

Some workflow options are missing.  Namely, Workflow Rerun Policy being very crucial to some of the way we are trying to use temporal.  Request_id is also quite useful to deduplicate possible failure scenarios for client calls. 

1. Closes  https://github.com/temporalio/sdk-core/issues/364

2. How was this tested:

- ran tests and integration tests
- Added tests for Async Activity Completion
- Added tests for describe_namespace
- Added tests for listing workflows

3. Any docs updates needed?

None


sdk/src/workflow_context.rs
```diff
@@ -370,6 +367,29 @@ impl WfContext {
...
 
+/// Helper Wrapper that can drain the channel into a Vec<SignalData> in a blocking way.  Useful
+/// for making sure channels are empty before ContinueAsNew-ing a workflow
+pub struct DrainableSignalStream(UnboundedReceiverStream<SignalData>);
+
+impl DrainableSignalStream {
+    pub fn drain_all(mut self) -> Vec<SignalData> {
+        let mut cx = std::task::Context::from_waker(futures::task::noop_waker_ref());
+        let mut signals = vec![];
+        while let Poll::Ready(Some(s)) = self.0.poll_next_unpin(&mut cx) {
```
Sushisource, line 378: I think you can avoid the no-op waker if you turn the stream back into a receiver https://docs.rs/tokio-stream/latest/tokio_stream/wrappers/struct.UnboundedReceiverStream.html#method.into_inner and then just call `try_recv` in a loop

tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -793,3 +795,15 @@ async fn one_activity_abandon_cancelled_after_complete() {
...
+#[tokio::test]
+async fn it_can_complete_async() {
+    let mut starter = CoreWfStarter::new("bah");
+    let mut worker = starter.worker().await;
+
+    async fn activity(_: ActContext, _: String) -> anyhow::Result<ActExitValue<()>> {
+        Ok(ActExitValue::WillCompleteAsync)
+    }
+
+    worker.register_activity("async_activity", activity);
```
Sushisource, line 808: Test should get fleshed out with a workflow that runs this, and then complete the activity with a client.

client/src/lib.rs
```diff
@@ -684,9 +686,11 @@ pub trait WorkflowClientTrait {
         task_queue: String,
         workflow_id: String,
         workflow_type: String,
+        request_id: Option<String>,
```
Sushisource, line 689: This argument list is starting to get pretty outta control.

I'm thinking it makes sense to accept arg structs here and make builders for them or something like that. Not in this PR I think, but, looks like that's starting to become a good idea.

dt665m: agreed, shouldn't be too bad with derive_builder

sdk/src/workflow_context.rs
dt665m: fixed

tests/integ_tests/workflow_tests/activities.rs
dt665m: done

client/src/lib.rs
```diff
@@ -714,9 +776,22 @@ pub trait WorkflowClientTrait {
 /// Optional fields supplied at the start of workflow execution
 #[derive(Debug, Clone, Default)]
 pub struct WorkflowOptions {
+    /// Set the policy for reusing the workflow id
+    pub id_reuse_policy: WorkflowIdReusePolicy,
+
+    /// Optionally set the execution timeout for the workflow
+    /// https://docs.temporal.io/workflows/#workflow-execution-timeout
```
Sushisource, line 783: CI is complaining about auto-link. You can use `cargo doc --workspace --all-features --no-deps` to see doc issues

client/src/lib.rs
```diff
@@ -714,9 +776,22 @@ pub trait WorkflowClientTrait {
...
+    pub execution_timeout: Option<Duration>,
+
+    /// Optionally indicates the default run timeout for workflow tasks
```
Sushisource, line 786:
```suggestion
    /// Optionally indicates the default timeout for a workflow run
```

Sushisource (approved):
Other CI failure is just linting. `cargo lint` will run the same stuff CI does.

Thanks again!

client/src/lib.rs
dt665m: fixed

client/src/lib.rs
dt665m: fixed

dt665m:
> 

I've fixed some of the project lint warnings, but it's also complaining about proto generated code not implementing `Eq` but implementing `PartialEq`.

dt665m:
Added integration tests for the series of client workflow listing calls.   The `list_workflow_executions` function isn't tested as the current integration tests environment doesn't include elasticsearch.

PR #374, opened 2022-08-12
More client refactoring & add versioning-opt-in config flag

Sushisource:
## What was changed
* Shuffled a bunch of stuff around to make the worker client easier to use / enable this feature
* Added bool flag to `WorkerConfig` for opting in to versioning feature

## Why?
Will be needed to test versioning from a real lang SDK.

Relates to https://github.com/temporalio/sdk-features/issues/84

### DIFF
--- a/client/src/raw.rs
+++ b/client/src/raw.rs
@@ -4,7 +4,6 @@
 
 use crate::{
     metrics::{namespace_kv, task_queue_kv},
-    raw::sealed::RawClientLike,
     LONG_POLL_TIMEOUT,
 };
 use futures::{future::BoxFuture, FutureExt};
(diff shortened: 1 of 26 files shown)


Sushisource:
Literally everything is all-internal (or super-low external impact) refactoring besides the last commit to add the flag: https://github.com/temporalio/sdk-core/pull/374/commits/4eb253443cf0ec4446e63f14c99ba328fa673deb

client/src/raw.rs
```diff
@@ -52,110 +50,110 @@ pub(super) mod sealed {
...
-        fn client(&mut self) -> &mut WorkflowServiceClient<Self::SvcType> {
-            self.inner.client()
-        }
+impl RawClientLike for Client {
+    type SvcType = InterceptedMetricsSvc;
 
-        fn operator_client(&mut self) -> &mut OperatorServiceClient<Self::SvcType> {
-            self.inner.operator_client()
-        }
+    fn client(&mut self) -> &mut WorkflowServiceClient<Self::SvcType> {
```
bergundy, line 147: Might as well call this `workflow_client`, we ended up changing `service` to `workflow_service` in TS and Python when adding the other services.

bergundy (commented):
Wondering if Core should protect when asked to use buildId based versioning and server doesn't have that capability

cretz (approved):

Sushisource:
> Wondering if Core should protect when asked to use buildId based versioning and server doesn't have that capability

Makes sense, added issue here https://github.com/temporalio/temporal/issues/3230

client/src/raw.rs
Sushisource: Done

Issue #331, opened 2022-05-25
[Bug] `CancelUnstartedChildWorkflowExecution` command might fail if Workflow tries to cancel child workflow before getting notified of start event


bergundy:
E.g. you might try to cancel the child workflow in a signal handler in which case the child workflow may or may not have been started as far as core is concerned.

This hasn't been confirmed yet but it looks like it could very well happen.

Recommending that instead Core supports cancelling the child workflow by `StartChildWorkflowExecution.seq` and not have Lang track the child workflow state (which could be outdated).

PR #379, opened 2022-08-18
Fix cancelling of started-but-lang-doesn't-know workflows

Sushisource:
## What was changed
* Created a facility for core to generate new commands internally
* Used that to allow child wf machine to generate a cancel external wf command when cancel is called on it, and the child WF has already started
* Various bits of cleanup as I went

I am reasonably sure this is history compat, but unfortunately that comes at the cost of lang continuing to issue a cancel external command if it does know the child is already started. Otherwise, it's possible command ordering could be slightly different. Core will generate the same cancel external command in that situation now, but that could show up in the command list in a slightly different place than it does when lang does it directly. We would need to test in TS.

## Why?
Fix the associated bug. Also allows us to remove the timer backoff hack, and that could now be done totally internally in core (however, not easy to do and maintain history compat...).

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/331
2. How was this tested:
Added UTs, existing tests


core/src/core_tests/child_workflows.rs
```diff
@@ -97,3 +103,68 @@ async fn cancel_child_workflow() {
...
+    // Make sure that a resolve for the "request cancel external workflow" command does *not* appear
+    // since lang didn't actually issue one. The only job should be resolving the child workflow.
+    assert_matches!(
+        act.jobs.as_slice(),
+        [WorkflowActivationJob {
+            variant: Some(workflow_activation_job::Variant::ResolveChildWorkflowExecution(_)),
+        }]
+    );
+    // Request cancel external is technically fallible, but the only reasons relate to targeting
+    // a not-found workflow, which couldn't happen in this case.
```
bergundy, line 169: Should we test other cases like when the child workflow is already complete? What happens in this case for example if we're processing a WFT while the child completes or we already know that the child completed?

bergundy: I'd test some more scenarios

core/src/core_tests/child_workflows.rs
```diff
@@ -97,3 +103,68 @@ async fn cancel_child_workflow() {
     wfm.process_all_activations().await.unwrap();
     wfm.shutdown().await.unwrap();
 }
+
+#[rstest::rstest]
+#[case::abandon(ChildWorkflowCancellationType::Abandon)]
+#[case::try_cancel(ChildWorkflowCancellationType::TryCancel)]
+#[case::wait_cancel_completed(ChildWorkflowCancellationType::WaitCancellationCompleted)]
```
bergundy, line 110: nit: could also test WaitCancellationRequested

core/src/core_tests/child_workflows.rs
```diff
@@ -97,3 +103,68 @@ async fn cancel_child_workflow() {
...
+#[tokio::test]
+async fn cancel_child_workflow_lang_thinks_not_started_but_is(
```
bergundy, line 112: Test cancellation of not started child too? Is this interesting enough to test?

bergundy: Test cancellation of already completed child, core should ignore this (we do the same for activities AFAIR).

bergundy (commented):
While I definitely agree this is better than what we had I wouldn't merge it before verifying it doesn't break TS.

Sushisource:
I verified that this does not break compat when used from TS using this branch:
https://github.com/temporalio/sdk-typescript/tree/test-child-wf-changes

core/src/core_tests/child_workflows.rs
Sushisource: Added

bergundy (approved):

Sushisource:
Will merge after https://github.com/temporalio/sdk-core/pull/396

PR #381, opened 2022-08-20
Correct API definition link

shanginn:
## What was changed

Old link is outdated, updated it to the correct (I think) one

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Sushisource (approved):
Thanks!

PR #389, opened 2022-09-12
Ephemeral server lazy-downloader and runner

### LINKED ISSUE #388: [Feature Request] Ephemeral server support

Every SDK keeps writing their own download-and-run code for the test server and sometimes Temporalite.

Would like:

* `EphemeralServer` trait with:
  * Frontend host:port getter
  * Namespace getter
  * Ability to shutdown
  * Getter for whether it supports the gRPC test service
* Function that returns a `EphemeralServer` implementation for Temporalite, e.g. `start_embedded` that:
  * Accepts a path to the Temporalite binary but if it doesn't exist:
    * Downloads and extracts single Temporalite binary (renamed to add version suffix) to overridable temporary location if it doesn't already exist using a version that can be supplied by lang. Note, this download can untar/unzip in memory and just extract binary out, no temp space needed.
  * Accepts desired port but if not present, defaults to a locally determined free port
  * Accepts namespace
  * Accepts generic Temporalite CLI arguments
  * Accepts stdin/stdout visibility options
  * Starts Temporalite and returns once confirmed up
* Function that returns a `EphemeralServer` implementation for test server, e.g. `start_test_server` that:
  * Accepts most of the things the above Temporalite one does (but no namespace for example)
  * Works the same way as the above Temporalite one in almost every way wrt lazy downloading, process running, etc

We should not try to be clever here and return a Core client from this trait or anything. This needs to be the most basic download-and-start abstraction so lang can choose what to do with it, when to shut it down, how to make gRPC connections to it, etc.

cretz:
## What was changed

Added support to download and run Temporalite or Test Server to `temporal-sdk-core`. Care was taken to work as well as possible w/ cross-process caching and zip/tar.gz extraction while downloading.

## Why?

All SDKs need to be able to start an ephemeral server directly.

## Checklist

1. Closes #388


core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,454 @@
...
+    _child: tokio::process::Child,
+}
+
+impl EphemeralServer {
+    async fn start(config: EphemeralServerConfig) -> anyhow::Result<EphemeralServer> {
+        // Start process
+        // TODO(cretz): Offer stdio suppression?
+        let child = tokio::process::Command::new(config.exe_path)
+            .args(config.args)
+            .kill_on_drop(true)
```
cretz, line 152: I'm now thinking we need an explicit shutdown that waits for the process to die. People from lang side expect a completed shutdown to mean the port is free.

cretz: Fixed

core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+    pub port: Option<u16>,
+    /// Sqlite DB filename if persisting or non-persistent if none.
+    #[builder(default)]
+    pub db_filename: Option<String>,
+    /// Whether to enable the UI.
+    #[builder(default)]
+    pub ui: bool,
+    /// Log format and level
+    #[builder(default = "(\"pretty\".to_owned(), \"warn\".to_owned())")]
+    pub log: (String, String),
```
bergundy, line 42: nit: Could have a separate struct for log instead of this tuple.

core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+    pub target: String,
+    /// Whether the target implements the gRPC TestService
+    pub has_test_service: bool,
+    child: tokio::process::Child,
+}
+
+impl EphemeralServer {
+    async fn start(config: EphemeralServerConfig) -> anyhow::Result<EphemeralServer> {
+        // Start process
+        // TODO(cretz): Offer stdio suppression?
```
bergundy, line 149: Yes please, we support this in TS today

bergundy: Is the default here to inherit the current proc's stdio?

core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+        let client_options = ClientOptionsBuilder::default()
+            .identity("online_checker".to_owned())
+            .target_url(Url::parse(&target_url)?)
+            .client_name("online-checker".to_owned())
+            .client_version("0.1.0".to_owned())
+            .build()?;
+        for _ in 0..50 {
+            sleep(Duration::from_millis(100)).await;
+            if client_options
+                .connect_no_namespace(None, None)
```
bergundy, line 176: Is this enough? I don't remember what this does internally, does it issue a `getSystemInfo` call?

core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+                }
+            }
+        }
+    }
+}
+
+fn get_free_port() -> u16 {
+    // Can just ask OS to give us a port then close socket. OS's don't give that
+    // port back to anyone else anytime soon.
+    std::net::TcpListener::bind("127.0.0.1:0")
```
bergundy, line 319: You should use the IP address provided in the config

tests/integ_tests/ephemeral_server_tests.rs
```diff
@@ -0,0 +1,138 @@
...
+    assert_ephemeral_server(&server).await;
+}
+
+#[tokio::test]
+async fn temporalite_shutdown_port_reuse() {
+    // Start, test shutdown, do again immediately on same port to ensure we can
+    // reuse after shutdown
+    let config = TemporaliteConfigBuilder::default()
+        .exe(default_cached_download())
+        .port(Some(10123))
```
bergundy, line 37: nit: In rare cases this port will already be taken, allocate a free port instead

bergundy (approved):
No major comments, overall LGTM

core/src/ephemeral_server/mod.rs
cretz: There wasn't much value to expose a whole new module-level type for this.

core/src/ephemeral_server/mod.rs
cretz: Yes it does issue that call. It is enough.

core/src/ephemeral_server/mod.rs
cretz: This is just for obtaining a free port from the current system. The user-defined IP is not needed from my understanding of how OS's allocate ports. I guess I could use `0.0.0.0` but I fear locked down Linux capabilities of doing that.

tests/integ_tests/ephemeral_server_tests.rs
cretz: I didn't really want to expose that API or copy it over. Not too concerned here.

bergundy: It should rarely matter but eventually we'll hit this, IMHO the effort of allocating a random port is low enough to make it worth addressing this.

Leaving it up to you.

core/src/ephemeral_server/mod.rs
bergundy: This is good enough for now.

core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+        .unwrap()
+        .local_addr()
+        .unwrap()
+        .port()
+}
+
+// Returns false if we successfully waited for another download to complete, or
+// true if the destination is known to exist. Should call again if false is
+// returned.
```
Sushisource, line 328: Can/should be docstring

core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+) -> anyhow::Result<bool> {
+    // If it already exists, do not extract
+    if dest.exists() {
+        return Ok(true);
+    }
+
+    // We only want to download if we're not already downloading. To avoid some
+    // kind of global lock, we'll just create the file eagerly w/ a temp
+    // filename and delete it on failure or move it on success. If the temp file
+    // already exists, we'll wait a bit and re-run this.
```
Sushisource, line 343: That is a global lock :wink:

core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+    let temp_dest_str = format!("{}{}", dest.to_str().unwrap(), ".downloading");
+    let temp_dest = Path::new(&temp_dest_str);
+    // This match only gets Ok if the file was downloaded and extracted to the
+    // temporary path
+    match new_executable_file(temp_dest) {
+        Err(err) if err.kind() == std::io::ErrorKind::AlreadyExists => {
+            // Since it already exists, we'll try once a second for 20 seconds
+            // to wait for it to be done, then recurse
```
Sushisource, line 351: No recursion actually happens here, so the comment should get adjusted

20 seconds seems maybe a bit short of a timeout as well, and the error might not make much sense when it bubbles all the way up to the user, probably worth adding more context.

core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+
+    // We only want to download if we're not already downloading. To avoid some
+    // kind of global lock, we'll just create the file eagerly w/ a temp
+    // filename and delete it on failure or move it on success. If the temp file
+    // already exists, we'll wait a bit and re-run this.
+    let temp_dest_str = format!("{}{}", dest.to_str().unwrap(), ".downloading");
+    let temp_dest = Path::new(&temp_dest_str);
+    // This match only gets Ok if the file was downloaded and extracted to the
+    // temporary path
+    match new_executable_file(temp_dest) {
```
Sushisource, line 348: I would actually inline this function since the behavior here depends heavily on `create_new` being set true and it's more obvious when that's right here.

You can use the `cfg` guards on different `let` statements binding the same variable name

core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+    // tarball or zip instead of loading into memory for Cursor/Seek.
+    let resp = client.get(uri).send().await?;
+    // We have to map the error type to an io error
+    let stream = resp
+        .bytes_stream()
+        .map(|item| item.map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err)));
+
+    // Since our tar/zip impls use sync IO, we have to create a bridge and run
+    // in a blocking closure. We have to extend the lifetime of some references
+    // we are given for this.
```
Sushisource, line 422:
```suggestion
    // in a blocking closure.
```
Not totally accurate (you don't really extend any lifetimes here, that isn't really possible in safe code - it's just a matter of shifting ownership), but I think fine to leave off this bit of comment anyway.

core/src/ephemeral_server/mod.rs
```diff
@@ -0,0 +1,461 @@
+use anyhow::anyhow;
+use flate2::read::GzDecoder;
+use futures::StreamExt;
+use serde::Deserialize;
+#[cfg(target_family = "unix")]
+use std::os::unix::fs::OpenOptionsExt;
```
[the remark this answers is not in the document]
cretz: Ah, I was just counting on `cargo fmt` on nightly to fix this all up. Ok, will move.

core/src/ephemeral_server/mod.rs
cretz: Heh, by definition but not by name. I just meant no static mutex.

core/src/ephemeral_server/mod.rs
cretz: > No recursion actually happens here, so the comment should get adjusted

Yeah, can change (it was recursion at one time).

> 20 seconds seems maybe a bit short of a timeout as well

The downloads are <= 22MB for Temporalite and <= 17MB for test server. I struggled coming up with the right number that wouldn't have a user sitting and waiting _too_ long if some crash caused the `.downloading` file to stay around.

> probably worth adding more context.

I do give the file path. I guess I can add that the temp file should be removed before retrying?

core/src/ephemeral_server/mod.rs
cretz: The reason I don't inline this is because it has different behavior per OS (I set it as executable for unix, doesn't apply for windows)

core/src/ephemeral_server/mod.rs
cretz: Can leave it off. Yeah, I didn't mean whatever is technically meant by "extending lifetimes", I just meant I had to do things to make these live longer.

tests/integ_tests/ephemeral_server_tests.rs
cretz: The goal of the test is to confirm I can create a server on the same port right after I shutdown the previous one. Basically it confirms that after `shutdown` is returned, the port is still available. So I need the same port for both, but technically I could use a "free port" helper if I had to.

tests/main.rs
```diff
@@ -11,6 +11,7 @@ mod integ_tests {
         };
     }
     mod client_tests;
+    mod ephemeral_server_tests;
```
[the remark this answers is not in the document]
cretz: You'd only download binaries the first time or new version. I don't think it's that unreasonable considering we're doing that on every other SDK using this when _they_ run integration tests locally.

Also, you might want to change up your integration tests to use this Temporalite downloader. It's more self-contained than having to have the environment prep a server (and often quite a bit faster too).

core/src/ephemeral_server/mod.rs
Sushisource: Yeah, that context is good, as well as just mentioning what's actually being downloaded. I'm not sure that'd show up obviously.

tests/main.rs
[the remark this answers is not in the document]
Sushisource: Sure, that makes sense. I'll do that next time I get a chance.

core/src/ephemeral_server/mod.rs
cretz: We aren't given a port for Java test server. I will pass in the user-supplied one for Temporalite and use 0.0.0.0 for Java test server.

bergundy: > We aren't given a port for Java test server

If we support user defined port (arguably not needed) for temporalite, we should for Java test server as well.

core/src/lib.rs
```diff
@@ -28,6 +29,10 @@ mod test_help;
 
 pub(crate) use temporal_sdk_core_api::errors;
 
+pub use ephemeral_server::{
+    EphemeralExe, EphemeralExeVersion, EphemeralServer, TemporaliteConfig,
+    TemporaliteConfigBuilder, TestServerConfig, TestServerConfigBuilder,
+};
```
Sushisource, line 35: Just noticed this - let's just pub the module rather than each thing individually. I think this one is nice to have behind a namespace.

Also ideally `EphemeralExe` and `EphemeralExeVersion` don't need to be pub, since they're largely an internal detail.

core/src/ephemeral_server/mod.rs
cretz: Sorry, I meant we aren't given an _ip_

core/src/lib.rs
cretz: > let's just pub the module rather than each thing individually

Ok

> Also ideally EphemeralExe and EphemeralExeVersion don't need to be pub, since they're largely an internal detail.

They are used for configuration. They are needed for a user to choose existing path, download default, or download specific version (I already use them from Python).

core/src/ephemeral_server/mod.rs
cretz: Added port for free port obtaining

core/src/ephemeral_server/mod.rs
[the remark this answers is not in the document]
cretz: Moved

core/src/ephemeral_server/mod.rs
cretz: Done

core/src/ephemeral_server/mod.rs
cretz: Done

core/src/ephemeral_server/mod.rs
cretz: Done

core/src/lib.rs
cretz: Made the module public

Sushisource: Ah, sure, makes sense

Sushisource (approved):

core/src/ephemeral_server/mod.rs
cretz: Done

PR #392, opened 2022-09-15
Fix for ephemeral test server zombie

### LINKED ISSUE #391: [Bug] Ephemeral test server leaves zombie process on shutdown sometimes
### Describe the bug

For some as-of-yet unknown reason, only on the test server, and only for parent process of Python 3.7 (and a few others), the spawned test server is defunct when using https://docs.rs/tokio/latest/tokio/process/struct.Child.html#method.kill. So that never returns.

Unix keeps the zombie around until its exit code is asked for which I guess Tokio is not properly doing. After much trial and error I've noticed that if we call https://docs.rs/tokio/latest/tokio/process/struct.Child.html#method.start_kill and manually call https://docs.rs/nix/latest/nix/sys/wait/fn.waitpid.html it is cleaned up properly. Of course this blocks which is less than ideal, but it is still better than the defunct process.

The Tokio process reaper seems to do child process waiting async via a reaper and something is causing it not to wait for this ID after sending SIGKILL. This does not happen for Temporalite.

I am also investigating other options...

cretz:
## What was changed

For the Java-based-Graal-built ephemeral test server, we are manually doing https://linux.die.net/man/2/waitpid because for whatever reason Tokio is not doing it properly leaving a zombie process around. Unfortunately I cannot replicate in Rust integration tests.

## Why?

See #391. Basically `child.kill().await` leaves zombies around in certain cases.

## Checklist

1. Closes #391

### DIFF
--- a/core/src/ephemeral_server/mod.rs
+++ b/core/src/ephemeral_server/mod.rs
@@ -89,6 +89,10 @@ impl TemporaliteConfig {
             port,
             args,
             has_test_service: false,
+            // Zombie prevention not needed on Temporalite, it shuts down
+            // properly
+            #[cfg(target_family = "unix")]
+            prevent_zombie_on_unix_shutdown: false,
         })
         .await
     }
@@ -126,6 +130,8 @@ impl TestServerConfig {
             port,
             args,
             has_test_service: true,
+            #[cfg(target_family = "unix")]
+            prevent_zombie_on_unix_shutdown: true,
         })
         .await
     }
@@ -136,6 +142,8 @@ struct EphemeralServerConfig {
     port: u16,
     args: Vec<String>,
     has_test_service: bool,
+    #[cfg(target_family = "unix")]
+    prevent_zombie_on_unix_shutdown: bool,
 }
 
 /// Server that will be stopped when dropped.
@@ -145,6 +153,8 @@ pub struct EphemeralServer {
     /// Whether the target implements the gRPC TestService
     pub has_test_service: bool,
     child: tokio::process::Child,
+    #[cfg(target_family = "unix")]
+    prevent_zombie_on_unix_shutdown: bool,
 }
 
 impl EphemeralServer {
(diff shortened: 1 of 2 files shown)


core/src/ephemeral_server/mod.rs
```diff
@@ -187,9 +199,45 @@ impl EphemeralServer {
...
+    /// Shutdown the server (i.e. kill the child process). This does not attempt
+    /// a kill if the child process appears completed, but such a check is not
+    /// atomic so a kill could still fail as completed if completed just before
+    /// kill.
     pub async fn shutdown(&mut self) -> anyhow::Result<()> {
-        Ok(self.child.kill().await?)
+        #[cfg(target_family = "unix")]
+        if self.prevent_zombie_on_unix_shutdown {
+            return self.shutdown_prevent_zombie().await;
+        }
```
Sushisource, line 210: Is the flag really necessary? Seems we could just always do this since waiting on the pid is mostly harmless and avoid the clutter of piping the flag in a bunch of places. Makes it easier to remove later too if that happens.

cretz: I am trying to localize the problem as much as possible. Temporalite doesn't need it, Windows doesn't need it. If you want me to use it for non-Windows-Temporalite though, I can.

cretz: I am going to try to do this with Temporalite too and switch to SIGINT to have a more graceful shutdown

cretz: Fixed

core/src/ephemeral_server/mod.rs
```diff
@@ -153,9 +153,7 @@ impl EphemeralServer {
         // TODO(cretz): Offer stdio suppression?
         let child = tokio::process::Command::new(config.exe_path)
             .args(config.args)
-            // We kill this on drop in case of errors or other issues. We still
-            // encourage users to manually shutdown.
-            .kill_on_drop(true)
+            .stdin(std::process::Stdio::null())
```
cretz, line 156: This was also needed by @bergundy because on M1 environments the non-blocking stdin was preventing startup for whatever reason

Sushisource (approved):

core/src/ephemeral_server/mod.rs
bergundy: nit: I'd consider writing this comment in the code

bergundy (approved):
lgtm

PR #396, opened 2022-09-16
Protect worker from more network errors

Sushisource:
## What was changed
No completion errors will ever crash the worker
Long poll permits "fatal" errors for up to a minute

## Why?
Sometimes infrastructure pieces return stupid error codes, and we shouldn't crash in that case, even if things are telling us the error is fatal. Those things are stupid and wrong, but, so be it.

## Checklist

1. Closes 

2. How was this tested:
Existing UTs

3. Any docs updates needed?

### DIFF
--- a/client/src/retry.rs
+++ b/client/src/retry.rs
@@ -2,7 +2,7 @@ use crate::{
     ClientOptions, ListClosedFilters, ListOpenFilters, Namespace, Result, RetryConfig,
     StartTimeFilter, WorkflowClientTrait, WorkflowOptions,
 };
-use backoff::{backoff::Backoff, ExponentialBackoff};
+use backoff::{backoff::Backoff, exponential::ExponentialBackoff, Clock, SystemClock};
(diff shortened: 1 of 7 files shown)


client/src/retry.rs
```diff
@@ -178,6 +187,10 @@ impl ErrorHandler<tonic::Status> for TonicErrorHandler {
                     }
                 }
             }
+        } else if is_long_poll && self.backoff.get_elapsed_time() <= LONG_POLL_FATAL_GRACE {
+            // We permit "fatal" errors while long polling for a while, because some proxies return
+            // stupid error codes while getting ready, among other weird infra issues
+            RetryPolicy::WaitRetry(self.backoff.max_interval)
```
bergundy, line 193: Hmm... should we wait for "that long"? How long will this be in practice?

bergundy (approved):

client/src/retry.rs
Sushisource: 5s

PR #397, opened 2022-09-16
Don't dispatch eager activity if task queue is not the "current"

bergundy:
Checklist:
- [x] Add a test

### DIFF
--- a/core/src/worker/workflow/mod.rs
+++ b/core/src/worker/workflow/mod.rs
@@ -84,6 +84,7 @@ type BoxedActivationStream = BoxStream<'static, Result<ActivationOrAuto, PollWfE
 
 /// Centralizes all state related to workflows and workflow tasks
 pub(crate) struct Workflows {
+    task_queue: String,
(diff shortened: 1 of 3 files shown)


core/src/worker/workflow/mod.rs
```diff
@@ -460,8 +464,16 @@ impl Workflows {
...
+                    // ever turn *off* eager execution if a slot is not available or the activity
+                    // is scheduled on a different task queue.
+                    if attrs.request_eager_execution
+                        && attrs
+                            .task_queue
+                            .as_ref()
+                            .map(|q| q.name == self.task_queue)
+                            .unwrap_or_default()
+                    {
+                        // if attrs.request_eager_execution {
```
Sushisource, line 476: Leftover comment, and this logic is wrong. You aren't setting `request_eager_execution` to false if the task queue is mismatched

bergundy: Yeah, I just saw this, fixing

cretz (approved):

PR #400, opened 2022-09-21
Implement multiple histories support for replaying

Sushisource:
## What was changed
Implemented ability to provide multiple histories when constructing a replay worker

## Why?
More efficient and easier to use than creating many workers lang side

## Checklist

1. Closes 

2. How was this tested:
Existing / new tests

3. Any docs updates needed?

### DIFF
--- a/core-api/src/worker.rs
+++ b/core-api/src/worker.rs
@@ -95,6 +95,12 @@ pub struct WorkerConfig {
     /// todo: link to feature docs
     #[builder(default = "false")]
     pub use_worker_versioning: bool,
+
+    /// If set true, shutdown will not finish until all pending evictions have been issued and
+    /// replied to. If not set (the default) shutdown will be considered complete when the only
+    /// remaining work is pending evictions.
+    #[builder(default = "false")]
+    pub issue_evicts_before_shutdown: bool,
 }
 
 impl WorkerConfig {
--- a/core-api/src/worker.rs
+++ b/core-api/src/worker.rs
@@ -96,11 +96,15 @@ pub struct WorkerConfig {
     #[builder(default = "false")]
     pub use_worker_versioning: bool,
 
-    /// If set true, shutdown will not finish until all pending evictions have been issued and
-    /// replied to. If not set (the default) shutdown will be considered complete when the only
+    /// If set false (default), shutdown will not finish until all pending evictions have been
+    /// issued and replied to. If set true shutdown will be considered complete when the only
     /// remaining work is pending evictions.
+    ///
+    /// This flag is useful during tests to avoid needing to deal with lots of uninteresting
+    /// evictions during shutdown. Alternatively, if a lang implementation finds it easy to clean
+    /// up during shutdown, setting this true saves some back-and-forth.
     #[builder(default = "false")]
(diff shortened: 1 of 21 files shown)


core/src/lib.rs
```diff
@@ -102,25 +102,27 @@ where
...
 /// Create a worker for replaying a specific history. It will auto-shutdown as soon as the history
-/// has finished being replayed. The provided client should be a mock, and this should only be used
-/// for workflow testing purposes.
-pub fn init_replay_worker(
+/// has finished being replayed.
+pub fn init_replay_worker<I>(
     mut config: WorkerConfig,
-    history: &History,
-) -> Result<Worker, anyhow::Error> {
+    histories: I,
```
cretz, line 108: I think a cleaner API would be to return a struct here that has a worker and a public function to be given a history. That way lang can send individual histories at any time after init instead of maintaining an iterator. The "replayer" is modeled as a class in Python which makes a lot of sense and probably deserves to be a struct in Rust.

cretz (approved):

core/src/lib.rs
```diff
@@ -102,25 +102,27 @@ where
...
+) -> Result<Worker, anyhow::Error>
+where
+    I: IntoIterator<Item = History> + 'static,
```
bergundy, line 111: The sync iterator approach is not going to work well for TS / Python.
IMHO we need a way to asynchronously feed histories, the "feed" method should block if concurrently replays exceed a configurable threshold to avoid process memory growing unbounded.

core-api/src/worker.rs
```diff
@@ -95,6 +95,12 @@ pub struct WorkerConfig {
     /// todo: link to feature docs
     #[builder(default = "false")]
     pub use_worker_versioning: bool,
+
+    /// If set true, shutdown will not finish until all pending evictions have been issued and
+    /// replied to. If not set (the default) shutdown will be considered complete when the only
+    /// remaining work is pending evictions.
```
bergundy, line 101: Any reason not to set it to true by default? I thought the only reason was core tests.

bergundy: I'd also document why this flag is here

sdk-core-protos/src/history_info.rs
```diff
@@ -166,7 +160,13 @@ impl HistoryInfo {
...
             }),
             workflow_execution_task_queue: Some(TaskQueue {
-                name: task_q.into(),
+                name: self
+                    .wf_exe_started_attrs
+                    .task_queue
+                    .as_ref()
+                    .unwrap()
+                    .name
+                    .clone(),
```
bergundy, line 169: This is so Rust lol

bergundy (approved):
Overall lgtm, didn't go into the implementation details.

sdk-core-protos/src/history_info.rs
Sushisource: You got me there :laughing:

core-api/src/worker.rs
Sushisource: Yeah, I can make it the default.

cretz (approved):

PR #403, opened 2022-09-24
Auto-start of temporalite & testsrv for integ tests

Sushisource:
## What was changed
Temporalite is now used by default when running integ tests. CI still uses docker (for now, I will add temporalite additionally shortly). Java Test server is supported, but breaks many tests.

## Why?
No more docker or manually setting things up / stopping them when running integ tests, yay!

## Checklist

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?

### DIFF
--- a/core/src/ephemeral_server/mod.rs
+++ b/core/src/ephemeral_server/mod.rs
@@ -20,6 +20,7 @@ use zip::read::read_zipfile_from_stream;
 
 #[cfg(target_family = "unix")]
 use std::os::unix::fs::OpenOptionsExt;
+use std::process::Stdio;
 
 /// Configuration for Temporalite.
 #[derive(Debug, Clone, derive_builder::Builder)]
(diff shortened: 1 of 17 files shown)


core/src/ephemeral_server/mod.rs
```diff
@@ -217,7 +233,7 @@ impl EphemeralServer {
         if let Some(pid) = self.child.id() {
             let nix_pid = nix::unistd::Pid::from_raw(pid as i32);
             Ok(spawn_blocking(move || {
-                nix::sys::signal::kill(nix_pid, nix::sys::signal::Signal::SIGINT)?;
+                nix::sys::signal::kill(nix_pid, nix::sys::signal::Signal::SIGKILL)?;
```
cretz, line 236: This was breaking on older Python and leaving zombies. Empirical evidence showed on Python this was the Java test server process would get cleaned up properly when running in Python 3.7. I should clarify:

> Do not consider this fixed without running that scenario.

To mean:

> Do not alter this section without running that scenario.

If a Python 3.7 Linux integration test passes with this change, no prob. I do not understand why SIGKILL is leaving zombies and/or not terminating the test server here (or maybe I mistested that scenario or am forgetting which scenarios I tested). I would welcome any assistance debugging it.

cretz (commented):
Curious if you noticed any speedups from Temporalite? (I assume docker image download and run is probably a good bit slower than Temporalite download and run, but curious about runtime speedups)

core/src/ephemeral_server/mod.rs
Sushisource: Oh, weird. OK, I'll change it back.

cretz (approved):

PR #405, opened 2022-09-26
Limit max eager activities to 3

bergundy:
See https://github.com/temporalio/sdk-java/pull/1455

### DIFF
--- a/core/src/core_tests/activity_tasks.rs
+++ b/core/src/core_tests/activity_tasks.rs
@@ -662,23 +662,13 @@ async fn activity_tasks_from_completion_are_delivered() {
             num_eager_requested_clone.store(count, Ordering::Relaxed);
             Ok(RespondWorkflowTaskCompletedResponse {
                 workflow_task: None,
-                activity_tasks: vec![
-                    PollActivityTaskQueueResponse {
-                        task_token: vec![1],
-                        activity_id: "act_id_1_same_queue".to_string(),
+                activity_tasks: (1..4)
+                    .map(|i| PollActivityTaskQueueResponse {
+                        task_token: vec![i],
+                        activity_id: format!("act_id_{}_same_queue", i).to_string(),
                         ..Default::default()
-                    },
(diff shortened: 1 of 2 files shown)


core/src/worker/workflow/mod.rs
```diff
@@ -475,6 +477,11 @@ impl Workflows {
                         if same_task_queue {
                             if let Some(p) = at_handle.reserve_slot() {
                                 reserved.push(p);
+                                if reserved.len()
+                                    >= MAX_EAGER_ACTIVITY_RESERVATIONS_PER_WORKFLOW_TASK
+                                {
+                                    break;
+                                }
```
Sushisource, line 484: This won't work. Commands come in with `request_eager_execution` set true. You need to set it false for the remaining commands in the list.

A test should prove this works

bergundy: Oh, I forgot about that part, thanks

bergundy:
Fixed @Sushisource

core/src/core_tests/activity_tasks.rs
```diff
@@ -658,15 +662,27 @@ async fn activity_tasks_from_completion_are_delivered() {
...
+                        task_token: vec![2],
+                        activity_id: "act_id_2_same_queue".to_string(),
+                        ..Default::default()
+                    },
+                    PollActivityTaskQueueResponse {
+                        task_token: vec![3],
+                        activity_id: "act_id_3_same_queue".to_string(),
+                        ..Default::default()
+                    },
+                ],
```
Sushisource, line 681:
```suggestion
                activity_tasks: (1..3).map(|i| 
                    PollActivityTaskQueueResponse {
                        task_token: vec![i],
                        activity_id: format!("act_id_{i}_same_queue"),
                        ..Default::default()
                    }).collect()
```

Sushisource (approved):
Would be nice to save the copy-pastes

bergundy:
I hope I made you happy with the last commit @Sushisource

Issue #406, opened 2022-09-28
[Feature Request] Backoff more on resource exhausted errors


Sushisource:
See https://github.com/temporalio/sdk-features/issues/96

PR #408, opened 2022-10-04
Backoff more if we receive ResourceExhausted error

Quinn-With-Two-Ns:
Backoff more aggressively if we receive a  ResourceExhausted error from the server. implementation based on the suggestion from the server team .

```
Here is an example of how it can be handled just to give some idea:
https://github.com/temporalio/temporal/blob/3182dc2cde75e3e49fccb1de2905db5d30db58ab/common/backoff/retry.go#L204-L206
```

See also: https://github.com/temporalio/sdk-features/issues/96

closes https://github.com/temporalio/sdk-core/issues/406

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -180,6 +180,17 @@ impl RetryConfig {
         }
     }
 
+    pub(crate) const fn throttle_retry_policy() -> Self {
+        Self {
+            initial_interval: Duration::from_secs(1),
+            randomization_factor: 0.2,
+            multiplier: 2.0,
+            max_interval: Duration::from_secs(10),
+            max_elapsed_time: None,
+            max_retries: 0,
+        }
+    }
+
     pub(crate) fn into_exp_backoff<C>(self, clock: C) -> exponential::ExponentialBackoff<C> {
         exponential::ExponentialBackoff {
             current_interval: self.initial_interval,
--- a/client/src/retry.rs
+++ b/client/src/retry.rs
@@ -101,32 +101,49 @@ impl<SG> RetryClient<SG> {
         F: FnMut() -> Fut + Unpin,
         Fut: Future<Output = Result<R>>,
     {
-        FutureRetry::new(factory, TonicErrorHandler::new(rtc, call_name))
+        FutureRetry::new(
+            factory,
+            TonicErrorHandler::new(rtc, RetryConfig::throttle_retry_policy(), call_name),
+        )
     }
 }
 
 #[derive(Debug)]
 pub(crate) struct TonicErrorHandler<C: Clock> {
     backoff: ExponentialBackoff<C>,
+    throttle_backoff: ExponentialBackoff<C>,
     max_retries: usize,
     call_type: CallType,
     call_name: &'static str,
 }
 impl TonicErrorHandler<SystemClock> {
-    fn new(cfg: RetryConfig, call_name: &'static str) -> Self {
-        Self::new_with_clock(cfg, call_name, SystemClock::default())
+    fn new(cfg: RetryConfig, throttle_cfg: RetryConfig, call_name: &'static str) -> Self {
+        Self::new_with_clock(
+            cfg,
+            throttle_cfg,
+            call_name,
+            SystemClock::default(),
+            SystemClock::default(),
+        )
     }
 }
 impl<C> TonicErrorHandler<C>
 where
     C: Clock,
 {
-    fn new_with_clock(cfg: RetryConfig, call_name: &'static str, clock: C) -> Self {
+    fn new_with_clock(
+        cfg: RetryConfig,
+        throttle_cfg: RetryConfig,
+        call_name: &'static str,
+        clock: C,
+        throttle_clock: C,
+    ) -> Self {
         Self {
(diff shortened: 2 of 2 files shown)


client/src/retry.rs
```diff
@@ -529,6 +531,16 @@ mod tests {
     use std::{ops::Add, time::Instant};
     use tonic::Status;
 
+    /// Predefined retry configs with low durations to make unit tests faster
+    const TEST_RETRY_CONFIG: RetryConfig = RetryConfig {
```
Sushisource, line 535: This should be guarded by `#[cfg(test)]` so we make sure it's never accidentally used in real code

Quinn-With-Two-Ns: isn't the whole module under `#[cfg(test)]` so it can't be used?

client/src/retry.rs
```diff
@@ -184,10 +184,12 @@ where
...
                 Some(backoff) => {
-                    if cfg!(test) {
-                        // Allow unit tests to do lots of retries quickly. This does *not* apply
-                        // during integration testing, importantly.
-                        RetryPolicy::WaitRetry(Duration::from_millis(1))
+                    // We treat ResourceExhausted as a special case and backoff more
+                    // so we don't overload the server
+                    if e.code() == Code::ResourceExhausted {
+                        let extended_backoff =
+                            backoff.max(self.backoff.next_backoff().unwrap_or_default());
```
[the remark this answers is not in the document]
bergundy: Backoff more doesn't mean use the next interval from the same retry policy.
You need to use a separate retry policy for this.
See the Go and TS PRs for how we implemented this. They're slightly different implementations but the basic idea is different from what you did here.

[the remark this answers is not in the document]
Sushisource: We want to use the value of the 2x backoff for this request, but it's not immediately obvious to me we want to advance the backoff state 2x. That said, it also makes some good sense, so honestly this seems fine

Sushisource (approved):

client/src/retry.rs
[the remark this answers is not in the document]
Sushisource: I mean either way we're sorta just making things up here without more info from the server. A different policy isn't strictly necessary I think, and complicates the implementation quite a bit.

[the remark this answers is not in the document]
Quinn-With-Two-Ns: I can do another retry policy so we line up with what go did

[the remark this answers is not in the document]
bergundy: Yes please, we’d like to keep the behavior as close as possible between the SDKs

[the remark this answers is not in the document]
bergundy: The SDK team and the server team are in agreement about the desired behavior. We discussed the contract and decided that if there’ll be any more special cases for retries we will propagate this information from the server.

client/src/lib.rs
```diff
@@ -181,6 +181,17 @@ impl RetryConfig {
         }
     }
 
+    pub(crate) const fn throttle_retry_policy() -> Self {
+        Self {
```
Quinn-With-Two-Ns, line 185: this should match what is done on the server and go SDK

client/src/retry.rs
```diff
@@ -184,10 +201,12 @@ where
             match self.backoff.next_backoff() {
```
bergundy, line 201: You should switch on the code before consuming the next backoff.
Something like:

```
let backoff = if e.code() == Code::ResourceExhausted {
    self.throttle_backoff.next_backoff()
} else {
    self.backoff.next_backoff()
}
match backoff {
   // cases for None / Some
}
```

bergundy: If you want to match TS, you could consume both backoffs and then do the switch.

client/src/retry.rs
```diff
@@ -641,15 +672,53 @@ mod tests {
...
+                initial_interval: Duration::from_millis(2),
+                randomization_factor: 0.0,
+                multiplier: 4.0,
+                max_interval: Duration::from_millis(10),
+                max_elapsed_time: None,
+                max_retries: 10,
+            }
+            .into_exp_backoff(FixedClock(Instant::now())),
+        };
+        let result = err_handler.handle(1, Status::new(Code::ResourceExhausted, "leave me alone"));
```
bergundy, line 692:
```suggestion
        let result = err_handler.handle(1, Status::new(Code::ResourceExhausted, "leave Britney alone"));
```

client/src/retry.rs
bergundy: Hmm.. yeah, I see it matches Go now.

I like the TS approach where the attempt increments in both throttlers but it's really not critical IMHO.

Quinn-With-Two-Ns: I tried to match the server behavior and the go sdk behavior

```
if next = r.NextBackOff(); next == done {
	return err
}

if err == ctx.Err() || !isRetryable(err) {
	return err
}

if _, ok := err.(*serviceerror.ResourceExhausted); ok {
	next = util.Max(next, t.NextBackOff())
}
```

bergundy (approved):
Had a single non-blocking comment.

PR #412, opened 2022-10-14
Update API to latest revision of Schedule API

mjameswh:
## What was changed

- Update upstream API
- Fix some minor build issues resulting from the update

## Why?

- Implementing Schedule API in TypeScript SDK
- The Typescript SDK proto package piggy back on sdk-core's copy of upstream API

## Checklist 💣 💣 💣 
1. I don't have enough knowledge of this package at this point to determine if anything else needs to be done
2. I don't have much experience in writing rust; please review carefully
3. Some API changes could potentially be breaking, but according to the corresponding tickets in temporal/api, they should not be in use.


bergundy (approved):

protos/api_upstream/.github/CODEOWNERS
```diff
@@ -1,4 +1,4 @@
 # Syntax is here:
 # https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
 
-*                @temporalio/server @temporalio/sdk
\ No newline at end of file
+* @temporalio/server @temporalio/sdk
```
Sushisource, line 4: Github seems unhappy about this for some reason

Sushisource (approved):
:+1:

protos/api_upstream/.github/CODEOWNERS
mjameswh: Yeah... That's because group 'temporal/server [doesn't have write access on this repo](https://github.com/orgs/temporalio/teams/server/repositories). I think we can safely ignore this.

PR #415, opened 2022-10-17
Fix replaying same run id multiple times possibly hanging replay

Sushisource:
## What was changed
Because of the expectations we have around WF tasks it was possible during replay to feed the same history in multiple times and hang the worker, since it tried to apply it as a task instead of a brand new run.

## Why?
Make sure users accidentally feeding in dupe histories doesn't break replay

## Checklist

1. Closes 

2. How was this tested:
Added UT

3. Any docs updates needed?

### DIFF
--- a/core/src/replay/mod.rs
+++ b/core/src/replay/mod.rs
@@ -10,7 +10,7 @@ use futures::{FutureExt, Stream, StreamExt};
 use once_cell::sync::OnceCell;
 use parking_lot::Mutex;
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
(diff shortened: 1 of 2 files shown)


core/src/replay/mod.rs
```diff
@@ -184,6 +186,15 @@ impl Stream for Historator {
                                  execution started events",
                     )
                     .to_string();
+                if self.seen_run_ids.contains(&run_id) {
```
cretz, line 189: What is the situation where we run the same run ID over and over?

I wonder if we should break here. People may have a history file they are altering a few ways to run on the same replayer. I'd rather break on them than let this invalid set through.

Sushisource: If they literally pass in multiple histories with the same run id

cretz (approved):

bergundy:
TBH not sure why we need to track all of the run IDs, why do we need to dedupe here?
I'd let users choose if they want to dedupe based on run ID.

PR #422, opened 2022-10-21
Ignorable events

Sushisource:
## What was changed
Ignore events which server indicates are explicitly ignorable by the SDK.

## Why?
Future proofing

## Checklist

1. Part of https://github.com/temporalio/sdk-features/issues/143

2. How was this tested:
Added UT

3. Any docs updates needed?

### DIFF
--- a/core/src/core_tests/workflow_tasks.rs
+++ b/core/src/core_tests/workflow_tasks.rs
@@ -2054,3 +2054,36 @@ async fn continue_as_new_preserves_some_values() {
         .await
         .unwrap();
 }
+
+#[rstest]
+#[tokio::test]
+async fn ignorable_events_are_ok(#[values(true, false)] attribs_unset: bool) {
+    let mut t = TestHistoryBuilder::default();
+    t.add_by_type(EventType::WorkflowExecutionStarted);
+    let id = t.add_get_event_id(
+        EventType::Unspecified,
+        Some(
+            history_event::Attributes::WorkflowPropertiesModifiedExternallyEventAttributes(
+                Default::default(),
+            ),
+        ),
+    );
+    t.modify_event(id, |e| e.worker_may_ignore = true);
+    if attribs_unset {
+        t.modify_event(id, |e| {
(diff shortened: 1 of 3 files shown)


Sushisource:
For reviewers: Skip the first commit

core/src/core_tests/workflow_tasks.rs
```diff
@@ -2054,3 +2054,36 @@ async fn continue_as_new_preserves_some_values() {
...
         .unwrap();
 }
+
+#[rstest]
+#[tokio::test]
+async fn ignorable_events_are_ok(#[values(true, false)] attribs_unset: bool) {
+    let mut t = TestHistoryBuilder::default();
+    t.add_by_type(EventType::WorkflowExecutionStarted);
+    let id = t.add_get_event_id(
+        EventType::Unspecified,
```
bergundy, line 2064: Do event types that aren't in the compiled proto definitions end up as `Unspecified`?

core/src/core_tests/workflow_tasks.rs
```diff
@@ -2054,3 +2054,36 @@ async fn continue_as_new_preserves_some_values() {
...
+            history_event::Attributes::WorkflowPropertiesModifiedExternallyEventAttributes(
+                Default::default(),
+            ),
+        ),
+    );
+    t.modify_event(id, |e| e.worker_may_ignore = true);
+    if attribs_unset {
+        t.modify_event(id, |e| {
+            e.event_type = EventType::WorkflowPropertiesModifiedExternally as i32;
+            e.attributes = None;
```
bergundy, line 2075: It makes sense to me for unknown enum variants end up as `None`s.

bergundy (approved):
LGTM.
Just making sure that we've tested what adding a new enum variant for the attributes and type would actually look like.

cretz (approved):

core/src/core_tests/workflow_tasks.rs
Sushisource: Yes

PR #423, opened 2022-10-28
Super duper tracing upgrade

Sushisource:
## What was changed
* Log forwarding mega-upgrade, including context field variables
* Per layer filtering
* Console and export no longer exclusive

## Why?
Much more configurable

## Checklist

1. Closes 

2. How was this tested:
Existing tests and added one

3. Any docs updates needed?


cretz (approved):

core/src/telemetry/log_export.rs
```diff
@@ -0,0 +1,183 @@
...
+        let top_span = span!(Level::INFO, "yayspan", huh = "wat");
+        let _guard = top_span.enter();
+        info!("Whata?");
+        instrumented("hi");
+        info!("Donezo");
+
+        let logs = fetch_global_buffered_logs();
+        for log in logs {
+            println!("{:?}", log);
+        }
```
bergundy, line 181: Should this be tested?

bergundy: That is with assertions instead of printing?

core/src/telemetry/mod.rs
```diff
@@ -255,18 +232,43 @@ pub fn telemetry_init(opts: &TelemetryOptions) -> Result<&'static GlobalTelemDat
...
+                                    .with_filter(EnvFilter::new(filter)),
+                            )
                         }
                     }
-                    Logger::Forward(filter) => {
-                        log::set_max_level(*filter);
-                        globaldat.core_export_logger = Some(CoreExportLogger::new(*filter));
+                    Logger::Forward { filter } => {
+                        let (export_layer, logs_out) = CoreLogExportLayer::new();
+                        globaldat.logs_out = Some(Mutex::new(logs_out));
```
bergundy, line 270: Not new in this PR but is there a way to avoid the mutex here?

bergundy (approved):

core/src/telemetry/mod.rs
Sushisource: Nope. Tracing implementations need to access self immutably. The locked operation is very, very cheap.

core/src/telemetry/log_export.rs
Sushisource: Yeah, may as well verify a few things at least. Will update.

Issue #426, opened 2022-09-14
Legacy query provided at same time as other activation jobs


cretz:
## Expected Behavior

Server sends legacy query and activity resolution at the same time and workflow should be able to send response to query and next command.

## Actual Behavior

This one is a bit hard to reproduce. You have to make an activity resolution happen at the same time as a query. Then the query comes over as a legacy query from the test server and the workflow responds to the query and does a start child workflow in the same response. Core responds to this with:

> Lang SDK sent us a malformed workflow completion for run (a067d736-1f05-4217-8c4c-4c4c5674c605): Workflow completion had a legacy query response along with other commands. This is not allowed and constitutes an error in the lang SDK

cretz:
How I replicated: Take the big test at https://github.com/temporalio/sdk-python/blob/656b77bf4d1176d5fca235f395c0c6a045b56cff/tests/worker/test_workflow.py#L1565-L1706 and just add `while True:` to the top of `test_workflow_dataclass_typed` and move all code beneath it. It seems one of the queries coincides with the LA completion, and LA completion causes child workflow command. So it errors when I send back this completion to the aforementioned/previously-pasted activation:

```
run_id: "d3f235c7-14de-47b4-b184-40fde9a88677"
successful {
  commands {
    start_child_workflow_execution {
      seq: 1
      namespace: "default"
      workflow_id: "workflow-a875e407-d9c2-4645-9130-5af85af3b572_child"
      workflow_type: "DataClassTypedWorkflow"
      task_queue: "d37fdb4b-67b4-4da8-920c-0166b773e088"
      input {
        metadata {
          key: "encoding"
          value: "json/plain"
        }
        data: "{\"field1\":\"some value\"}"
      }
      parent_close_policy: PARENT_CLOSE_POLICY_TERMINATE
      workflow_id_reuse_policy: WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE
      cancellation_type: WAIT_CANCELLATION_COMPLETED
    }
  }
  commands {
    respond_to_query {
      query_id: "legacy_query"
      succeeded {
        response {
          metadata {
            key: "encoding"
            value: "json/plain"
          }
          data: "{\"field1\":\"some value\"}"
        }
      }
    }
  }
}
```

PR #427, opened 2022-11-03
Fix and add test for sending activation w/ legacy query + other job

Sushisource:
## What was changed
Never send a legacy query with other jobs

## Why?
It's illegal. Other jobs can trigger new commands, and a legacy query activation can never be responded to with other commands, and I promise I'll never do it (oops)

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/426

2. How was this tested:
New test

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/managed_run.rs
+++ b/core/src/worker/workflow/managed_run.rs
@@ -288,6 +288,7 @@ impl ManagedRun {
         match outcome {
             Ok((None, data, me)) => Ok(Some(me.prepare_complete_resp(resp_chan, data, false))),
             Ok((Some((chan, start_t, wft_timeout)), data, me)) => {
+                warn!("Thinks waiting on las");
                 if let Some(wola) = me.waiting_on_la.as_mut() {
                     wola.heartbeat_timeout_task.abort();
                 }
--- a/core/src/worker/workflow/managed_run.rs
+++ b/core/src/worker/workflow/managed_run.rs
@@ -288,7 +288,6 @@ impl ManagedRun {
         match outcome {
             Ok((None, data, me)) => Ok(Some(me.prepare_complete_resp(resp_chan, data, false))),
             Ok((Some((chan, start_t, wft_timeout)), data, me)) => {
-                warn!("Thinks waiting on las");
                 if let Some(wola) = me.waiting_on_la.as_mut() {
                     wola.heartbeat_timeout_task.abort();
                 }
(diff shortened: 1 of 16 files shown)


Sushisource:
Drafting this because I realized it's really not an accurate repro:

I don't 100% love it because it's not an exact repro, but it's similar. An exact repro would require me to put hooks inside the real worker code to introduce delays in some spots I haven't exactly determined yet.

In Chad's log, an LA resolves for a task which had an LA and a normal query. This LA, because of :sparkles: contention :sparkles: , manages to hit exactly the right codepath such that we mark the LA as done, and mark the WFT as completed while at the same time getting a new WFT with a legacy query. This legacy query WFT is immediately applied and we end up sending out a resolution with both the activation and the legacy query.

This is ultra hard to repro because what normally happens is the WFT is not marked as completed yet, and we buffer the legacy WFT, or it is and we send the resolution to lang before the legacy WFT has made it into the machinery. You have to get exactly the right async behavior for the WFT to be marked completed and apply the next WFT before actually sending out the resolution.

As I wrote this, I realized it's not possible for the server to respond the way this test does, since I force new WFT and that forced WFT shouldn't/can't contain a legacy query. So I need to figure out how to introduce the delay.

core/src/worker/workflow/managed_run.rs
```diff
@@ -288,6 +288,7 @@ impl ManagedRun {
         match outcome {
             Ok((None, data, me)) => Ok(Some(me.prepare_complete_resp(resp_chan, data, false))),
             Ok((Some((chan, start_t, wft_timeout)), data, me)) => {
+                warn!("Thinks waiting on las");
```
cretz, line 291: Leftover debug code?

cretz (approved):

core/src/worker/workflow/managed_run.rs
```diff
@@ -366,19 +367,26 @@ impl ManagedRun {
...
+            || outgoing_cmds.replaying
+            || is_query_playback
+            || no_commands_and_evicting);
+        // If there are pending LA resolutions, and we're responding to a query here,
+        // we want to make sure to force a new task, as otherwise once we tell lang about
+        // the LA resolution there wouldn't be any task to reply to with the result of iterating
+        // the workflow.
+        if has_query_responses && self.wfm.machines.has_pending_la_resolutions() {
+            force_new_wft = true;
+        }
```
Sushisource, line 380: For reviewers: This is the meat of the fix

PR #429, opened 2022-11-03
Don't request eager activities when worker started with no remote activities

bergundy:


bergundy:
I failed to compile this locally, not sure why, trying in CI.

I'm not rebasing this on latest master so we can patch the TS 1.4.3 release.

Sushisource (approved):
Thank ye!

core/src/worker/workflow/mod.rs
```diff
@@ -459,32 +459,34 @@ impl Workflows {
...
+                    let same_task_queue = attrs
+                        .task_queue
+                        .as_ref()
+                        .map(|q| q.name == self.task_queue)
+                        .unwrap_or_default();
+                    if same_task_queue
+                        && reserved.len() < MAX_EAGER_ACTIVITY_RESERVATIONS_PER_WORKFLOW_TASK
+                    {
+                        if let Some(p) = self
+                            .activity_tasks_handle
```
cretz, line 480: Does `activity_tasks_handle` represent local too? If not, why wouldn't this be `None` if there is no activity worker?

cretz (approved):

core/src/worker/workflow/mod.rs
Sushisource: Nope. There's a totally separate manager specifically for local activities.

cretz: Ah, I see now, this didn't _unset_ before because that conditional was outside of the loop.

bergundy:
Merged into the ts-release branch.
I'll release a fix ASAP and then issue a new PR for master.

Issue #384, opened 2022-09-02
[Feature Request] Don't require telemetry_init to be global


cretz:
### Is your feature request related to a problem? Please describe.

Currently people cannot configure telemetry per "core" runtime, they have to do it per process and have to remember to do it before doing anything else with core.

### Describe the solution you'd like

I know this is hard to do, but ideally there is a "telemetry" object that can be woven via async/task-local contexts throughout the Rust code (or even Tokio runtime specific). That way we can still have this global default, but people can configure a different "runtime"/"telemetry" that's not global if they want.

Ideally, instead of things like `connect_no_namespace` on client options and top-level `init_worker`, "Core" can be an instantiable thing with those things as methods that share all telemetry and other core-level pieces.

This is low priority. I am just making sure it is on the backlog.

PR #432, opened 2022-11-05
Nonglobal Telemetry

Sushisource:
## What was changed
* Made telemetry nonglobal
* Introduced a core runtime concept, which more or less exists already in lang as tokio runtime

## Why?
Can be useful to possibly configure different runtimes & telemetry options in the same process

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/384
2. How was this tested:
Existing / new tests

3. Any docs updates needed?


bridge-ffi/src/lib.rs
```diff
@@ -794,6 +752,8 @@ impl tmprl_worker_t {
         self.worker.request_workflow_eviction(&req.run_id);
     }
 
+    // TODO: Fetch logs
```
cretz, line 755: Can probably even remove this TODO, I don't think .Net is gonna use it (at least not anytime soon, if we get there, I think .Net FFI is fast enough I may request a literal per-log callback)

core/src/lib.rs
```diff
@@ -174,3 +179,66 @@ mod sealed {
         }
     }
 }
+
+pub struct CoreRuntime {
```
cretz, line 183: This abstraction is great. Brings us one step closer to `new_remote` that would connect to an external/WASM'd core.

For TS and Python (and .Net and Ruby), I recommend we have an in-language `Runtime` concept backed by this, w/ a lazily loaded singleton. Then each client (and worker if it's separate from client) can optionally accept a _different_ runtime or leave unset to use the singleton.

cretz (approved):

core/src/lib.rs
Sushisource: Yeah, I agree. In fact, I am adding something like that for my tests just so that I don't unnecessarily spawn a bajillion telemetry instances while they run

PR #433, opened 2022-11-08
Fix "bug" caused by debug-assert friendly fire

Sushisource:
## What was changed
The debug assert I added for the last bug that hit this path turns out to be an OK thing to happen in a completely different circumstance. So, in a weird twist of events, the assert was very useful to discover the bug in https://github.com/temporalio/sdk-core/pull/427 but turns out to be a non-issue in reality, and in fact can be triggered by the legitimate path covered by the test added in this PR.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (approved):
Admit to not reading the test

cretz (approved):

PR #439, opened 2022-11-17
Add activity_task_received metric and validate minimum number of activity task pollers

bergundy:


core/src/worker/activities.rs
```diff
@@ -374,7 +375,21 @@ impl WorkerActivityTasks {
     }
 
     /// Called when there is a new act task about to be bubbled up out of the manager
-    fn about_to_issue_task(&self, task: PermittedTqResp) -> ActivityTask {
+    fn about_to_issue_task(&self, task: PermittedTqResp, is_eager: bool) -> ActivityTask {
+        if let Some(ref act_type) = task.resp.activity_type {
+            if let Some(ref wf_type) = task.resp.workflow_type {
+                self.metrics
+                    .with_new_attrs([
```
cretz, line 382: I assume task queue and namespace is already present?

cretz (approved):
Each language's SDK features harness will also need a way to expose its worker's metrics. In Java and Go this can be done w/ raw capturing using metric abstractions. In TS maybe you can have a Otel collector on a local URL, but in Python this is tricky due to proto issues right now.

core/src/worker/activities.rs
```diff
@@ -374,7 +375,21 @@ impl WorkerActivityTasks {
     }
 
     /// Called when there is a new act task about to be bubbled up out of the manager
-    fn about_to_issue_task(&self, task: PermittedTqResp) -> ActivityTask {
+    fn about_to_issue_task(&self, task: PermittedTqResp, is_eager: bool) -> ActivityTask {
+        if let Some(ref act_type) = task.resp.activity_type {
+            if let Some(ref wf_type) = task.resp.workflow_type {
```
Sushisource, line 380: You should be able to combine these with the new let chains feature! https://rust-lang.github.io/rfcs/2497-if-let-chains.html 

`if let Some(ref act_type) = task.resp.activity_type && let Some(ref wf_type) = task.resp.workflow_type {`

core/src/worker/activities.rs
Sushisource: Yeah that's part of the metrics context

Sushisource (approved):

bergundy:
> Each language's SDK features harness will also need a way to expose its worker's metrics. In Java and Go this can be done w/ raw capturing using metric abstractions. In TS maybe you can have a Otel collector on a local URL, but in Python this is tricky due to proto issues right now.

We can scrape the prom endpoint.

core/src/worker/activities.rs
bergundy: I didn't know that was stable, thanks!

bergundy: It's unstable:

https://github.com/rust-lang/rust/issues/53667

Sushisource: Bleh, they stabilized part of it but I have no idea which part lol because every time I try to use it I fail

PR #444, opened 2022-11-30
Fix heartbeats being counted as a logical WFT

Sushisource:
## What was changed
WFT heartbeats should not count as a "real" WFT, they should instead be ignored when scanning ahead to the next logical WFT.

## Why?
Treating heartbeats as a full logical WFT meant that when scanning ahead to pre-resolve local activities, they were not always resolved when they should have been. This led to weird nondeterministic behavior where we would try to newly execute LAs that should have already been resolved.

## Checklist

1. Closes 

2. How was this tested:
Added repro test cases, and one long regression integ test that is not on by default.

Will also manually run tests in lang sdks w/ updated core.

3. Any docs updates needed?


core/src/worker/workflow/history_update.rs
```diff
@@ -366,6 +400,46 @@ pub mod tests {
         assert_eq!(seq_2.last().unwrap().event_id, 8);
     }
 
+    #[tokio::test]
+    async fn history_ends_abruptly() {
```
bergundy, line 404: Do we have a test for WFT failed and timeout?
I was too lazy too look.

tests/integ_tests/workflow_tests/local_activities.rs
```diff
@@ -632,3 +637,139 @@ async fn repro_nondeterminism_with_timer_bug() {
...
+        activity_type: "delay".to_string(),
+        start_to_close_timeout: Some(Duration::from_secs(20)),
+        input: "hi!".as_json_payload().expect("serializes fine"),
+        ..Default::default()
+    })
+    .await;
+    Ok(().into())
+}
+
+// Expensive to run - worth enabling on a stress/regression pipeline.
```
bergundy, line 666: Can we make this cheaper by reducing the number iterations and cache size?

bergundy (approved):

core/src/worker/workflow/history_update.rs
Sushisource: Yes

tests/integ_tests/workflow_tests/local_activities.rs
Sushisource: Maybe, but we have a need for a nightly-esque core pipeline anyway. I'd prefer to just make that.

Sushisource:
After continuing to hammer away with weird tests I managed to trigger a panic around expecting run state to not suddenly vanish. Don't think it has anything to do with these changes, but obviously I want to fix that.

Spikhalskiy (approved):

PR #453, opened 2022-12-16
Cancel LAs when WFTs fail

Sushisource:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


core/src/core_tests/local_activities.rs
```diff
@@ -783,12 +776,19 @@ async fn test_schedule_to_start_timeout() {
     worker.run_until_done().await.unwrap();
 }
 
+#[rstest::rstest]
+#[case::sched_to_start(true)]
+#[case::sched_to_close(false)]
 #[tokio::test]
-async fn test_schedule_to_start_timeout_not_based_on_original_time() {
+async fn test_schedule_to_start_timeout_not_based_on_original_time(
```
mjameswh, line 783: You should probably rename this test name to reflect both test cases. Maybe something along the line of `test_schedule_timeouts_...`

core/src/worker/activities/local_activities.rs
```diff
@@ -474,11 +515,54 @@ impl LocalActivityManager {
         }
     }
 
-    pub(crate) async fn shutdown_and_wait_all_finished(&self) {
-        while !self.dat.lock().outstanding_activity_tasks.is_empty() {
+    pub(crate) fn workflows_have_shutdown(&self) {
+        // TODO: Cancel _all_ outstanding las
```
mjameswh, line 519: Just checking: is the intention to cover this in a distinct PR, or this a forgotten TODO?

mjameswh (approved):
LGTM

core/src/worker/workflow/mod.rs
```diff
@@ -869,6 +874,18 @@ enum ActivationCompleteOutcome {
...
+    /// The workflow task failed, but we shouldn't report it. EX: We have failed 2 or more attempts
+    /// in a row.
+    WFTFailedDontReport,
+}
+/// Did we report, or not, completion of a WFT to server?
+#[derive(Debug)]
+enum WFTReportStatus {
+    Reported,
+    NotReported,
+    /// We didn't report, but we want to clear the outstanding workflow task anyway. See
```
Spikhalskiy, line 886: I would add a comment to `NotReported`. Something like "we didn't report and the worker needs to continue working on this WFT task". Otherwise, the difference between `NotReported` and `DropWFT` is not very clear.

Spikhalskiy (approved):

core/src/worker/activities/local_activities.rs
Sushisource: Actually just not going to do this - LAs should be allowed to finish during shutdown and if lang wants to abort sooner than that, that's up to it.

PR #455, opened 2022-12-19
feat: add failure cause to anyhow conversion

baszalmstra:
## What was changed

I modified the conversion from `anyhow::Error` to include the full error chain in the `Failure::cause`. 

## Why?
I often use `anyhow::Context` to provide helpful context to error messages. However, if the cause of the error is not printed this is pretty useless. The `Failure` API does already include a `cause` field so this seems meant to be! 😄 

## Checklist

2. How was this tested: I ran several activities that failed, the error message now includes the cause.


Sushisource:
Sweet, thanks for contributing! This clever little rfold is neat, added a little test just because I wanted to see it work. Also got rid of the other conversion function which was pointless, not sure why I had it there.

Sushisource (approved):

PR #456, opened 2022-12-19
Remove always-log feature from sdk and core to prevent double logging

rossdylan:
## What was changed
I've Removed the `always-log` feature from all `tracing` dependencies in order to fix a double logging issue.

## Why?
If a consumer of the `temporal-sdk` or `temporal-sdk-core` crates initializes a tracing-subscriber it will begin double-logging everything. I'm still not 100% sure why it happens but it seems to be something link-time related from the tracing crate. You can reproduce the problem just by including `temporal-sdk-core` in Cargo.toml.

## Checklist

1. How was this tested:
Built a simple reproduction:
```rust
fn main() {
    tracing_subscriber::fmt::init();
    tracing::info!("should not be double");
}
```
With `always-log` enabled in sdk and core running this code will double log. If the feature is removed it will not.


baszalmstra:
Thank you! I was just about to file an issue! Im running into exactly this.

Issue #454, opened 2022-12-19
[Feature Request] Error message to Failure


baszalmstra:
We are using this alpha version of the Rust SDK and ran into the following issue:

### Is your feature request related to a problem? Please describe.

We have an activity that can fail for a number of reasons. We would like to retry the activity for most errors but not for a few specific ones. I know this is possible through the `non_retryable_error_types` field of the `RetryPolicy`. However, currently, a (Rust) activity will always return an `anyhow::Error` result in case of an error. Digging through the code, I think this is converted to an `api::failure::v1::Failure` in the `anyhow_to_fail` function. 

### Describe the solution you'd like

I would like to have more control over this behavior. Ideally, I would be able to return a custom error type from my activities that implements `Into<Failure>`.

### Additional context

I would be more than happy to implement this myself and create a PR but I would like a bit of direction before I start.

Sushisource:
It would be great if you could contribute this, thanks! First note - keep in mind this isn't even alpha, this is still "totally unreleased SDK I happen to use for testing" :)

I think the best way to do this is probably make a new `ActivityError` enum. It will need a variant for accepting cancellation, which is currently handled via [downcasting](https://github.com/temporalio/sdk-core/blob/13bcad663568babbcea543833ddc7f3aacef7874/sdk/src/lib.rs#L495). It would need a variant for application level errors, which would be a struct including the `non_retryable` field, as well as an anyhow error. Then the `IntoActivityFunc` trait would be changed to have the functions ideally return `impl Into<ActivityError>` rather than just a raw `anyhow` error. Getting that to work would be the most ergonomic in terms of not needing to `map_err(Into::into)` everywhere (until you need to explicitly mark something non-retryable). 

Alternatively - we just make a `NonRetryableActivityError` struct (with an inner anyhow error), and try downcasting like is already done for accepting cancels. Now that I've typed everything... that might be the most ergonomic option in terms of avoiding uglying things up with too many conversions. Some care will need to be taken to make sure the `non_retryable` field is properly propagated from only the top-level error.

I'm not sure I want to expose raw `Failure` returning as an option, since those protos are... a bit confusing.

I'm open to input here for sure. This is all subject to change later as well, once I get the chance to actually make the SDK intended for public consumption.

tdejager:
I'm a colleague of @baszalmstra we've made the PR here: https://github.com/temporalio/sdk-core/pull/457

PR #457, opened 2022-12-20
Added the option to have the activity return a non-retryable error.

tdejager:
## What was changed

Added a `NonRetryableActivityError` type so that an ActivityFunction can return this Error and the activity is not retried. Hence, the workflow will fail.

## Why?

To be able to accommodate for the use-case where you as the user *know* that the activity should not be retried.

## Checklist

1. Closes: https://github.com/temporalio/sdk-core/issues/454

2. How was this tested:
Test was added `activity_non_retryable_failure_with_error` that uses the new function: `Failure::application_failure_from_error`.
  Also, created a manual test locally: Where an activity was used that returns the **new** error type. Not sure if a test could be added to `local_activities` to properly test the error conversion code? Could use some pointers here if this is deemed necessary :) Thanks!

4. Any docs updates needed?
 Not sure, as there is no currently published Rust docs.

### DIFF
--- a/sdk-core-protos/src/lib.rs
+++ b/sdk-core-protos/src/lib.rs
@@ -1127,6 +1127,26 @@ pub mod coresdk {
             }
         }
 
+        pub fn application_failure_from_error(ae: anyhow::Error, non_retryable: bool) -> Self {
+            Self {
+                failure_info: Some(FailureInfo::ApplicationFailureInfo(
+                    ApplicationFailureInfo {
+                        non_retryable,
+                        ..Default::default()
+                    },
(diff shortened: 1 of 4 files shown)


tdejager:
Seems to be failing. Is there any way I can check the output?

Sushisource (approved):
Awesome! LGTM. Thank you!

The CI is only failing b/c of a formatting issue. Run cargo fmt and it'll be good to go :+1:

tdejager:
Formatted :)

Sushisource:
Thanks for contributing!

PR #468, opened 2023-01-12
Don't panic on unusual history fetching responses from server

Sushisource:
## Why?
Turns out server can in fact respond in some ways I did not think it could. Handle those situations and evict the workflow rather than panicking. 

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/history_update.rs
+++ b/core/src/worker/workflow/history_update.rs
@@ -20,9 +20,13 @@ use std::{
 use temporal_sdk_core_protos::temporal::api::{
     enums::v1::EventType,
     history::v1::{History, HistoryEvent},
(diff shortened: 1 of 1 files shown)


cretz (approved):
Can you confirm this doesn't error on every empty page, only on empty pages when you know there should be more events? I know they can give a next page token to an empty page sometimes on purpose (because they can't know they are on the last event because they don't fetch +1 IIRC).

Sushisource:
> Can you confirm this doesn't error on every empty page, only on empty pages when you know there should be more events? I know they can give a next page token to an empty page sometimes on purpose (because they can't know they are on the last event because they don't fetch +1 IIRC).

One of the added tests covers that situation

PR #469, opened 2023-01-13
Initial fuzzy wf implementation & internal state replay

Sushisource:
Fuzzy workflow for stressing as well as ability to record all changes to state and play them back for debugging purposes.

There is also some boring dedupe in tests.


cretz (approved):
Will be interested to see how this is used, but in the meantime seems to have no public-facing effects, so LGTM.

Issue #443, opened 2022-11-29
[Bug] Changing activity type name should result in non deterministic error


bergundy:
The following change should result in non deterministic error:

```diff
const activities = proxyActivities({ startToCloseTimeout: '1m' })
export async function myWorkflow(): Promise<void> {
-  await activities.foo();
+  await activities.bar();
}
```

Today, the workflow can be replayed and no error is emitted.

mjameswh:
Validating activity type name would also help detect many cases where significant changes to workflow code structure does not result in non-determinism error on replay because former history events get perniciously matched to totally unrelated commands in the new code.

I have seen this situation myself a few time, but here is a very interesting example submitted by a slack user. He starts a workflow execution using this code, then, while the workflow is waiting on the final `sleep` (at `E`), he stops the worker, and modifies `parseWebhook` from being a workflow function to being an activity (ie. comment line `B` and uncomment line `C`), _without using a patch_ (the user was actually trying to figure out if a patch was needed in that case), and restarts the worker to test.

The user gets no non-determinism error, but he's puzzled observing that the workflow is now stuck _inside_ the main loop. Turns out that the newly added activity at line `C` is incorrectly matched with the historical execution of activity at line `D`. `webhookType` is now `undefined`, so the code takes the `continue` statement, and the implicit `sleep` done by the condition at line `A` now get incorrectly matched to the `sleep` at line `E`.

```
export async function testWorkflow(): Promise<void> {
  let status = 'pending';
  let webhook: object | null = null;
  let webhookType;

  setHandler(webhookSignal, (wbh: object) => {
    webhook = wbh;
  });

  while(!isFinalStatus(status)) {
    if (await condition(() => webhook !== null, WEBHOOK_TIMEOUT)) {       // *A*
      // Before
      ({ status, webhookType } = parseWebhook(webhook));                  // *B*

      // After
      // ({ status, webhookType } = await coreAct.parseWebhook(webhook)); // *C*

      webhook = null;
      if (webhookType !== 'test') continue;
    } else {
      status = await coreAct.getTransactionStatus();
    }

    await coreAct.updateTransaction({ status });                          // *D*
  }

  await sleep(120 * 1000);                                                // *E*
}
```

jcm300:
Slack user here 👋 
If somehow there's the need of the full code example, I can provide it.

Sushisource:
This always should have been considered nondeterminism. Surprised this one slipped by somehow. I'll fix shortly.

[...]

PR #475, opened 2023-01-27
Make activity type / id mismatches a nondeterministic change

Sushisource:
## What was changed
If an activity scheduled command has a different type from the scheduled event in history matching that sequence number, we'll now fail w/ nondeterminism as we always should've.

Turns out, the check that their ids match is also missing, which is less surprising than it sounds since we already do sequence number checks and in almost all cases the ID == the sequence number, but that's been fixed as well.

Also did a bunch of semi-related mechanical cleanup to test construction. You can look at only the first & last commit diffs to see relevant changes.

## Why?
Previous behavior was wrong.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/443
2. How was this tested:
Added test, existing tests.

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/heavy.yml
+++ b/.github/workflows/heavy.yml
@@ -8,6 +8,7 @@ on: # rebuild any PRs and main branch changes
 
 jobs:
   build-and-test:
+    timeout-minutes: 20
     runs-on: ubuntu-latest-16-cores
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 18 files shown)


Sushisource:
Turns out, the check that their ids match is also missing, which is less surprising than it sounds since we already do sequence number checks and in almost all cases the ID == the sequence number.

I'm implementing a fix for that too.

cretz (approved):

mjameswh:
Seems like there is a high probability that this change will be breaking for a lot of people...

PR #478, opened 2023-01-30
Fix failed fetch request not unblocking completion

Sushisource:
## What was changed
If fetching pages failed while in the middle of an activation completion, that could leave the completion stuck because the eviction wouldn't unblock it. This fixes that.

## Why?
Bug

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/history_update.rs
+++ b/core/src/worker/workflow/history_update.rs
@@ -26,6 +26,8 @@ use tracing::Instrument;
 lazy_static::lazy_static! {
     static ref EMPTY_FETCH_ERR: tonic::Status
         = tonic::Status::data_loss("Fetched empty history page");
+    static ref EMPTY_TASK_ERR: tonic::Status
+        = tonic::Status::data_loss("Received an empty workflow task with no queries or history");
 }
 
 /// Represents one or more complete WFT sequences. History events are expected to be consumed from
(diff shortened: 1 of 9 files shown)


core/src/worker/workflow/history_update.rs
```diff
@@ -147,12 +152,13 @@ impl HistoryPaginator {
             npt,
             client,
         );
+        if empty_hist && wft.legacy_query.is_none() && wft.query_requests.is_empty() {
+            return Err(EMPTY_TASK_ERR.clone());
```
mjameswh, line 156: Q: Does lang needs to expect and catch this error?

Sushisource: Nope. You don't need to do anything special, it's just going to show up in evictions / logs

core/src/worker/workflow/mod.rs
```diff
@@ -829,6 +829,7 @@ struct RequestEvictMsg {
     run_id: String,
     message: String,
     reason: EvictionReason,
+    because_fetch_failed: bool,
```
mjameswh, line 832: Could that special purpose field have been avoided? I don't know if adding a new `EvictionReason` would be better?

Sushisource: Good point, I'll change that

Spikhalskiy (approved):

PR #482, opened 2023-02-07
SDK Internal Patches

Sushisource:
Requires https://github.com/temporalio/api/pull/259 to be merged first

## What was changed
* Implemented internal patching mechanism for adjusting SDK behavior based on what version of the SDK has previously processed the workflow
* Auto-fail new workflow tasks which encounter a problem upon their application, but before any activation has been issued to lang. This may fix some scenarios where previously a WFT would simply time out.
* Activity, LocalActivity, and ChildWorkflow determinism checks all guarded by new mechanism. **NOTE!!!** this won't actually do anything until server supports the patching mechanism.
* `EphemeralServer::Default` is renamed to avoid clashing with stdlib

### Non-lang affecting changes
* Adjusted replay flag behavior to be more accurate
* Run workflow state stream inside dedicated single thread, removing `Send` requirement from everything that lives in it

## Why?
In support of https://github.com/temporalio/features/issues/200

## Checklist

1. Closes 

2. How was this tested:
Added various tests

3. Any docs updates needed?


core/src/internal_flags.rs
```diff
@@ -0,0 +1,84 @@
...
+/// When a flag has existed long enough the version it was introduced in is no longer supported, it
+/// may be removed from the enum. *Importantly*, all variants must be given explicit values, such
+/// that removing older variants does not create any change in existing values. Removed flag
+/// variants must be reserved forever (a-la protobuf), and should be called out in a comment.
+#[repr(u32)]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Copy, Clone, Debug)]
+pub(crate) enum CoreInternalFlags {
+    /// In this flag, additional checks were added to a number of state machines to ensure that
+    /// the ID and type of activities, local activities, and child workflows match during replay.
+    IdAndTypeDeterminismChecks = 0,
```
cretz, line 21: To confirm, basically when this version of core is incorporated into the SDK, we are gonna add this flag to the next wft completion for every workflow using the new SDK?

I wonder if you want to pack a few more features in here pre-launch, e.g. the one for patch search attr or any others. I guess lang can choose when to update core and the next wft complete at that point will have all of them, so not in this PR necessarily.

protos/local/temporal/sdk/core/workflow_completion/workflow_completion.proto
```diff
@@ -20,10 +21,14 @@ message WorkflowActivationCompletion {
 message Success {
     // A list of commands to send back to the temporal server
     repeated workflow_commands.WorkflowCommand commands = 1;
+    // Any internal flags which the lang SDK used in the processing of this activation
+    repeated uint32 used_internal_flags = 6;
```
cretz, line 25: I am probably just misreading or not finding a macro somewhere or something, but with a quick find-text I don't see where these are put on the WFT completion to the server.

core/src/internal_flags.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// We received a value higher than this code can understand.
+    TooHigh = u32::MAX,
+}
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub(crate) struct InternalFlags {
+    core: BTreeSet<CoreInternalFlags>,
```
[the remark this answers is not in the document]
Sushisource: Yeah, definitely not worth it. This could be 1000s of things which would never happen and it'll still be super cheap.

core/src/internal_flags.rs
Sushisource: No, only the first WFT in which the check is used, so it's easy to add them whenever needed, the comment in the upstream proto describes the precise behavior.

protos/local/temporal/sdk/core/workflow_completion/workflow_completion.proto
Sushisource: Here: https://github.com/temporalio/sdk-core/pull/482/files#diff-db8d7ea3d79aee3cbe4019c0292acc44a5f4e55de5aa225be9d2285b4cc62724R995

Sushisource: Oh wait, you're right, inside that it's not gathering it up right. Good catch!

core/src/internal_flags.rs
cretz: > only the first WFT in which the check is used

Ok, I think I have it now. But I am still confused about something.

Do you ever check ID/type mismatches during non-replay? How is it possible to check during non-replay when you don't even have the event to check against? Basically, when would `replaying` ever be false for `try_use` when you're testing against an event during replay? (sorry, reading code but not clear)

The reason I ask is because I want to confirm the "replaying" bool when this flag applies to confirm it won't do an ID/type for older tasks (which obviously it is built not to, I just need understanding).

cretz (approved):
Approving, but have an outstanding question just to confirm understanding

core/src/worker/workflow/machines/activity_state_machine.rs
```diff
@@ -394,19 +401,24 @@ impl ScheduleCommandCreated {
         dat: SharedState,
         sched_dat: ActTaskScheduledData,
     ) -> ActivityMachineTransition<ScheduledEventRecorded> {
-        if sched_dat.act_id != dat.attrs.activity_id {
-            return TransitionResult::Err(WFMachinesError::Nondeterminism(format!(
-                "Activity id of scheduled event '{}' does not \
+        if dat.internal_flags.borrow_mut().try_use(
+            CoreInternalFlags::IdAndTypeDeterminismChecks,
+            sched_dat.last_task_in_history,
```
cretz, line 406: Scenario I just thought up:

* I have a workflow started a while back that has started an activity and is waiting on it (so schedule activity is in the last task)
* I've since updated my workflow code to rename the activity type
* I issue a query against the this long-since running workflow and it works against my older SDK because this check is not in place
* I upgrade my SDK
* I issue a query again and it breaks now because the activity is technically the last task in history

Any concerns with this scenario?

mjameswh: Just noting that having the check here means that the flag will essentially be added on every workflows calling an activity from this point on. Alternatively, it would be possible to do this check only if one of the check below fails, thus adding the flag only on workflows that actually have a problem.

Nothing bad here, just mentioning because I think it somehow contradicts something else you wrote elsewhere, but can't find back now, so feel free to ignore.

core/src/worker/workflow/machines/child_workflow_state_machine.rs
```diff
@@ -124,13 +135,32 @@ impl StartCommandCreated {
     pub(super) fn on_start_child_workflow_execution_initiated(
         self,
         state: SharedState,
-        initiated_event_id: i64,
+        event_dat: ChildWorkflowInitiatedData,
     ) -> ChildWorkflowMachineTransition<StartEventRecorded> {
+        if state.internal_flags.borrow_mut().try_use(
+            CoreInternalFlags::IdAndTypeDeterminismChecks,
+            dbg!(event_dat.last_task_in_history),
```
mjameswh, line 142: Why `dbg!` here?

Sushisource: leftover, thanks

core/src/worker/workflow/machines/activity_state_machine.rs
Sushisource: @cretz I'm OK with that scenario. The workflow hasn't yet ever "verified" the activity, because it's never processed that latest task. In fact, I'm not even totally sure your scenario can happen, because the first query would I believe cause the old worker to process and complete the task containing the schedule event, and then then new worker wouldn't even run the check.

mjameswh:
Do we have any way of determining either the server supports SDK metadata or not?

Sushisource:
@mjameswh No, we don't, but at least for this particular flag it isn't yet relevant. We could potentially add a server capability though, if/when we introduce a flag which would explode on that behavior.

mjameswh:
> No, we don't, but at least for this particular flag it isn't yet relevant. We could potentially add a server capability though, if/when we introduce a flag which would explode on that behaviour.

That specific change will not be problematic, but I will have to move TS SDK's internal patches to this new mechanism sooner than later, and the `condition(fn, 0)` patch would definitely explose in that situation.

Sushisource:
@mjameswh Makes sense. I have to go make server changes to propagate this new proto anyway. I can add the capability while I'm at it.

Issue #483, opened 2023-02-10
[Bug] Some metrics missing "temporal_" prefix


cretz:
### Describe the bug

In Python temporalio 1.0.0 (so older core), running a client/worker w/ Prometheus enabled gives some metrics like `temporal_num_pollers` with the `temporal_` prefix present and other metrics like `long_request` and `request` without them present.

This could be an issue with older core. If it is thought to be fixed, let me know and I will upgrade and try to replicate.

PR #484, opened 2023-02-10
Fix client metrics lacking the `temporal_` prefix

Sushisource:
## What was changed
Client metrics were emitted missing this prefix, by accident.

## Why?
Not what we say on the tin.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/483

2. How was this tested:
Existing test

3. Any docs updates needed?
Potentially a breaking change for users depending on the incorrect names. There is an existing flag to not use the prefix, but that'll apply to _all_ metrics, including ones that maybe they expected to find the prefix on. 

IMO, we should go with this and just call it out in release notes, as emitting multiple metrics is uggo - but if needed we can add _another_ flag for specifically this issue.

### DIFF
--- a/tests/integ_tests/metrics_tests.rs
+++ b/tests/integ_tests/metrics_tests.rs
@@ -28,7 +28,7 @@ async fn prometheus_metrics_exported() {
     let addr = rt.telemetry().prom_port().unwrap();
     let opts = get_integ_server_options();
     let mut raw_client = opts
-        .connect_no_namespace(rt.metric_meter(), None)
+        .connect_no_namespace(rt.metric_meter().as_deref(), None)
         .await
         .unwrap();
     assert!(raw_client.get_client().capabilities().is_some());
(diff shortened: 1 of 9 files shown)


tests/integ_tests/metrics_tests.rs
```diff
@@ -28,7 +28,7 @@ async fn prometheus_metrics_exported() {
     let addr = rt.telemetry().prom_port().unwrap();
     let opts = get_integ_server_options();
     let mut raw_client = opts
-        .connect_no_namespace(rt.metric_meter(), None)
+        .connect_no_namespace(rt.metric_meter().as_deref(), None)
```
Sushisource, line 31: Note for SDK authors - you'll need to add this small change to your bridge code.

cretz:
Would like @bergundy's opinion on if there are concerns just fixing this.

bergundy (commented):
TBH I don't know how much users are relying on these client metrics but to be on the safe side we might want to add a separate prefix for these.

We had a similar issue with worker metrics and we added this flag in case users wanted to preserve the buggy behavior.

https://typescript.temporal.io/api/interfaces/worker.TelemetryOptions#notemporalprefixformetrics

Sushisource:
> TBH I don't know how much users are relying on these client metrics but to be on the safe side we might want to add a separate prefix for these.
> 
> We had a similar issue with worker metrics and we added this flag in case users wanted to preserve the buggy behavior.
> 
> https://typescript.temporal.io/api/interfaces/worker.TelemetryOptions#notemporalprefixformetrics

Right, that same flag (of course) exists in Core, and I mentioned in the description ("There is an existing flag to not use the prefix...")

I think another flag is overkill and uglies up our options for an indeterminate period of time to little benefit, but we can add it if we feel the extra safety is needed.

bergundy (approved):

PR #486, opened 2023-02-14
Activity poller refactor

bergundy:
Refactor the activity poller to ensure it only shuts down once there are no more activity starts or cancellations to deliver.

The previous behavior was to shut down the poller on the first source to complete.
There is an outstanding issue that I did not fix where local activities shut down too early, need to follow up on that.

For Core based SDKs, the implication here is that the graceful shutdown timer should be started immediately after shutdown has been initiated instead of after the poller returns shutdown.
The main reason for this requirement is so activity cancellations can still be delivered to Lang while there are inflight activities (local or non-local).


core/src/core_tests/activity_tasks.rs
```diff
@@ -956,6 +947,13 @@ async fn activity_tasks_from_completion_reserve_slots() {
         .await
         .unwrap();
         barr.wait().await;
+        // TODO: This is ugly, find a better way to ensure that we don't shutdown until the workflow completes.
+        // This is required because after shutting down eager activities will not be requested.
+        tokio::time::sleep(Duration::from_millis(50)).await;
+        core.initiate_shutdown();
+        // While this test requests eager activity tasks, none are returned in poll responses.
+        let err = core.poll_activity_task().await.unwrap_err();
+        assert_matches!(err, PollActivityError::ShutDown);
```
bergundy, line 956: Can remove this now.

core/src/worker/mod.rs
```diff
@@ -69,6 +72,10 @@ pub struct Worker {
     /// Will be called at the end of each activation completion
     #[allow(clippy::type_complexity)] // Sorry clippy, there's no simple way to re-use here.
     post_activate_hook: Option<Box<dyn Fn(&Self, &str, usize) + Send + Sync>>,
+    /// Set when non-local activities are complete and should stop being polled
+    non_local_activities_complete: Arc<AtomicBool>,
+    /// Set when local activities are complete and should stop being polled
+    local_activities_complete: Arc<AtomicBool>,
```
bergundy, line 78: Ended up going with these atomics since I didn't want to do another refactor.
Can revisit in a follow up PR.

core/src/worker/mod.rs
```diff
@@ -539,23 +581,25 @@ fn build_wf_basics(
...
 
-        let cfg = test_worker_cfg()
-            .max_outstanding_activities(5_usize)
-            .build()
-            .unwrap();
-        let worker = Worker::new_test(cfg, mock_client);
-        assert_eq!(worker.activity_poll().await.unwrap(), None);
-        assert_eq!(worker.at_task_mgr.unwrap().remaining_activity_capacity(), 5);
-    }
+    // TODO: this test is stuck now that the activity poller loops until a task is available
```
bergundy, line 585: I commented this out a while ago and the code has changed significantly since then.
I'll check this again after the recent changes.

core/src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -74,15 +71,116 @@ pub enum ActivityHeartbeatError {
...
 }
 
 /// Manages activity heartbeating for a worker. Allows sending new heartbeats or requesting and
 /// awaiting for the shutdown. When shutdown is requested, signal gets sent to all processors, which
 /// allows them to complete gracefully.
 impl ActivityHeartbeatManager {
+    /// Creates a new instance of an activity heartbeat manager and returns a handle to the user,
+    /// which allows to send new heartbeats and initiate the shutdown.
+    /// Returns the manager and a channel that buffers cancellation notifications to be sent to Lang.
+    pub fn new(client: Arc<dyn WorkerClient>) -> (Self, UnboundedReceiver<PendingActivityCancel>) {
```
bergundy, line 83: I just moved the constructor from where it was before, there was a duplicate `impl` block.

core/src/worker/activities.rs
```diff
@@ -145,87 +139,216 @@ impl WorkerActivityTasks {
...
-        self.poller.notify_shutdown();
+        self.poller_shutdown_token.cancel();
+        self.eager_activities_semaphore.close();
     }
 
     /// Wait for all outstanding activity tasks to finish
-    pub(crate) async fn wait_all_finished(&self) {
-        while !self.outstanding_activity_tasks.is_empty() {
-            self.complete_notify.notified().await
+    async fn wait_all_finished(
```
bergundy, line 305: TODO: this seems to require a better name now that it also notifies shutdown.
Maybe `wait_drain_and_notify_heartbeat_manager`...

core/src/core_tests/activity_tasks.rs
bergundy: Just the last poll.

core/src/worker/mod.rs
bergundy: Local and non-local activities should be aggregated into a single stream instead.

core/src/worker/mod.rs
bergundy: Checked, we'd have to revisit this test.

tests/integ_tests/workflow_tests/local_activities.rs
```diff
@@ -242,12 +244,31 @@ async fn cancel_immediate(#[case] cancel_type: ActivityCancellationType) {
...
+        if !self.cancel_on_workflow_completed {
+            return;
+        }
+        if let Some(workflow_activation_completion::Status::Successful(
+            workflow_completion::Success { commands, .. },
+        )) = completion.status.as_ref()
+        {
+            if !commands.is_empty() {
+                if let Some(Variant::CompleteWorkflowExecution(_)) =
+                    commands[commands.len() - 1].variant.as_ref()
```
Sushisource, line 261: `.last().unwrap()` is a bit more idiomatic

But I think this whole thing may not be needed. Just call the manual cancel token right at the end of the workflow function.

test-utils/src/lib.rs
```diff
@@ -302,6 +303,11 @@ impl CoreWfStarter {
         self
     }
 
+    pub fn no_remote_activities(&mut self) -> &mut Self {
```
Sushisource, line 306: I don't love how this ended up everywhere, but I don't want to hold things up on getting rid of it either, so I can see if I can do some dedupe that combines options & wf/act registration all in one go so this gets automatically set when needed, but I don't wanna block the PR on it.

Sushisource (commented):
Overall I think this makes good sense, just some opportunities to simplify / clean up

core/src/abstractions.rs
```diff
@@ -87,6 +89,99 @@ impl MeteredSemaphore {
     }
 }
 
+/// A version of MeteredSemaphore that can be closed and supports waiting for close to complete.
+/// Once closed, no permits will be handed out.
+/// Close completes when all permits have been returned.
+pub(crate) struct ClosableMeteredSemaphore {
+    inner: Arc<MeteredSemaphore>,
+    outstanding_permits: Arc<AtomicUsize>,
+    close_requested: Arc<AtomicBool>,
```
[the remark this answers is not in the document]
bergundy: Do you see a problem with the current implementation or is it just your opinion that mutex is better than 2 atomics?

core/src/abstractions.rs
```diff
@@ -87,6 +89,99 @@ impl MeteredSemaphore {
...
+        res.map(|permit| TrackedOwnedMeteredSemPermit {
+            inner: Some(permit),
+            on_drop: self.on_permit_dropped(),
+        })
+    }
+
+    fn on_permit_dropped(&self) -> Box<dyn Fn() + Send + Sync> {
+        let outstanding_permits = self.outstanding_permits.clone();
+        let close_requested = self.close_requested.clone();
+        let close_token = self.close_complete_token.clone();
```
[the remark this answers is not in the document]
bergundy: Yeah, I like this, I'll change

core/src/abstractions.rs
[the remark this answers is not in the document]
Sushisource: > help you avoid the dance where you increment before acquiring.

That's the main thing I think. There's nothing objectively wrong with it - but that bit of code made me squint for a few minutes

core/src/abstractions.rs
```diff
@@ -87,6 +89,99 @@ impl MeteredSemaphore {
...
+    }
+}
+
+/// Tracks an OwnedMeteredSemPermit and calls on_drop when dropped.
+pub(crate) struct TrackedOwnedMeteredSemPermit {
+    inner: Option<OwnedMeteredSemPermit>,
+    on_drop: Box<dyn Fn() + Send + Sync>,
+}
+impl From<TrackedOwnedMeteredSemPermit> for OwnedMeteredSemPermit {
+    fn from(mut value: TrackedOwnedMeteredSemPermit) -> Self {
```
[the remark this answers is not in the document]
bergundy: That's what I tried initially but I got:
```
cannot move out of type `TrackedOwnedMeteredSemPermit`, which implements the `Drop` trait
```

core/src/core_tests/activity_tasks.rs
```diff
@@ -518,8 +511,8 @@ async fn can_heartbeat_acts_during_shutdown() {
 
     let act = core.poll_activity_task().await.unwrap();
     // Make sure shutdown has progressed before trying to record heartbeat / complete
-    let shutdown_fut = core.shutdown();
-    advance_fut!(shutdown_fut);
+    core.initiate_shutdown();
+    tokio::time::sleep(Duration::from_millis(50)).await;
```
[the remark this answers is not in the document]
bergundy: I forgot to revert this change after I fix some stuff. Thanks.

core/src/core_tests/mod.rs
```diff
@@ -57,7 +60,7 @@ async fn shutdown_interrupts_both_polls() {
             async move {
                 BARR.wait().await;
                 sleep(Duration::from_secs(1)).await;
-                Ok(Default::default())
+                Err(tonic::Status::cancelled(NO_MORE_WORK_ERROR_MSG))
```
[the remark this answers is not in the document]
bergundy: Fixed, forgot to clean this up after I fixed some issues.

core/src/test_help/mod.rs
```diff
@@ -896,3 +897,34 @@ macro_rules! prost_dur {
             .expect("test duration fits")
     };
 }
+
+/// Initiate shutdown, drain the pollers, and wait for shutdown to complete.
+pub(crate) async fn drain_pollers_and_shutdown(worker: Worker) {
```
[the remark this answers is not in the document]
bergundy: Yeah.. I thought about it and just went with this because it was easier and I was in the phase of getting tests to pass.
I'll change, I like the extension trait.

core/src/worker/activities.rs
```diff
@@ -42,11 +44,15 @@ use temporal_sdk_core_protos::{
...
+use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
+use tokio::sync::{Mutex, Notify};
+use tokio::task::JoinHandle;
+use tokio_stream::wrappers::UnboundedReceiverStream;
+use tokio_util::sync::CancellationToken;
 use tracing::Span;
 
 #[derive(Debug, derive_more::Constructor)]
-struct PendingActivityCancel {
+pub(crate) struct PendingActivityCancel {
```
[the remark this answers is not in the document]
bergundy: 👍

core/src/worker/activities.rs
```diff
@@ -96,43 +102,31 @@ impl RemoteInFlightActInfo {
     }
 }
 
-struct NonPollActBuffer {
-    tx: async_channel::Sender<PermittedTqResp>,
-    rx: async_channel::Receiver<PermittedTqResp>,
```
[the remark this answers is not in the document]
bergundy: 👍

core/src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -121,17 +216,14 @@ impl ActivityHeartbeatManager {
         completed.notified().await;
     }
 
-    /// Returns a future that resolves any time there is a new activity cancel that must be
-    /// dispatched to lang
-    pub(super) async fn next_pending_cancel(&self) -> Option<PendingActivityCancel> {
-        self.incoming_cancels.lock().await.recv().await
+    pub(super) fn notify_shutdown(&self) {
+        self.shutdown_token.cancel();
```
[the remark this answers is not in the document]
bergundy: I'll add a docstring.
The name seems right to me.

core/src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -121,17 +216,14 @@ impl ActivityHeartbeatManager {
...
     }
 
     // TODO: Can own self now!
```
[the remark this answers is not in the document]
bergundy: Fixed the name as you suggested, updated the docstring, and removed the TODO.

core/src/worker/mod.rs
bergundy: I just didn't want this to drag on longer than it already is.
I agree a stream is the right approach.

core/src/worker/mod.rs
```diff
@@ -140,6 +147,10 @@ impl WorkerTrait for Worker {
         if let Some(atm) = self.at_task_mgr.as_ref() {
             atm.notify_shutdown();
         }
+        let lam = self.local_act_mgr.clone();
```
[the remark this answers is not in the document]
bergundy: Oops, leftover from some of my experiments. Thanks.

core/src/worker/mod.rs
```diff
@@ -275,8 +286,10 @@ impl Worker {
                 config.default_heartbeat_throttle_interval,
             )
         });
+        let mut poll_on_non_local_activities = true;
         if at_task_mgr.is_none() {
             info!("Activity polling is disabled for this worker");
+            poll_on_non_local_activities = false;
```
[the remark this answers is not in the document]
bergundy: Oh thanks!

core/src/worker/mod.rs
bergundy: I'm leaving this TODO open

test-utils/src/lib.rs
bergundy: Agree, don't love it either but I didn't want this PR to get out of hand.

tests/integ_tests/workflow_tests/local_activities.rs
bergundy: Changed to use `last` and got rid of the `is_empty` check.

core/src/worker/activities.rs
```diff
@@ -96,43 +102,31 @@ impl RemoteInFlightActInfo {
...
-    outstanding_activity_tasks: DashMap<TaskToken, RemoteInFlightActInfo>,
-    /// Buffers activity task polling in the event we need to return a cancellation while a poll is
-    /// ongoing.
-    poller: BoxedActPoller,
+    outstanding_activity_tasks: Arc<DashMap<TaskToken, RemoteInFlightActInfo>>,
     /// Holds activity tasks we have received by non-polling means. EX: In direct response to
     /// workflow task completion.
-    non_poll_tasks: NonPollActBuffer,
+    eager_activities_tx: UnboundedSender<TrackedPermittedTqResp>,
     /// Ensures we stay at or below this worker's maximum concurrent activity limit
```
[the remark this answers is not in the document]
bergundy: 👍

core/src/worker/activities.rs
```diff
@@ -145,87 +139,216 @@ impl WorkerActivityTasks {
...
+    pub(crate) async fn shutdown(&self) {
+        self.notify_shutdown();
+        self.shutdown_complete().await;
+    }
 
-        tokio::select! {
-            biased;
+    pub(crate) async fn finalize_shutdown(self) {
+        self.shutdown_complete().await;
+    }
```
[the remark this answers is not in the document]
bergundy: I'm fine with this suggestion 🤷

core/src/worker/activities.rs
bergundy: I'm just worried about races if we do this synchronously.
I would agree that your approach is better if we had a single place that manages all of the state but that's not the case.

bergundy: Another option is to feed this future into the heartbeat manager.

core/src/worker/mod.rs
Sushisource: I think we can fix this pretty easily by just making sure the mock gets called 1 or more times (which is as easy as adding `.times(1..)`) and then also setting a timeout on the poll. 

So, poll for a bit, see mock returned default, abort polling, assert slots are unchanged.

core/src/worker/activities.rs
bergundy: I think I managed to get rid of this by using `take_until` on the cancelled stream.

core/src/worker/mod.rs
```diff
@@ -471,7 +509,7 @@ impl Worker {
         // about to happen anyway. Tell the local activity manager that, so that it can know to
         // cancel any remaining outstanding LAs and shutdown.
         if matches!(r, Err(PollWfError::ShutDown)) {
-            self.initiate_shutdown();
+            self.initiate_shutdown(); // TODO: Why is this here? Wasn't this supposed to have been called already?
```
Sushisource, line 512: This is covering the situation where WFT pollers dying is the reason for shutdown

core/src/worker/mod.rs
Sushisource: Would still like to fix this before merge

Sushisource (approved):
I think this looks good! I would like to address the commented out test, but other than that I'm fine with merging. I'll probably take a pass through it after

core/src/worker/mod.rs
bergundy: Hmm.. but timeouts do eat up permits now.
We don't return anymore, we keep polling in a loop until we get a non-timeout response.
I can change the test to verify that we keep the permit alive as long as the poller times out.

bergundy: Changed.

PR #487, opened 2023-02-15
Attach nonfirst LA attempt count to WFTs

Sushisource:
## What was changed
Attach nonfirst LA attempt count to WFT completes

## Why?
Support metering changes

## Checklist

1. Closes 

2. How was this tested:
Added two UTs

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/activities/local_activities.rs
+++ b/core/src/worker/activities/local_activities.rs
@@ -110,12 +110,17 @@ impl Debug for NewLocalAct {
 pub(crate) enum LocalActRequest {
     New(NewLocalAct),
     Cancel(ExecutingLAId),
+    #[from(ignore)]
     CancelAllInRun(String),
     StartHeartbeatTimeout {
         send_on_elapse: HeartbeatTimeoutMsg,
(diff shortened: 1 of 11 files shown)


core/src/worker/activities/local_activities.rs
```diff
@@ -564,6 +594,16 @@ impl LocalActivityManager {
         }
     }
 
+    pub(crate) fn get_nonfirst_attempt_count(&self, for_run_id: &str) -> usize {
+        let dlock = self.dat.lock();
+        dlock
+            .la_info
+            .iter()
+            .filter(|(id, info)| id.run_id == for_run_id && info.first_wft_has_ended)
```
bergundy, line 602: Are we concerned with local activities that were started for this run ID but their workflow was previously evicted?
Do we keep those around? If so, can we detect them?

Sushisource: If that run was evicted then those LAs are invalidated and need to start over again anyway

bergundy (approved):

PR #488, opened 2023-02-17
Add Ruby package option to coresdk protos

mjameswh:
## What changed

- Add `option ruby_package` on all coresdk proto files

## Why

- Ruby convention for module name is camel case. This was previously fixed manually in sdk-ruby

### DIFF
--- a/protos/local/temporal/sdk/core/activity_result/activity_result.proto
+++ b/protos/local/temporal/sdk/core/activity_result/activity_result.proto
@@ -1,7 +1,7 @@
 syntax = "proto3";
 
 package coresdk.activity_result;
-option ruby_package = "Coresdk::ActivityResult";
+option ruby_package = "Temporalio::Bridge::Api::ActivityResult";
 
 import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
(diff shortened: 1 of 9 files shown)


protos/local/temporal/sdk/core/activity_result/activity_result.proto
```diff
@@ -1,6 +1,7 @@
 syntax = "proto3";
 
 package coresdk.activity_result;
+option ruby_package = "Coresdk::ActivityResult";
```
cretz, line 4: Wouldn't you want this something like `Temporalio::Bridge::Proto::ActivityResult`?

mjameswh: I agree that would be nicer.

Right now TS and Ruby both use the "official" package name, except for camel case in Ruby. Is that what they are named in .Net and Python?

cretz: Mostly. It is `temporalio.bridge.proto.activity_result` in Python and `Temporalio.Bridge.Api.ActivityResult` in .NET. I find putting everything under "bridge" and making all that internal/private and/or excluded from docs is helpful.

mjameswh: Updated.

cretz (approved):

Sushisource (approved):

Issue #492, opened 2023-02-22
Add search attribute upsert on patch calls


Sushisource:
Each patch call should include a command to upsert the `TemporalChangeVersion` search attribute, including the patch ID if it didn't previously exist in the attribute.

If the attribute size would be exceeded by such an upsert, log a warning and drop the new ID.

cretz:
> If the attribute size would be exceeded by such an upsert, log a warning and drop the new ID.

Hrmm, I wonder if the better approach would be to log a warning and drop the oldest IDs in the keyword list until it's back under size with this new ID.

Quinn-With-Two-Ns:
We discussed both approaches and thought not adding new IDs the lesser of two evils. Mostly because dropping old IDs is dangerous for users, and you most likely have to drop more then 1 old ID for a new one so you loose more information .

PR #495, opened 2023-03-02
Auto-upsert versions SA on patched calls & nondeterminism check for upserts

Sushisource:
## What was changed
* Auto-create an upsert command for the `TemporalChangeVersion` SA upon patched calls
* Warn & don't do that if it would exceed SA size limit
* Nondeterminism checking for upsert calls

## Why?
Parity + Safety

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/492

2. How was this tested:
Added UTs / existing UTs & ITs

3. Any docs updates needed?
We will need lang release notes to mention the determinism changes & the fact that writing/reading the `TemporalChangeVersion` attribute from workflow code is a no-no.

### DIFF
--- a/core/src/worker/workflow/machines/patch_state_machine.rs
+++ b/core/src/worker/workflow/machines/patch_state_machine.rs
@@ -123,16 +123,17 @@ pub(super) fn has_change<'a>(
     OnEventWrapper::on_event_mut(&mut machine, PatchMachineEvents::Schedule)
         .expect("Patch machine scheduling doesn't fail");
 
-    // Possibly produce an upsert SA command for this patch.
-    let maybe_upsert_cmd = if replaying_when_invoked && !seen_in_peekahead
-        || !internal_flags.borrow_mut().try_use(
-            CoreInternalFlags::UpsertSearchAttributeOnPatch,
-            !replaying_when_invoked,
-        ) {
-        // If we're replaying but this patch isn't in the peekahead, then we wouldn't have
-        // upserted either, and thus should not create the machine
+    // If we're replaying but this patch isn't in the peekahead, then we wouldn't have
+    // upserted either, and thus should not create the machine
+    let replaying_and_not_in_history = replaying_when_invoked && !seen_in_peekahead;
+    let cannot_use_flag = !internal_flags.borrow_mut().try_use(
+        CoreInternalFlags::UpsertSearchAttributeOnPatch,
+        !replaying_when_invoked,
+    );
+    let maybe_upsert_cmd = if replaying_and_not_in_history || cannot_use_flag {
         vec![]
     } else {
+        // Produce an upsert SA command for this patch.
         let mut all_ids = HashSet::<_, RandomState>::from_iter(existing_patch_ids);
         all_ids.insert(machine.shared_state.patch_id.as_str());
         let serialized = all_ids
--- a/core/src/worker/workflow/machines/patch_state_machine.rs
+++ b/core/src/worker/workflow/machines/patch_state_machine.rs
@@ -21,18 +21,28 @@ use super::{
     workflow_machines::MachineResponse, Cancellable, EventInfo, NewMachineWithCommand,
     OnEventWrapper, WFMachinesAdapter, WFMachinesError,
 };
-use crate::{protosext::HistoryEventExt, worker::workflow::machines::HistEventData};
+use crate::{
+    internal_flags::CoreInternalFlags,
+    protosext::HistoryEventExt,
+    worker::workflow::{machines::HistEventData, InternalFlagsRef},
+};
+use anyhow::Context;
 use rustfsm::{fsm, TransitionResult};
-use std::convert::TryFrom;
+use std::{collections::HashMap, convert::TryFrom};
 use temporal_sdk_core_protos::{
     constants::PATCH_MARKER_NAME,
-    coresdk::common::build_has_change_marker_details,
+    coresdk::{common::build_has_change_marker_details, AsJsonPayloadExt},
     temporal::api::{
-        command::v1::{Command, RecordMarkerCommandAttributes},
+        command::v1::{
+            Command, RecordMarkerCommandAttributes, UpsertWorkflowSearchAttributesCommandAttributes,
+        },
+        common::v1::SearchAttributes,
         enums::v1::CommandType,
         history::v1::HistoryEvent,
     },
 };
+
+pub(crate) const VERSION_SEARCH_ATTR_KEY: &str = "TemporalChangeVersion";
 
 fsm! {
     pub(super) name PatchMachine;
@@ -59,10 +69,14 @@ fsm! {
 #[derive(Clone)]
 pub(super) struct SharedState {
     patch_id: String,
+    internal_flags: InternalFlagsRef,
+    seen_in_peekahead: bool,
 }
 
 #[derive(Debug, derive_more::Display)]
-pub(super) enum PatchCommand {}
+pub(super) enum PatchCommand {
+    UpsertSearchAttr { is_executing: bool },
+}
 
 /// Patch machines are created when the user invokes `has_change` (or whatever it may be named
 /// in that lang).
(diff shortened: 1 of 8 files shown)


core/src/worker/workflow/machines/patch_state_machine.rs
```diff
@@ -71,53 +89,91 @@ pub(super) enum PatchCommand {}
...
+    let maybe_upsert_cmd = if replaying_when_invoked && !seen_in_peekahead
+        || !internal_flags.borrow_mut().try_use(
+            CoreInternalFlags::UpsertSearchAttributeOnPatch,
+            !replaying_when_invoked,
+        ) {
+        // If we're replaying but this patch isn't in the peekahead, then we wouldn't have
+        // upserted either, and thus should not create the machine
+        vec![]
+    } else {
+        let mut all_ids = HashSet::<_, RandomState>::from_iter(existing_patch_ids);
```
cretz, line 136: Is this set ordered by insertion? I think this should be a vec. Even if Elastic/visibility doesn't care, we should have user-predictable order of SA values.

core/src/worker/workflow/machines/patch_state_machine.rs
```diff
@@ -656,4 +750,75 @@ mod tests {
 
         wfm.shutdown().await.unwrap();
     }
+
+    const SIZE_OVERFLOW_PATCH_AMOUNT: usize = 180;
+    #[rstest]
+    #[case::happy_path(50)]
+    // We start exceeding the 2k size limit at 180 patches with this format
```
cretz, line 757: I wonder if it's worth an integration test to ensure server behavior (e.g. confirm they aren't using, say 2000 as a limit). We can do this lang side if needed.

core/src/worker/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -65,21 +126,39 @@ pub(super) struct Created {}
...
+        self,
+        shared: SharedState,
+        dat: CmdRecDat,
+    ) -> UpsertSearchAttributesMachineTransition<Done> {
+        if shared.internal_flags.borrow_mut().try_use(
+            CoreInternalFlags::UpsertSearchAttributeOnPatch,
+            !dat.replaying,
+        ) {
+            let sa = dat.sa_map.unwrap_or_default();
+            if !shared.should_skip_determinism && shared.sa_map != sa {
```
cretz, line 145: Was a little concerned about this equality check, but I can't currently think of a situation where byte-for-byte value check is harmful (we were bitten in Go by [this commit](https://github.com/protocolbuffers/protobuf-go/commit/582ab3de426ef0758666e018b422dd20390f7f26) which randomly adds whitespace to prevent proto byte checks, but this is checking json-encoded byte arrays not proto-encoded byte arrays so it should be ok).

core/src/worker/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -65,21 +126,39 @@ pub(super) struct Created {}
...
+
+impl CommandIssued {
+    pub(super) fn on_command_recorded(
+        self,
+        shared: SharedState,
+        dat: CmdRecDat,
+    ) -> UpsertSearchAttributesMachineTransition<Done> {
+        if shared.internal_flags.borrow_mut().try_use(
+            CoreInternalFlags::UpsertSearchAttributeOnPatch,
+            !dat.replaying,
```
cretz, line 142: Wouldn't the only time to check this be on replay? Or am I misunderstanding this field? Can you confirm there is a test testing this nondeterminism error?

core/src/worker/workflow/machines/patch_state_machine.rs
Sushisource: None of that really matters since this isn't how they're going to show up on the lang side when reading. That's up to the lang implementation.

core/src/worker/workflow/machines/patch_state_machine.rs
Sushisource: I read through the code and the 2k value limit is quite clear - in any case this is sort of a "best effort" since in theory users could change the value.

core/src/worker/workflow/machines/upsert_search_attributes_state_machine.rs
Sushisource: There is a test that verifies this works. The replay flag is false the first time you get a new WFT with the recorded command event. Replay only becomes true once we've caught up to the first non-command event in the most recent WFT.

core/src/worker/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -238,12 +346,61 @@ mod tests {
...
+                MachineError::Underlying(WFMachinesError::Nondeterminism(_))
+            );
+        } else {
+            recorded_res.expect("CommandRecorded should transition CommandIssued -> Done");
+            assert_eq!(Done {}.to_string(), sm.state().to_string());
+        }
+    }
+
+    #[tokio::test]
+    async fn upserting_change_version_directly_does_nothing() {
```
bergundy, line 371: Arguably it should only do nothing if the server supports SDK metadata and Core can upsert the search attribute.

bergundy (commented):
I'd just want to see tests for histories generated without the internal patch and server that supports the new behavior that we can replay these workflows:
- Upsert the TemporalChangeVersion marker
- Patch and then set the TemporalChangeVersion marker

core/src/worker/workflow/machines/patch_state_machine.rs
```diff
@@ -21,19 +21,37 @@ use super::{
...
+        command::v1::{
+            Command, RecordMarkerCommandAttributes, UpsertWorkflowSearchAttributesCommandAttributes,
+        },
+        common::v1::SearchAttributes,
         enums::v1::CommandType,
         history::v1::HistoryEvent,
     },
 };
 
+pub(crate) const VERSION_SEARCH_ATTR_KEY: &str = "TemporalChangeVersion";
```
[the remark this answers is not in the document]
Sushisource: Agreed

core/src/worker/workflow/machines/patch_state_machine.rs
```diff
@@ -71,53 +89,91 @@ pub(super) enum PatchCommand {}
...
+            CoreInternalFlags::UpsertSearchAttributeOnPatch,
+            !replaying_when_invoked,
+        ) {
+        // If we're replaying but this patch isn't in the peekahead, then we wouldn't have
+        // upserted either, and thus should not create the machine
+        vec![]
+    } else {
+        let mut all_ids = HashSet::<_, RandomState>::from_iter(existing_patch_ids);
+        all_ids.insert(machine.shared_state.patch_id.as_str());
+        let serialized = all_ids
```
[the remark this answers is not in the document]
Sushisource: It could. I don't think it matters at all. The backend doesn't care that's for sure.

core/src/worker/workflow/machines/patch_state_machine.rs
cretz: I think it matters because we should send the command in the same order every time. If we are asking people to be deterministic in their upsert we should too (and not just that it happens to be deterministic to iterate a set in Rust, but predictably deterministic).

core/src/worker/workflow/machines/patch_state_machine.rs
cretz: :+1: On the .NET or Python side I may have an integration test that pushes SA past the limit.

core/src/worker/workflow/machines/upsert_search_attributes_state_machine.rs
Sushisource: Done

core/src/worker/workflow/machines/patch_state_machine.rs
Sushisource: Changed

core/src/worker/workflow/machines/upsert_search_attributes_state_machine.rs
```diff
@@ -238,12 +355,94 @@ mod tests {
...
+            t.set_flags_first_wft(
+                &[CoreInternalFlags::UpsertSearchAttributeOnPatch as u32],
+                &[],
+            );
+        }
+        if with_patched_cmd {
+            t.add_has_change_marker(&patch_id, false);
+        }
+        t.add_upsert_search_attrs_for_patch(&[patch_id.clone()]);
+        t.add_full_wf_task();
```
bergundy, line 399: nit: this history wouldn't be generated by a real workflow, there would need to be something that prevents the workflow from completing in the first task but that probably wouldn't change the outcome of the test.

Sushisource: Not true. Waiting on a signal can result in a history which looks just like this

bergundy (approved):

core/src/worker/workflow/machines/upsert_search_attributes_state_machine.rs
bergundy: But then a signal would be delivered to unblock the workflow

Sushisource: Lol, yeah fair point. It doesn't matter for the test though indeed.

PR #496, opened 2023-03-02
Stop cloning state machines on every transition

Sushisource:
This had just been bugging me for a while. Needless copying going on. Will improve perf, probably not by much, but in some pathological cases it could make a real difference.

Side bonus, ended up with less code.


bergundy (approved):
Admit to not looking closely at the changes, should be fine if tests pass.

Issue #493, opened 2023-02-28
[Feature Request] Graceful activity worker shutdown


bergundy:
### Is your feature request related to a problem? Please describe.

Currently, all Lang SDKs implement their own graceful shutdown for activities.
The way it works is that once shutdown has been requested, after a user specified time, the worker cancels any inflight activities.
These activities cannot end up as cancelled since the server does not accept that, instead, they should end up as failed.

### Describe the solution you'd like

Move all of this logic into Core.

PR #499, opened 2023-03-03
Graceful shutdown for activities

Sushisource:
## What was changed
Auto-issue cancels for activities still running past a grace period

## Why?
Lang layer is implementing this right now

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/493

2. How was this tested:
UT/IT

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/activities.rs
+++ b/core/src/worker/activities.rs
@@ -47,7 +47,7 @@ use temporal_sdk_core_protos::{
         ActivityHeartbeat,
     },
     temporal::api::{
-        failure::v1::{failure::FailureInfo, CanceledFailureInfo, Failure},
+        failure::v1::{failure::FailureInfo, ApplicationFailureInfo, CanceledFailureInfo, Failure},
         workflowservice::v1::PollActivityTaskQueueResponse,
     },
 };
@@ -203,7 +203,7 @@ impl WorkerActivityTasks {
         let activity_task_stream = Self::merge_source_streams(
             source_stream,
             outstanding_activity_tasks.clone(),
-            start_tasks_stream_complete.clone(),
+            start_tasks_stream_complete,
             complete_notify.clone(),
             graceful_shutdown,
             cancels_tx,
@@ -418,9 +418,16 @@ impl WorkerActivityTasks {
                             act_info.issued_cancel_to_lang,
                             Some(ActivityCancelReason::WorkerShutdown),
                         ) {
-                            // We don't tell server about cancels for graceful shutdown since they
-                            // were never requested
-                            None
+                            // We don't report cancels for graceful shutdown as failures, so we
+                            // don't wait for the whole timeout to elapse, which is what would
+                            // happen anyway.
+                            client
+                                .fail_activity_task(
+                                    task_token.clone(),
+                                    Some(worker_shutdown_failure()),
+                                )
+                                .await
+                                .err()
                         } else {
                             let details = if let Some(Failure {
                                 failure_info:
@@ -577,6 +584,23 @@ pub(crate) struct PermittedTqResp {
 pub(crate) struct TrackedPermittedTqResp {
     pub permit: TrackedOwnedMeteredSemPermit,
     pub resp: PollActivityTaskQueueResponse,
+}
+
+fn worker_shutdown_failure() -> Failure {
+    Failure {
+        message: "Worker is shutting down and this activity did not complete in time".to_string(),
+        source: "".to_string(),
+        stack_trace: "".to_string(),
+        encoded_attributes: None,
+        cause: None,
+        failure_info: Some(FailureInfo::ApplicationFailureInfo(
+            ApplicationFailureInfo {
+                r#type: "WorkerShutdown".to_string(),
+                non_retryable: false,
+                details: None,
+            },
+        )),
+    }
 }
 
 #[cfg(test)]
--- a/core/src/worker/activities.rs
+++ b/core/src/worker/activities.rs
@@ -59,6 +59,8 @@ use tokio_stream::wrappers::UnboundedReceiverStream;
 use tokio_util::sync::CancellationToken;
 use tracing::Span;
 
+type OutstandingActMap = Arc<DashMap<TaskToken, RemoteInFlightActInfo>>;
+
 #[derive(Debug, derive_more::Constructor)]
 struct PendingActivityCancel {
     task_token: TaskToken,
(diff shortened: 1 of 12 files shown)


core/src/worker/activities.rs
```diff
@@ -255,68 +288,82 @@ impl WorkerActivityTasks {
...
-                        // outstanding activity task. This is fine because it means that we no
-                        // longer need to cancel this activity, so we'll just ignore such orphaned
-                        // cancellations.
-                        if let Some(mut details) = outstanding_tasks.get_mut(&next_pc.task_token) {
-                            if details.issued_cancel_to_lang {
-                                // Don't double-issue cancellations
-                                return None
+        source_stream
+            .filter_map(move |source| {
+                let outstanding_tasks = outstanding_tasks.clone();
```
Sushisource, line 293: Nothing changed here, autoformat just indented a bunch

core/src/worker/activities.rs
```diff
@@ -360,22 +407,33 @@ impl WorkerActivityTasks {
...
-                        }) = failure
-                        {
-                            details
+                        if matches!(
+                            act_info.issued_cancel_to_lang,
+                            Some(ActivityCancelReason::GracefulShutdown),
+                        ) {
+                            // We don't tell server about cancels for graceful shutdown since they
+                            // were never requested
+                            None
```
bergundy, line 416: We currently transform these into failures in lang (mentioned in the tests too), the rationale is that we don't want to wait for timeout until the activity can be restarted.

core/src/worker/activities.rs
```diff
@@ -514,6 +572,54 @@ pub(crate) struct TrackedPermittedTqResp {
...
+                outstanding,
+            }),
+        }
+    }
+
+    fn initiate(&self) {
+        if let Some(state) = self.once.take_once() {
+            let grace_deadline = Instant::now() + state.grace_period;
+            // We don't really need to join this
+            tokio::task::spawn(async move {
```
bergundy, line 605: I would make this a stream like the others and not have to worry about spawning this in the background or executing exactly once.

core/src/worker/activities.rs
```diff
@@ -514,6 +572,54 @@ pub(crate) struct TrackedPermittedTqResp {
...
+                // Don't issue cancels until we know we won't be getting any more tasks to avoid
+                // missing the cancellation of any just-starting activities.
+                state.start_tasks_stream_complete.cancelled().await;
+                // Make sure we've waited at least the grace period. This way if waiting for
+                // starts to finish took a while, we subtract that from the grace period.
+                tokio::time::sleep_until(grace_deadline.into()).await;
+                for mapref in state.outstanding.iter() {
```
bergundy, line 612: Also, don't need the map here since you have it in the merged starts and cancels stream above.

protos/local/temporal/sdk/core/activity_task/activity_task.proto
```diff
@@ -75,6 +75,8 @@ enum ActivityCancelReason {
     CANCELLED = 1;
     /// Activity timed out
     TIMED_OUT = 2;
+    /// Core is shutting down and the graceful timeout has elapsed
+    GRACEFUL_SHUTDOWN = 3;
```
bergundy, line 79: Would help TS if we kept the reason name: https://github.com/temporalio/sdk-typescript/blob/87e6a158f2623b89ed829c94f6b825eb61dd9d99/packages/worker/src/activity.ts#L22

core/src/worker/activities.rs
```diff
@@ -42,12 +51,16 @@ use temporal_sdk_core_protos::{
...
-use tokio::sync::{Mutex, Notify};
+use tokio::sync::{
+    mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
+    Mutex, Notify,
+};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 use tokio_util::sync::CancellationToken;
 use tracing::Span;
 
+type OutstandingActMap = Arc<DashMap<TaskToken, RemoteInFlightActInfo>>;
```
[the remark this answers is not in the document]
Sushisource: There's a bunch of other places we use this abbreviation (in this same file... in this same line even). I think it's about as common as WFT  - there are some things for which I'd really like to be able to use shorthand and I think this is probably one that qualifies.

core/src/worker/activities.rs
Sushisource: Ah, sure good point

core/src/worker/activities.rs
Sushisource: Yeah, makes sense. Forgot about that.

protos/local/temporal/sdk/core/activity_task/activity_task.proto
Sushisource: Will do

core/src/worker/activities.rs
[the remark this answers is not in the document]
bergundy: I know, you probably noticed that I started changing that in my PRs.
I'm requesting this because this is the only codebase we have in the team where we use these abbreviations and I personally would like to see this change mostly for the reason I mentioned above.
If you feel that strongly about it, I'm fine keeping as is, ultimately it's your decision.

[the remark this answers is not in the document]
Sushisource: The way I look at it is it'll take a reader a few seconds to figure this out (we're in a file named `activities` after all), but the cumulative line length additions leading to more line breaks etc are probably more harmful than the tiny warmup time it takes to learn the two abbreviations.

For anything in the public API, I definitely agree

tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -900,3 +902,66 @@ async fn it_can_complete_async() {
...
+        // Wait for all acts to be started before initiating shutdown
+        let _ = ACTS_STARTED.acquire_many(10).await;
+        handle();
+        // Kill workflow once all acts are cancelled. This also ensures we actually see all the
+        // cancels, otherwise run_until_done will hang since the workflow won't complete.
+        let _ = ACTS_DONE.acquire_many(10).await;
+        client
+            .terminate_workflow_execution(wf_name.to_owned(), None)
+            .await
+            .unwrap();
```
bergundy, line 961: Shouldn't need to terminate the workflow anymore since all of the activities should fail after one attempt.

bergundy (approved):
Got one final comment, but this is already good to merge.

tests/integ_tests/workflow_tests/activities.rs
Sushisource: We still do - because the next activity task won't be taken because that part is already shut down and hence the workflow no longer completes.

bergundy: 👍

bergundy: I would actually add more retries to get an additional check that the worker doesn't pick up any more activity tasks after shutdown is initiated.

Issue #498, opened 2023-03-03
[Feature Request] Sort activation jobs by desired execution order


bergundy:
We do this in every lang implementation, better to ensure the right order in Core.

See for reference: 
https://github.com/temporalio/sdk-dotnet/pull/29/files/1af1fe8bd952dd0020055e4e8f6b0cb0bebc9eed#diff-7d5de4dd86e4fda729feeeac183fc71ca2f35e30324a3a631228d311e150f1d8R195

Issue #497, opened 2023-03-02
[Feature Request] Discard post-terminal non-query commands


cretz:
### Describe the solution you'd like

Having to do this in lang now, should do in core. Can't remember the situation why we need to do this as arguably it shows something wrong in lang. But if we consider this wrong, we can warn on the core side.

Re https://github.com/temporalio/sdk-dotnet/pull/29#discussion_r1123810234

PR #502, opened 2023-03-06
Drop all post-terminal commands & sort activation jobs

Sushisource:
## What was changed
Drop & warn about commands following a terminal workflow command.
Also ensure jobs are sorted in the order langs are expecting anyways before sending them out.

## Why?
We'll end up blowing up server-side right now, which is a bit annoying for a user to fix when we can do this instead - and not require lang layer to do it.
De-dupe lang job-sorting work and move it into core.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/497 and closes https://github.com/temporalio/sdk-core/issues/498

2. How was this tested:
Added UTs. Existing tests.

4. Any docs updates needed?


core/src/worker/workflow/mod.rs
```diff
@@ -1005,6 +1005,18 @@ fn validate_completion(
                 });
             }
 
+            // Any commands after a terminal command should be ignored
+            if let Some(term_cmd_pos) = commands.iter().position(|c| c.is_terminal()) {
+                let remaining = commands.split_off(term_cmd_pos + 1);
+                if !remaining.is_empty() {
+                    warn!(
+                        "Activation included commands after a terminal workflow command. These \
```
cretz, line 1013: Can you confirm query results are still ok?

bergundy: Not sure we want the warning here, maybe debug.
We don't warn in lang today, we accept that this can happen if workflow is concurrent.

Sushisource: Yeah, good point. I have to double check if that scenario is even possible, but either way I'll protect for it.

Sushisource: I think the warning is worth keeping. Pretty much by definition users are writing somewhat sloppy WF code if they are doing this, but I don't hugely care. We can git rid of it, I'm just not sure exactly who getting rid of it benefits.

Sushisource: I got rid of it, not like anyone will use it anyway lol

core/src/worker/workflow/mod.rs
```diff
@@ -1273,3 +1304,22 @@ impl LocalActivityRequestSink for LAReqSink {
         res
     }
 }
+
+/// Sorts jobs in an activation to be in the order lang expects:
+/// `patches -> signals -> other -> queries`
+fn sort_act_jobs(wfa: &mut WorkflowActivation) {
+    wfa.jobs.sort_by(|j1, j2| {
```
cretz, line 1311: Can you confirm this is a "stable" sort (meaning equal things aren't reordered)?

core/src/worker/workflow/mod.rs
```diff
@@ -1273,3 +1304,22 @@ impl LocalActivityRequestSink for LAReqSink {
...
+        let j1v = j1.variant.as_ref().unwrap();
+        let j2v = j2.variant.as_ref().unwrap();
+        if discriminant(j1v) == discriminant(j2v) {
+            return Ordering::Equal;
+        }
+        match (j1v, j2v) {
+            (workflow_activation_job::Variant::NotifyHasPatch(_), _) => Ordering::Less,
+            (workflow_activation_job::Variant::SignalWorkflow(_), _) => Ordering::Less,
+            (workflow_activation_job::Variant::QueryWorkflow(_), _) => Ordering::Greater,
+            _ => Ordering::Equal,
```
cretz, line 1322: Can you confirm just for my reading that this case should never be hit because of the `discriminant(j1v) == discriminant(j2v)` check above? (fine to remain for exhaustion, I just want to confirm that the match clause never compares equal elements)

core/src/worker/workflow/mod.rs
```diff
@@ -1273,3 +1304,22 @@ impl LocalActivityRequestSink for LAReqSink {
...
+    wfa.jobs.sort_by(|j1, j2| {
+        // Unwrapping is fine here since we'll never issue empty variants
+        let j1v = j1.variant.as_ref().unwrap();
+        let j2v = j2.variant.as_ref().unwrap();
+        if discriminant(j1v) == discriminant(j2v) {
+            return Ordering::Equal;
+        }
+        match (j1v, j2v) {
+            (workflow_activation_job::Variant::NotifyHasPatch(_), _) => Ordering::Less,
+            (workflow_activation_job::Variant::SignalWorkflow(_), _) => Ordering::Less,
```
cretz, line 1320: What if `j1v` is `SignalWorkflow` and `j2v` is `NotifyHasPatch`? I think you should give integer ordinals to each job and then `cmp` them.

core/src/worker/workflow/mod.rs
Sushisource: Yes, it's stable.

core/src/worker/workflow/mod.rs
Sushisource: No, it never does.

core/src/worker/workflow/mod.rs
Sushisource: Yeah, that's a reasonable way to do it. Changed.

cretz (approved):

PR #504, opened 2023-03-07
Fix flag lookahead in next workflow task completion

bergundy:
Found and fixed an NDE with patches during replay with latest server in the TS SDK integration tests.

https://github.com/temporalio/sdk-typescript/actions/runs/4349517127/jobs/7599273593#step:13:11327

### DIFF
--- a/core/src/worker/workflow/machines/workflow_machines.rs
+++ b/core/src/worker/workflow/machines/workflow_machines.rs
@@ -426,7 +426,10 @@ impl WorkflowMachines {
         // Update observed patches with any that were used in the task
         if let Some(next_complete) = self
             .last_history_from_server
-            .peek_next_wft_completed(self.last_processed_event)
+            // Our last_processed event is always a WorkflowTaskStarted, we don't care about
+            // internal flags in the current workflow task since we already know about those.
+            // Peek past the WorkflowTaskCompleted event for the current workflow task.
+            .peek_next_wft_completed(self.last_processed_event + 2)
         {
             (*self.observed_internal_flags)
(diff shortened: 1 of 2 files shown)


bergundy:
I think I have a proper fix for this and ended up with a simpler repro.

core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -426,7 +426,10 @@ impl WorkflowMachines {
         // Update observed patches with any that were used in the task
         if let Some(next_complete) = self
             .last_history_from_server
-            .peek_next_wft_completed(self.last_processed_event)
+            // Our last_processed event is always a WorkflowTaskStarted, we don't care about
+            // internal flags in the current workflow task since we already know about those.
+            // Peek past the WorkflowTaskCompleted event for the current workflow task.
+            .peek_next_wft_completed(self.last_processed_event + 2)
```
Sushisource, line 432: I don't love adding 2 explicitly here. I think it'd be better to add a bool argument like `skip_next` or something (hard to come up with a good name there). And then encode logic into the peek function directly to skip past the next immediate completed and look for the next one after that. I think it's just a slightly safer / more intentional approach.

bergundy: I agree, there are probably some edge cases that it doesn't cover like when the current WFT failed.
Need a better solution to ensure we look ahead past the "previous" (or current depending on perspective) task.

Sushisource:
See https://github.com/temporalio/sdk-core/pull/505

PR #505, opened 2023-03-08
Fix internal flags not always reflecting correct state

Sushisource:
Replaces https://github.com/temporalio/sdk-core/pull/504


bergundy (approved):
Should we test a history that has a WFT failure or timeout before a patch is used?

PR #506, opened 2023-03-09
Ensure LA tasks can be delivered during shutdown

Sushisource:
## What was changed
Fix shutting down the LA manager a bit too early. Now we wait for workflows to shutdown first.

## Why?
If you completed a task with a new LA request, that would all just get dropped, now it doesn't.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/core_tests/local_activities.rs
+++ b/core/src/core_tests/local_activities.rs
@@ -23,12 +23,17 @@ use temporal_client::WorkflowOptions;
 use temporal_sdk::{
     ActContext, ActivityCancelledError, LocalActivityOptions, WfContext, WorkflowResult,
 };
-use temporal_sdk_core_api::Worker;
+use temporal_sdk_core_api::{
+    errors::{PollActivityError, PollWfError},
+    Worker,
+};
 use temporal_sdk_core_protos::{
     coresdk::{
         activity_result::ActivityExecutionResult,
         workflow_activation::{workflow_activation_job, WorkflowActivationJob},
-        workflow_commands::{ActivityCancellationType, QueryResult, QuerySuccess},
+        workflow_commands::{
+            ActivityCancellationType, QueryResult, QuerySuccess, ScheduleLocalActivity,
+        },
         workflow_completion::WorkflowActivationCompletion,
         ActivityTaskCompletion, AsJsonPayloadExt,
     },
@@ -43,7 +48,7 @@ use temporal_sdk_core_protos::{
 use temporal_sdk_core_test_utils::{
     schedule_local_activity_cmd, start_timer_cmd, WorkerTestHelpers,
 };
-use tokio::sync::Barrier;
+use tokio::{join, sync::Barrier};
 
 async fn echo(_ctx: ActContext, e: String) -> anyhow::Result<String> {
     Ok(e)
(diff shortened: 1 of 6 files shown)


core/src/core_tests/local_activities.rs
```diff
@@ -1044,3 +1049,68 @@ async fn local_act_records_nonfirst_attempts_ok() {
     assert!(nonfirst_counts.pop().unwrap() > 0);
     assert!(nonfirst_counts.pop().unwrap() > 0);
 }
+
+#[tokio::test]
+async fn las_can_be_delivered_during_shutdown() {
```
bergundy, line 1054: Please spell out local activities as a personal favor to me :)

core/src/worker/workflow/mod.rs
```diff
@@ -261,10 +262,12 @@ impl Workflows {
             activity_tasks_handle,
             wft_semaphore,
             local_act_mgr,
+            ever_polled: AtomicBool::new(false),
         }
     }
 
     pub(super) async fn next_workflow_activation(&self) -> Result<WorkflowActivation, PollWfError> {
+        self.ever_polled.store(true, atomic::Ordering::Release);
```
bergundy, line 270: Probably simpler to just have a worker option that tells the worker that it's an activity only worker.

bergundy: Replace `no_remote_activities` with something like `mode: All | Workflows | Actvities`

bergundy: We know this ahead of time since workflows are registered before worker construction in newer SDKs.

core/src/worker/mod.rs
```diff
@@ -357,6 +361,7 @@ impl Worker {
 
     /// Finish shutting down by consuming the background pollers and freeing all resources
     async fn finalize_shutdown(self) {
+        self.shutdown().await;
         if let Some(b) = self.at_task_mgr {
```
bergundy, line 365: We do this in `shutdown` already.

core/src/worker/mod.rs
```diff
@@ -340,15 +343,16 @@ impl Worker {
     /// completed
     async fn shutdown(&self) {
```
bergundy, line 344: We should just get rid of this `shutdown` method.
It's basically finalize at this point.

core/src/worker/workflow/mod.rs
Sushisource: That can be screwed up, though. I'd have to return a new error type from workflow polling or something like "You messed up and said you wouldn't do this".

I like this, it's simple, effective, and nothing new for lang.

core/src/worker/mod.rs
Sushisource: Yeah, I wanted to do that anyway. If we're fine with that then I'll get rid of it.

Sushisource: Though, that said, it's quite annoying to change for many of the tests. Many of them like to see when shutdown completes while doing something else, and that can't be done with finalize since it consumes.

It's probably worth me reconciling all that, though... but I'm not sure we want to do it right now as it's a decent chunk of work.

core/src/worker/mod.rs
Sushisource: Oh, you mean this bit with the task manager. This is different in some sense, it's consuming the activity manager.

bergundy: does it matter? We consume the entire worker here.

core/src/worker/workflow/mod.rs
bergundy: I'm fine with your approach if you prefer it this way.

bergundy (approved):

PR #507, opened 2023-03-13
Add support for metrics over UDS

ytaben:
Disclaimer:
I'm not a Rust developer and this draft is more of a proposal to check for two things:
1. Would this be valuable for the sdk-core? We are trying to use sockets in our organization and already do it for DogStatsd. Using TCP would require doing so at an org level.
2. Whether this is a sound implementation. It looks fairly reliable logic-wise but it does add one more code branch. Maybe there's a better way to do this more easily.

I also only focused on metrics because that's what we care about.

## What was changed
Handle 'unix' schema for metrics

## Why?
Currently, sdk-core only supports grpc over tcp while some collectors allow UDS usage.
Sadly `tonic` does not support connecting to `unix` schema natively.

## Checklist

1. Closes N/A

4. How was this tested:
Testing in progress (no longer getting invalid schema but have to verify metrics still reach destination)

5. Any docs updates needed?
No

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -62,6 +62,7 @@ tokio = { version = "1.1", features = ["rt", "rt-multi-thread", "parking_lot", "
 tokio-util = { version = "0.7", features = ["io", "io-util"] }
 tokio-stream = "0.1"
 tonic = { version = "0.8", features = ["tls", "tls-roots"] }
+tower = { version="0.4.13" }
 tracing = "0.1"
 tracing-futures = "0.2"
 tracing-opentelemetry = "0.18"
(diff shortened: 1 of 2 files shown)


Sushisource:
Thanks for this! I simplified it slightly and added some comments. As soon as you sign the CLI I can merge it.

ytaben:
@Sushisource 
Thank you for a quick response!

I've tested and saw that data _is_ being sent over the socket.
I'm not seeing these metrics on datadog's end (we use DD agent with OTel receiver) but it may be on our end given the data is being sent.

core/Cargo.toml
```diff
@@ -62,6 +62,7 @@ tokio = { version = "1.1", features = ["rt", "rt-multi-thread", "parking_lot", "
 tokio-util = { version = "0.7", features = ["io", "io-util"] }
 tokio-stream = "0.1"
 tonic = { version = "0.8", features = ["tls", "tls-roots"] }
+tower = { version="0.4.13" }
```
bergundy, line 65: Where do we use this?

bergundy (approved):
Thanks!

core/Cargo.toml
ytaben: mod.rs:253
`tower::service_fn`

Afaik I can't just use some sort of a lambda for it

PR #508, opened 2023-03-15
Fix possible incorrect history length value

Sushisource:
It was possible to hit a case where core would fail to realize it needed to fetch more events and instead apply partial history, with an incorrect length.

This fixes that.

### DIFF
--- a/.buildkite/docker/Dockerfile
+++ b/.buildkite/docker/Dockerfile
@@ -3,7 +3,7 @@ FROM rust:1.68
 RUN rustup component add rustfmt && \
 	rustup component add clippy
 
-RUN cargo install cargo-tarpaulin
+# RUN cargo install cargo-tarpaulin
 RUN apt-get update && apt-get install -y protobuf-compiler
 
 WORKDIR /sdk-core
--- a/.buildkite/docker/Dockerfile
+++ b/.buildkite/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.65
+FROM rust:1.68
 
 RUN rustup component add rustfmt && \
 	rustup component add clippy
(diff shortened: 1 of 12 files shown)


.buildkite/docker/Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM rust:1.65
+FROM rust:1.68
 
 RUN rustup component add rustfmt && \
 	rustup component add clippy
 
-RUN cargo install cargo-tarpaulin
+# RUN cargo install cargo-tarpaulin
```
bergundy, line 6: Did you leave this here intentionally?

core/src/worker/workflow/history_update.rs
```diff
@@ -225,31 +240,42 @@ impl HistoryPaginator {
...
-                // If next page fetching happened, and we still ended up with no events, something
-                // is wrong. We're expecting there to be more events to be able to extract this
-                // update, but server isn't giving us any. We have no choice except to give up and
-                // evict.
+            let seen_enough_events = current_events
+                .back()
+                .map(|e| e.event_id)
+                .unwrap_or_default()
+                >= self.wft_started_event_id;
+            if current_events.is_empty() || (no_next_page && !seen_enough_events) {
```
bergundy, line 250: I'm missing context to understand this condition.
Could use an explanation, I'll hit you up for that.

bergundy (approved):
Mostly blindly approving since I do not have enough context on this part of the codebase.

.buildkite/docker/Dockerfile
Sushisource: Ya, I want to bring it back if it works again - or swap it for a different one maybe.

PR #509, opened 2023-03-16
Revert "Add support for metrics over UDS"

ytaben:
Reverts temporalio/sdk-core#507

For context:
In additional testing I found that there is a bug in opentelemetry-otlp library.
It does not respect `with_channel` for _metrics_ as of now (and falls back to default endpoint)

I may try to fix it there if I have resources but I want to revert this for now to avoid the confusion I've gone through (I have a local bandaid but a proper fix needs much more polish).

To clarify: the traffic I saw in my initial testing turned out to be the connection opening with an immediate 'GOAWAY' frame (as the channel presumably gets immediately "collected").
I mistakenly thought that the traffic I saw initially contained metrics that we failed to ingest on _our_ end.

I will revisit this if\when opentelemetry-otlp has a fix for this.

Thank you again for a speedy response with the initial PR!


Sushisource (approved):
Ah, too bad. Thanks for coming back and mentioning this!

Issue #323, opened 2022-05-05
[Bug] Local activity cancelled failure not wrapped in same way as regular activity cancelled failure


cretz:
### Describe the bug

This is returned to workflow on try-cancel-based regular activity cancel:

```
jobs {
  resolve_activity {
    seq: 1
    result {
      cancelled {
        failure {
          message: "Activity cancelled"
          cause {
            canceled_failure_info {
            }
          }
          activity_failure_info {
            scheduled_event_id: 5
            activity_type {
              name: "wait_cancel"
            }
            activity_id: "1"
            retry_state: RETRY_STATE_CANCEL_REQUESTED
          }
        }
      }
    }
  }
}
```

But this is returned for local activity:

```
jobs {
  resolve_activity {
    seq: 1
    result {
      cancelled {
        failure {
          message: "Activity cancelled"
          canceled_failure_info {
          }
        }
      }
    }
  }
}
```

Need to just make those the same

cretz:
This same issue exists for LA timeout and maybe other LA exceptions, have not checked.

PR #517, opened 2023-03-21
Wrap local activity cancel & timeout failures the same way as normal activities

Sushisource:
## What was changed
LA cancel & timeout failures are now wrapped inside an activity failure, as with normal activities. This has the downside that the activity failure info is largely occupied by nonsense (started/scheduled IDs that don't apply and identity), which is annoying, but it keeps the structure consistent.

## Why?
Consistency

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/323
2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/machines/local_activity_state_machine.rs
+++ b/core/src/worker/workflow/machines/local_activity_state_machine.rs
@@ -683,25 +683,18 @@ impl WFMachinesAdapter for LocalActivityMachine {
                         ),
                     }
                 } else {
-                    match result {
-                        LocalActivityExecutionResult::Completed(c) => ActivityResolution {
-                            status: Some(c.into()),
-                        },
-                        LocalActivityExecutionResult::Failed(f)
-                        | LocalActivityExecutionResult::TimedOut(f) => ActivityResolution {
-                            status: Some(f.into()),
-                        },
-                        LocalActivityExecutionResult::Cancelled(mut cancel) => {
-                            // Cancels are to be wrapped with an activity failure
-                            let mut fail = cancel.failure.take();
+                    // Cancels and timeouts are to be wrapped with an activity failure
+                    macro_rules! wrap_fail {
+                        ($me:ident, $fail:ident, $msg:expr, $info:pat) => {
+                            let mut fail = $fail.failure.take();
                             let fail_info = fail.as_ref().and_then(|f| f.failure_info.as_ref());
-                            if matches!(fail_info, Some(FailureInfo::CanceledFailureInfo(_))) {
+                            if matches!(fail_info, Some($info)) {
                                 fail = Some(Failure {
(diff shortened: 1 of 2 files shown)


core/src/worker/workflow/machines/local_activity_state_machine.rs
```diff
@@ -117,6 +120,7 @@ impl From<CompleteLocalActivityData> for ResolveDat {
             result: match d.result {
                 Ok(res) => LocalActivityExecutionResult::Completed(Success { result: Some(res) }),
                 Err(fail) => {
+                    // TODO: or cause is cancelled - unpack properly
```
bergundy, line 123: Leftover TODO?

core/src/worker/activities/local_activities.rs
```diff
@@ -70,7 +74,18 @@ impl LocalActivityExecutionResult {
         Self::Cancelled(Cancellation::from_details(None))
     }
     pub(crate) fn timeout(tt: TimeoutType) -> Self {
-        Self::TimedOut(ActFail::timeout(tt))
+        Self::TimedOut(ActFail {
+            failure: Some(APIFailure {
+                message: "Activity timed out".to_string(),
+                failure_info: Some(failure::FailureInfo::TimeoutFailureInfo(
```
bergundy, line 80: AFAIU this should be an ActivityFailure with TimeoutFailure set as the cause.

Sushisource: Thanks - this isn't quite the right spot but I added it to the right place

sdk/src/workflow_context.rs
```diff
@@ -556,16 +556,16 @@ impl<'a> Future for LATimerBackoffFut<'a> {
...
-
             if let Some(activity_resolution::Status::Backoff(b)) = r.status.as_ref() {
+                // If we've already said we want to cancel, don't schedule the backoff timer. Just
+                // return cancel status. This can happen if cancel comes after the LA says it wants
+                // to back off but before we have scheduled the timer.
+                if self.did_cancel.load(Ordering::Acquire) {
+                    return Poll::Ready(ActivityResolution {
+                        status: Some(activity_resolution::Status::Cancelled(Default::default())),
+                    });
+                }
```
cretz, line 567: Any consequences to moving this under the conditional? Seems before we'd return cancel if cancel was requested whereas now we only will if cancel was requested _and_ it's backing off.

cretz (approved):

sdk/src/workflow_context.rs
Sushisource: This is in the Rust SDK, but yeah it was actually wrong before. This was only supposed to deal with early return in specifically the backoff case, and it returning a Default value was hiding real cancel reasons.

PR #518, opened 2023-03-24
Add support for global metric and tracing tags

### LINKED ISSUE #467: [Feature Request] Support custom tags/labels in TelemetryOptions

People need a way to set tags/labels to all metrics (global is the only current requirement, nothing lower level yet)

Support providing metrics tags/labels to be passed with `TelemetryOptions`

ytaben:
## What was changed

Add support for passing a map of tags for metrics and traces

## Why?

Currently - core SDK uses a single `service.name` tag (that cannot be overriden)
Many users, however, may want to supply their own (for example, to identify service or host running a worker or making client requests)

## Checklist

1. Closes #467

2. How was this tested:

Tested with a local otel collector and temporal SDK python.
Note: I used an older commit because python does not seem to compile with latest SDK.
Ensured a custom tag set in Python gets propagated to the collector (along service.name tag)

Did not test prometheus as we don't use it but can upon request

3. Any docs updates needed?
Probably not

We don't use opentelemetry so maybe variables aren't most aptly named so open to feedback on that (in datadog these are just tags)

Also, I added the functionality to both traces and metrics. I don't see why not do both but let me know if we want to do this for metrics-only.

### DIFF
--- a/core-api/src/telemetry.rs
+++ b/core-api/src/telemetry.rs
@@ -41,6 +41,10 @@ pub struct TelemetryOptions {
     /// Specifies the aggregation temporality for metric export. Defaults to cumulative.
     #[builder(default = "MetricTemporality::Cumulative")]
     pub metric_temporality: MetricTemporality,
+
+    // A map of tags to be applied to all metrics
+    #[builder(default)]
+    pub global_tags: HashMap<String, String>
 }
 
 /// Options for exporting to an OpenTelemetry Collector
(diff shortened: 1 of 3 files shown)


Sushisource (approved):
Cool! Thanks for this. Makes sense to me.

You'll want to run `cargo fmt` which is what the CI is complaining about.

ytaben:
Ah, though it was lint (tried clippy)
Sadly contributors can't see buildkite output

PR #522, opened 2023-03-29
Fix replay hanging on empty WFT complete @ end of history

Sushisource:
## What was changed
Fixed possible hang if replaying a history that ended in an empty WFT complete.

This had to do with the way WFT sequences work. The WFT sequence ends at started, and replay is considered "over" because there's no meaningful next sequence to handle (just an empty WFT complete doesn't count).

I fixed this by just checking if replay was over (rather than seeing if we hit the end of history) to determine if replay is... over (by that I mean the replay worker's replay).

## Why?
Hang bad.

## Checklist

1. Closes 

2. How was this tested:
Added UT

3. Any docs updates needed?


core/src/worker/workflow/managed_run.rs
```diff
@@ -988,7 +988,8 @@ impl ManagedRun {
         let should_respond = !(self.wfm.machines.has_pending_jobs()
             || outgoing_cmds.replaying
             || is_query_playback
-            || data.activation_was_only_eviction);
+            || data.activation_was_only_eviction
+            || self.wfm.machines.have_seen_terminal_event);
```
Sushisource, line 992: This was added because, now that replaying gets flipped to false upon seeing the workflow execution's terminal event, a couple unit tests that were asserting things about completions saw different results on those final completions.

Thing is, those completions are in "response" to impossible tasks from the server. Server will never send a WFT with a terminal event in it (except queries, which are permitted just below). So, now such completions are never "sent", which all probably makes more sense than it used to anyway.

core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -420,6 +420,8 @@ impl WorkflowMachines {
         // then we don't need to do anything here, and in fact we need to avoid re-applying the
         // final WFT.
         if self.have_seen_terminal_event {
+            // Replay clearly counts as done now, since we return here and never do anything else.
+            self.replaying = false;
```
bergundy, line 424: What's the implication of setting this to false?

Sushisource: See my other comment

Issue #523, opened 2023-03-29
[Bug] Cancelling a child workflow with CancellationType.Abandon after cancelling the parent leads to an invalid state


PenguinToast:
### What are you really trying to do?

We have a workflow that kicks off a child workflow that shouldn't have a lifecycle tied to the parent -- so we start it with
```
    parentClosePolicy: ParentClosePolicy.PARENT_CLOSE_POLICY_ABANDON,
    cancellationType: ChildWorkflowCancellationType.ABANDON,
```
We'd like to be able to cancel these workflows independently of each other -- and these workflows might run cleanup activities after being cancelled (so the parent workflow can continue running long enough after receiving a cancellation signal such that it can observe the cancellation of the child as well).

### Describe the bug

If we cancel the parent workflow first, and then cancel the child workflow before the parent workflow has finished cleanup, we end up with the following error:
```
WARN complete_workflow_activation: temporal_sdk_core::worker::workflow: Failing workflow task run_id=f163e4ca-6cb5-4043-b579-4562d4e10865 failure=Failure { failure: Some(Failure { message: "Fatal(\"ChildWorkflowMachine in state Cancelled says the transition is invalid during event EventInfo { event_id: 379, event_type: ChildWorkflowExecutionCanceled, has_next_event: false }\")", source: "", stack_trace: "", encoded_attributes: None, cause: None, failure_info: Some(ApplicationFailureInfo(ApplicationFailureInfo { r#type: "", non_retryable: false, details: None })) }) } completion=WorkflowActivationCompletion(run_id: f163e4ca-6cb5-4043-b579-4562d4e10865, status: Success())
```

Reading over the code, it looks like we put the child workflow state machine in state `Cancelled` once the parent workflow receives a cancellation signal:
https://github.com/temporalio/sdk-core/blob/edb8f4eb9e4ffbc0ec052659e37231460c4429e1/core/src/worker/workflow/machines/child_workflow_state_machine.rs#L294
Then, when the `ChildWorkflowExecutionCanceled` event comes in, there's no valid `Cancelled --(ChildWorkflowExecutionCancelled)` transition so the state machine errors.

I'm not familiar enough with this code to understand whether the right fix is to stop the child workflow from transitioning to `Cancelled` when the `CancellationType` is `Abandon`, or if it's to add a valid transition from `Cancelled --(ChildWorkflowExecutionCancelled) --> Cancelled`, or something else entirely -- happy to help out as needed to get a fix for this issue though!

[...]

PR #524, opened 2023-03-29
Fix bug where an abandoned child which was cancelled externally errors

Sushisource:
## What was changed
Fixed not handling a legitimate scenario where an abandoned child workflow which has been "cancelled" internally then processes an _external_ cancellation event and blows up. This scenario should be permitted.

## Why?
Bug

## Checklist

1. Closes  https://github.com/temporalio/sdk-core/issues/523

2. How was this tested:
Added integ test

3. Any docs updates needed?


Quinn-With-Two-Ns:
I think I fixed this exact same scenario in Go as well

Quinn-With-Two-Ns (approved):

cretz (approved):

PR #528, opened 2023-04-12
Allow overriding origin for core connections

Sushisource:
## What was changed
Allow overriding origin for connections

## Why?
When connecting via a proxy to cloud, just changing the domain for TLS verification is only part of the puzzle. Additionally, our stuff expects to see the right `:authority` and/or `host` headers set. Setting the origin value for the connection handles that.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -124,7 +124,8 @@ pub struct TlsConfig {
     /// Cloud offering).
     pub server_root_ca_cert: Option<Vec<u8>>,
     /// Sets the domain name against which to verify the server's TLS certificate. If not provided,
-    /// the domain name will be extracted from the URL used to connect.
+    /// the domain name will be extracted from the URL used to connect. Will also set the `host`
+    /// header key to this value if not otherwise specified by passed-in headers.
     pub domain: Option<String>,
     /// TLS info for the client. If specified, core will attempt to use mTLS.
     pub client_tls_config: Option<ClientTlsConfig>,
@@ -318,6 +319,16 @@ impl ClientOptions {
             })
             .service(channel);
         let headers = headers.unwrap_or_default();
+        // Explicitly set host header to tls domain name override if present
+        if let Some(tls_cfg) = self.tls_cfg.as_ref() {
+            if let (Some(dname), false) =
+                (tls_cfg.domain.as_ref(), headers.read().contains_key("host"))
+            {
+                headers
+                    .write()
+                    .insert("host".to_string(), dname.to_string());
+            }
+        }
         let interceptor = ServiceCallInterceptor {
             opts: self.clone(),
             headers: headers.clone(),
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -33,7 +33,7 @@ use crate::{
     workflow_handle::UntypedWorkflowHandle,
 };
 use backoff::{exponential, ExponentialBackoff, SystemClock};
-use http::uri::InvalidUri;
+use http::{uri::InvalidUri, Uri};
 use once_cell::sync::OnceCell;
 use parking_lot::RwLock;
 use std::{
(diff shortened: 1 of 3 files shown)


client/src/lib.rs
```diff
@@ -114,6 +114,11 @@ pub struct ClientOptions {
     /// Retry configuration for the server client. Default is [RetryConfig::default]
     #[builder(default)]
     pub retry_config: RetryConfig,
+
+    /// If set, override the origin used when connecting. May be useful in conjunction with
+    /// [TlsConfig::domain] when routing connections through some form of proxy.
+    #[builder(default)]
+    pub override_origin: Option<Uri>,
```
cretz, line 121: Often this would be just overriding authority (e.g. https://pkg.go.dev/google.golang.org/grpc#WithAuthority) and would not have to be a full URL with scheme and path and port and such. On lang side we hide the full client URL that Tonic uses in favor of a simple host:port. I wonder how this value will look lang side.

cretz (approved):
LGTM. One comment about how this should manifest lang side, but nothing blocking.

mjameswh:
I get the feeling we might be tackling this from the wrong side. Instead of exposing certificate and SNI host override, we may instead want to override the connection endpoint address (or gateway?). Essentially equivalent to a proxy in HTTP terminology. I believe this be closer to users understanding of the subject (ie. I conceptually connect to Temporal server X, but using endpoint Y).

mjameswh:
Also, given that this will affect some of our customers (ie. those connecting through Private Link and similars), we should involve Cloud and/or Infra teams.

client/src/lib.rs
Sushisource: Yeah, I can't do that unfortunately. I've cleaned it up so you don't need to specify this and we'll just construct it from the domain override.

Tonic doesn't let me set the authority header directly which is... annoying. I'll see about opening an item there for this.

Sushisource:
> Also, given that this will affect some of our customers (ie. those connecting through Private Link and similars), we should involve Cloud and/or Infra teams.

They are

client/src/lib.rs
```diff
@@ -361,6 +371,13 @@ impl ClientOptions {
 
             if let Some(domain) = &tls_cfg.domain {
                 tls = tls.domain_name(domain);
+
+                // This song and dance ultimately is just to make sure the `:authority` header ends
+                // up correct on requests while we use TLS. Setting the header directly in our
+                // interceptor doesn't work since seemingly it is overridden at some point by
+                // something lower level.
+                let uri: Uri = format!("https://{}", domain).parse()?;
+                channel = channel.origin(uri);
```
bergundy, line 380: Do we need to override this when `domain` is set? Can't we leave the origin untouched?

Sushisource: That's literally the whole fix

bergundy: Oh is it just for the case where `domain` is different than the URL?

Sushisource: Exactly

bergundy (approved):

tests/main.rs
```diff
@@ -58,35 +58,23 @@ mod integ_tests {
...
-        )
-        .await
-        .unwrap();
+        init_integ_telem();
+        let root = tokio::fs::read("../.cloud_certs/ca.pem").await.unwrap();
+        let client_cert = tokio::fs::read("../.cloud_certs/client.pem").await.unwrap();
+        let client_private_key = tokio::fs::read("../.cloud_certs/client.key").await.unwrap();
         let sgo = ClientOptionsBuilder::default()
-            .target_url(Url::from_str("https://localhost:7233").unwrap())
+            .target_url(Url::from_str("https://spencer.temporal-dev.tmprl.cloud:7233").unwrap())
```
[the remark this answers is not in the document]
Sushisource: I will make it work with our sdk namespace in CI in another PR.

client/src/lib.rs
bergundy: I see, we call it `serverNameOverride` in TS so I was confused by the name.

PR #531, opened 2023-04-13
 Ensure cancels get delivered after grace, even if poll stream not done

Sushisource:
## What was changed
Begin sending activity cancellations as soon as shutdown has initiated and the grace period has elapsed, even if the poller stream isn't yet terminated. This allows the cancellation of activities to proceed even if we are already at the maximum outstanding tasks permitted.

## Why?
Currently, we'd hang in a situation where there are long-running activities that need to be told to cancel while shutting down and all the task slots are eaten

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


core/src/worker/activities.rs
```diff
@@ -199,22 +199,21 @@ impl WorkerActivityTasks {
             starts_stream.map(ActivityTaskSource::from),
             |_: &mut ()| PollNext::Left,
         );
-        // Create a task stream composed of (in poll preference order):
-        //  cancels_stream ------------------------------+--- activity_task_stream
-        //  eager_activities_rx ---+--- starts_stream ---|
-        //  server_poll_stream  ---|
```
bergundy, line 205: I liked this comment :(

bergundy (approved):

PR #532, opened 2023-04-14
Fix internal flags not always being written when they should be

Sushisource:
## What was changed
Internal flags weren't always written in the first WFT completion where you might actually care about that. Rather than littering the code with a bunch of accesses just to make sure that it always shows up where needed, given these are just ints, I'm going back on what I thought earlier where writing them lazily is best. Seems totally fine to just spit out all the known ones on the first non-replay WFT.

Also restructured the flags structure a bit to be somewhat safer.

## Why?
Right now if you, ex, have a workflow with one activity which sched/starts/completes in one WFT, determinism type checking will not work for it, since the flag wasn't written until the scheduled event is first seen.

## Checklist

1. Closes 

2. How was this tested:
Added tests, changed existing tests

3. Any docs updates needed?


cretz:
Mind updating api_upstream as part of this to get https://github.com/temporalio/api/pull/270?

Sushisource:
> Mind updating api_upstream as part of this to get [temporalio/api#270](https://github.com/temporalio/api/pull/270)?

Done

core/src/internal_flags.rs
```diff
@@ -28,83 +32,125 @@ pub(crate) enum CoreInternalFlags {
...
     /// [Self::gather_for_wft_complete].
     pub fn try_use(&mut self, core_patch: CoreInternalFlags, should_record: bool) -> bool {
-        if !self.enabled {
+        match self {
+            Self::Enabled {
+                core,
+                core_since_last_complete,
+                ..
+            } => {
+                if should_record {
```
mjameswh, line 92: I don't understand the reasoning here. Shouldn't we check if the flag already exists _first_? As it is now, a patch might be registered multiple time...

Sushisource: The dedupe just happens on the way out, rather than the way in - in `gather_for_wft_complete`

mjameswh: Clever. Might be worth adding a few comments as this is definitely not obvious, but not blocking this PR.

mjameswh (approved):
LGTM

bergundy (approved):
Didn't review, just stamping to get codeowner approval.
@mjameswh already reviewed this.

PR #537, opened 2023-04-25
Fix possibly dropping final events in a WFT during cache miss

Sushisource:
## What was changed
It was possible to drop events that were in a WFT but not in the history fetch request in the event that pagination ends up being required.

## Why?
Bug

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/sdk/src/workflow_context.rs
+++ b/sdk/src/workflow_context.rs
@@ -11,7 +11,7 @@ use crate::{
     SignalExternalWfResult, TimerResult, UnblockEvent, Unblockable,
 };
 use crossbeam::channel::{Receiver, Sender};
-use futures::{task::Context, FutureExt, Stream};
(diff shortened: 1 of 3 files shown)


sdk/src/workflow_context.rs
```diff
@@ -398,6 +398,14 @@ impl DrainableSignalStream {
         }
         signals
     }
+
+    pub fn drain_ready(&mut self) -> Vec<SignalData> {
```
mjameswh, line 402: Intentionnal?

mjameswh (approved):

sdk/src/workflow_context.rs
Sushisource: Yeah, it's useful for something else I didn't include in the commit but is also useful generally, might as well throw it in

bergundy (approved):

PR #538, opened 2023-04-26
Fix more pagination insanity

Sushisource:
## What was changed
More fixes for everyone's favorite spot of code. This time, we handle the fact that we could consider workflow task sequences as complete if we just happen to have history which ends on a WFT started event - but that's not sufficient. If we know there might be more events, we need to peek beyond the started event to see if the next event is a task failure, or a heartbeat, etc.

## Why?
Bugs!

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/history_update.rs
+++ b/core/src/worker/workflow/history_update.rs
@@ -1,4 +1,5 @@
 use crate::{
+    abstractions::dbg_panic,
     protosext::ValidPollWFTQResponse,
     worker::{
         client::WorkerClient,
@@ -329,6 +330,9 @@ impl HistoryPaginator {
             }
             self.id_of_last_event_in_last_extracted_update =
                 update.events.last().map(|e| e.event_id);
+            if cfg!(debug_assertions) {
+                update.assert_contiguous();
+            }
             return Ok(update);
         }
     }
@@ -460,6 +464,21 @@ impl HistoryUpdate {
     }
     pub fn first_event_id(&self) -> Option<i64> {
         self.events.get(0).map(|e| e.event_id)
+    }
+
+    #[cfg(debug_assertions)]
+    fn assert_contiguous(&self) -> bool {
+        for win in self.events.as_slice().windows(2) {
+            match &win {
+                &[e1, e2] => {
+                    if e2.event_id != e1.event_id + 1 {
+                        dbg_panic!("HistoryUpdate isn't contiguous! {:?} -> {:?}", e1, e2);
+                    }
+                }
+                _ => {}
+            }
+        }
+        true
     }
 
     /// Create an instance of an update directly from events. If the passed in event iterator has a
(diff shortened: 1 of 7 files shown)


core/src/worker/workflow/history_update.rs
```diff
@@ -698,13 +713,21 @@ fn find_end_index_of_next_wft_seq(
                     if let Some(next_next_event) = events.get(ix + 2) {
                         if next_next_event.event_type() == EventType::WorkflowTaskScheduled {
                             continue;
-                        } else {
-                            saw_any_non_wft_event = true;
+                        } else if next_next_event.is_command_event() {
+                            saw_any_command_event = true;
```
mjameswh, line 717: I'm confused by the utility of `saw_any_command_event` in this logic.

Suppose that we get here and set saw_any_command_event to true, assuming it was false before. That means that `ix` is pointing to a `WorkflowTaskStarted`, followed by `WorkflowTaskCompleted`, followed by some command event. That command relates to the `WFTCompleted` at `ix + 1`. We'll end up returning a `NextWFTSeqEndIndex::Complete` at `ix` (ie. on the `WFTStarted`), which means that the sequence may actually contain no command at all.

Now, we could instead have found a command _before_ the `WFTStarted` event (ie. attached to a previous `WFTCompleted`). In that case, we'll end up returning a `NextWFTSeqEndIndex::Complete` at `ix` (ie. on the `WFTStarted`), which means that the sequence will contain those commands.

Why is that needed? And why specifically 'commands' (vs 'signals', 'timer fired', 'activity completed', etc)?

Sushisource: Yeah, so this is the intent. The "next sequence" includes the last completion, it's commands, and any non-command things that happened after than until the next WFT started. So in your two examples, things actually end up working as we want them to.

The reason the command distinction matters is that it allows us to avoid accidentally categorizing a WFT that ends in failure as being a real WFT sequence because it happens to contain a signal, for example. This is a bit closer to how the Java code works for it's version of this.

mjameswh: > Yeah, so this is the intent. The "next sequence" includes the last completion, it's commands, and any non-command things that happened after than until the next WFT started. So in your two examples, things actually end up working as we want them to.

Ok... Then what if I get some non-commands (let say incoming signals), followed by a `WorkflowTaskStarted`, followed by WorkflowTaskCompleted, no command, some more signals, and then again, a `WorkflowTaskStarted` (`ix` is pointing on that one), `WorkflowTaskCompleted`, no command, signals...

That means we won't split at `ix` in this case, even though we technically accumulated a complete WFT at that point. Can this be problematic in some cases? Shouldn't we simply send a `Complete` at line 717, now that we have ruled out the possibility that the WFT starting at `ix+1` could be either a failure/timeout or a heartbeat?

Sushisource: Hmm, I thought that particular case still worked properly, but I confirmed it doesn't. Indeed, that's not right. Will update.

Sushisource:
Ugh, a suspicion I had that this could cause nonsense orderings is true. Fixing that.

bergundy (approved):
Blindly approving so this isn't blocked. @mjameswh is reviewing.

mjameswh (approved):

Issue #539, opened 2023-05-01
[Feature Request] Workflow definitions


h7kanna:
### Is your feature request related to a problem? Please describe.

Create workflow definitions in idiomatic Rust code similar to other SDKs.

### Describe the solution you'd like

Extend the SDK with abstractions for creating type-safe workflow definitions with API similar to other SDKs, but whatever possible within Rust.

Issue #222, opened 2021-11-15
Rust SDK Prototype Release


Sushisource:
We've had a number of people ask about Rust SDK progress. This issue serves for now as a place to track that progress.

Currently there is **no official timeline for releasing a Rust SDK**. However, I do plan to make the prototype SDK publicly exposed with **no API compatibility guarantees**. This issue will track that. It directly depends on #176

bdbelevate:
Any thoughts on this? I can see you guys are working on it, but if there's a specific timeline to it or ways that we can help contribute, let me know. Thanks!

Sushisource:
> Any thoughts on this? I can see you guys are working on it, but if there's a specific timeline to it or ways that we can help contribute, let me know. Thanks!

Still no timelines around official release. However, if your question is more "can you just publish this to crates.io?" then yes, I can prioritize that so at least it's a bit easier to use, if still alpha.

bdbelevate:
Thanks for the update. I don't mind forking and keeping our own version to use locally or within our own cargo workspace. But in trying to us it, what I do see is that it _seems_ as if the rust SDK has no way to send a response to temporal for a workflow. I don't see any tests where this happens and it also seems like the exit for a workflow must be `WorkflowResult<()>`. I'm not sure if this is intentional or just upcoming changes. Maybe I'm missing something but it appears the typescript sdk gets around this by sending it's own grpc result.

Sushisource:
That's correct. The workflow authoring APIs in general are still very rudimentary in the Rust SDK. The underlying functionality is there (that's what typescript uses), it's just not all exposed yet.

We'll want to publish a proposal to our proposals repo to get some ideas around what we want a final workflow API to look like in the Rust SDK before putting in the work to polish it up. It'll definitely happen but other things are higher priority right now.

There seems to be enough demand generally that getting discussions going around that proposal might at least be good to get started though. So, I'll see about finding some time for that and will update here with anything relevant.

bdbelevate:
Thanks - that would be awesome. Like I mentioned our team would be happy to help contribute once there's a proposal.

Sushisource:
I forgot to mention in this issue that the first draft of the workflow api is here https://github.com/temporalio/sdk-core/pull/550

mkarbo:
Hello! I was just wondering if anything has changed on your side regarding commitment to a stable SDK release for rust?

[...]

PR #540, opened 2023-05-01
WIP: Workflow defintions

h7kanna:
## What was changed

1) Modified SDK to handle 'Payload' as the return value of workflow function instead of an empty tuple.
2) Minor changes to SDK workflow and activity functions to accommodate generic user-defined functions.
3) Added 'workflow' module to create abstractions to define workflow definitions in idiomatic Rust(Async)
4) Added 'prelude' module for easy importing of types in workflow definitions.
5) Updated tests

I use Rust a lot and I write workflows a lot and a lot.  So created the abstraction functions needed to define workflows similar to other SDKs (Go). Submitting this PR to receive feedback on design and approach.

TODO

- [ ] Factor in the feedback
- [ ] Move the repeating and boilerplate into Macros once the API is good enough
- [ ] Create more convenient functions similar to other SDKs. 
- [ ] Add more examples

## Why?

A step towards https://github.com/temporalio/sdk-core/issues/222

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/539

2. How was this tested:
Updated tests and added new tests

3. Any docs updates needed?
Updated crate documentation and added new module documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ You can buld and test the project using cargo:
 Run integ tests with `cargo integ-test`. You will need to already be running the server:
 `docker-compose -f .buildkite/docker/docker-compose.yaml up`
 
+Run load tests with `cargo test --features=save_wf_inputs --test heavy_tests`.
+
 ## Formatting
 To format all code run:
 `cargo fmt --all`
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -95,6 +95,8 @@ bimap = "0.6.1"
 clap = { version = "4.0", features = ["derive"] }
 criterion = "0.4"
 rstest = "0.17"
+serde = "1.0"
+serde_json = "1.0"
 temporal-sdk-core-test-utils = { path = "../test-utils" }
 temporal-sdk = { path = "../sdk" }
 
--- a/core/benches/workflow_replay.rs
+++ b/core/benches/workflow_replay.rs
@@ -1,9 +1,9 @@
 use criterion::{criterion_group, criterion_main, Criterion};
 use futures::StreamExt;
 use std::time::Duration;
-use temporal_sdk::{WfContext, WorkflowFunction};
+use temporal_sdk::{WfContext, WfExitValue, WorkflowFunction};
 use temporal_sdk_core::replay::HistoryForReplay;
-use temporal_sdk_core_protos::DEFAULT_WORKFLOW_TYPE;
+use temporal_sdk_core_protos::{coresdk::AsJsonPayloadExt, DEFAULT_WORKFLOW_TYPE};
 use temporal_sdk_core_test_utils::{canned_histories, replay_sdk_worker};
 
 pub fn criterion_benchmark(c: &mut Criterion) {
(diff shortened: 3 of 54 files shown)


Sushisource:
Thanks! I really appreciate your work here. Adding support for returning Payloads from workflows is something I think we could take first and separately, which would be great. More on this after.

As for the workflow & activity function abstractions, I think the direction we've had in mind is a bit different. You mentioned converting things to macros later, I think that's something we'll want to leverage, and I've also mulled over the idea of using a trait-based implementation for workflows. Big picture wise, we want to make sure we've nailed down an API that aligns with what we want a future Rust SDK to look like. Importantly, we want to make sure that once we do start moving forward on a more complete Rust SDK that it aligns there, and it's something we're going to be comfortable supporting in the long run. In fact, the main reason we haven't moved forward here is mainly a question of support rather than implementation.

So all that in mind, at the moment I think the kind of PRs I'm happy to merge at the moment are ones that target specific bits of missing functionality.

When we do feel that we have the bandwidth to move forward with a non-prototype API we'll start with a proposal here: https://github.com/temporalio/proposals

---
Re: Separating out returning payloads from workflow functions. It'd be great to add this. In particular, what I had in mind was being able to return anything that `impl TemporalSerializable`, or something similar, where:

```rust
pub trait TemporalSerializable {
    fn into_payload(self) -> anyhow::Result<Payload>
}

impl<T> TemporalSerializable for T where T: serde::Serialize {
    // blah
}
```

This way, the return site in the workflow need not do any conversion directly, and this can fit into a data converter type facility later on like the other SDKs have.


Thanks again! Let me know if you have any questions

h7kanna:
Hi @Sushisource, Thanks.

The reason I opened the PR is to get the first commit(Payloads) accepted and merged so that I can iterate on the Workflow abstractions without constant rebasing.

I will create a separate PR for it and make use of the TemporalSerializable you suggested.

Regarding the Workflow definitions, I will keep on improving them as I need some boilerplate reduction for my project which is a lot of Rust workflows.

h7kanna:
Hi @Sushisource,

On second thought, It might take a while to design for return 'impl TemporalSerializable' instead of Payload from the workflow function. At the same time, I want to keep on working on abstractions for my project workflows.

I will break this into 3 parts.
1) Return Payload from workflow functions (PR-1)
2) Modify the Activity function similar to Workflow Function and accept Vec<Payload>. Need this for the abstractions. Will help with rebasing as a lot of unit tests are updated (PR-2)
3) Figure returning the TemporalSerializable(PR-3)
3) I will continue working on workflow definitions in my fork and close this one.

Is it ok with you?

Sushisource:
I'd like to have the PR introducing `Payload` go straight to the more complete abstraction, largely because it allows avoiding touching almost every single test file to add a fake returned payload. With the complete abstraction, they might still need to be touched, but it'll be to _remove_ an `into()` call rather than adding more which would be great.

Supporting sending multiple payloads when scheduling activities makes sense, and is eventually needed regardless, so yeah I'm happy to have that addition too. Related, I'm still not sure I _ever_ want to support anything other than 0 and 1 arity workflow and activity definitions in Rust. It's generally an anti-pattern to take more than one argument in your definitions because it makes maintaining input & output compatibility difficult. We need to support scheduling activities and child WFs with 2+ arities, because you might be targeting things defined in other languages, but my inclination is to avoid defining them entirely in a Rust SDK.

h7kanna:
Closing this and continue in parts in https://github.com/temporalio/sdk-core/pull/542

PR #542, opened 2023-05-01
Updated workflow function to return Payload instead of empty tuple

h7kanna:
## What was changed

1) Modified SDK to handle 'Payload' as the return value of workflow function instead of an empty tuple.
2) Updated tests

## Why?

A step towards https://github.com/temporalio/sdk-core/issues/222

## Checklist

1. This part of work in https://github.com/temporalio/sdk-core/issues/539

2. How was this tested:
Updated tests and added new tests

3. Any docs updates needed?
No

### DIFF
--- a/core/src/core_tests/local_activities.rs
+++ b/core/src/core_tests/local_activities.rs
@@ -21,7 +21,8 @@ use std::{
 };
 use temporal_client::WorkflowOptions;
 use temporal_sdk::{
-    ActContext, ActivityCancelledError, LocalActivityOptions, WfContext, WorkflowResult,
+    ActContext, ActivityCancelledError, LocalActivityOptions, WfContext, WorkflowFunction,
+    WorkflowResult,
 };
 use temporal_sdk_core_api::{
     errors::{PollActivityError, PollWfError},
@@ -964,17 +965,18 @@ async fn resolved_las_not_recorded_if_wft_fails_many_times() {
     mh.num_expected_completions = Some(0.into());
     let mut worker = mock_sdk_cfg(mh, |w| w.max_cached_workflows = 1);
 
+    #[allow(unreachable_code)]
     worker.register_wf(
         DEFAULT_WORKFLOW_TYPE.to_owned(),
-        |ctx: WfContext| async move {
+        WorkflowFunction::new::<_, _, ()>(|ctx: WfContext| async move {
             ctx.local_activity(LocalActivityOptions {
                 activity_type: "echo".to_string(),
                 input: "hi".as_json_payload().expect("serializes fine"),
                 ..Default::default()
             })
             .await;
-            panic!("Oh nooooo")
-        },
+            panic!()
+        }),
     );
     worker.register_activity(
         "echo",
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ You can buld and test the project using cargo:
 Run integ tests with `cargo integ-test`. You will need to already be running the server:
 `docker-compose -f .buildkite/docker/docker-compose.yaml up`
 
+Run load tests with `cargo test --features=save_wf_inputs --test heavy_tests`.
+
 ## Formatting
 To format all code run:
 `cargo fmt --all`
--- a/core/src/worker/workflow/managed_run/managed_wf_test.rs
+++ b/core/src/worker/workflow/managed_run/managed_wf_test.rs
@@ -61,7 +61,7 @@ pub struct ManagedWFFunc {
     activation_tx: UnboundedSender<WorkflowActivation>,
     completions_rx: UnboundedReceiver<WorkflowActivationCompletion>,
     completions_sync_tx: crossbeam::channel::Sender<WorkflowActivationCompletion>,
-    future_handle: Option<JoinHandle<WorkflowResult<()>>>,
+    future_handle: Option<JoinHandle<WorkflowResult<Payload>>>,
     was_shutdown: bool,
 }
 
@@ -175,7 +175,7 @@ impl ManagedWFFunc {
         Ok(last_act)
     }
 
-    pub async fn shutdown(&mut self) -> WorkflowResult<()> {
+    pub async fn shutdown(&mut self) -> WorkflowResult<Payload> {
         self.was_shutdown = true;
         // Send an eviction to ensure wf exits if it has not finished (ex: feeding partial hist)
         let _ = self.activation_tx.send(create_evict_activation(
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -65,6 +65,7 @@ use crate::{
 use anyhow::{anyhow, bail, Context};
 use app_data::AppData;
 use futures::{future::BoxFuture, FutureExt, StreamExt, TryFutureExt, TryStreamExt};
+use serde::Serialize;
 use std::{
     any::{Any, TypeId},
     cell::RefCell,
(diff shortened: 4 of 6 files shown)


Sushisource:
Thanks! Per my comment in https://github.com/temporalio/sdk-core/pull/540 I think I'd prefer it if we went straight to the more general abstraction here :smile:

h7kanna:
Sure, I will work towards updating this. Closing for now.

h7kanna:
Hi @Sushisource,

I am actually scratching my head to understand why you are asking for the Payload abstraction.  :grin:

I designed the workflow module to make sure we can return any type as long as it implements `Serialize`. So in my workflow code, I am just returning structs wrapped in WfExitValue without manual payload conversions.

Indeed I just postponed implementing the Payload converter as well in the workflow module.

The thing is in all the tests, Workflow functions are created manually using Workflow. new, wherein I am just using the `worker.register` function with a [wrapper](https://github.com/h7kanna/sdk-core/blob/workflow-definitions/sdk/src/workflow.rs#L740) to convert any supported function to BoxedFuture that the WorkflowFunction needs. :grimacing:

I just need to update the `WorkflowFunction.new` to match the wrapper :relieved:
Also, as a bonus, I will go ahead and modify the wrapper to use the generic payload converter trait instead of the `AsJsonPayloadExt` and `FromJsonPayloadExt` I am using now.

Sushisource:
I'm a bit confused - very few of the tests (3 by my count) directly use `WorkflowFunction::new`, most are just using `register_wf` directly.

What you have looks good to me now in terms of not touching many files - the main reason to introduce an abstraction is the `JsonPayloadExt` traits are not meant to be the end-goal for what workflow/activity functions can return. They should be able to return anything `Serialize`able, and eventually that can get hooked up to a payload converters concept.

So for now we could do that, just make the bounds [here](https://github.com/temporalio/sdk-core/pull/542/files#diff-d8b2bc9f3303d943e5be413a2c54965e6faa5836ccd597d12fa2a557df826608R713) be `Serialize` instead.

Eventually `WorkflowFuture` itself needs to be generic over the output type, rather than always being `WorkflowResult<Payload>`, but I'm fine with that side of things just giving you `Payload`s for now.

core/src/core_tests/local_activities.rs
```diff
@@ -973,7 +973,8 @@ async fn resolved_las_not_recorded_if_wft_fails_many_times() {
                 ..Default::default()
             })
             .await;
-            panic!("Oh nooooo")
+            let _ = "panic".parse::<i32>().unwrap();
+            Ok(().into())
```
Sushisource, line 977: This test needs to stay as a panic, it's important to the goal of the test. You'll have to just add a type hint for the closure's return type.

h7kanna: I tried a lot to replace 'Payload' the Workflow Future with a trait before starting the discussion 😄.
It ends with the requirement of WorkflowFunction becoming generic and then bounds won't add up correctly and the From implementation fails with 'unconstrained' bounds etc. I will try again though.

Regarding the Payload being generic. I changed the trait a bit for now [here](https://github.com/temporalio/sdk-core/commit/bb612333dc4cdc11e4db4999e2109fb9a00f8659), next PR. I think we need to pass the Payload converter into both WfContext and ActContext and use it in a wrapper function like [this](https://github.com/h7kanna/sdk-core/blob/payload-converter/sdk/src/lib.rs#L765).

I will work on how we can make this dynamic without help from the generics. Currently looks like this

```move |ctx: WfContext| async move {
    let activity = ActivityOptions {
        activity_id: Some(activity_id.to_string()),
        ..Default::default()
    };
    let result = ctx.activity(activity).await.unwrap_ok_payload();
    let result = String::from_payload(None, &result).expect("serializes fine");
    Ok(WfExitValue::Normal(result))
}
```

h7kanna:
I tried a lot to replace 'Payload' the Workflow Future with a trait before starting the discussion 😄.
It ends with the requirement of WorkflowFunction becoming generic and then bounds won't add up correctly and the From implementation fails with 'unconstrained' bounds etc. I will try again though.

Regarding the Payload being generic. I changed the trait a bit for now [here](https://github.com/temporalio/sdk-core/commit/bb612333dc4cdc11e4db4999e2109fb9a00f8659), next PR. I think we need to pass the Payload converter into both WfContext and ActContext and use it in a wrapper function like [this](https://github.com/h7kanna/sdk-core/blob/payload-converter/sdk/src/lib.rs#L765).

I will work on how we can make this dynamic without help from the generics. Currently looks like this

```move |ctx: WfContext| async move {
    let activity = ActivityOptions {
        activity_id: Some(activity_id.to_string()),
        ..Default::default()
    };
    let result = ctx.activity(activity).await.unwrap_ok_payload();
    let result = String::from_payload(None, &result).expect("serializes fine");
    Ok(WfExitValue::Normal(result))
}
```

core/src/core_tests/local_activities.rs
h7kanna: Type inference won't work and the code becomes unreachable after `panic!()`. So panicking with unwrap

h7kanna:
Though, the end result with wrapper functions the workflow looks like this, without any manual payload processing

```
pub async fn workflow_shell(ctx: WfContext, _input: (String,)) -> Result<ExecOutput, Error> {
    let activity_timeout = Duration::from_secs(5);
    let output = execute_activity_1_args_with_errors(
        &ctx,
        ActivityOptions {
            activity_id: Some("shell_activity".to_string()),
            activity_type: "activity_shell::execute".to_string(),
            ..Default::default()
        },
        activity_shell::exec_command2,
        activity_shell::ExecInput {
            command: "sort".to_string(),
            ..Default::default()
        },
    )
    .await;
    match output {
        Ok(output) => Ok(output),
        Err(e) => Err(e),
    }
}
```

core/src/core_tests/local_activities.rs
Sushisource: Ah right, I missed the unwrap - regardless this can keep the more obvious panic by providing an explicit return type for the closure.

Sushisource:
Cool, for the purposes of this PR, I think just changing the bounds [here](https://github.com/temporalio/sdk-core/pull/542/files#diff-d8b2bc9f3303d943e5be413a2c54965e6faa5836ccd597d12fa2a557df826608R713) be Serialize instead would be sufficient, and keeps things pretty much the same but without explicitly coupling things to JSON.

As for the further generic-ification, I like what you have in your final example there, but I also want to avoid any explicit arity-based wrapper functions. More likely than not, I don't think we'll be merging anything that makes changes that substantial until we're prepared to really start investing in productionizing and supporting the Rust SDK.

That said, your interest has been great, and in the spirit of us being transparent I'm going to make some time to at least publish a [proposal](https://github.com/temporalio/proposals) outlining the direction I want to go with the Rust SDK, even though we still may not actually invest in the implementation immediately.

h7kanna:
Hi,

The multiple arity functions are just wrappers in my code. They will not be in the SDK. But I would need the possibility of passing Vec<Payload> in the ActivityOptions. Are you okay with that?

h7kanna:
Thanks for deciding to spend time on the proposal 😄 . I can use that for direction instead.

core/src/core_tests/local_activities.rs
h7kanna: Fixed

Sushisource (approved):
Sweet, thanks!

PR #543, opened 2023-05-04
Updated Activity/Workflow Function bounds to allow a more generic use…

h7kanna:
## What was changed
1) Updated workflow and activity functions to take 0 or 1 arguments and accommodate generic functions as extensions (loosen up bounds)

2) Unfortunately, it touches activity tests, but they are only updating the function signature, all tests pass including load tests.

## Why?
To enable creating a type-safe API using more generic user functions. 
For example, a user function that does not return `anyhow::Error`

With this, we can register
1) 0 arg function similar to how it is done currently

```
worker.register_activity("echo", |_: ActContext| async {
    Result::<ActExitValue<()>, _>::Err(anyhow!("Oh no I failed!"))
});
```

2) 1 arg function as shown below, similar to how workflow function is registered currently.

```
worker.register_activity(
    "echo_activity",
    ActivityFunction::from(|_ctx: ActContext, echo_me: String| async move { Ok(echo_me) }),
);
```

## Checklist

1. Closes 

2. How was this tested:
1) Updated tests
2) Ran unit tests
3) Ran integration tests
4) Ran heavy tests(load)
All tests passed.

3. Any docs updates needed?
Updated function documentation

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -9,7 +9,7 @@
 //! An example of running an activity worker:
 //! ```no_run
 //! use std::{str::FromStr, sync::Arc};
-//! use temporal_sdk::{sdk_client_options, ActContext, Worker};
+//! use temporal_sdk::{sdk_client_options, ActContext, ActExitValue, ActivityFunction, Worker};
 //! use temporal_sdk_core::{init_worker, Url, CoreRuntime};
 //! use temporal_sdk_core_api::{worker::WorkerConfigBuilder, telemetry::TelemetryOptionsBuilder};
 //!
@@ -32,7 +32,7 @@
 //!     let mut worker = Worker::new_from_core(Arc::new(core_worker), "task_queue");
 //!     worker.register_activity(
 //!         "echo_activity",
-//!         |_ctx: ActContext, echo_me: String| async move { Ok(echo_me) },
+//!         ActivityFunction::from(|ctx: ActContext, echo: String| async move { Ok(echo) }),
 //!     );
 //!
 //!     worker.run().await?;
@@ -207,17 +207,14 @@ impl Worker {
 
     /// Register an Activity function to invoke when the Worker is asked to run an activity of
     /// `activity_type`
-    pub fn register_activity<A, R, O>(
+    pub fn register_activity(
         &mut self,
         activity_type: impl Into<String>,
-        act_function: impl IntoActivityFunc<A, R, O>,
+        act_function: impl Into<ActivityFunction>,
     ) {
-        self.activity_half.activity_fns.insert(
-            activity_type.into(),
-            ActivityFunction {
-                act_func: act_function.into_activity_fn(),
-            },
-        );
+        self.activity_half
+            .activity_fns
+            .insert(activity_type.into(), act_function.into());
     }
 
     /// Insert Custom App Context for Workflows and Activities
@@ -477,7 +474,7 @@ impl ActivityHalf {
                 self.task_tokens_to_cancels
                     .insert(task_token.clone().into(), ct.clone());
 
-                let (ctx, arg) = ActContext::new(
+                let ctx = ActContext::new(
                     worker.clone(),
                     app_data,
                     ct,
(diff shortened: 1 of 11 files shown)


h7kanna:
Hi @Sushisource, This is the last PR I will submit until further update from you on the proposal.
This will enable me to work independently of the SDK in my fork without any large rebasing in the future.
Changes are in line with what you have suggested in the previous comments till now, with only 0/1 arguments API in the SDK.

Thanks for taking the time, I do know this is not planned work for you.

sdk/src/lib.rs
```diff
@@ -743,6 +740,39 @@ impl WorkflowFunction {
             }),
         }
     }
+
+    /// Build a workflow function from a closure or function pointer which accepts a [WfContext] and takes one argument
+    pub fn from<A, F, Fut, R, O>(f: F) -> Self
```
Sushisource, line 745: So ideally both this and the activity version are implemented generically (as actual `From` impls) for anything that matches these bounds -- allowing us to avoid `XXXFunction::from` calls at definition sites, since the registration functions already accept `impl Into<XXXFunction>`. I think this is a really nice piece of ergonomics to maintain.

Sushisource (commented):
Sorry it took me a bit to get around to this

sdk/src/lib.rs
h7kanna: Hi, Sorry, did not understand this comment completely.

Sushisource: The idea would be to do something like this:

```
impl< ... generics ... > From<F> where: bounds for ActivityFunction {
   ...
}
```
For both activity and wf functions

So there's no need to call `ActivityFunction::from` at the same place you're defining or registering the function - everything just slots in.

h7kanna: Haa,

I tried to do that, but the From implementation cannot be implemented for function with generics, it fails with '[unconstrained](https://stackoverflow.com/questions/68639821/unconstrained-type-parameter-for-impl-fromstruct-for-t-when-impl-struct) bounds'

So the only From implementation is the one [existing](https://github.com/temporalio/sdk-core/blob/master/sdk/src/lib.rs#L710). Activity/Workflow Function that takes only Context.

For functions that take an argument, We need the `ActivityFunction::from/WorkflowFunction::new` [wrapper](https://github.com/temporalio/sdk-core/blob/master/core/src/core_tests/child_workflows.rs#L103).

But in the production SDK, we will actually not use these functions(I think). There should be higher-level 'execute_activity' or 'execute_childworkflow' functions that exist in other SDKs but that check the types of the arguments like [this](https://github.com/h7kanna/sdk-core/blob/workflow-definitions/tests/integ_tests/workflow_tests/examples.rs#L296) 

Essentially, the PR is to allow any function that can be converted into 

```
dyn Fn(WfContext) -> BoxFuture<'static, Result<WfExitValue<Payload>, anyhow::Error>>
    + Send
    + Sync
    + 'static
```
or
```
    dyn Fn(ActContext) -> BoxFuture<'static, Result<ActExitValue<Payload>, anyhow::Error>>
        + Send
        + Sync,
```

Sushisource: So this https://github.com/temporalio/sdk-core/blob/master/sdk/src/lib.rs#L840 implements an arg-accepting automatic implementation for activities - and, yeah, clearly we can't make that work for both 0 and 1 input-arity functions without some extra boilerplate or macros, etc. That said, there's nothing preventing implementing `IntoActivityFunc` for 0 and 1 input-arity functions and having this all work properly. Same could be done with workflows.

I think that'd be the best place to leave things for now, no need to worry about higher-arity functions at the moment.

You can see a preview of what I'm thinking of for later here: https://github.com/temporalio/sdk-core/pull/550 which will lean on procmacros for making this all seamless.

h7kanna:
Closed this as per [Comments](https://github.com/temporalio/sdk-core/pull/543#discussion_r1192684019)

sdk/src/lib.rs
h7kanna: I will look up the proposal and adapt this accordingly.

PR #544, opened 2023-05-08
Lang metrics export

Sushisource:
This takes a crack at what it will look like to export metrics to lang.

DRAFT since it's still subject to change, though the existing implementation being backed by the new interface does work.

### DIFF
--- a/core-api/src/telemetry.rs
+++ b/core-api/src/telemetry.rs
@@ -10,6 +10,8 @@ use std::{
 use tracing_core::Level;
 use url::Url;
 
+pub static METRIC_PREFIX: &str = "temporal_";
+
 /// Each core runtime instance has a telemetry subsystem associated with it, this trait defines the
 /// operations that lang might want to perform on that telemetry after it's initialized.
 pub trait CoreTelemetry {
@@ -35,22 +37,11 @@ pub struct TelemetryOptions {
     pub logging: Option<Logger>,
     /// Optional metrics exporter - set as None to disable.
     #[builder(setter(into, strip_option), default)]
-    pub metrics: Option<MetricsExporter>,
-
-    /// If set true, do not prefix metrics with `temporal_`. Will be removed eventually as
-    /// the prefix is consistent with other SDKs.
+    pub metrics: Option<Arc<dyn CoreMeter>>,
+    /// If set true, strip the prefix `temporal_` from metrics, if present. Will be removed
+    /// eventually as the prefix is consistent with other SDKs.
     #[builder(default)]
     pub no_temporal_prefix_for_metrics: bool,
-
-    // TODO: Will move inside constructors for otel/prom
-    /// Specifies the aggregation temporality for metric export. Defaults to cumulative.
-    #[builder(default = "MetricTemporality::Cumulative")]
-    pub metric_temporality: MetricTemporality,
-
-    // A map of tags to be applied to all metrics
-    #[builder(default)]
-    pub global_tags: HashMap<String, String>,
-
     /// If set true (the default) explicitly attach a `service_name` label to all metrics. Turn this
     /// off if your collection system supports the `target_info` metric from the OpenMetrics spec.
     /// For more, see
@@ -60,7 +51,7 @@ pub struct TelemetryOptions {
 }
 
 /// Options for exporting to an OpenTelemetry Collector
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, derive_builder::Builder)]
 pub struct OtelCollectorOptions {
     /// The url of the OTel collector to export telemetry and metrics to. Lang SDK should also
     /// export to this same collector.
@@ -68,7 +59,29 @@ pub struct OtelCollectorOptions {
     /// Optional set of HTTP headers to send to the Collector, e.g for authentication.
     pub headers: HashMap<String, String>,
     /// Optionally specify how frequently metrics should be exported. Defaults to 1 second.
-    pub metric_periodicity: Option<Duration>,
+    #[builder(default = "Duration::from_secs(1)")]
+    pub metric_periodicity: Duration,
+    /// Specifies the aggregation temporality for metric export. Defaults to cumulative.
+    #[builder(default = "MetricTemporality::Cumulative")]
+    pub metric_temporality: MetricTemporality,
+    // A map of tags to be applied to all metrics
+    #[builder(default)]
+    pub global_tags: HashMap<String, String>,
+    /// A prefix to be applied to all metrics. Defaults to "temporal_".
+    #[builder(default = "METRIC_PREFIX")]
+    pub metric_prefix: &'static str,
+}
+
+/// Options for exporting metrics to Prometheus
+#[derive(Debug, Clone, derive_builder::Builder)]
+pub struct PrometheusExporterOptions {
+    pub socket_addr: SocketAddr,
+    // A map of tags to be applied to all metrics
+    #[builder(default)]
+    pub global_tags: HashMap<String, String>,
+    /// A prefix to be applied to all metrics. Defaults to "temporal_".
+    #[builder(default = "METRIC_PREFIX")]
+    pub metric_prefix: &'static str,
 }
 
 /// Configuration for the external export of traces
@@ -86,18 +99,6 @@ pub enum TraceExporter {
     // TODO: Remove
     /// Export traces to an OpenTelemetry Collector <https://opentelemetry.io/docs/collector/>.
     Otel(OtelCollectorOptions),
-}
-
-/// Control where metrics are exported
-#[derive(Debug, Clone)]
-pub enum MetricsExporter {
-    /// Export metrics to an OpenTelemetry Collector <https://opentelemetry.io/docs/collector/>.
-    Otel(OtelCollectorOptions),
-    /// Expose metrics directly via an embedded http server bound to the provided address.
-    Prometheus(SocketAddr),
-    /// Export metrics calls to lang where it can inject them into whatever metrics system the
-    /// user likes
-    Lang(Arc<dyn CoreMeter>),
 }
 
 /// Control where logs go
--- a/core-api/src/telemetry.rs
+++ b/core-api/src/telemetry.rs
@@ -1,6 +1,10 @@
+pub mod metrics;
+
+use crate::telemetry::metrics::CoreMeter;
 use std::{
     collections::HashMap,
     net::SocketAddr,
+    sync::Arc,
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 use tracing_core::Level;
@@ -90,6 +94,9 @@ pub enum MetricsExporter {
     Otel(OtelCollectorOptions),
     /// Expose metrics directly via an embedded http server bound to the provided address.
     Prometheus(SocketAddr),
+    /// Export metrics calls to lang where it can inject them into whatever metrics system the
+    /// user likes
+    Lang(Arc<dyn CoreMeter>),
 }
 
 /// Control where logs go
--- a/core-api/src/telemetry.rs
+++ b/core-api/src/telemetry.rs
@@ -82,6 +82,13 @@ pub struct PrometheusExporterOptions {
     /// A prefix to be applied to all metrics. Defaults to "temporal_".
     #[builder(default = "METRIC_PREFIX")]
     pub metric_prefix: &'static str,
+    /// If set true, all counters will include a "_total" suffix
+    #[builder(default = "false")]
+    pub counters_total_suffix: bool,
+    /// If set true, all histograms will include the unit in their name as a suffix.
+    /// Ex: "_milliseconds".
+    #[builder(default = "false")]
+    pub unit_suffix: bool,
 }
 
 /// Configuration for the external export of traces
--- a/core-api/src/telemetry.rs
+++ b/core-api/src/telemetry.rs
@@ -42,6 +42,7 @@ pub struct TelemetryOptions {
     #[builder(default)]
     pub no_temporal_prefix_for_metrics: bool,
 
+    // TODO: Will move inside constructors for otel/prom
     /// Specifies the aggregation temporality for metric export. Defaults to cumulative.
     #[builder(default = "MetricTemporality::Cumulative")]
     pub metric_temporality: MetricTemporality,
--- a/core-api/src/telemetry/metrics.rs
+++ b/core-api/src/telemetry/metrics.rs
@@ -1,0 +1,263 @@
+use std::{fmt::Debug, sync::Arc};
+
+/// Implementors of this trait are expected to be defined in each language's bridge.
+/// The implementor is responsible for the allocation/instantiation of new metric meters which
+/// Core has requested.
+pub trait CoreMeter: Send + Sync + Debug {
+    fn new_attributes(&self, attribs: MetricsAttributesOptions) -> MetricAttributes;
+    fn counter(&self, name: &str) -> Arc<dyn Counter>;
+    fn histogram(&self, name: &str) -> Arc<dyn Histogram>;
+    fn gauge(&self, name: &str) -> Arc<dyn Gauge>;
+}
+
+#[derive(Debug)]
+pub struct NoOpCoreMeter;
+impl CoreMeter for NoOpCoreMeter {
+    fn new_attributes(&self, _: MetricsAttributesOptions) -> MetricAttributes {
+        MetricAttributes::Lang {
+            id: 0,
+            new_attributes: vec![],
+        }
+    }
+
+    fn counter(&self, _: &str) -> Arc<dyn Counter> {
+        Arc::new(NoOpInstrument)
+    }
+
+    fn histogram(&self, _: &str) -> Arc<dyn Histogram> {
+        Arc::new(NoOpInstrument)
+    }
+
+    fn gauge(&self, _: &str) -> Arc<dyn Gauge> {
+        Arc::new(NoOpInstrument)
+    }
+}
+
+#[derive(Debug)]
+pub struct PrefixedMetricsMeter<CM>(&'static str, CM);
+impl<CM> PrefixedMetricsMeter<CM> {
+    pub fn new(prefix: &'static str, core_meter: CM) -> Self {
+        PrefixedMetricsMeter(prefix, core_meter)
+    }
+}
+impl<CM: CoreMeter> CoreMeter for PrefixedMetricsMeter<CM> {
+    fn new_attributes(&self, attribs: MetricsAttributesOptions) -> MetricAttributes {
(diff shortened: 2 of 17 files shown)


core-api/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,263 @@
+use std::{fmt::Debug, sync::Arc};
+
+/// Implementors of this trait are expected to be defined in each language's bridge.
+/// The implementor is responsible for the allocation/instantiation of new metric meters which
+/// Core has requested.
+pub trait CoreMeter: Send + Sync + Debug {
```
Sushisource, line 6: Most public API stuff is in this file

core-api/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,263 @@
...
+    fn new_attributes(&self, attribs: MetricsAttributesOptions) -> MetricAttributes;
+    fn counter(&self, name: &str) -> Arc<dyn Counter>;
```
cretz, line 8: Are there any perf concerns with the cost of `dyn` for metrics? Is this dynamically dispatched for each metrics call?

core-api/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+/// Histogram buckets
+#[derive(Clone, Debug)]
+pub struct Buckets {
+    buckets: Vec<f64>,
+}
+
+pub trait Counter: Send + Sync {
+    fn add(&self, value: u64, attributes: &MetricAttributes);
```
cretz, line 171: Providing attributes at metric record time can get a bit expensive for naive implementations that need to group by collective attributes. There is a cost associated with not pre-binding attributes. Can we consider an alteration in the API (even if the primary impl of that doesn't use it that way) that associates metric attributes with a `CoreMeter` instead? And you can have `with_new_attrs` that returns `Self` on that instead? Or at least have `with_new_attrs` on each of these metrics instead? This way implementers can implement these metric impls having already created the counter with its attributes. And it also discourages users of these metrics from adding any attribute willy-nilly at record time.

core-api/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,263 @@
...
+            }
+        }
+    }
+
+    impl CoreMeter for Meter {
+        fn new_attributes(&self, attribs: MetricsAttributesOptions) -> MetricAttributes {
+            MetricAttributes::OTel {
+                kvs: Arc::new(attribs.attributes.into_iter().map(KeyValue::from).collect()),
+                // TODO: Pass in? Hard to make work with trait
+                ctx: opentelemetry::Context::current(),
```
cretz, line 222: Just documenting that users can use thread-locals or thread contexts or whatever is enough I think.

core-api/src/telemetry/metrics.rs
cretz: Having thought about this, I now fear that the way the metrics are recorded may preclude this. While `counter.with_new_attrs(attrs).add(123)` is just as cheap as `counter.add(123, attrs)`, if you are always calling `with_new_attrs` on every record time anyways then there's no value here. So this can be ignored.

Just know that many non-OTel implementations are required to group by these attributes before applying the value. But otherwise, this can be ignored.

bergundy: Looks like this will work with the Otel JS metrics API: https://github.com/open-telemetry/opentelemetry-js/blob/b8b63083869fb5850fd4830ded0fc509a2cd0f0a/api/src/metrics/Metric.ts#L65

bergundy: We'll lose the otel context but other than that everything else should work the way it is.

core-api/src/telemetry/metrics.rs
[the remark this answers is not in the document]
Sushisource: I can't. I actually had quite the struggle with the type system on trying to get this to be as little overhead as possible. Traits cannot return existential types (can't return `impl Counter`) yet. I would need to make the trait generic, or have an associated type, but that would infect the entire rest of the code with generics just for this purpose which would be no bueno.

Thus, these methods have to return some kind of pointer, so might as well be `Arc` here since `Box` is just going to get shoved in one anyway

Sushisource: See other answer. Kinda need to to keep things sane. That's not the overhead I'm worried about so much as attribute passing.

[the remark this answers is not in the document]
Sushisource: The only other option would make the instrument types also be enums, which is another spot where there can be a mis-match between the attribute type and the instrument type, which is sorta sad. But, given that can already happen, I'd probably be fine with enum-ing them too.

bergundy (approved):
Overall this LGTM, I'll try to integrate this with TS when I get some non-versioning time.
Maybe let's leave this PR open until we have the first use case.

core-api/src/telemetry/metrics.rs
cretz: That lang controls creating this set (e.g. w/ pre-built hashing) does seem to help, so long as the Rust side doesn't commonly create this attribute set on each recording (here and there can make sense though).

[Go Prometheus](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus) for instance takes the labels (what we call attributes here) and uses fnv1-a to hash them for grouping purposes, but is optimized to do that ahead of time if the caller does it ahead of time. And so they don't take labels at record time to discourage dynamic labelling.

cretz (approved):

core-api/src/telemetry/metrics.rs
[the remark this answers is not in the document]
Sushisource: I don't mean enum instrument, I mean enum counter which is either otel-in-core backed, or lang-backed. 

The lang-backed version would essentially do what you suggest there, except it would store an internal `u64` or something which would be used for a pre-hashed or indexed lookup lang side rather than hashing the name every time.

core-api/src/telemetry.rs
```diff
@@ -82,6 +86,9 @@ pub enum MetricsExporter {
     Otel(OtelCollectorOptions),
     /// Expose metrics directly via an embedded http server bound to the provided address.
     Prometheus(SocketAddr),
+    /// Export metrics calls to lang where it can inject them into whatever metrics system the
+    /// user likes
+    Lang(Arc<dyn CoreMeter>),
 }
```
cretz, line 92: I'm not sure if we need to rework this, but we need to be able to access the `CoreMeter` for the OTel/Prom impls too. Langs need to be able to record custom metrics on the existing metric impls. With the API as it is today, even if the OTel/Prom impls are made public, langs will never call the `Otel` and `Prometheus` enums again on `MetricsExporter`.

Maybe we can remove `MetricsExporter` and just make `CoreMeter` the trait accepted in Telemetry and have functions to create `CoreMeter` for OTel/Prom (and of course lang can just impl `CoreMeter` themselves). Otherwise, can we have some other way to access the `CoreMeter`? I wonder if we should also make standard tag names available.

Sushisource: Standard tag names available as constants? Yeah, that makes sense to me.

I can see getting rid of `MetricsExporter` and making `CoreMeter` the only thing, and having a way to get one for Otel & Prom. @mjameswh What do you think? I will make this work if we all like it.

cretz: :+1: Yeah, anything that will let us both (and separately) create custom meter _and_ create custom metrics on whatever meter is created will work great!

mjameswh: I agree: the `MetricsExporter` enum doesn't add much value. Not sure however this is blocker for lang->core custom metrics, as we can get the configured `CoreMeter` object from `CoreRuntime.metric_meter`. But still, just to keep things simpler, we should probably remove the useless enum.

One minor thing however is that both the Prometheus and the Otel metric initialization code currently depends on the Tokio `runtime` object; seems like we use that to run the Prometheus server on a distinct thread. Moving that initialization

Sushisource: Done

core-api/src/telemetry/metrics.rs
bergundy: I'm missing how lang is supposed to maintain this cache, aren't most attributes defined core side?

core/src/telemetry/metrics.rs
```diff
@@ -454,18 +413,191 @@ impl AggregatorSelector for SDKAggSelector {
...
+    Ok(StartedPromServer {
+        meter: Arc::new(mp.meter(TELEM_SERVICE_NAME)),
+        bound_addr,
+        abort_handle: handle.abort_handle(),
+    })
+}
+
+/// Buffers [MetricEvent]s for periodic consumption by lang
+#[derive(Debug)]
+pub struct MetricsCallBuffer {
```
cretz, line 481: Is this visible outside the crate?

Sushisource: Indeed it is

Sushisource:
Needs https://github.com/temporalio/sdk-core/pull/544 to get merged first

cretz:
> Needs https://github.com/temporalio/sdk-core/pull/544 to get merged first

I assume you mean #584

mjameswh (approved):

core-api/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,250 @@
...
+    fn add(&self, value: u64, attributes: &MetricAttributes);
+}
+
+pub trait Histogram: Send + Sync {
+    // When referring to durations, this value is in millis
+    fn record(&self, value: u64, attributes: &MetricAttributes);
+}
+
+pub trait Gauge: Send + Sync {
+    // When referring to durations, this value is in millis
```
cretz, line 174: Just crawling this PR again...is there a use of gauge for durations?

core-api/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,250 @@
...
+        MetricValue::String(value.to_string())
+    }
+}
+
+pub trait Counter: Send + Sync {
+    fn add(&self, value: u64, attributes: &MetricAttributes);
+}
+
+pub trait Histogram: Send + Sync {
+    // When referring to durations, this value is in millis
```
cretz, line 169: Is there a case where we use histograms for anything but durations? In the Go SDK, we were able to get away with three metric types: counter, gauge, and timer (which is histogram, but we called it timer and clearly used durations). I wonder if a similar thing would make sense here.

I am worried that people will want to know whether you are recording a duration or a numeric value.

core-api/src/telemetry/metrics.rs
Sushisource: I don't think so at the moment

core-api/src/telemetry/metrics.rs
Sushisource: They very easily could be non-durations, and either way it's a histogram. The language is clear with "when" - I don't think this is an issue. At the end of the day it's still a histogram regardless of what unit you've chosen.

cretz: But I think it's helpful for our users to know whether they're recording a duration or not. We can say "may be milliseconds or just some other number" but they they're gonna have some extra conditionals checking string name to know it's a duration. I think we should have a separate concept for duration-based histograms and non-duration-based histograms (and if there are only duration-based histograms ever used, we can wait until we need the non-duration-based to add it).

core-api/src/telemetry/metrics.rs
cretz: :+1: I vote we remove the comment and when we do have duration based gauges (I can't think of when) we can revisit how to expose.

core-api/src/telemetry/metrics.rs
Sushisource: I disagree. There are basically zero other metrics APIs that do this. There are metric types, which generally correlate to how things are aggregated, and there are units. Confusing units with the method of aggregation is I think not an appropriate design.

If we would like to accept units, sure, that's totally reasonable, but IMO it's a definite no-no to consider two metric types to be different simply because they are using different units.

cretz: > If we would like to accept units, sure, that's totally reasonable, but IMO it's a definite no-no to consider two metric types to be different simply because they are using different units.

:+1: I will happily take units. I just figured separate traits was the best way to represent separate units instead of int + enum or something. I was never different metric types on the downstream system (they both still become the same histograms). For example, with Prometheus in some languages people create either a "timer" or a "histogram", but the former is just sugar over the latter with the type as duration.

Sushisource: Cool. I will update things to allow passing units and descriptions in when building instruments. They are just wrappers around strings at the end of the day.

cretz: :+1: If at all possible would like an enumerate or similar for units instead of an arbitrary value. For any lang-side impl of this, they can abstract them to different overloads (I suspect only .NET will be fast enough to support a lang-side impl).

Sushisource: Certainly possible but the problem is it only has the meaning we give it. It ends up becoming a string, at least using OTel as a backend. I can make one up, but then users are limited only to what we've made up, where they might have some unit they care about we can never imagine... So I think we should just keep it a string.

cretz: > users are limited only to what we've made up

This is what drove the metrics abstraction choices in Go. We didn't want a general purpose metrics abstraction for all uses, we wanted a Temporal metrics abstraction for only how Temporal uses metrics. Yes, this limits users reusing that abstraction inside of their workflow/activity to just what we also use, but what we've needed has been enough so far and if/when we offer custom metrics impls, they will use their more powerful impl instead of the weaker Temporal abstraction in super advanced cases.

I prefer the strongly typed int or duration. If we can't have that, I prefer the strongly typed int with strongly typed unit enum. If we can't have that, I guess I can settle for int with arbitrary string, but it's the least preferred.

Sushisource: They're still limited to the kinds we expose, which seems to me the important part. Units are just a string description exported to the backend. Limiting users to what that string can be doesn't seem to help them in any material way. If the language layer wants to offer some "TimingHistogram" abstraction, it is still absolutely free to do that, and adding a typed enum on the core side can't meaningfully enforce anything interesting, we just don't gain anything. I can't somehow enforce that when you pick a `Milliseconds` variant, that the number you're passing actually makes any sense.

Sushisource:
I realized the underling library update seems to have imposed some naming changes on gauges too. I want to make sure those are resolved before I merged this.

core-api/src/telemetry/metrics.rs
cretz: > Units are just a string description exported to the backend

That assumes a lot about the backend. We always prefer strongly typed, even if the backend is stringly typed. For example, an implementation may want to record durations and let a config decide the duration unit. Or they may have to translate (for example some backends use milliseconds and [some use seconds](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#NewTimer) by default).

> If the language layer wants to offer some "TimingHistogram" abstraction, it is still absolutely free to do that, and adding a typed enum on the core side can't meaningfully enforce anything interesting, we just don't gain anything.

:+1: Can we at least document all the string units _we_ set for this value so that users can convert your value back to a duration to have the freedom of deciding unit?

> adding a typed enum on the core side can't meaningfully enforce anything interesting

It just signals intent. That you can only pass a duration to the `TimingHistogram` makes it clearer what the value really is. Kinda the same reason these languages even have duration types instead of just integer types even though they technically can be expressed as integers in their greatest-supported resolution.

mjameswh: FYIW I feels like Prometheus encourage us to take the wrong direction, and we should not follow it's lead simply on the principle that "most engines out there" are doing it that way. Metric emitters should focus on what those values really means (ie to them), rather than how they will be stored or presented.

The fact is that some place in the code is either emitting time values, or something else, but in the end, the unit for that metric is unambiguous for the caller and should therefore also be clearly specified as part the "metric" contract. The data is a duration, expressed in ms. The fact that it will likely be presented as an histogram is of lesser significancy.

cretz: > FYIW I feels like Prometheus encourage us to take the wrong direction, and we should not follow it's lead simply on the principle that "most engines out there" are doing it that way.

It's not about a specific backend, that was just an example. It's not taking me in a certain direction. Very simply, we should type our datatypes as specifically as is useful. Knowing that something is a duration and not just an integer is useful. People need to translate durations.

>  the unit for that metric is unambiguous for the caller

Not true if someone has to `== "milliseconds"` in their code. That's pretty ambiguous. Hope we don't add "seconds" next.

>  The data is a duration, expressed in ms.

More like the data is an integer, expressed with an arbitrary string you must know to assume ms.

core-api/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,280 @@
...
+pub struct MetricParameters {
+    /// The name for the new metric/instrument
+    #[builder(setter(into))]
+    pub name: Cow<'static, str>,
+    /// A description that will appear in metadata if the backend supports it
+    #[builder(setter(into), default = "\"\".into()")]
+    pub description: Cow<'static, str>,
+    /// Unit information that will appear in metadata if the backend supports it
+    #[builder(setter(into), default = "\"\".into()")]
+    pub unit: Cow<'static, str>,
```
cretz, line 23: Can we document the unit string core uses for its durations? We will have to document this lang side so that when users use durations or when users write a backend, they know to match (though lang side is probably going to be more strongly typed than Rust here I suspect, which is a weird statement).

core-api/src/telemetry/metrics.rs
Sushisource: > Can we at least document all the string units we set
Core doesn't even set any units, right now, but I can add that to this PR.

Anyway, my main point here is it doesn't really make much sense to expose the unit type safety stuff at the Core level, because lang can do that, and the translation to calls into core is quite easy to get right. I agree of course that if users want to specify unit & associated value types in lang that that makes sense.

The only thing I could do in Core would maybe be something like accepting `std::time::Duration` for some kinda time-specific histogram, but even then I'm just immediately going to toss that out the window when I make the real metric call, and lang should've already done that it's type.

cretz: > The only thing I could do in Core would maybe be something like accepting std::time::Duration for some kinda time-specific histogram, but even then I'm just immediately going to toss that out the window when I make the real metric call,

But I, the language dev, am not going to toss that out the window. I am going to give it to the user's metric impl and they're gonna be happy I differentiated.

But we've discussed enough, string is good enough I suppose. Let's just document the string core uses for its durations so I can have a hardcoded conditional for stronger typing lang side.

PR #545, opened 2023-05-09
Poll concurrently at will

Sushisource:
## What was changed
Remove the semaphore that restricted concurrent polls based on requests. This in practice meant we were typically limited to one concurrent poll, except in some situations where polls get put in buffers.

## Why?
Closer to what other SDKs are doing, allows for some concurrency scale-up without getting clever yet, which we'll do later.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/546

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/core_tests/workflow_tasks.rs
+++ b/core/src/core_tests/workflow_tasks.rs
@@ -18,7 +18,7 @@ use rstest::{fixture, rstest};
 use std::{
     collections::{HashMap, HashSet, VecDeque},
     sync::{
-        atomic::{AtomicU64, Ordering},
+        atomic::{AtomicU64, AtomicUsize, Ordering},
         Arc,
     },
     time::Duration,
@@ -60,6 +60,7 @@ use temporal_sdk_core_test_utils::{fanout_tasks, start_timer_cmd, WorkerTestHelp
 use tokio::{
     join,
     sync::{Barrier, Semaphore},
+    time,
 };
 
 #[fixture(hist_batches = &[])]
@@ -1535,8 +1536,6 @@ async fn failing_wft_doesnt_eat_permit_forever() {
             .unwrap();
     }
     assert_eq!(worker.outstanding_workflow_tasks().await, 0);
-    // 1 permit is in use because the next task is buffered and has re-used the permit
-    assert_eq!(worker.available_wft_permits().await, 1);
     // We should be "out of work" because the mock service thinks we didn't complete the last task,
     // which we didn't, because we don't spam failures. The real server would eventually time out
     // the task. Mock doesn't understand that, so the WFT permit is released because eventually a
@@ -1545,7 +1544,7 @@ async fn failing_wft_doesnt_eat_permit_forever() {
     outstanding_mock_tasks.unwrap().release_run(&run_id);
     let activation = worker.poll_workflow_activation().await.unwrap();
     // There should be no change in permits, since this just unbuffered the buffered task
-    assert_eq!(worker.available_wft_permits().await, 1);
+    assert_eq!(worker.available_wft_permits(), 1);
     worker
         .complete_workflow_activation(WorkflowActivationCompletion::from_cmd(
             activation.run_id,
@@ -1554,7 +1553,7 @@ async fn failing_wft_doesnt_eat_permit_forever() {
         .await
         .unwrap();
     worker.shutdown().await;
-    assert_eq!(worker.available_wft_permits().await, 2);
+    assert_eq!(worker.available_wft_permits(), 2);
 }
 
 #[tokio::test]
@@ -1875,7 +1874,7 @@ async fn poll_faster_than_complete_wont_overflow_cache() {
 
     join!(blocking_poll, complete_evict);
     // p5 outstanding and final poll outstanding -- hence one permit available
-    assert_eq!(core.available_wft_permits().await, 1);
+    assert_eq!(core.available_wft_permits(), 1);
     assert_eq!(core.cached_workflows().await, 3);
 }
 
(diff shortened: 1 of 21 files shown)


Sushisource:
This may need a bit of work to ensure it doesn't take tasks when we're already full up -- I'll add some testing around that

core/src/pollers/poll_buffer.rs
```diff
@@ -127,7 +115,6 @@ where
     /// Returns `None` if the poll buffer has been shut down
     #[instrument(name = "long_poll", level = "trace", skip(self))]
     async fn poll(&self) -> Option<pollers::Result<T>> {
-        self.polls_requested.add_permits(1);
         if let Some(fun) = self.num_pollers_changed.as_ref() {
```
bergundy, line 118: Should this metric reflect the number of active polls? Why only emit it when lang polls?

Sushisource: Yup, this needs to change now

core/src/pollers/poll_buffer.rs
```diff
@@ -63,31 +59,24 @@ where
         FT: Future<Output = pollers::Result<T>> + Send,
     {
         let (tx, rx) = channel(buffer_size);
```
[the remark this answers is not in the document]
Sushisource: Irrelevant now

core/src/pollers/poll_buffer.rs
Sushisource: This is changed

core/src/core_tests/workflow_tasks.rs
```diff
@@ -2588,3 +2587,72 @@ async fn history_length_with_fail_and_timeout(
...
+    let hung_poll = async {
+        // This should end up getting shut down after the other routine finishes tasks
+        assert_matches!(
+            worker.poll_workflow_activation().await.unwrap_err(),
+            PollWfError::ShutDown
+        );
+    };
+    // Wait for a bit concurrently with above, verify no extra tasks got taken, shutdown
+    let ender = async {
+        time::sleep(Duration::from_millis(300)).await;
```
bergundy, line 2643: Why sleep for 300 ms?
Isn't it enough to initiate shutdown to ensure that the next activation poll will return a shutdown error?

core/src/pollers/poll_buffer.rs
```diff
@@ -16,102 +17,110 @@ use temporal_sdk_core_protos::temporal::api::workflowservice::v1::{
...
+        mpsc::{unbounded_channel, UnboundedReceiver},
+        Mutex,
     },
     task::JoinHandle,
 };
 use tokio_util::sync::CancellationToken;
 
 pub struct LongPollBuffer<T> {
-    buffered_polls: Mutex<Receiver<pollers::Result<T>>>,
+    buffered_polls: Mutex<UnboundedReceiver<pollers::Result<(T, OwnedMeteredSemPermit)>>>,
```
bergundy, line 28: `OwnedMeteredSemPermit` strikes again

core/src/core_tests/workflow_tasks.rs
Sushisource: Just to make sure that, if things weren't working right, there would be enough time for the pollers to start buffering things (though they should not). Otherwise possibly shutdown could just happen too fast

core/src/pollers/poll_buffer.rs
```diff
@@ -16,102 +17,110 @@ use temporal_sdk_core_protos::temporal::api::workflowservice::v1::{
...
+                let nph = nph.as_ref().map(|a| a.as_ref());
                 loop {
                     if shutdown.is_cancelled() {
                         break;
                     }
-                    let sp = tokio::select! {
-                        sp = polls_requested.acquire() => sp.expect("Polls semaphore not dropped"),
+                    let permit = tokio::select! {
+                        p = poll_semaphore.acquire_owned() => p,
                         _ = shutdown.cancelled() => continue,
```
bergundy, line 95: I think you can just break here.

core/src/pollers/poll_buffer.rs
```diff
@@ -16,102 +17,110 @@ use temporal_sdk_core_protos::temporal::api::workflowservice::v1::{
...
                     };
-                    let _active_guard = ActiveCounter::new(ap.as_ref());
+                    let permit = if let Ok(p) = permit {
```
bergundy, line 97: When is permit not `Ok`? Is it only on shutdown?
If that's the case you might be able to remove the check for `shutdown.is_cancelled()` at the top of this loop.

core/src/worker/workflow/managed_run.rs
```diff
@@ -1039,13 +1039,26 @@ impl ManagedRun {
...
         outcome: ActivationCompleteOutcome,
         chan: Option<oneshot::Sender<ActivationCompleteResult>>,
     ) {
         if let Some(chan) = chan {
-            chan.send(self.build_activation_complete_result(outcome))
-                .expect("Rcv half of activation reply not dropped");
+            if chan
+                .send(self.build_activation_complete_result(outcome))
+                .is_err()
+            {
```
Sushisource, line 1050: I did see this happen once running things against cloud - but I wouldn't be too worried about it. I think it's actually likely happening because of a rare problem in the Rust SDK where it might just drop a future that is calling an activation completion suddenly, thus triggering this problem. AFAICT that's kinda the only way it'd happen.

In any case, it's mitigated now at minimum.

core/src/pollers/poll_buffer.rs
```diff
@@ -16,102 +17,110 @@ use temporal_sdk_core_protos::temporal::api::workflowservice::v1::{
...
+                        p
+                    } else {
+                        break;
+                    };
+                    let _active_guard = ActiveCounter::new(ap.as_ref(), nph);
                     let r = tokio::select! {
                         r = pf() => r,
                         _ = shutdown.cancelled() => continue,
```
bergundy, line 105: can break here too

core/src/pollers/poll_buffer.rs
Sushisource: It'd be exceedingly hard to reason about since that will only happen when the semaphore is dropped. Given it's refcounted all over the place I don't want to have to think about that.

core/src/pollers/poll_buffer.rs
```diff
@@ -16,102 +17,110 @@ use temporal_sdk_core_protos::temporal::api::workflowservice::v1::{
...
-        Self(a)
+struct ActiveCounter<'a, F: Fn(usize)>(&'a AtomicUsize, Option<F>);
+impl<'a, F> ActiveCounter<'a, F>
+where
+    F: Fn(usize),
+{
+    fn new(a: &'a AtomicUsize, change_fn: Option<F>) -> Self {
+        let v = a.fetch_add(1, Ordering::Relaxed) + 1;
+        if let Some(cfn) = change_fn.as_ref() {
+            cfn(v);
```
bergundy, line 44: Wondering if this needs to be throttled, emitting a gauge every time a poll is started or completed seems a bit excessive to me.

bergundy (approved):
That seemed easy

core/src/pollers/poll_buffer.rs
Sushisource: This is why all metrics libraries do some internal buffering. My expectation is with the export-to-lang stuff we'll need to do the same thing. Having to do this at the callsite anytime you have a hot metric is pretty annoying.

PR #547, opened 2023-05-11
fix: add 5 minute tcp keepalive to client

andrewtoth:
An idle client connection is dropped server side every 5 minutes. This adds a keepalive on the client side to prevent this.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -80,6 +80,7 @@ static LONG_POLL_METHOD_NAMES: [&str; 2] = ["PollWorkflowTaskQueue", "PollActivi
 /// The server times out polls after 60 seconds. Set our timeout to be slightly beyond that.
 const LONG_POLL_TIMEOUT: Duration = Duration::from_secs(70);
 const OTHER_CALL_TIMEOUT: Duration = Duration::from_secs(30);
+const TCP_KEEPALIVE: Duration = Duration::from_secs(300);
 
 type Result<T, E = tonic::Status> = std::result::Result<T, E>;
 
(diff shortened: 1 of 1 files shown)


Sushisource:
Thanks! You'll need to run rustfmt before CI will pass, then we'll get it merged :)

andrewtoth:
@Sushisource whoops, fixed!

Sushisource:
@andrewtoth  On second thought - two things probably make sense to happen here, it should probably be a bit below the 5 minute server-side timeout to avoid races, and ideally we'll make this configurable. If you want to take a crack at plumbing that through, awesome, otherwise I can add to the PR.

andrewtoth:
@Sushisource there are a lot more options to configure the `tonic::transport::Endpoint`. Not sure if you want to pass more through as well, and if so how? I'll let you add to the PR if you already have a set idea.

Sushisource:
@andrewtoth Can you check the little box on the right of this PR which allows me to make changes to the PR? Then I can update this and it'll still have your commit in it :)

andrewtoth:
@Sushisource sorry I don't see that option on this PR. Not sure where it is supposed to be, it doesn't show a screenshot at https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests. But either I don't see that checkbox anywhere 🤷

Sushisource:
Hmm. You don't see it at the bottom of the righthand column? 
image


If not, I'll just move this to my own branch and merge it from there then.

andrewtoth:
@Sushisource ahh it's because my branch is from `@buildable/temporal-sdk-core` instead of my own fork.

Sushisource:
Re made it over here https://github.com/temporalio/sdk-core/pull/549

PR #549, opened 2023-05-11
Add configurable TCP keepalive

Sushisource:
Re-creation of https://github.com/temporalio/sdk-core/pull/547

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -80,6 +80,7 @@ static LONG_POLL_METHOD_NAMES: [&str; 2] = ["PollWorkflowTaskQueue", "PollActivi
 /// The server times out polls after 60 seconds. Set our timeout to be slightly beyond that.
 const LONG_POLL_TIMEOUT: Duration = Duration::from_secs(70);
 const OTHER_CALL_TIMEOUT: Duration = Duration::from_secs(30);
+const TCP_KEEPALIVE: Duration = Duration::from_secs(300);
 
 type Result<T, E = tonic::Status> = std::result::Result<T, E>;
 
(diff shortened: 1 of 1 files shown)


andrewtoth (approved):

Sushisource:
@andrewtoth In Go/Java we're setting an HTTP2 keepalive rather than at the TCP level. Would that solve whatever you had in mind as well? If so I may switch things to that just for consistency with the other existing SDKs.

andrewtoth:
@Sushisource unfortunately the http2 keep alive does not prevent the connection from being dropped. I ran 3 clients at the same time, one currently in master, one using this branch, and one using this branch and the diff
```
-            Channel::from_shared(self.target_url.to_string())?.tcp_keepalive(Some(TCP_KEEPALIVE));
+            Channel::from_shared(self.target_url.to_string())?.http2_keep_alive_interval(TCP_KEEPALIVE);
```
I then set a timer for 6 minutes, and ran a request through each of them. The only one with a good connection that did not fail was the one with `tcp_keepalive`.

cretz:
@andrewtoth - Hrmm, the goal of keep alive for our SDKs is to identify server outage and drop/reconnect when it does. I don't think it's a good tool for preventing idle connection termination. For HA and load balancing reasons (and general connection pool health), I think we intentionally have a max idle connection timeout. Are you hosting your own Temporal server or using Temporal cloud? If hosting your own, you can set the dynamic config value of `frontend.keepAliveMaxConnectionAge` server side to control this value. But I don't think client side should use TCP-level keep alive to prevent the connection from being seen as idle server side.

andrewtoth:
@cretz thanks for the informative reply. You caused me to look a little closer at my code, because we are using a `RetryClient<Client>` so I thought the bug could be in there. But alas, we were calling `RetryClient::get_client()` before calling `start_workflow` which unwrapped the client from the retry logic. Removing the `get_client` fixed the issue, and the connection now succeeds after an idle 5 minutes. So I agree this fix is not required and may actually be a bad idea.

Thanks!

PR #550, opened 2023-05-12
[NOMERGE] Rust SDK future proposal draft

Sushisource:
Just serves as a place we can discuss Rust SDK vision before turning it into an actual proposal.

For the community: This (unfortunately, sorry) is not meant to imply we'll be implementing this any time soon - rather it will serve to illustrate where we intend to go once we have decided to take on productionization of the Rust SDK.


workflow-api/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+    // Workflow implementation example
+    struct MyWorkflow {
+        foo: u64,
+        bar: HashMap<String, u64>,
+    }
+
+    impl Workflow for MyWorkflow {
+        type Input = String;
+        type Output = u64;
+        const NAME: &'static str = "MyWorkflowType";
```
cretz, line 127: I wonder if this would be better as a static getter, e.g. `fn get_name() -> &str` (or maybe `String` since we should only ask once anyways). If you do leave as const, might as well use the `#[workflow]` attribute.

workflow-api/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        fn new(input: Self::Input, _ctx: SafeWfContext) -> Self {
```
cretz, line 129:
```suggestion
        fn new(input: Self::Input, _ctx: PureWorkflowContext) -> Self {
```

Two suggestions:

* No more abbreviations (boo "Wf")
* Something besides safe because I fear it's too generic of an adjective. I don't think "Pure" is that good either tbh, I'd have to think on it. But it's basically "CommandsDisallowedWorkflowContext".

workflow-api/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+                // The into() is unfortunately unavoidable without making C-A-N and confirm cancel
+                // be errors instead. Personally, I don't love that and I think it's not idiomatic
+                // Rust, whereas needing to `into()` something is. Other way would be macros, but
+                // it's slightly too much magic I think.
+                Ok(self.foo.into())
+            }
+            .boxed()
+            // TODO: The need to box here is slightly unfortunate, but it's either that or require
+            //  users to depend on `async_trait` (which just hides the same thing). IMO this is the
+            //  best option until more language features stabilize and this can go away.
```
cretz, line 153: Can you just require `run` be an `async fn`?

workflow-api/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        fn new(input: Self::Input, _ctx: SafeWfContext) -> Self {
+            let mut bar = HashMap::new();
+            bar.insert(input, 10);
+            Self { foo: 0, bar }
+        }
+
+        fn run(
+            &mut self,
+            ctx: WfContext,
```
cretz, line 137: When you do get around to proposal time, I think it'd be worth a discussion of the tradeoff of making this statically accessible via async-local or similar. What I've found in the last two SDKs is that I can just grab the current coroutine scheduler.

Calls like https://docs.rs/tokio/latest/tokio/task/fn.spawn.html#panics are ok assuming the runtime, I wonder if we can be too. But also I could understand from a function-grouping perspective that having it on a context may be clearer.

One benefit of not having a manually propagated context is that users aren't tempted to store it to call from a query handler or something.

Hrmm, now I wonder if we can have the trait have `fn context(&mut self) -> WorkflowContext` and `fn safe_context(&self) -> SafeWorkflowContext` on it. But that doesn't help the `new` call though.

workflow-api/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    /// Defines the actual workflow logic. The function must return a future, and this future is
+    /// cached and polled as updates to the workflow history are received.
+    ///
+    /// `ctx` should be used to perform various Temporal commands like starting timers and
+    /// activities.
+    fn run(
+        &mut self,
+        ctx: WfContext,
+    ) -> BoxFuture<Result<WfExitValue<Self::Output>, anyhow::Error>>;
```
cretz, line 42: Is there any way we can avoid forcing `anyhow` on users here? Can we just have `TemporalFailure` enum here and some into() or whatever-similar-to-anyhow helper that's clear it converts to `ApplicationFailure`?

workflow-api/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+            //  users to depend on `async_trait` (which just hides the same thing). IMO this is the
+            //  best option until more language features stabilize and this can go away.
+        }
+    }
+
+    // #[workflow] miiiight be necessary here, but, ideally is not.
+    impl MyWorkflow {
+        // Attrib commented out since it's nonexistent for now, but that's what it'd look like.
+        // #[signal]
+        pub fn my_signal(&mut self, arg: String) {
```
cretz, line 161: Can I have the context here?

workflow-api/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+            self.bar.insert(arg, 1);
+        }
+        // #[query]
+        pub fn my_query(&self, arg: String) -> Option<u64> {
```
cretz, line 165: Can I have the "safe" context here?

workflow-api/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        fn run(
+            &mut self,
+            ctx: WfContext,
+        ) -> BoxFuture<Result<WfExitValue<Self::Output>, anyhow::Error>> {
+            async move {
+                ctx.timer(Duration::from_secs(1)).await;
+                self.foo = 1;
+                // See activity definitions file
+                ctx.my_act_fn("Hi!").await.unwrap();
```
cretz, line 143: Concern here is how to pass in options. One reason that Java's approach of reusing actual activity signature was so rough is that you were forced to make options a separate thing.

workflow-api/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+                // The into() is unfortunately unavoidable without making C-A-N and confirm cancel
+                // be errors instead. Personally, I don't love that and I think it's not idiomatic
+                // Rust, whereas needing to `into()` something is. Other way would be macros, but
+                // it's slightly too much magic I think.
```
cretz, line 147: Concerning the `into()`, I wonder if maybe instead of `Result<WfExitValue<Self::Output>, anyhow::Error>` you have `Result<Self::Output, TemporalNonSuccess>` of some form. That way `Ok` _only_ represents literal success, not cancellation or continue as new.

I think it's worth optimizing for the normal/happy path of `Ok` being the literal result and all else being the `Err`.

workflow-api/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+    use futures::FutureExt;
+    use std::{collections::HashMap, marker::PhantomData};
+
+    // Workflow implementation example
+    struct MyWorkflow {
+        foo: u64,
+        bar: HashMap<String, u64>,
+    }
+
+    impl Workflow for MyWorkflow {
```
cretz, line 124:
```suggestion
    #[workflow]
    impl Workflow {
```

I wonder if there's value in not requiring trait impl and instead marking the run method with `#[workflow_run]` to extract input/output types. Only reason I was thinking this is because I kinda like that I can put my signal/query/update functions in the same `impl` block as my run stuff. But otherwise, I see how it hurts `new()` (unless you put the attribute there too).

cretz (commented):
Initial thoughts, I may be way off on some but they are easily dismissable. They at least provide alternatives for the proposal to justify not using any them.

workflow-api/src/lib.rs
Sushisource: Nah, that's the thing that can't work in a trait yet.

workflow-api/src/lib.rs
Sushisource: We could. I would be OK with it I think. This feels somewhat more obvious to me, and I like that you don't have these error paths for "called in some context where it doesn't make sense".

workflow-api/src/lib.rs
Sushisource: It can certainly be our own thing yes

workflow-api/src/lib.rs
Sushisource: Yep

workflow-api/src/lib.rs
Sushisource: Also yep

workflow-api/src/lib.rs
cretz: May make it worth doing it attribute style. PyO3 are heavy users of this approach over trait-based (e.g. https://pyo3.rs/v0.18.3/class and https://pyo3.rs/v0.18.3/function).

workflow-api/src/lib.rs
Sushisource: The problem is CAN _is_ a form of success. The semantics of it are my real issue. `TemporalNonSuccess` as a name is a perfect example. CAN isn't "NonSuccess".

Now, to be fair, currently I have WillRespondAsync as an error in activities. My rationale is that's less common than CAN. I get optimizing for the happy path, but `into` is easy and the compiler error will make it super obvious if the user hasn't done it.

I could live with it, for sure, but it just doesn't feel correct to me or particularly idiomatic for Rust

workflow-api/src/lib.rs
Sushisource: That's just async trait at that point, but if it ends up being useful for other reasons I'm definitely OK with a procmacro

cretz: Maybe something like:

```rust
#[workflow]
struct MyWorkflow {}

impl MyWorkflow {
    #[workflow_run]
    async fn run(&mut self, some_arg: String) -> TemporalResult<Strnig> {
        // ...
    }
    
    async fn my_signal(&mut self, some_arg) {
        // ...
    }
}
```

workflow-api/src/lib.rs
Sushisource: People are pretty used to multiple impl blocks in Rust. That said, I can get behind a procmacro like this too if it's providing value like auto-naming. It just needs to not do too much magic, because messing up IDEs etc is where things get really bad.

workflow-api/src/lib.rs
cretz: (and a `#[workflow_init]` or `#[workflow_new]` or whatever named attribute on a func that accepts workflow args to create the struct if it's not "default"...we built this attribute for .NET constructors to allow for readonly field initialization)

Sushisource: The `#[workflow]` attribute would just go over the impl block I think. There's no need for it on the struct definition, and that allows you to just require a `new` and a `run` method without them needing separate attributes.

cretz: Nice. May be worth exploring at proposal time if there aren't too many drawbacks. If caught between two impls that both may kinda work, just list pros/cons of each and then start one of our classic design Zoom calls to pick a winner :-)

workflow-api/src/lib.rs
cretz: Since continue as new has to be invoked top level to use this type anyways, I wonder if you can make it a macro (so they don't have to `?` or `Err` it), e.g. `continue_as_new!()`. That can kinda hide it. Otherwise, I agree that continue as new and cancel aren't technically errors, it's just a bit rough on the simple path. I can see both arguments here.

workflow-api/src/lib.rs
Sushisource: Mega necropost

Preparing for pres at offsite... I'm now pretty convinced `#[workflow]` attribute macro is the best way to go for sure. Fixes `async` problem, makes signals/queries automatically work, enforces `new` & `run`... It's the way to go. People who want to could still do it all by hand if they hate it for whatever reason

cretz: And it's so pretty!

PR #553, opened 2023-05-16
Add the `sticky_cache_total_forced_eviction` metric

Sushisource:
Exists in Go / Java and useful for determining if cache size is too small

### DIFF
--- a/core/src/telemetry/metrics.rs
+++ b/core/src/telemetry/metrics.rs
@@ -87,6 +87,7 @@ struct Instruments {
     sticky_cache_hit: Counter<u64>,
     sticky_cache_miss: Counter<u64>,
     sticky_cache_size: Histogram<u64>,
+    sticky_cache_evictions: Counter<u64>,
 }
 
 impl MetricsContext {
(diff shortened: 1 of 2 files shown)


core/src/telemetry/metrics.rs
```diff
@@ -327,6 +335,7 @@ impl Instruments {
             sticky_cache_hit: meter.counter("sticky_cache_hit"),
             sticky_cache_miss: meter.counter("sticky_cache_miss"),
             sticky_cache_size: meter.histogram(STICKY_CACHE_SIZE_NAME),
+            sticky_cache_evictions: meter.counter("sticky_cache_total_forced_eviction"),
```
bergundy, line 338: I saw this in the metrics doc and could figure out what "forced" is in this context, looks like here it's just "normal" eviction.

https://docs.temporal.io/references/sdk-metrics#sticky_cache_total_forced_eviction

bergundy (approved):

core/src/telemetry/metrics.rs
Sushisource: Yeah, it's a kinda bad name but can't really change it just for this

PR #556, opened 2023-05-18
Add the ability to use the schedule APIs from the rust SDK library client.

### LINKED ISSUE #555: [Feature Request] Add the ability to manipulate the schedule from the Rust SDK

It is not possible to change or initiate a schedule from the Rust SDK directly. 

See attached PR. Open to suggestions though on how to improve. I didn't know if creating direct structs vs using the proto messages was preferred but happy to change.

### Additional context

While I tried to follow some of the Go examples, it looks like the Rust SDK is a bit different in terms of the logic.

bdbelevate:
## What was changed
Added the ability to use the schedule APIs from the rust SDK client library.

## Why?
Because there's currently no way to create, describe, delete, update or patch the schedule.

1. Closes #555 

2. How was this tested:
I tested it using a client in this example library: https://github.com/cosm-eng/temporal-samples-rust/blob/main/schedule/src/bin/client.rs

3. Any docs updates needed?
Seeing as there aren't many docs, no ;)?

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -55,8 +55,8 @@ use temporal_sdk_core_protos::{
         query::v1::WorkflowQuery,
         replication::v1::ClusterReplicationConfig,
         schedule::v1::{
-            BackfillRequest, Schedule, ScheduleAction, SchedulePolicies, ScheduleSpec,
-            ScheduleState,
+            BackfillRequest, Schedule, ScheduleAction, SchedulePatch, SchedulePolicies,
+            ScheduleSpec, ScheduleState, TriggerImmediatelyRequest,
         },
         taskqueue::v1::TaskQueue,
         testservice::v1::test_service_client::TestServiceClient,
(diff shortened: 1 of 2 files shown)


Sushisource:
Hi @bdbelevate thanks for contributing!

So this is pretty close to the direct gRPC calls themselves, and I've mentioned in https://github.com/temporalio/sdk-core/issues/555 that you can just use those directly if desired.

If we want to introduce a higher-level API that's nice too, but I think we probably want to do it in a manner similar to the one that's been added to the Go SDK which provides more familiar types than the stuff directly from the protos: https://github.com/temporalio/sdk-go/blob/269b55b8d184849209c6973fe913a427fafc07e3/internal/schedule_client.go

Except we can probably do better than the Go client as well, with nicer enums, etc.

If you're willing to take that on that'd be amazing! Otherwise I'd say it probably makes sense to just use the raw client directly for now.

bdbelevate:
Great. I'll go down that path then. I started with that originally, but when I saw how it was different than the Go client I worried that you were keeping them different for a reason.

PR #557, opened 2023-05-19
Apply rate limiter at the poller level

Sushisource:
## What was changed
Move ratelimiter capability inside pollers to make activity ratelimiting work properly with concurrent pollers

## Why?
Recent change to make pollers genuinely concurrent could cause task timeouts in the event users have enabled ratelimiting

## Checklist

1. Closes 

2. How was this tested:
Existing UT made more accurate w/ mock client rather than mock poller

4. Any docs updates needed?

### DIFF
--- a/core/src/pollers/poll_buffer.rs
+++ b/core/src/pollers/poll_buffer.rs
@@ -247,6 +247,7 @@ pub(crate) fn new_workflow_task_buffer(
 }
 
 pub type PollActivityTaskBuffer = LongPollBuffer<PollActivityTaskQueueResponse>;
+#[allow(clippy::too_many_arguments)]
 pub(crate) fn new_activity_task_buffer(
     client: Arc<dyn WorkerClient>,
     task_queue: String,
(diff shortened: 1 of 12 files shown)


core/src/pollers/poll_buffer.rs
```diff
@@ -241,7 +255,12 @@ pub(crate) fn new_activity_task_buffer(
     max_tps: Option<f64>,
     shutdown: CancellationToken,
     num_pollers_handler: Option<impl Fn(usize) + Send + Sync + 'static>,
+    mac_worker_aps: Option<f64>,
```
bergundy, line 258: What's mac_worker_aps :-p? Sounds like Mac OS apps or something.
Seriously, can you please name this in a way that doesn't require guessing the meaning?

bergundy (approved):

bergundy:
We need to apply rate limits for eager activities too but that doesn't have to be in this PR, we've had that debt ever since they were introduced.

core/src/telemetry/mod.rs
```diff
@@ -297,6 +297,9 @@ pub fn telemetry_init(opts: TelemetryOptions) -> Result<TelemetryInstance, anyho
             .with(forward_layer)
             .with(export_layer);
 
+        #[cfg(feature = "tokio-console")]
+        let reg = reg.with(console_subscriber::spawn());
```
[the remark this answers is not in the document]
Sushisource: Enables https://github.com/tokio-rs/console if I want to use it for debugging

core/src/pollers/poll_buffer.rs
Sushisource: It's just a typo, it should be max

PR #560, opened 2023-05-25
WIP: Added feature flag for ephemeral server module

h7kanna:
## What was changed
Added a feature flag for ephemeral server functionality.
Made it default to not break the existing usage.

## Why?
To optimize compilation time for pure Rust usage of SDK.

## Checklist

1. Closes #559

3. How was this tested:
Run the existing unit/integration/load tests, 

4. Any docs updates needed?
   No

PR #561, opened 2023-05-25
Added feature flag for ephemeral server module

### LINKED ISSUE #559: [Feature Request] Move ephemeral server behind feature flag

As ephemeral_server is only used for testing, Can it be moved into a separate crate to reduce deps in the core and optimize compilation time for user code for using client and worker?

### Additional context

I am trying to optimize compilation time for worker code, and zip dependency of ephemeral_server is big contributor

h7kanna:
## What was changed
Added a feature flag for ephemeral server functionality.

## Why?
To optimize compilation time for pure Rust usage of SDK.
This is same as https://github.com/temporalio/sdk-core/pull/560 but without default

## Checklist

1. Closes #559

3. How was this tested:
Run the existing unit/integration/load tests, 

4. Any docs updates needed?
   No

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -13,11 +13,13 @@ categories = ["development-tools"]
 [lib]
 
 [features]
+default = []
 # Do not enable this feature when building production SDKs. If we ever want a user in the field to
 # record WF input data, we can build them a custom SDK or they can build - it adds significant extra
 # code size in the form of [de]serializers.
 save_wf_inputs = ["rmp-serde", "temporal-sdk-core-protos/serde_serialize"]
 tokio-console = ["console-subscriber"]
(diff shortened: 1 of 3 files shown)


cretz:
All "official" uses of this crate need this feature that they'll now opt-in to. Will wait for @Sushisource to weigh in here and on #559. Maybe we can use find cheaper-to-compile zip dependency or another dependency or acknowledge this crate is unstable for external use currently.

Sushisource:
@cretz We can talk about this in standup next week. It is super easy to adjust the lang-side to just set this option, so perhaps that makes sense. OTOH, yeah, those are the primary consumer. If you & the rest of the team are fine just passing the flag, it seems reasonable to do this.

Sushisource:
@h7kanna We decided we're fine with having to set this flag in our language layers. Thanks!

Sushisource (approved):

PR #566, opened 2023-06-07
Build id updates

Sushisource:
Various changes to handle the latest updates to the versioning API


client/src/lib.rs
```diff
@@ -527,6 +529,8 @@ pub struct Client {
     namespace: String,
     /// If set, attach as the worker build id to relevant calls
     bound_worker_build_id: Option<String>,
+    /// If set, task completions will include the appropriate version stamp
+    use_versioning: bool,
```
bergundy, line 533: It's a bit weird that all of this is on the client and not the worker.

bergundy, line 533: For someone advocating for making illegal states unrepresentable in code, I would have expected you to use an enum such as:
```
Unversioned,
UnversionedWithBuildId(string),
Versioned(string),
```

client/src/lib.rs
```diff
@@ -1095,6 +1119,14 @@ impl WorkflowClientTrait for Client {
             binary_checksum: self.bound_worker_build_id.clone().unwrap_or_default(),
```
bergundy, line 1119: You shouldn't send both binary checksum and build ID, decide which one to send based on server capabilities.

core/src/core_tests/workflow_tasks.rs
```diff
@@ -2688,3 +2683,88 @@ async fn poller_wont_poll_until_lang_polls() {
...
+    });
+    let compat_flag_expected = match intent {
+        VersioningIntent::Unspecified => !different_tq,
+        VersioningIntent::Compatible => true,
+        VersioningIntent::Default => false,
+    };
+    mock_client
+        .expect_complete_workflow_task()
+        .returning(move |mut c| {
+            let can_cmd = c.commands.pop().unwrap().attributes.unwrap();
```
bergundy, line 2718: `can_cmd`?

core/src/core_tests/workflow_tasks.rs
```diff
@@ -2688,3 +2683,88 @@ async fn poller_wont_poll_until_lang_polls() {
...
+    let worker = Worker::new_test(test_worker_cfg().build().unwrap(), mock_client);
+    let r = worker.poll_workflow_activation().await.unwrap();
+    let task_queue = if different_tq {
+        "enchi cat!".to_string()
+    } else {
+        "".to_string()
+    };
+    let cmd = match command_type {
+        "continue_as_new" => ContinueAsNewWorkflowExecution {
+            workflow_type: "meow".to_string(),
```
bergundy, line 2743: The best workflow type I've seen to this day.

core/src/worker/client.rs
```diff
@@ -196,6 +190,7 @@ impl WorkerClient for WorkerClientBag {
             worker_version_stamp: Some(WorkerVersionStamp {
                 build_id: self.versioning_build_id(),
                 bundle_id: "".to_string(),
+                use_versioning: self.use_versioning,
             }),
             messages: vec![],
             binary_checksum: self.worker_build_id.clone(),
```
bergundy, line 196: Don't send binary checksum unnecessarily.

core/src/worker/mod.rs
```diff
@@ -262,8 +265,11 @@ impl Worker {
                     let sticky_metrics = metrics.with_new_attrs([workflow_sticky_poller()]);
                     new_workflow_task_buffer(
                         client.clone(),
-                        sqn.clone(),
-                        true,
+                        TaskQueue {
+                            name: sqn.clone(),
```
bergundy, line 269: Looks like we still embed the "normal" name in the sticky queue name, no need for that anymore (especially if the server supports build id based versioning and is aware of the normal_name field).

Sushisource: Huh? That's explicitly when we _do_ need it -- otherwise why does the field even exist at all? We added normal name for this purpose.

core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -1050,8 +1053,16 @@ impl WorkflowMachines {
                 }
                 WFCommand::AddActivity(attrs) => {
                     let seq = attrs.seq;
+                    let use_compat = self.determine_use_compatible_flag(
```
bergundy, line 1056: Would have been cleaner if the workflow machines context (with the task queue) propagated to the internal state machines and more of the logic moved in there.

core/src/worker/mod.rs
bergundy: I mean this: https://github.com/temporalio/sdk-core/blob/31099415d28f983350d3cf1d005705de037ccb8d/core/src/lib.rs#L152

core/src/worker/workflow/machines/workflow_machines.rs
Sushisource: I agree but I want to deal with that more generally. I want to do a change where we stop cloning a bunch of these strings everywhere and instead pass some refcounted propery bag type things. I made an item https://github.com/temporalio/sdk-core/issues/567

core/src/worker/mod.rs
Sushisource: Oooooh, that. Yeah... I suppose that should be fine to change.

core/src/core_tests/workflow_tasks.rs
bergundy: Looks like some commit was lost, I saw this fixed already.

core/src/worker/client.rs
```diff
@@ -48,11 +48,39 @@ impl WorkerClientBag {
             use_versioning,
         }
     }
-    fn versioning_build_id(&self) -> String {
-        if self.use_versioning {
+
+    fn default_capabilities(&self) -> Capabilities {
```
bergundy, line 52: nit: Why don't you just initialize it as empty and overwrite when system info resolves instead of dealing with optionals?

core/src/worker/client.rs
```diff
@@ -239,6 +248,11 @@ impl WorkerClient for WorkerClientBag {
                 result,
                 identity: self.identity.clone(),
                 namespace: self.namespace.clone(),
+                worker_version: Some(WorkerVersionStamp {
```
bergundy, line 251: self.version_stamp

core/src/worker/client.rs
```diff
@@ -275,6 +289,11 @@ impl WorkerClient for WorkerClientBag {
                 details,
                 identity: self.identity.clone(),
                 namespace: self.namespace.clone(),
+                worker_version: Some(WorkerVersionStamp {
```
bergundy, line 292: `self.worker_version_stamp()`

core/src/worker/client.rs
```diff
@@ -311,9 +335,14 @@ impl WorkerClient for WorkerClientBag {
             cause: cause as i32,
             failure,
             identity: self.identity.clone(),
-            binary_checksum: self.worker_build_id.clone(),
+            binary_checksum: self.binary_checksum(),
             namespace: self.namespace.clone(),
             messages: vec![],
+            worker_version: Some(WorkerVersionStamp {
```
bergundy, line 341: Same

core/src/worker/client.rs
Sushisource: There are other places which like to know if it was ever set by reading from server

bergundy (approved):

Issue #564, opened 2023-06-07
[Feature Request] Expose history_size_in_bytes and should_continue_as_new


lorensr:
https://github.com/temporalio/features/issues/16

PR #570, opened 2023-06-14
Expose history size in bytes and CAN suggestion

Sushisource:
## What was changed
Expose these attributes from WFT started events

## Why?
Seeing these can be desirable when writing entity workflows

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/564
2. How was this tested:
Verified in UT

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/machines/workflow_machines.rs
+++ b/core/src/worker/workflow/machines/workflow_machines.rs
@@ -116,7 +116,7 @@ pub(crate) struct WorkflowMachines {
     /// Set on each WFT started event, the most recent size of history in bytes
     history_size_bytes: u64,
     /// Set on each WFT started event
-    continue_as_new_suggseted: bool,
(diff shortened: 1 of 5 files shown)


core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -110,6 +113,10 @@ pub(crate) struct WorkflowMachines {
     /// The internal flags which have been seen so far during this run's execution and thus are
     /// usable during replay.
     observed_internal_flags: InternalFlagsRef,
+    /// Set on each WFT started event, the most recent size of history in bytes
+    history_size_bytes: u64,
+    /// Set on each WFT started event
+    continue_as_new_suggseted: bool,
```
cretz, line 119:
```suggestion
    continue_as_new_suggested: bool,
```

cretz (approved):

Issue #575, opened 2023-06-27
[Bug] Local activity fails after first attempt if start-to-close timeout expires


bergundy:
Simple repro here:
https://github.com/temporalio/sdk-typescript/commit/e37ae758d671fa05e0e898e3eb05aa1b5b98a914

Expected behavior:
Activity keeps retrying.

PR #576, opened 2023-06-28
La start close retry

Sushisource:
## What was changed
Fix start-to-close LA timeouts not being retryable like they should be.

## Why?
Incorrect behavior.

## Checklist

- [ ] Address one remaining TODO.
- [ ] Fix a few integ test issues, should be easy.

1. Closes https://github.com/temporalio/sdk-core/issues/575

2. How was this tested:
Added unit test & existing tests

3. Any docs updates needed?

### DIFF
--- a/core/src/core_tests/local_activities.rs
+++ b/core/src/core_tests/local_activities.rs
@@ -887,16 +887,11 @@ async fn test_schedule_to_start_timeout_not_based_on_original_time(
 #[tokio::test]
 async fn start_to_close_timeout_allows_retires() {
     crate::telemetry::test_telem_console();
+    // TODO: Also fail completely
     let mut t = TestHistoryBuilder::default();
     t.add_by_type(EventType::WorkflowExecutionStarted);
     t.add_full_wf_task();
-    t.add_local_activity_marker(
-        1,
-        "1",
-        None,
-        Some(Failure::application_failure("la failed".to_string(), false)),
-        |_| {},
-    );
+    t.add_local_activity_marker(1, "1", Some("hi".into()), None, |_| {});
     t.add_full_wf_task();
     t.add_workflow_execution_completed();
 
@@ -918,13 +913,13 @@ async fn start_to_close_timeout_allows_retires() {
                     activity_type: DEFAULT_ACTIVITY_TYPE.to_string(),
                     input: "hi".as_json_payload().expect("serializes fine"),
                     retry_policy: RetryPolicy {
-                        initial_interval: Some(prost_dur!(from_millis(50))),
+                        initial_interval: Some(prost_dur!(from_millis(20))),
                         backoff_coefficient: 1.0,
                         maximum_interval: None,
                         maximum_attempts: 5,
                         non_retryable_error_types: vec![],
                     },
-                    start_to_close_timeout: Some(prost_dur!(from_millis(50))),
+                    start_to_close_timeout: Some(prost_dur!(from_millis(25))),
                     ..Default::default()
                 })
                 .await;
@@ -937,7 +932,7 @@ async fn start_to_close_timeout_allows_retires() {
         DEFAULT_ACTIVITY_TYPE,
         move |_ctx: ActContext, _: String| async move {
             // Timeout the first 4 attempts
-            if ATTEMPTS.fetch_add(1, Ordering::AcqRel) < 4 {
+            if dbg!(ATTEMPTS.fetch_add(1, Ordering::AcqRel)) < 4 {
                 tokio::time::sleep(Duration::from_millis(100)).await;
             }
             Ok(())
--- a/core/src/core_tests/local_activities.rs
+++ b/core/src/core_tests/local_activities.rs
@@ -885,6 +885,79 @@ async fn test_schedule_to_start_timeout_not_based_on_original_time(
 }
 
 #[tokio::test]
+async fn start_to_close_timeout_allows_retires() {
+    crate::telemetry::test_telem_console();
+    let mut t = TestHistoryBuilder::default();
+    t.add_by_type(EventType::WorkflowExecutionStarted);
+    t.add_full_wf_task();
+    t.add_local_activity_marker(
+        1,
+        "1",
+        None,
+        Some(Failure::application_failure("la failed".to_string(), false)),
+        |_| {},
+    );
+    t.add_full_wf_task();
+    t.add_workflow_execution_completed();
+
+    let wf_id = "fakeid";
+    let mock = mock_workflow_client();
+    let mh = MockPollCfg::from_resp_batches(
+        wf_id,
+        t,
+        [ResponseType::ToTaskNum(1), ResponseType::AllHistory],
+        mock,
+    );
+    let mut worker = mock_sdk_cfg(mh, |w| w.max_cached_workflows = 1);
+
+    worker.register_wf(
+        DEFAULT_WORKFLOW_TYPE.to_owned(),
+        move |ctx: WfContext| async move {
+            let la_res = ctx
+                .local_activity(LocalActivityOptions {
+                    activity_type: DEFAULT_ACTIVITY_TYPE.to_string(),
+                    input: "hi".as_json_payload().expect("serializes fine"),
+                    retry_policy: RetryPolicy {
+                        initial_interval: Some(prost_dur!(from_millis(50))),
+                        backoff_coefficient: 1.0,
+                        maximum_interval: None,
+                        maximum_attempts: 5,
+                        non_retryable_error_types: vec![],
+                    },
+                    start_to_close_timeout: Some(prost_dur!(from_millis(50))),
+                    ..Default::default()
+                })
+                .await;
+            assert!(la_res.completed_ok());
+            Ok(().into())
+        },
+    );
+    static ATTEMPTS: AtomicUsize = AtomicUsize::new(0);
+    worker.register_activity(
+        DEFAULT_ACTIVITY_TYPE,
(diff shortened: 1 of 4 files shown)


core/src/worker/activities/local_activities.rs
```diff
@@ -531,21 +533,67 @@ impl LocalActivityManager {
...
-                    ) {
+                    ),
+                    is_timeout: false,
+                },
+                LocalActivityExecutionResult::TimedOut(fail)
+                    if matches!(status.get_timeout_type(), Some(TimeoutType::StartToClose)) =>
+                {
+                    // Start to close timeouts are retryable, other timeout types aren't.
+                    Outcome::FailurePath {
+                        backoff: info.la_info.schedule_cmd.retry_policy.should_retry(
```
Sushisource, line 558: Also dedupe this stupid bit

core/src/worker/activities/local_activities.rs
```diff
@@ -368,46 +382,30 @@ impl LocalActivityManager {
...
+                                seq_num: resolution.seq,
+                            })
+                            .as_ref()
+                            .map(|lai| lai.task_token.clone());
+                        if let Some(task_token) = tt {
+                            Some(NextPendingLAAction::Autocomplete(
+                                self.complete(&task_token, resolution.result),
+                            ))
                         } else {
+                            // This timeout is for a no-longer tracked activity, so, whatever
```
paulnpdev, line 406:
```suggestion
                            // This timeout is for a no-longer-tracked activity, so, whatever
```
is this a nit or what? :)

core/src/worker/activities/local_activities.rs
```diff
@@ -993,10 +1080,10 @@ mod tests {
         let tt = TaskToken(next.task_token);
         let res = lam.complete(
             &tt,
-            &LocalActivityExecutionResult::Failed(Default::default()),
+            LocalActivityExecutionResult::Failed(Default::default()),
         );
-        assert_matches!(res, LACompleteAction::LangDoesTimerBackoff(dur, info)
-            if dur.seconds == 10 && info.attempt == 5
+        assert_matches!(res, LACompleteAction::Report{resolution,..}
```
paulnpdev, line 1085:
```suggestion
        assert_matches!(res, LACompleteAction::Report{resolution, ..}
```

core/src/worker/activities/local_activities.rs
```diff
@@ -855,15 +933,24 @@ mod tests {
...
                 _ => {
                     panic!("Non-dispatched action returned")
                 }
             }
         }
+
+        fn is_timeout(&self, with_task: bool) -> bool {
+            matches!(
+                self,
+                Self::Autocomplete(LACompleteAction::Report{ task, resolution, ..})
```
paulnpdev, line 949:
```suggestion
                Self::Autocomplete(LACompleteAction::Report{task, resolution, ..})
```

core/src/worker/activities/local_activities.rs
```diff
@@ -368,46 +382,30 @@ impl LocalActivityManager {
...
+                            .lock()
+                            .la_info
+                            .get(&ExecutingLAId {
+                                run_id,
+                                seq_num: resolution.seq,
+                            })
+                            .as_ref()
+                            .map(|lai| lai.task_token.clone());
+                        if let Some(task_token) = tt {
+                            Some(NextPendingLAAction::Autocomplete(
```
bergundy, line 402: What does autocomplete do here?
We should cancel the lang activity if it does not complete within the given timeout.

core/src/worker/activities/local_activities.rs
```diff
@@ -531,37 +533,92 @@ impl LocalActivityManager {
...
-                    self.complete_notify.notify_one();
-                    LACompleteAction::Report(info)
-                }
-                LocalActivityExecutionResult::Failed(f) => {
-                    if let Some(backoff_dur) = info.la_info.schedule_cmd.retry_policy.should_retry(
+            enum Outcome {
+                FailurePath { backoff: Option<Duration> },
+                JustReport,
+            }
+            macro_rules! calc_backoff {
```
bergundy, line 540: Wy do you need a macro for this, seems like overkill.

core/src/worker/activities.rs
```diff
@@ -3,9 +3,8 @@ mod activity_task_poller_stream;
 mod local_activities;
 
 pub(crate) use local_activities::{
-    DispatchOrTimeoutLA, ExecutingLAId, LACompleteAction, LocalActRequest,
-    LocalActivityExecutionResult, LocalActivityManager, LocalActivityResolution,
-    LocalInFlightActInfo, NewLocalAct,
+    ExecutingLAId, LACompleteAction, LocalActRequest, LocalActivityExecutionResult,
```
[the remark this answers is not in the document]
cretz: `cargo fmt` should do it for you IIRC

core/src/worker/activities/local_activities.rs
bergundy: Hmmm... it seems like it does get cancelled below

core/src/core_tests/local_activities.rs
```diff
@@ -884,6 +884,96 @@ async fn test_schedule_to_start_timeout_not_based_on_original_time(
     worker.run_until_done().await.unwrap();
 }
 
+#[rstest::rstest]
+#[tokio::test]
+async fn start_to_close_timeout_allows_retires(#[values(true, false)] la_completes: bool) {
```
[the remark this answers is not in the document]
Sushisource: Nice catch

core/src/worker/activities.rs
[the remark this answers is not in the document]
Sushisource: Formatter does what it does, we don't question it. Everything is autoformatted. These fall into the category of "I never look at this ever".

core/src/worker/activities/local_activities.rs
Sushisource: Interesting point, these actually _don't_ get autoformatted, since they're inside a macro.

core/src/worker/activities/local_activities.rs
Sushisource: That's exactly what happens. This is just a mechanism for feeding the timeout back as a completion so it's easier to re-use code.

core/src/worker/activities/local_activities.rs
Sushisource: There's no way to simply re-use the code besides this because things would be double-moved.

core/src/worker/activities/local_activities.rs
```diff
@@ -531,37 +533,92 @@ impl LocalActivityManager {
...
-                                info,
-                            );
+                            // This la needs to write a failure marker, and then we will tell lang how
+                            // long of a timer to schedule to back off for. We do this because there are
+                            // no other situations where core generates "internal" commands so it is
+                            // much simpler for lang to reply with the timer / next LA command than to
+                            // do it internally. Plus, this backoff hack we'd like to eliminate
+                            // eventually.
+                            resolution.backoff =
+                                Some(backoff_dur.try_into().expect("backoff fits into proto"));
```
bergundy, line 616: When would try_into fail?

core/src/core_tests/local_activities.rs
```diff
@@ -884,6 +884,96 @@ async fn test_schedule_to_start_timeout_not_based_on_original_time(
...
+            Ok(().into())
+        },
+    );
+    let attempts: &'static _ = Box::leak(Box::new(AtomicUsize::new(0)));
+    let cancels: &'static _ = Box::leak(Box::new(AtomicUsize::new(0)));
+    worker.register_activity(
+        DEFAULT_ACTIVITY_TYPE,
+        move |ctx: ActContext, _: String| async move {
+            // Timeout the first 4 attempts, or all of them if we intend to fail
+            if attempts.fetch_add(1, Ordering::AcqRel) < 4 || !la_completes {
```
bergundy, line 949: Attempt should be on the activity context

core/src/core_tests/local_activities.rs
```diff
@@ -884,6 +884,96 @@ async fn test_schedule_to_start_timeout_not_based_on_original_time(
     worker.run_until_done().await.unwrap();
 }
 
+#[rstest::rstest]
+#[tokio::test]
+async fn start_to_close_timeout_allows_retries(#[values(true, false)] la_completes: bool) {
```
bergundy, line 889: Why not make this an integration test instead of crafting a history?

bergundy (approved):

core/src/worker/activities/local_activities.rs
Sushisource: It won't, you'd have to have a backoff of thousands of years. Proto internally uses i64 and Duration uses u64

core/src/core_tests/local_activities.rs
Sushisource: Yeah, eventually.

core/src/core_tests/local_activities.rs
Sushisource: A few reasons:
* Speed, it's nice to keep integration test suite small, because that means it's fast. If it doesn't need to be an IT, I can keep that. Running this test by itself is (slightly) faster too w/o the server start/stop.
* Easier to use w/ a debugger
* What needs to happen here is clear, so I can pretty easily test just that

core/src/worker/activities/local_activities.rs
Sushisource: Yeah, probably truncate down is the way to go.

Issue #568, opened 2023-06-08
[Bug] A child workflow with CancellationType.Abandon completing after cancelling the parent leads to an invalid state


PenguinToast:
### What are you really trying to do?

See #523

### Describe the bug

A pretty similar case as #523 -- if a child workflow finishes after the parent workflow has already received a cancellation request, then the workflow ends up in stuck state with the error:
```
Fatal(\"ChildWorkflowMachine in state Cancelled says the transition is invalid during event EventInfo { event_id: 57, event_type: ChildWorkflowExecutionCompleted }\")
```

It seems like we should either make all transitions for the child workflow state machine out of `Cancelled` valid (for child workflow completion, cancellation, failure, or termination), or not put the child workflow state machine in the state `Cancelled` pre-emptively if the `CancellationType` is `Abandon`

### Environment/Versions

Running on Temporal version 1.18 with Typescript SDK 1.7.0

PenguinToast:
@Sushisource happy to help implement a fix here -- just need direction on what you think is the right way to fix this!

Sushisource:
@PenguinToast Probably the right approach here is either a separate `CancelledAbandoned` state, or to just look at the cancel type and be OK with swallowing a transition like this while in `Cancelled` if we had the abandoned type.

I should be able to fix this fairly soon (next week sometime), but if you want to try a fix, by all means go for it :)

PR #580, opened 2023-07-21
Fix abandoned children nondeterminism if they complete

Sushisource:
* Also fix arch diagram & compiler err changes / rust ver in CI

## Why?
Abandoned children can complete/fail/timeout/terminate & that's fine. Don't barf.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/568

2. How was this tested:
IT & UT

3. Any docs updates needed?
nawp


core/src/worker/workflow/machines/child_workflow_state_machine.rs
```diff
@@ -79,6 +82,15 @@ fsm! {
     Cancelled --(Cancel) --> Cancelled;
     Cancelled --(ChildWorkflowExecutionCancelled,
         on_child_workflow_execution_cancelled) --> Cancelled;
+    // Completions of any kind after cancellation are acceptable for abandoned children
+    Cancelled --(ChildWorkflowExecutionCompleted(Option<Payloads>),
```
cretz, line 86: Wonder if one day you may want to expand this macro format to build in "guards" so you can just add an if-abandoned conditional

core/src/worker/workflow/machines/child_workflow_state_machine.rs
```diff
@@ -120,6 +132,14 @@ pub(super) struct Cancelled {
     seen_cancelled_event: bool,
 }
 
+fn completion_of_not_abandoned_err() -> WFMachinesError {
+    WFMachinesError::Nondeterminism(
+        "Child workflows which don't have the ABANDON cancellation type cannot complete after \
```
cretz, line 137: Hrmm, I figured `ABANDON` was a parent-close-only thing. Arguably this is a lang bug if they send you a cancel-after-complete of any child/activity regardless of abandon setting. What situation/lang is sending you cancel of completed-yet-abandoned children?

core/src/worker/workflow/machines/child_workflow_state_machine.rs
Sushisource: Maybe. I don't want to make the DSL to complex when you can handle that in the handler easily.

core/src/worker/workflow/machines/child_workflow_state_machine.rs
Sushisource: Lang shouldn't have to think about that, that's the point of core. And this isn't cancel after-complete, this is complete after cancel, which is outside of SDK control entirely. The activity never sees a cancel, and thus never knows it shouldn't complete.

cretz (approved):

PR #584, opened 2023-08-02
Update otel and tonic

Sushisource:
## What was changed
Updated deps & made a bunch of fixes due to OTel breaking changes in metrics API

Notably, I had to add a `attach_service_name` field to TelemetryOptions which defaults to true to maintain backwards compatibility w.r.t. the `service_name` label. See the docstring for it.

## Why?
Been waiting to update these for a while
Allow recording min/max for histograms

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core-api/src/telemetry.rs
+++ b/core-api/src/telemetry.rs
@@ -45,6 +45,13 @@ pub struct TelemetryOptions {
     // A map of tags to be applied to all metrics
     #[builder(default)]
     pub global_tags: HashMap<String, String>,
+
+    /// If set true (the default) explicitly attach a `service_name` label to all metrics. Turn this
+    /// off if your collection system supports the `target_info` metric from the OpenMetrics spec.
+    /// For more, see
+    /// [here](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems)
+    #[builder(default = "true")]
+    pub attach_service_name: bool,
 }
 
 /// Options for exporting to an OpenTelemetry Collector
@@ -71,6 +78,7 @@ pub struct TraceExportConfig {
 /// Control where traces are exported.
 #[derive(Debug, Clone)]
 pub enum TraceExporter {
+    // TODO: Remove
     /// Export traces to an OpenTelemetry Collector <https://opentelemetry.io/docs/collector/>.
     Otel(OtelCollectorOptions),
 }
--- a/core/src/telemetry/metrics.rs
+++ b/core/src/telemetry/metrics.rs
@@ -1,16 +1,15 @@
 use crate::telemetry::TelemetryInstance;
 use opentelemetry::{
+    self,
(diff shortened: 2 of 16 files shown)


core-api/src/telemetry.rs
```diff
@@ -71,6 +78,7 @@ pub struct TraceExportConfig {
 /// Control where traces are exported.
 #[derive(Debug, Clone)]
 pub enum TraceExporter {
+    // TODO: Remove
```
Sushisource, line 81: Will be done in metrics export PR

cretz (approved):
Safe to assume no user-visible changes? (I know we will expose the attach service name option, but just want to confirm no incompatible changes)

core/src/telemetry/prometheus_server.rs
```diff
@@ -1,52 +1,47 @@
...
+            },
+            exporter.build()?,
+        ))
     }
 
     pub async fn run(self) -> hyper::Result<()> {
         // Spin up hyper server to serve metrics for scraping. We use hyper since we already depend
         // on it via Tonic.
-        let expclone = self.exporter.clone();
+        let regclone = self.registry.clone();
```
dandavison, line 41: This clone isn't needed.

dandavison (approved):
Nice! Looks like it was way more of a pain than one would have hoped. (This helps with ginepro, if we end up using that, since latest ginepro needs latest tonic.)

core/src/telemetry/metrics.rs
```diff
@@ -428,44 +405,104 @@ static TASK_SCHED_TO_START_MS_BUCKETS: &[f64] =
...
+        )?)
+        .with_view(histo_view(ACT_EXEC_LATENCY_NAME, ACT_EXE_MS_BUCKETS)?))
+}
+
+/// OTel has no built-in synchronous Gauge. Histograms used to be able to serve that purpose, but
+/// they broke that. Lovely. So, we need to implement one by hand.
+pub(crate) struct MemoryGaugeU64 {
+    labels_to_values: Arc<RwLock<HashMap<AttributeSet, u64>>>,
+}
+
```
dandavison, line 477: I didn't immediately see test coverage for this. Worth adding some?

core/src/telemetry/prometheus_server.rs
Sushisource: It is - otherwise there's no way to move the registry into the closure w/o capturing `self`

core/src/telemetry/metrics.rs
Sushisource: It's covered by some existing tests which verify that the prometheus endpoint has the expected metrics.

Sushisource:
@cretz No there should not be any breaking changes for users that I could tell. I will probably add a few other prometheus endpoint tests with the exporter PR just to make extra sure

core/src/telemetry/prometheus_server.rs
Sushisource: NM, you're right - forgot this fn is consuming self

Issue #535, opened 2023-04-18
[Feature Request] Permit the removal + change of adjacent deprecated patches


Sushisource:
### Is your feature request related to a problem? Please describe.

User had 2 consecutive patches in history `deprecatePatch('patch-1')` followed by `patched('patch-2')` and deployed a change removing the first deprecated patch while deprecating patch-2 and tripped the check in Core that verifies that a patch ID doesn't change during replay.

### Describe the solution you'd like

Ideally, it should be possible to make this change. A couple options:

* When it comes to nondeterminism checking specifically for deprecated patches, if there is a patch-id conflict, assume that this was because of the removal of a previously-adjacent deprecated patch, and allow it.
* Allow patches to be emitted / applied in any order. This is generally safe (since we perform lookahead anyway, and activated patches are sent in a batch for a given WFT), but a more invasive change.

Sushisource:
@rupalivohra For now you should be able to do this as long as you only change one patch at a time. I believe (I haven't had time to confirm this with tests yet though) that this problem only arises if you make simultaneous changes to two adjacent patches.

rupalivohra:
> this problem only arises if you make simultaneous changes to two adjacent patches.

Unfortunately it's difficult to guarantee that there aren't two `deprecatePatch()` calls in all cases, and safer to simply not delete the code given that there's no recovery from the bad state.

PR #587, opened 2023-08-08
Fix removal of deprecated patch call adjacent to other patch

Sushisource:
## What was changed
Permit the removal of deprecated patch calls which are immediately adjacent to another patch call.

This was done by allowing patch-id-mistmaches in when checking a patch event against a patch machine, if the event is deprecated. I do not see this as a risky change primarily because nondeterminism checking of this kind purely serves as an "early warning" against other subsequent nondeterminism which would happen as a result of taking the incorrect branch. So, although this change technically means you could swap out one deprecated patch for another and there would be no error on that line - you'd still likely encounter another one later as if you hadn't used patching at all.

I think this is an acceptable tradeoff, since I cannot think of any other way to make this work and at least a couple people have run into it.

## Why?
Without this, it can be quite challenging to remove deprecated patches in some circumstances.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/535
2. How was this tested:
Added IT, existing tests.

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/machines/workflow_machines.rs
+++ b/core/src/worker/workflow/machines/workflow_machines.rs
@@ -1409,14 +1409,12 @@ fn patch_marker_handling(
                 )))
             }
         }
+    } else if patch_machine.is_some() {
+        debug!("Skipping non-matching event against patch machine");
+        Ok(EventHandlingOutcome::SkipCommand)
     } else {
-        if patch_machine.is_some() {
-            debug!("Skipping non-matching event against patch machine");
-            Ok(EventHandlingOutcome::SkipCommand)
-        } else {
-            // Not a patch machine or a patch event
-            Ok(EventHandlingOutcome::Normal)
-        }
+        // Not a patch machine or a patch event
+        Ok(EventHandlingOutcome::Normal)
     }
 }
 
--- a/core/src/worker/workflow/machines/workflow_machines.rs
+++ b/core/src/worker/workflow/machines/workflow_machines.rs
@@ -138,7 +138,7 @@ pub(crate) struct WorkflowMachines {
     current_wf_task_commands: VecDeque<CommandAndMachine>,
 
     /// Information about patch markers we have already seen while replaying history
-    encountered_change_markers: HashMap<String, ChangeInfo>,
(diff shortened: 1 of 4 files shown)


cretz:
I admit not understanding the problem well enough (but my understanding may not be required, but if my understanding is required, problematic TS/Python code on patch change could help me). I'll defer review.

> Permit the removal [...] I do not see this as a risky change [...]

Sounds like this is a harmless change simply because you're loosening restrictions, correct?

Sushisource:
> Sounds like this is a harmless change simply because you're loosening restrictions, correct?

Yes. I wouldn't call it completely harmless. The potential harm is in making the reason something has gone wrong somewhat less obvious - but it's not harmful from a "bad things will happen" perspective.

core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -1360,45 +1360,62 @@ enum EventHandlingOutcome {
...
-    if !mach.matches_event(event) {
-        // Version markers can be skipped in the event they are deprecated
-        if let Some((patch_name, deprecated)) = event.get_patch_marker_details() {
+    let patch_machine = match mach {
+        Machines::PatchMachine(pm) => Some(pm),
+        _ => None,
+    };
+    let patch_details = event.get_patch_marker_details();
+    fn skip_one_or_two_events(next_event: Option<&HistoryEvent>) -> Result<EventHandlingOutcome> {
+        // Also ignore the subsequent upsert event if present
```
bergundy, line 1374: Not introduced in this PR, but...
Are there cases where upsert search attributes follows a patch but the patch machine is not the one that issued that command?

I _think_ it's only possible if a patch was emitted before we started emitting SAs on patches but still worth supporting for correctness.

bergundy (approved):

core/src/worker/workflow/machines/workflow_machines.rs
Sushisource: It's probably technically possible, but, this check looks to see for the specific attribute key for versioning. So this could only happen if the user explicitly did that themselves somehow, which: A) would be super unlikely and B) means I have no way to differentiate anyways so nothing can be done.

Issue #595, opened 2023-08-21
[Bug] Missing cfg feature line in core-api/src/telemetry/metrics.rs


lee-aaron:
### What are you really trying to do?
I'm just playing around with the SDK in Rust and using the client to connect to a temporal server

### Describe the bug
I'm seeing a compile error if I don't have the otel_impls feature.

```
#[cfg(feature = "otel_impls")]
```

in core-api/src/telemetry/metrics.rs line 118

    error[E0599]: no variant named `OTel` found for enum `MetricAttributes`
       --> /Users/aaronlee/.cargo/git/checkouts/sdk-core-3d5ba50f52602518/8625016/core-api/src/telemetry/metrics.rs:118:32
        |
    96  | pub enum MetricAttributes {
        | ------------------------- variant `OTel` not found here
    ...
    118 |             (MetricAttributes::OTel { ref mut kvs }, MetricAttributes::OTel { kvs: other_kvs }) => {
        |                                ^^^^ variant not found in `MetricAttributes`
    
    error[E0599]: no variant named `OTel` found for enum `MetricAttributes`
       --> /Users/aaronlee/.cargo/git/checkouts/sdk-core-3d5ba50f52602518/8625016/core-api/src/telemetry/metrics.rs:118:72
        |
    96  | pub enum MetricAttributes {
        | ------------------------- variant `OTel` not found here
    ...
    118 |             (MetricAttributes::OTel { ref mut kvs }, MetricAttributes::OTel { kvs: other_kvs }) => {
        |                                                                        ^^^^ variant not found in `MetricAttributes`I believe the solution is just adding that cfg feature above the match case?

### Minimal Reproduction
Just import the sdk core repo in another Rust project and specify the client package.

```
Workspace Cargo.toml
[workspace.dependencies]
temporal-client = { git = "https://github.com/temporalio/sdk-core" }

Subpackage Cargo.toml
[dependencies]
temporal-client = { workspace = true }
```

### Environment/Versions

* OS and processor: M1 Mac
* Temporal SDK Version: Latest
* Are you using Docker or Kubernetes or building Temporal from source? N/A

### Additional context
I have a fix on my branch so let me know if this is ok to make a PR for / merge. Thanks!

Sushisource:
Thanks for the catch, I'd be happy to take your PR

lee-aaron:
I linked the PR. Let me know if there's anything else to be done. Thanks

PR #596, opened 2023-08-21
Add missing cfg feature tag

lee-aaron:
## What was changed

Adds a cfg feature tag in the match statement

## Why?

If you don't have the otel_impls feature and you're using the Rust SDK client, it will fail to compile.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/595
1. How was this tested:

`cargo check`
1. Any docs updates needed?

None


Sushisource (approved):
Thanks for this!

Issue #589, opened 2023-08-10
[Bug] Activation only ever get called with `is_replaying = true`


mjameswh:
### Describe the bug
It is expected that the very first time a workflow event is ever dispatched, the flag `is_replaying` must be `false`. Yet, in some particular cases involving a patch and a query right after a cache eviction, Core may dispatch an activation with `is_replaying = true`, even though that activation was never dispatched before.

### Minimal Reproduction
See https://github.com/mjameswh/sdk-typescript/tree/issue-core-589 for a demo.

```
npm i
npm run build
cd package/test
RUN_INTEGRATION_TESTS=true npm exec ava -- lib/test-sinks.js --match 'Core issue 589'
```

Also see [core-log-589.txt](https://github.com/temporalio/sdk-core/files/12317130/core-log-589.txt) for Core's debug log output for execution of this test on my machine (trimmed to include only log messages from the second worker, as log messages from the first worker seems irrelevant).

Note in particular the following log entry, which is the only activation ever sent for the `FireTimer` event. Yet, `is_replaying` is unexpectedly `true`.

```
2023-08-10T20:49:11.923276Z DEBUG next_workflow_activation: temporal_sdk_core::worker::workflow: Sending activation to lang activation=WorkflowActivation(run_id: dd99eed3-e761-414f-b15a-3dbed9730f3f, is_replaying: true, jobs: FireTimer(1)) task_queue=MyQueue run_id="dd99eed3-e761-414f-b15a-3dbed9730f3f"
```

This is the only activation ever sent

PR #597, opened 2023-08-21
Fix query related is_replay flag error

Sushisource:
## What was changed
In some situations server could issue WFTs with queries that also contained the last WFT complete & associated commands. This would confuse the code that determines if we are replaying or not.

Also make all activations containing only queries always be replay.

## Why?
Fixes incorrect semantics

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/589

2. How was this tested:
Added UT, verified existing tests. Fixes James' repro in the bug.

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/machines/workflow_machines.rs
+++ b/core/src/worker/workflow/machines/workflow_machines.rs
@@ -367,9 +367,17 @@ impl WorkflowMachines {
     /// "no work" situation. Possibly, it may know about some work the machines don't, like queries.
     pub(crate) fn get_wf_activation(&mut self) -> WorkflowActivation {
         let jobs = self.drive_me.drain_jobs();
+        // Even though technically we may have satisfied all the criteria to be done with replay,
+        // query only activations are always "replaying" to keep things sane.
+        let all_query = jobs.iter().all(|j| {
+            matches!(
+                j.variant,
+                Some(workflow_activation_job::Variant::QueryWorkflow(_))
+            )
+        });
         WorkflowActivation {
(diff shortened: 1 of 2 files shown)


core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -367,9 +367,17 @@ impl WorkflowMachines {
...
+        let all_query = jobs.iter().all(|j| {
+            matches!(
+                j.variant,
+                Some(workflow_activation_job::Variant::QueryWorkflow(_))
+            )
+        });
         WorkflowActivation {
             timestamp: self.current_wf_time.map(Into::into),
-            is_replaying: self.replaying,
+            is_replaying: self.replaying || all_query,
```
mjameswh, line 380: I technically agree with the idea that queries should always be considered as replay (or _never_, that's debatable, but it should be consistent), yet I'm a bit worried that this specific change may eventually end up making future bugs much harder to diagnose.

Not blocking at this time, but I think we should consider refactoring the activation protocol so that there no need for assumptions. Like have core presort jobs and attach the `is_replaying` flag at the job level, or make it so that an activation may never contain both a query and some non-query jobs.

mjameswh (approved):

core/src/worker/workflow/machines/workflow_machines.rs
Sushisource: Yeah, I agree. I think we should do the latter option, where queries always get their own activation independent of other jobs. It might even make sense to segregate the activation type entirely.

Issue #598, opened 2023-08-23
[Feature Request] Do not prefix user-supplied metric names


cretz:
### Describe the solution you'd like
Today we prefix user-supplied metrics with the same prefix we do for core metrics (`temporal_`) but we probably shouldn't

PR #599, opened 2023-08-23
Fix Metrics prefixing & remove trace export configs

Sushisource:
## What was changed
Metrics prefixing works properly, shuffled where it is around a bit. Removed options for trace export which already did nothing.

## Why?
Make it possible to export un-prefixed user metrics / simplify API.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/598
2. How was this tested:
Existing & updated tests

3. Any docs updates needed?

### DIFF
--- a/tests/integ_tests/metrics_tests.rs
+++ b/tests/integ_tests/metrics_tests.rs
@@ -3,7 +3,10 @@ use std::{net::SocketAddr, sync::Arc, time::Duration};
 use temporal_client::{WorkflowClientTrait, WorkflowOptions, WorkflowService};
 use temporal_sdk_core::{init_worker, telemetry::start_prometheus_metric_exporter, CoreRuntime};
 use temporal_sdk_core_api::{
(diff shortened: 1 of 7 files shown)


tests/integ_tests/metrics_tests.rs
```diff
@@ -88,6 +91,17 @@ async fn prometheus_metrics_exported() {
...
+    let mm = rt.telemetry().get_metric_meter().unwrap();
+    let g = mm.inner.gauge(MetricParameters::from("mygauge"));
+    g.record(
+        42,
+        &MetricAttributes::OTel {
+            kvs: Arc::new(vec![]),
+        },
+    );
+    let body = get_text(format!("http://{addr}/metrics")).await;
+    assert!(body.contains("mygauge 42"));
```
bergundy, line 104: This doesn't verify that the prefix isn't added.

Sushisource: Ah, yeah good point thanks.

cretz (approved):

Issue #600, opened 2023-08-25
[Bug] Queries should always come in their own activation


Sushisource:
Reason being otherwise `is_replaying` is ill defined. This also means we can always reject additional commands that were sent in the reply, because queries can never advance state.

PR #603, opened 2023-09-01
Separate activations for queries

Sushisource:
## What was changed
Queries now always go in their own activation

## Why?
We agreed this is the most reasonable way to disambiguate the `is_replaying` flag.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/600

2. How was this tested:
Existing / new tests

3. Any docs updates needed?


mjameswh (approved):
LGTM

PR #608, opened 2023-09-25
Allow `CoreMeter` implementations to use their own custom attribute reference

Sushisource:
## What was changed
Added `CustomMetricAttributes` trait / variant to `MetricAttributes` to allow langs to have a simple reference based approach to allocating their own attributes.

## Why?
Allows destroying backing attributes on `Drop` of the thing implementing `CustomMetricAttributes`

## Checklist
It may no longer be worth keeping a buffer at all, or, if we do, I need to re-implement it to work in terms of refs.

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?

### DIFF
--- a/core-api/src/telemetry/metrics.rs
+++ b/core-api/src/telemetry/metrics.rs
@@ -4,7 +4,13 @@ use std::{borrow::Cow, collections::HashSet, fmt::Debug, sync::Arc};
 /// The implementor is responsible for the allocation/instantiation of new metric meters which
 /// Core has requested.
 pub trait CoreMeter: Send + Sync + Debug {
+    /// Given some k/v pairs, create a return a new instantiated instance of metric attributes.
+    /// Only [MetricAttributes] created by this meter can be used when calling record on instruments
+    /// created by this meter.
     fn new_attributes(&self, attribs: NewAttributes) -> MetricAttributes;
+    /// Extend some existing attributes with new values. Implementations should create new instances
+    /// when doing so, rather than mutating whatever is backing the passed in `existing` attributes.
+    /// Ideally that new instance retains a ref to the extended old attribute, promoting re-use.
     fn extend_attributes(
         &self,
         existing: MetricAttributes,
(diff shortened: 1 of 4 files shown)


cretz (approved):

mjameswh (approved):

core-api/src/telemetry/metrics.rs
```diff
@@ -197,14 +164,92 @@ pub trait Gauge: Send + Sync {
...
+}
+
+/// A lazy reference to some metrics buffer attributes
+pub type BufferAttributes = LazyRef<Arc<dyn CustomMetricAttributes + 'static>>;
+
+/// Types lang uses to contain references to its lang-side defined instrument references must
+/// implement this marker trait
+pub trait BufferInstrumentRef {}
+/// A lazy reference to a metrics buffer instrument
+pub type LazyBufferInstrument<T> = LazyRef<Arc<T>>;
```
cretz, line 214: Can we put a type constraint on `T` that it has to implement `BufferInstrumentRef`? Or maybe `I` can have that constraint above? Or rather, is it at least expected that `T` implements `BufferInstrumentRef`?

core/src/telemetry/metrics.rs
```diff
@@ -637,44 +628,66 @@ pub fn start_prometheus_metric_exporter(
...
         }
     }
 }
-impl CoreMeter for MetricsCallBuffer {
-    fn new_attributes(&self, opts: MetricsAttributesOptions) -> MetricAttributes {
-        let id = self.attribute_ids.fetch_add(1, Ordering::AcqRel);
+
+impl<I> CoreMeter for MetricsCallBuffer<I>
+where
+    I: Debug + Send + Sync + Clone + 'static,
```
cretz: Hrmm, so I guess if I'm tying my non-cloneable Python metric's instrument instance to this, I will have to wrap it in Arc or similar? It doesn't seem like `CustomMetricAttributes` has to implement `Clone`. I don't think it's a problem though.

cretz (approved):

core-api/src/telemetry/metrics.rs
cretz: EDIT: Nevermind, I see this constraint is now available on the `CoreMeter` impl

core/src/telemetry/metrics.rs
Sushisource: The trait doesn't require it, but the places where it's used do (when they need it) which is fairly typical practice

PR #609, opened 2023-09-26
Client keep alive, enabled by default

### LINKED ISSUE #585: [Feature Request] Add client keep-alive support
We need to support client keep alive. Up for debate:

* Whether it remains off by default and opt-in or we turn it on by default w/ a fairly high value
  * I'm leaning to on by default (EDIT: Seems it was decided to turn on by default in https://github.com/temporalio/features/issues/123)
* Whether to use TCP keep alive or HTTP/2 keep alive
  * I'm leaning HTTP/2 keep alive, it's the higher level gRPC one that does a better job and is more configurable
* Which options to expose to users
  * I'm leaning to interval, timeout, and while-idle

cretz:
## What was changed

Exposed keep alive options, and defaulted them to enabled with 30s interval and 15s timeout by default. See https://github.com/temporalio/features/issues/123.

## Checklist

1. Closes #585

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -122,6 +122,10 @@ pub struct ClientOptions {
     /// override.
     #[builder(default)]
     pub override_origin: Option<Uri>,
+
+    /// If set (which it is by default), HTTP2 gRPC keep alive will be enabled.
+    #[builder(default = "Some(ClientKeepAliveConfig::default())")]
+    pub keep_alive: Option<ClientKeepAliveConfig>,
 }
 
 /// Configuration options for TLS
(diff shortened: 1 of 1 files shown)


client/src/lib.rs
```diff
@@ -1441,4 +1471,27 @@ mod tests {
         let next_req = iceptor.call(req).unwrap();
         assert_eq!(next_req.metadata().get("enchi").unwrap(), "cat");
     }
+
+    #[test]
+    fn keep_alive_defaults() {
```
Sushisource, line 1476: I dunno how useful this is really, but, since you wrote it I'm fine keeping it too.

client/src/lib.rs
```diff
@@ -1441,4 +1471,27 @@ mod tests {
...
+        let mut builder = ClientOptionsBuilder::default();
+        builder
+            .identity("enchicat".to_string())
+            .target_url(Url::parse("https://smolkitty").unwrap())
```
Sushisource, line 1480: I am glad you have maintained the traditions

Sushisource (approved):

client/src/lib.rs
cretz: I was scared of my Rust string in the default builder and I wanted to confirm it worked just for me, so I figured why not a test, but I agree in general this is a silly test

Issue #582, opened 2023-07-31
[Feature Request] Support for workflow update


cretz:
### Describe the solution you'd like
This has been implemented in Go and Java and is needed here for the other SDKs. See https://github.com/temporalio/features/issues/296.

PR #613, opened 2023-10-02
Workflow Update

Sushisource:
## What was changed
Implement workflow update in Core

## Why?
The people want the feature!

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/582
2. How was this tested:
Added a variety of tests

3. Any docs updates needed?

TODO:

- [x] Speculative WFT last wft started ID reset isn't handled yet
- [x] Remove `ManagedWFFunc` - this thing has always annoyed me by using APIs that should be more internal, and thus makes searching for usages of some important machines-internal functions hard. It's also pointless now, and I want to remove it. Not really related to update specifically but it annoyed me more than once working on this.
- [x] Add more testcases


core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -456,6 +456,31 @@ impl WorkflowMachines {
...
+        // WFT started, which the id given to us to reset to must be, and we need to avoid
+        // re-applying the WFT Completed event, so we make sure to consider that processed
+        self.last_processed_event = id + 1;
+        // Then, we have to drop any state machines (which should only be one workflow task machine)
+        // we may have created when servicing the speculative task.
+        // Remove when https://github.com/rust-lang/rust/issues/59618 is stable
+        let remove_these: Vec<_> = self
+            .machines_by_event_id
+            .iter()
+            .filter(|(mid, _)| **mid >= id)
```
Quinn-With-Two-Ns, line 474: Nit: Don't think you need the equals because you `id` is a WFT started event so no state machine can be associated with that event

protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto
```diff
@@ -309,3 +311,19 @@ message ModifyWorkflowProperties {
     // used as the value for the key being deleted.
     temporal.api.common.v1.Memo upserted_memo = 1;
 }
+
+// A reply to a `DoUpdate` job - lang must run the update's validator, and then immediately run
+// the handler, if the update was accepted.
+message UpdateResponse {
```
cretz, line 317: Can you clarify, in the case of immediately completed update code, are you expecting two update responses, one for accepted and one for completed/rejected, or if you get just one do you assume completed/rejected implies accepted?

Sushisource: Two. The comments below I was hoping clarified that, since both accepted and completed say they must be sent when the validator and handler pass, respectively.

LMK if you want me to change the comments

cretz: Thanks, I was confused about failure, I'll ask on that field

protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto
```diff
@@ -309,3 +311,19 @@ message ModifyWorkflowProperties {
...
+    // The protocol message instance ID
+    string protocol_instance_id = 1;
+    oneof response {
+        // Must be sent if the update's validator has passed.
+        google.protobuf.Empty accepted = 2;
+        // Must be sent if the update's validator does not pass, or after acceptance if the update
+        // handler fails.
+        temporal.api.failure.v1.Failure rejected = 3;
```
cretz, line 325: I think validator rejection and an update failure are different things. The former doesn't store in history but the latter does. How can core differentiate between the two situations in an immediately-completing workflow?

protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto
cretz: I see, I was confused by the term "rejected" being for update failure, will ask there.

protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
```diff
@@ -242,6 +245,29 @@ message ResolveRequestCancelExternalWorkflow {
     temporal.api.failure.v1.Failure failure = 2;
 }
 
+// Lang is requested to invoke an update handler on the workflow. Lang should invoke the update
+// validator first (if requested). If it accepts the update, immediately invoke the update handler.
+// Lang must reply to the activation containing this job with an `UpdateResponse`.
+message DoUpdate {
+    // A workflow-unique identifier for this update
+    string id = 1;
```
cretz, line 253: I am assuming there is no need to surface this to the user? I wonder if we need to expose to lang.

protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
```diff
@@ -242,6 +245,29 @@ message ResolveRequestCancelExternalWorkflow {
...
+    string protocol_instance_id = 2;
+    // The name of the update handler
+    string name = 3;
+    // The input to the update
+    repeated temporal.api.common.v1.Payload input = 4;
+    // Headers attached to the update
+    map<string, temporal.api.common.v1.Payload> headers = 5;
+    // Remaining metadata associated with the update. The `update_id` field is stripped from here
+    // and moved to `id`, since it is guaranteed to be present.
+    temporal.api.update.v1.Meta meta = 6;
```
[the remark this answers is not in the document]
Sushisource: I would probably just for the client identity string, which may be useful.

protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
Sushisource: Yes, I think they'll expect/want to have it.

protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto
Sushisource: We could have separate commands, but either way you've got to send the expected sequence and this just has fewer thingies to deal with, so I figure let's stick with this unless there's some compelling reason.

As for the question - no, actually - you always send the same sequence. Check out the Rust SDK implementation as reference: https://github.com/temporalio/sdk-core/blob/8e0e0e010d6abd5b8c70830ed7a02243ebb2f1b0/sdk/src/workflow_future.rs#L245

cretz: > As for the question - no, actually - you always send the same sequence

Ok, I think that's definitely worth comment update/clarification, because it says:

> if the update's validator has passed

but it may not even run. Maybe change that to:

> if the update's validator has passed or it was asked not to run

Then at the top of the proto or somewhere general add something like:

> If a `DoUpdate` is received, there must always immediately be an `UpdateResponse` with `accepted` or `rejected` set. `accepted` is validation success/skipped, `rejected` is validation failure. Only if the update was `accepted` can/should the update handler be run and any time later there can be _another_ `UpdateResponse` with either `completed` or `rejected` set depending on success/failure respectively.

cretz (approved):
LGTM from a proto perspective. I have no comment on the Rust code.

protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto
Sushisource: Sure, sounds good I'll add some extra info

client/src/lib.rs
```diff
@@ -1402,6 +1413,40 @@ impl WorkflowClientTrait for Client {
...
+            .update_workflow_execution(UpdateWorkflowExecutionRequest {
+                namespace: self.namespace.clone(),
+                workflow_execution: Some(WorkflowExecution {
+                    workflow_id,
+                    run_id,
+                }),
+                wait_policy: Some(wait_policy),
+                request: Some(update::v1::Request {
+                    meta: Some(update::v1::Meta {
+                        update_id: "".into(),
```
Quinn-With-Two-Ns, line 1435: Why are you not passing an update ID?

core/src/protosext/protocol_messages.rs
```diff
@@ -0,0 +1,102 @@
...
+        let v = v.ok_or_else(|| anyhow!("Protocol message body must be populated"))?;
+        // Undo explicit type url checks when https://github.com/fdeantoni/prost-wkt/issues/48 is
+        // fixed
+        Ok(match v.type_url.as_str() {
+            "type.googleapis.com/temporal.api.update.v1.Request" => {
+                IncomingProtocolMessageBody::UpdateRequest(
+                    v.unpack_as(update::v1::Request::default())?.try_into()?,
+                )
+            }
+            o => bail!("Could not understand protocol message type {}", o),
```
Quinn-With-Two-Ns, line 68: Will this error fail the WFT?

Quinn-With-Two-Ns (approved):

client/src/lib.rs
Sushisource: I just wanted to keep the signature small in this trait. Rust client users can go raw grpc if they need to specify it.

PR #616, opened 2023-10-17
Workflow Update small fixes

Sushisource:
## What was changed
The main fix resolves a problem where Core was sending a WFT complete too early when updates were included in the first WFT, causing a WFT failure spam loop if the update failed a task.

Also fix some dangling TODOs from the first PR

## Why?
Yummy fixes

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


core/src/worker/workflow/machines/workflow_machines.rs
cretz, line 1: Near the bottom of this file is a warning for an unnecessary `.borrow()` (can't comment on line 1493). May want to fix warning and enable warnings-as-errors in CI

cretz (approved):

PR #617, opened 2023-10-23
Fix bug where message replies to updates could be sent during replay

Sushisource:
## What was changed
Task failures after an update could trigger this path which would cause the sending of protocol messages that shouldn't have been sent. Fixes that.

## Why?
Bug bad

## Checklist

1. Closes 

2. How was this tested:
Added integ test

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/machines/update_state_machine.rs
+++ b/core/src/worker/workflow/machines/update_state_machine.rs
@@ -75,6 +75,7 @@ pub(super) struct SharedState {
     instance_id: String,
     event_seq_id: i64,
     request: UpdateRequest,
+    replaying_when_created: bool,
 }
 
 impl UpdateMachine {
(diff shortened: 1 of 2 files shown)


core/src/worker/workflow/machines/update_state_machine.rs
```diff
@@ -75,6 +75,7 @@ pub(super) struct SharedState {
     instance_id: String,
     event_seq_id: i64,
     request: UpdateRequest,
+    replaying_when_created: bool,
```
dandavison, line 78: Can you explain what the "when_created" part of the name means?

cretz (approved):

core/src/worker/workflow/machines/update_state_machine.rs
Sushisource: When the state machine is created

PR #619, opened 2023-10-25
Attach current Build ID to workflow activations

Sushisource:
## What was changed
Add current build id as a field on workflow activations

## Why?
So users can tell during replay what build id was used to process the current task.

## Checklist

1. Closes https://github.com/temporalio/features/issues/253

2. How was this tested:
Added unit test

3. Any docs updates needed?

### DIFF
--- a/core/src/protosext/mod.rs
+++ b/core/src/protosext/mod.rs
@@ -139,6 +139,8 @@ impl TryFrom<PollWorkflowTaskQueueResponse> for ValidPollWFTQResponse {
 pub(crate) trait WorkflowActivationExt {
     /// Returns true if this activation has one and only one job to perform a legacy query
     fn is_legacy_query(&self) -> bool;
+    /// Augment the activation with the worker's Build ID if appropriate
+    fn attach_build_id_if_needed(&mut self, build_id: &str);
 }
 
 impl WorkflowActivationExt for WorkflowActivation {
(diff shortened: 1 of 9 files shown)


core/src/protosext/mod.rs
```diff
@@ -147,6 +149,11 @@ impl WorkflowActivationExt for WorkflowActivation {
                     variant: Some(workflow_activation_job::Variant::QueryWorkflow(qr))
                 }] if qr.query_id == LEGACY_QUERY_ID)
     }
+    fn attach_build_id_if_needed(&mut self, build_id: &str) {
```
bergundy, line 152: This is a bit confusing, why would this logic be on a proto struct, it doesn't know when attaching is needed.
I would put it in the worker logic before it hands over the activation or better yet when the activation is created to avoid mutating unnecessarily.

bergundy (approved):
Overall LGTM, I just have a nit about the structuring of the code.

core/src/protosext/mod.rs
Sushisource: It can't really be when the activation is created without passing the build id down a few layers where it's not currently needed (and would get cloned a bunch of times). I can just inline the method where it's used though.

cretz (approved):
LGTM. 

Lang impl note: In Python and .NET we have tried to make properties on the info immutable for the life of the run (even if the contents of that property's object may change). So we'll either be exposing this as a method on the info (Python only if so), or top-level on the workflow context. Other languages however may not have this immutability rule.

core/src/protosext/mod.rs
Sushisource: Inlined it w/ comment

PR #620, opened 2023-10-27
Local timeouts for remote activities

Sushisource:
## What was changed
Core now will send cancellations if an activity would have already timed out according to server but we haven't learned of that (typically, because it isn't heartbeating). There is a configurable buffer defaulting to 5s, to avoid races with server (we would prefer to learn from server).

## Why?
Nice to have your activities get cancelled when they would have been even if they don't heartbeat

## Checklist

1. Closes https://github.com/temporalio/features/issues/170

2. How was this tested:
Unit & Integ

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/activities.rs
+++ b/core/src/worker/activities.rs
@@ -108,6 +108,8 @@ struct RemoteInFlightActInfo {
     pub known_not_found: bool,
     /// Handle to the task containing local timeout tracking, if any.
     pub local_timeouts_task: Option<JoinHandle<()>>,
+    /// Used to reset the local heartbeat timeout every time we record a heartbeat
+    timeout_resetter: Option<Arc<Notify>>,
     /// The permit from the max concurrent semaphore
     _permit: UsedMeteredSemPermit,
 }
@@ -126,6 +128,7 @@ impl RemoteInFlightActInfo {
             issued_cancel_to_lang: None,
             known_not_found: false,
             local_timeouts_task: None,
+            timeout_resetter: None,
             _permit: permit,
         }
     }
@@ -397,10 +400,11 @@ impl WorkerActivityTasks {
         details: ActivityHeartbeat,
     ) -> Result<(), ActivityHeartbeatError> {
         // TODO: Propagate these back as cancels. Silent fails is too nonobvious
-        let heartbeat_timeout: Duration = self
+        let at_info = self
             .outstanding_activity_tasks
             .get(&TaskToken(details.task_token.clone()))
-            .ok_or(ActivityHeartbeatError::UnknownActivity)?
+            .ok_or(ActivityHeartbeatError::UnknownActivity)?;
+        let heartbeat_timeout: Duration = at_info
             .heartbeat_timeout
             .clone()
             // We treat None as 0 (even though heartbeat_timeout is never set to None by the server)
@@ -420,7 +424,8 @@ impl WorkerActivityTasks {
         };
         let throttle_interval =
             std::cmp::min(throttle_interval, self.max_heartbeat_throttle_interval);
-        self.heartbeat_manager.record(details, throttle_interval)
+        self.heartbeat_manager
+            .record(details, throttle_interval, at_info.timeout_resetter.clone())
     }
 
     /// Returns a handle that the workflows management side can use to interact with this manager
@@ -564,24 +569,41 @@ where
                                     task.resp.start_to_close_timeout.clone(),
                                 ]
                                 .into_iter()
-                                .flatten()
-                                .map(Duration::try_from)
-                                .filter_map(Result::ok)
-                                .chain(sched)
-                                .filter(|d| !d.is_zero())
-                                .min();
-                                if let Some(timeout_at) = timeout_at {
+                                .enumerate()
+                                .filter_map(|(i, d)| {
+                                    d.and_then(|d| Duration::try_from(d).ok().map(|d| (i, d)))
+                                })
+                                .chain(sched.map(|s| (2, s)))
+                                .filter(|(_, d)| !d.is_zero())
+                                .min_by(|(_, d1), (_, d2)| d1.cmp(d2));
+                                if let Some((timeout_type, timeout_at)) = timeout_at {
+                                    let sleep_time = timeout_at + local_timeout_buffer;
                                     let cancel_tx = cancels_tx.clone();
+                                    let resetter = if timeout_type == 0 {
+                                        Some(Arc::new(Notify::new()))
+                                    } else {
+                                        None
+                                    };
+                                    let resetter_clone = resetter.clone();
                                     outstanding_info.local_timeouts_task =
                                         Some(tokio::task::spawn(async move {
-                                            tokio::time::sleep(timeout_at + local_timeout_buffer)
-                                                .await;
+                                            if let Some(rs) = resetter_clone {
+                                                loop {
+                                                    tokio::select! {
+                                                        _ = rs.notified() => continue,
+                                                        _ = tokio::time::sleep(sleep_time) => break,
+                                                    }
+                                                }
+                                            } else {
+                                                tokio::time::sleep(sleep_time).await;
+                                            }
                                             let _ = cancel_tx.send(PendingActivityCancel {
                                                 task_token: tt,
                                                 reason: ActivityCancelReason::TimedOut,
                                                 consider_not_found: true,
                                             });
                                         }));
+                                    outstanding_info.timeout_resetter = resetter;
                                 }
                             }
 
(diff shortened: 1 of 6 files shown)


core/src/worker/activities.rs
```diff
@@ -509,6 +527,53 @@ where
...
+                                    task.resp.heartbeat_timeout.clone(),
+                                    task.resp.start_to_close_timeout.clone(),
+                                ]
+                                .into_iter()
+                                .flatten()
+                                .map(Duration::try_from)
+                                .filter_map(Result::ok)
+                                .chain(sched)
+                                .filter(|d| !d.is_zero())
+                                .min();
```
bergundy, line 565: This is the rustiest piece of code I've ever seen 😆

Sushisource: :crab:

core/src/worker/activities.rs
```diff
@@ -509,6 +527,53 @@ where
...
+                                if let Some(timeout_at) = timeout_at {
+                                    let cancel_tx = cancels_tx.clone();
+                                    tokio::task::spawn(async move {
```
bergundy, line 568: It would be nice not to leave dangling tasks around and instead merge this work into some stream that can be aborted when no longer needed or at least have a reference to this task so it can be canceled when the activity resolves.

core/src/worker/activities.rs
```diff
@@ -509,6 +527,53 @@ where
...
+                                    .zip(task.resp.scheduled_time.clone())
+                                    .and_then(|(to, st)| {
+                                        let now = SystemTime::now();
+                                        let sched_time = SystemTime::try_from(st).unwrap_or(now);
+                                        let already_elapsed =
+                                            now.duration_since(sched_time).unwrap_or_default();
+                                        Duration::try_from(to).map(|to| to - already_elapsed).ok()
+                                    });
+                                let timeout_at = [
+                                    task.resp.heartbeat_timeout.clone(),
```
bergundy, line 556: You don't want heartbeat timeout here since you don't extend the timeout on heartbeat.
I'd just use the minimum of start to close and schedule to close (which I think the server already sets one of those fields as the minimum, have to verify).

Sushisource: Oh, that's a good point. Will change.

core/src/worker/activities.rs
Sushisource: I generally agree but there is no real need to track these tasks at all - nothing can go wrong in them. That said cancelling when the activity resolves would be useful for sure. I'll find a way to add that.

core/src/worker/activities.rs
```diff
@@ -509,6 +527,53 @@ where
...
+                                        // Must filter out zero before any subtraction math here
+                                        if to.seconds == 0 && to.nanos == 0 {
+                                            None
+                                        } else {
+                                            Some(to)
+                                        }
+                                    })
+                                    .zip(task.resp.scheduled_time.clone())
+                                    .and_then(|(to, st)| {
+                                        let now = SystemTime::now();
```
cretz, line 549: I'm scared of SystemTime and it not being monotonic. Should we use `Instant` here?

core/src/worker/activities.rs
```diff
@@ -509,6 +527,53 @@ where
...
                                     ActivityCancelReason::WorkerShutdown,
                                 ));
+                            } else {
+                                // Fire off task to keep track of local timeouts. We do this so that
+                                // activities can still get cleaned up even if the user isn't
+                                // heartbeating.
+                                let local_timeout_buffer = self.local_timeout_buffer;
+                                let sched = task
+                                    .resp
+                                    .schedule_to_close_timeout
```
cretz, line 537: I don't think schedule to close should ever come into play. I think only start to close because this is only about one attempt right? And does server set start to close on the task response even if a user didn't set it? I'm also scared of doing _anything_ with server provided times like scheduled_time. I think we need heartbeat and start-to-close only.

But maybe I didn't read/understand this algorithm close enough to understand what schedule time is for here.

core/src/worker/activities.rs
cretz: Got a 10-chainer at https://github.com/temporalio/sdk-python/blob/2c15ed30159242d7f93534e8263e221de1959e1e/temporalio/bridge/src/metric.rs#L273-L283

core/src/worker/activities.rs
Sushisource: We can't. System time is the only thing the protobuf stuff can be converted into. The unwrap_or / defaults provide protection against stuff not making sense

core/src/worker/activities.rs
Sushisource: Because we use the scheduled time coming in the task, we know if this attempt would hit the overall schedule timeout - so we can handle that knowing it will have already elapsed on the server in the same way.

As for heartbeat, I got rid of it per Roey's point, b/c there's a good amount of extra machinery involved to refresh it per-heartbeat, and that seems not worth it since the only situation it matters in is when you've specified a HB timeout but aren't actually heartbeating, which would be weird.

cretz: > Because we use the scheduled time coming in the task, we know if this attempt would hit the overall schedule timeout - so we can handle that knowing it will have already elapsed on the server in the same way.

I don't think we need to protect against that, only the attempt. I was under the impression start to close is always bound by the schedule to close time server side anyways and set here as a difference between the two.

> As for heartbeat, I got rid of it per Roey's point, b/c there's a good amount of extra machinery involved to refresh it per-heartbeat, and that seems not worth it since the only situation it matters in is when you've specified a HB timeout but aren't actually heartbeating, which would be weird.

It's not that weird, this code is very normal with a large start-to-close timeout:

```python
while True:
    activity.heartbeat()
    await my_http_client.do_hopefully_short_http_poll()
```

In fact, that's kinda the genesis of this whole issue, that users are not setting good short timeouts on HTTP calls they hoped were quick. If `do_hopefully_short_http_poll` hangs, the entire system gums up. I didn't read the logic in the PR, but ideally we can have a start to close timer that will fail task and a heartbeat timer that will fail the task, the latter simply getting reset every heartbeat call.

Sushisource: There is a start to close timer. So that example still gets ended by that. But, sure, I'll see how bad it is to refresh the heartbeat and have a timer around that too

cretz: :+1: If too much of a mess, can defer it. I wonder if the start-to-close timer is enough to never care about schedule-to-close.

Sushisource: I think we want to keep schedule to close because a user might only have a schedule timeout but not a start one.

cretz: Hrmm, I thought server _always_ set a start-to-close based on schedule-to-close if user does not explicitly have a start-to-close. But now I understand the logic, I wasn't reading ht close enough (zip, slice of two, min, etc threw me a bit)

bergundy: > I thought server always set a start-to-close based on schedule-to-close if user does not explicitly have a start-to-close

I mentioned this in one of the outdated comments, I would check what server does here and avoid this SDK complexity if possible.
I'm also a bit worried about clock skew when referencing system time, you will decrease the overall timeout if the local clock is ahead of the server clock.

cretz: Hrmm, do we need to up that buffer leeway time from 5s to 10s, because it needs to represent the skew between server and client _and_ give the server time to timeout first?

Sushisource: I added heartbeat timeouts back with resetting. Schedule doesn't really add any complexity compared to the others, so I don't see why not to keep it.

core/src/worker/activities.rs
```diff
@@ -509,6 +539,72 @@ where
...
+                                                loop {
+                                                    tokio::select! {
+                                                        _ = rs.notified() => continue,
+                                                        _ = tokio::time::sleep(sleep_time) => break,
+                                                    }
+                                                }
+                                            } else {
+                                                tokio::time::sleep(sleep_time).await;
+                                            }
+                                            let _ = cancel_tx.send(PendingActivityCancel {
```
cretz, line 600: Do we want to increment a metric here or log or somehow let lang know via the proto that this was due to local timeout? I'm just starting to get scared that a user's clock is gonna skew like 10s off of server's and things are gonna just start cancelling everywhere.

Or I guess we can look at all the activity failures and if they are failures for untriggered cancel we can assume it was either worker shutdown or this timeout? That's probably good enough.

cretz (approved):
A bit scared of local timeouts for users that have something like:

```python
def my_activity() -> None:
    do_sync_thing_that_never_fails_because_I_never_heartbeat()
    do_cleanup_guaranteed_to_run_even_on_server_timeout()
```

The second call used to run for them, but maybe not anymore. But I guess they could never gracefully shutdown works then. I guess we just need to be real clear in release notes.

core/src/worker/activities.rs
Sushisource: I suppose skew is only an issue with the scheduled timeout, since that's the only one that uses an absolute time. We could remove it for just that reason, since you're more likely to fail an attempt timeout or heartbeat anyway.

I don't want to add another metric... we've already got so many. I'll add a debug log line though.

cretz: :+1: Very happy to see schedule timeout removed. I think it's much safer to do single-attempt duration than clock-based. May be worth checking whether server _always_ sets start-to-close, I forget, but I do feel safer now w/ what is here.

PR #621, opened 2023-10-30
Make Client properly go through RawClient, expose more non-mut clients

Sushisource:
Needed to enable the work on eager wf start


cretz (approved):

cretz:
Since this is a trait for internal use mostly, I have no strong opinions on mut/non-mut abstractions.

antlai-temporal:
Going through the macros for all the integration tests

Issue #606, opened 2023-09-25
[Feature Request] Enable Eager Workflow Start


antlai-temporal:
### Is your feature request related to a problem? Please describe.

Support for Eager Workflow Start has been enabled by the server since 1.20, and is already available in Java and Go SDKs. It reduces latency to start a new workflow by allowing the client to directly schedule a first task to a local worker, bypassing matching and an extra transaction/roundtrip. See https://github.com/temporalio/features/issues/242 for details.

We want to support Eager Workflow Start for core-based languages, starting with Python.

### Describe the solution you'd like

The implementation relies on https://github.com/temporalio/sdk-core/issues/605, a worker registry that finds local compatible workers, 
and extends it with a `SlotManager` trait that provides a method to reserve a worker slot, and another one to schedule a task using this slot.

The client will also transparently use eager mode when it is enabled, and gracefully default to non-eager mode when either the server or the `SlotManager` rejects it.

PR #622, opened 2023-10-31
Eager Workflow Start (#606)

antlai-temporal:
## What was changed
Adding to clients the ability of tracking workers that are attached to them, so that we can implement Eager Workflow Start,
a latency optimization that allows the client to directly dispatch the first workflow task to the worker, eliminating a server round-trip.

## Why?
Significantly reduces the latency of short-lived workflows.

## Checklist

1. Closes 
#606


client/src/lib.rs
```diff
@@ -299,6 +301,11 @@ impl<C> ConfiguredClient<C> {
     pub fn capabilities(&self) -> Option<&get_system_info_response::Capabilities> {
         self.capabilities.as_ref()
     }
+
+    /// Returns a cloned reference to a registry with workers using this client instance
+    pub fn workers(&self) -> Arc<RwLock<SlotManager>> {
```
cretz, line 306: I wonder if RwLock should be an implementation detail of the `SlotManager` instead of exposed here

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+    fn task_queue(&self) -> String;
+    /// Try to reserve a slot with matching namespace and task queue.
+    fn try_reserve_wft_slot(&self) -> Option<Box<dyn Slot>>;
+}
+
+/// This trait represents a slot reserved for processing a WFT by a worker.
+#[cfg_attr(test, mockall::automock)]
+pub trait Slot: Send + Sync {
+    /// Consumes this slot by dispatching a WFT to its worker. This can only be called once.
+    fn schedule_wft(&mut self, task: PollWorkflowTaskQueueResponse) -> Result<(), anyhow::Error>;
```
cretz, line 26:
```suggestion
    fn schedule_wft(self, task: PollWorkflowTaskQueueResponse) -> Result<(), anyhow::Error>;
```

Might as well consume the slot if it's expected to be the only call on it (makes it clear to callers this slot cannot be reused)

client/src/raw.rs
```diff
@@ -435,10 +478,37 @@ proxier! {
...
+                match workers.read().try_reserve_wft_slot(namespace, task_queue) {
+                    Some(s) => slot = Some(s),
+                    None => req_mut.request_eager_execution = false
+                }
+            }
+            slot
+        },
+        start_workflow_execution, // Disambiguate the macro
+        |resp, slot| {
+            if let Some(mut s) = slot {
```
cretz, line 499: To confirm, slot is returned on drop?

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+use rand::{seq::SliceRandom, thread_rng};
+use std::{collections::HashMap, fmt::Debug};
+use temporal_sdk_core_protos::temporal::api::workflowservice::v1::PollWorkflowTaskQueueResponse;
+
+/// This trait wraps a worker, which provides WFT processing slots.
+#[cfg_attr(test, mockall::automock)]
+pub trait SlotProvider: Send + Sync + Debug {
+    /// A unique identifier for the worker.
+    fn uuid(&self) -> String;
```
cretz, line 13:
```suggestion
    fn id(&self) -> String;
```

Probably not required to be a UUID (can remove UUID from the name everywhere I assume)

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+        }
+    }
+
+    /// Try to reserve a compatible processing slot in any of the registered workers.
+    pub fn try_reserve_wft_slot(
+        &self,
+        namespace: String,
+        task_queue: String,
+    ) -> Option<Box<dyn Slot>> {
+        let key = to_key(&namespace, &task_queue);
```
cretz, line 72: Wonder if a simple two-field struct with `#[derive(PartialEq, Eq, Hash)]` as map key is better than making a string here

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+/// This trait represents a slot reserved for processing a WFT by a worker.
+#[cfg_attr(test, mockall::automock)]
+pub trait Slot: Send + Sync {
+    /// Consumes this slot by dispatching a WFT to its worker. This can only be called once.
+    fn schedule_wft(&mut self, task: PollWorkflowTaskQueueResponse) -> Result<(), anyhow::Error>;
+}
+
+/// This trait enables local workers to made themselves visible to a shared client instance.
+pub trait WorkerRegistry: Send + Sync {
```
cretz, line 30: Could also have a single `register` that returns a `WorkerPermit` with `unregister` on it. That way you could actually get rid of this trait and change `fn workers(&self) -> Arc<RwLock<dyn WorkerRegistry>> {` to something like `fn register_slot_provider(&self, Box<dyn SlotProvider>) -> WorkerPermit`. Also it will simplify your unregister logic instead of asking the caller to give you their ID again. Also may be worth choosing to use either "worker" or "slot provider" terminology instead of both here.

(none of my suggestions like these are required, just discussion starters)

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// Try to reserve a compatible processing slot in any of the registered workers.
+    pub fn try_reserve_wft_slot(
+        &self,
+        namespace: String,
+        task_queue: String,
+    ) -> Option<Box<dyn Slot>> {
+        let key = to_key(&namespace, &task_queue);
+        if let Some(all) = self.providers.get(&key) {
+            let index = random_permutation(all.len());
```
cretz, line 74: In a more performance-minded situation, I'd consider a specialization for a single-sized vector since 99% of the time that's what it will be and you're creating a single size vec of shuffled indices for no reason.

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+            for i in &index {
+                let slot = all[*i].try_reserve_wft_slot();
+                if let Some(s) = slot {
+                    return Some(s);
+                }
```
cretz, line 79: Nit
```suggestion
                if slot.is_some() {
                    return slot;
                }
```
(reads clearer to me to not recreate option, but no need to change)

client/src/raw.rs
```diff
@@ -356,10 +378,27 @@ macro_rules! proxy {
...
+                }.boxed()
+            };
+            self.call(stringify!($method), fact, request.into_request())
+        }
+    };
 }
 macro_rules! proxier {
     ( $trait_name:ident; $impl_list_name:ident; $client_type:tt; $client_meth:ident;
-      $(($method:ident, $req:ty, $resp:ty $(, $closure:expr)?  );)* ) => {
+      $(($method:ident, $req:ty, $resp:ty $(, $closure:expr)? $(, $_dummy:ident, $closure_after:expr)? );)* ) => {
```
Sushisource, line 401: Hmm, we should be able to avoid this dummy identifier. I think just the extra comma should be sufficient here since the whole clause is completed by a semicolon

client/src/raw.rs
```diff
@@ -435,10 +478,37 @@ proxier! {
...
+                    None => req_mut.request_eager_execution = false
+                }
+            }
+            slot
+        },
+        start_workflow_execution, // Disambiguate the macro
+        |resp, slot| {
+            if let Some(mut s) = slot {
+                if resp.is_ok() {
+                    let response_inner = resp.as_ref().unwrap().get_ref().clone();
```
Sushisource, line 501: You can avoid this unwrap here, you should use another `if let`

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
+//! This module enables the tracking of workers that are associated with a client instance.
+//! This is needed to implement Eager Workflow Start, a latency optimization in which the client,
+//!  after reserving a slot, directly forwards a WFT to a local worker.
+
+use rand::{seq::SliceRandom, thread_rng};
+use std::{collections::HashMap, fmt::Debug};
+use temporal_sdk_core_protos::temporal::api::workflowservice::v1::PollWorkflowTaskQueueResponse;
+
+/// This trait wraps a worker, which provides WFT processing slots.
```
Sushisource, line 9: Wraps is maybe a weird way to say this. More like "is implemented by"?

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+#[cfg_attr(test, mockall::automock)]
+pub trait SlotProvider: Send + Sync + Debug {
```
Sushisource, line 11: Typically you don't require supertraits that the trait itself doesn't actually need - rather, include the bounds in places where you need the trait to also satisfy those bounds.

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// A unique identifier for the worker.
+    fn uuid(&self) -> String;
+    /// The namespace for the WFTs that it can process.
+    fn namespace(&self) -> String;
+    /// The task queue this provider listens to.
+    fn task_queue(&self) -> String;
+    /// Try to reserve a slot with matching namespace and task queue.
```
Sushisource, line 18:
```suggestion
    /// Try to reserve a slot on this worker.
```
It's true that they're going to match, but, that's sort of up to the caller really

client/src/worker_registry/mod.rs
Sushisource, line 30: I don't think we even need this trait. It's only ever backed by the one implementation. Let's just use `SlotManager` directly.

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+    fn try_reserve_wft_slot(&self) -> Option<Box<dyn Slot>>;
+}
+
+/// This trait represents a slot reserved for processing a WFT by a worker.
+#[cfg_attr(test, mockall::automock)]
+pub trait Slot: Send + Sync {
```
Sushisource, line 24: Same here, the trait _itself_ doesn't need `Send + Sync`, but specific users of it sometimes need those implementations to be `Send + Sync`

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// Try to reserve a compatible processing slot in any of the registered workers.
+    pub fn try_reserve_wft_slot(
+        &self,
+        namespace: String,
+        task_queue: String,
+    ) -> Option<Box<dyn Slot>> {
+        let key = to_key(&namespace, &task_queue);
+        if let Some(all) = self.providers.get(&key) {
+            let index = random_permutation(all.len());
+            for i in &index {
```
Sushisource, line 75: You don't need this permuting helper, you can use: https://docs.rs/rand/latest/rand/seq/trait.IteratorRandom.html#method.choose

Sushisource: But, that said, I kinda question if we need any of this anyway. IMO I am 100% fine with the restriction that you can't have more than one worker pulling from the same task queue in the same process. That just doesn't make much sense to me.

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+pub struct SlotManager {
+    /// Maps keys, i.e., namespace#task_queue, to providers.
+    providers: HashMap<String, Vec<Box<dyn SlotProvider>>>,
+    /// Maps uuids to keys in `providers`.
+    index: HashMap<String, String>,
+}
+
+fn to_key(namespace: &str, task_queue: &str) -> String {
+    format!("{}#{}", namespace, task_queue)
+}
```
Sushisource, line 49: No need to make up this format string. Just make a type which has `namespace` and `task_queue` fields.

client/src/worker_registry/mod.rs
Sushisource: That also allows us to get rid of the UUID per-worker, which would be nice to not have I think.

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// Consumes this slot by dispatching a WFT to its worker. This can only be called once.
+    fn schedule_wft(&mut self, task: PollWorkflowTaskQueueResponse) -> Result<(), anyhow::Error>;
+}
+
+/// This trait enables local workers to made themselves visible to a shared client instance.
+pub trait WorkerRegistry: Send + Sync {
+    /// Register a local worker that can provide WFT processing slots.
+    fn register(&mut self, provider: Box<dyn SlotProvider>);
+    /// Unregister a provider, typically when its worker starts shutdown.
+    fn unregister(&mut self, uuid: String);
```
cretz, line 34:
```suggestion
    fn unregister(&mut self, uuid: &str);
```

Should be unnecessary to have an owned string here

core/src/worker/mod.rs
```diff
@@ -352,7 +365,17 @@ impl Worker {
...
+        let uuid = uuid::Uuid::new_v4().simple().to_string();
+        let provider = SlotProvider::new(
+            uuid.clone(),
+            config.namespace.clone(),
+            config.task_queue.clone(),
+            wft_semaphore.clone(),
+            external_wft_tx,
+        );
+        client.workers().write().register(Box::new(provider));
         Self {
```
cretz, line 377: Hrmm, wonder if it's clearer for the worker to just impl `SlotProvider` trait directly

client/src/worker_registry/mod.rs
Sushisource, line 26: The hint is already in the docstring - this should take `self` rather than `&mut self`, since it "consumes" the slot.

core/src/worker/slot_provider.rs
```diff
@@ -0,0 +1,199 @@
...
+use temporal_client::{Slot as SlotTrait, SlotProvider as SlotProviderTrait};
+use temporal_sdk_core_protos::temporal::api::workflowservice::v1::PollWorkflowTaskQueueResponse;
+use tokio::sync::mpsc::UnboundedSender;
+use tonic::Status;
+
+type WFTStreamSender =
+    UnboundedSender<Result<(ValidPollWFTQResponse, OwnedMeteredSemPermit), Status>>;
+
+pub struct Slot {
+    permit: Option<OwnedMeteredSemPermit>,
```
Sushisource, line 24: This doesn't need to be an option any more, once you make `schedule_wft` consume `self`

core/src/worker/slot_provider.rs
```diff
@@ -0,0 +1,199 @@
...
+
+pub struct SlotProvider {
+    uuid: String,
+    namespace: String,
+    task_queue: String,
+    wft_semaphore: Arc<MeteredSemaphore>,
+    external_wft_tx: WFTStreamSender,
+}
+
+impl Debug for SlotProvider {
```
Sushisource, line 57: Check out existing usages of `derive_more::DebugCustom` - a little bit less boilerplate

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+/// This trait wraps a worker, which provides WFT processing slots.
+#[cfg_attr(test, mockall::automock)]
+pub trait SlotProvider: Send + Sync + Debug {
+    /// A unique identifier for the worker.
+    fn uuid(&self) -> String;
+    /// The namespace for the WFTs that it can process.
+    fn namespace(&self) -> String;
+    /// The task queue this provider listens to.
+    fn task_queue(&self) -> String;
```
Sushisource, line 17: These should all return `&str` - up to the caller if they want to clone

client/src/worker_registry/mod.rs
cretz, line 13: There's a lot of string cloning going on, can you have a scalar identifier? If you switch to the `WorkerPermit` thing mentioned elsewhere, the worker registry can be responsible for ID creation and therefore can just keep an atomic counter of u64 or something.

client/src/lib.rs
antlai-temporal: I think RwLock makes sense here because the majority of accesses in the critical path are just readers, it is only when a worker starts or shutdown that we need a write lock...

cretz: But the caller shouldn't care when the implementation of `SlotManager` needs a lock. `SlotManager` impl should obtain the lock when it needs.

client/src/worker_registry/mod.rs
antlai-temporal: I did try to do that, it would be nicer that the compiler enforces that, and not the runtime check.
 The problem is that is a dyn Slot, and Rust complains that it cannot be moved because it does not know the size. Is there a way to get around that?

client/src/raw.rs
antlai-temporal: Yes, there is a unit test for that in core/src/worker/slot_provider.rs called `unused_slots_reclaimed()`. 
It is just dropping the semaphore inside the slot.

client/src/worker_registry/mod.rs
cretz: You might have to add a `: Sized` the trait or make `self: Box<dyn Trait>` maybe or something, I dunno

client/src/worker_registry/mod.rs
antlai-temporal: Makes sense, I changed uuid to id everywhere...

client/src/worker_registry/mod.rs
antlai-temporal: I did considered that, but it is also a value in the other hashmap, and since it is all internal, it was just easier to go with a string...

client/src/worker_registry/mod.rs
antlai-temporal: I agree, you could also cache the random indexes to avoid allocation/deallocation. But I think Spencer's comment of using  a random iterator will solve this in another way...

client/src/worker_registry/mod.rs
antlai-temporal: Much cleaner, thanks!

client/src/worker_registry/mod.rs
antlai-temporal: Yes, that saved me quite a few to_string() and clone()...

core/src/worker/mod.rs
antlai-temporal: At first I started that way, but I had to Arc workers + Weak refs, and the comparisons with the hashmap+vector were getting unnecessary ugly, and it required to do more intrusive changes because Workers::new() don't  return Arc workers refs.  
It was also more difficult to debug compared to just using simple uuid strings, so I think the indirection helps...

client/src/raw.rs
antlai-temporal: I did try to get rid of it with not much luck so far. The problem is that there seems to be  no look-ahead with the macros, so when it hits the first $(, $closure:expr)? it cannot tell whether it belongs to the first or second closure, even if I rewrite my second rule to explicitly have two closures. Is there a standard trick for this?

client/src/raw.rs
antlai-temporal: Much cleaner, done...

client/src/worker_registry/mod.rs
antlai-temporal: Yes, what I was trying to say is "this trait is implemented by an object that wraps a worker", i.e., the worker does not implement it for the reasons I mentioned above...

client/src/worker_registry/mod.rs
antlai-temporal: Done

client/src/worker_registry/mod.rs
antlai-temporal: I was trying to hide the try_reserve_wft_slot and others methods that are only called by the client, so that the "core" code only sees the WorkerRegistry relevant methods, i.e., register/unregister...

client/src/worker_registry/mod.rs
antlai-temporal: This is coming from the Go and Java implementations where you can have multiple workers to the same queue+namespace and you want to uniformly distribute the load with randomness.
I'm not sure we can always guarantee that there is only one worker per task queue for all core languages, and uniform distribution helps not only with load but also robustness...
 And if one day we will fix versioning, it may be easier if we already support many to one...

client/src/worker_registry/mod.rs
antlai-temporal: Yes, as I mentioned above, I tried to make it `self` but is a `Dyn Slot`, and Rust doesn't let me move it since it does not know the size. Is there a way to do this with Dyn?

client/src/worker_registry/mod.rs
antlai-temporal: Happy to do that, as I just replied above, it is also used as a value in another hashmap, and that's why I did that...

client/src/worker_registry/mod.rs
antlai-temporal: Got rid of Send+Sync, I had to keep Debug for SlotManager, which needs to implement  Debug for ConfiguredClient...

client/src/worker_registry/mod.rs
antlai-temporal: Done

core/src/worker/slot_provider.rs
antlai-temporal: Good tip! Done...

client/src/worker_registry/mod.rs
antlai-temporal: Makes sense, done...

client/src/worker_registry/mod.rs
antlai-temporal: Enough procrastination, I just added a typed map key, and it is much cleaner with the macros doing all the work...

client/src/worker_registry/mod.rs
antlai-temporal: Done, much cleaner with a tuple key and all the macros help...

client/src/lib.rs
antlai-temporal: In other languages I completely agree with you that you want to hide the mutex, but what I'm seeing in Rust is to explicitly use Arc<Mutex> to protect top level references to shared objects, I guess because the compiler keeps thing safe, and the mutex always owns the protected object. For example, in the same ConfiguredClient struct that contains SlotManager the headers field is  Arc<RwLock<HashMap<String, String>>>...

One of the things that worry me about Rust is that because "it is safe to share",  sometimes safe programs are much harder to understand because shared state is everywhere, and with few abstractions... But I guess that is another discussion, and I have an Erlang bias here...

client/src/worker_registry/mod.rs
antlai-temporal: I just checked the IteratorRandom but it doesn't give you a permutation, i.e., it can return the same element many times and miss another one for a very long time... The `choose_multiple` method  seems to have this problem too, it also says that for fully random  order you need to shuffle the result, which kind of defeats the purpose...

client/src/lib.rs
cretz: What is a top level reference? Why does core not make the user have a `Arc<RwLock<ConfiguredClient>>` instead of putting a lock around that headers field and hiding it from users? You should do the same. It is a small semantic difference, but callers can't possibly understand the internal expense of the try-reserve call and therefore can't be confident in knowing that they aren't deadlocking the system by taking a lock when calling it. Only the implementation of the `SlotManager` knows its implementation is acceptable to lock around.

client/src/worker_registry/mod.rs
Sushisource: Yeah, I just don't think we need multiple workers on the same queue in the same process. In fact, that's illegal already in Java: https://github.com/temporalio/sdk-java/blob/8af4a2647dacab2cb0d4a3a613482e00bc297a39/temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java#L167

So, let's avoid some complexity and forbid it. I can't think of any reasonable use-case.

client/src/worker_registry/mod.rs
Sushisource: Where does it require it? In any case, the same idea should apply, you only need to say `+ Debug` where needed in `ConfiguredClient`. But, it doesn't look like that's actually holding on to a `SlotProvider`?

client/src/worker_registry/mod.rs
Sushisource: You can't do that because `Sized` traits cannot be used as trait objects. I can't remember off the top of my head exactly what the best fix is here but, worst case you could use an enum of `Slot` implementors, since this will never be implemented outside of Core.

client/src/worker_registry/mod.rs
antlai-temporal: With the move to a single worker per namespace+queue+client it helps to have a worker id on register to detect multiple workers claiming the same namespace+queue+client, and ignore the registration. The logic on the caller worker gets easier too, they don't have to deal with whether they got a permit or not. So I think I'll stick to register(provider)/unregister(id)...

client/src/worker_registry/mod.rs
antlai-temporal: This not needed with the move to one worker per namespace+queue+client

client/src/worker_registry/mod.rs
cretz: Can't you just consider a second attempted registration of the same namespace+queue a failure? Why do you have to account for identifier at all?

antlai-temporal: No, I can't,  because registration will happen whether the programmer is using eager start or not in their code, and I want to make sure that I don't break anything in existing code... I also worry about a worker trying to register twice, and the id helps to decide to do a warn! or silently ignore it.

cretz: I think this kind of registry/registration should be eager-workflow-start specific and don't perform eager-workflow-enabled-worker-register if that's disabled.

antlai-temporal: The goal is that eager will always be enabled, and it is up to the app to add a flag to start_workflow to use eager. Since we don't know that beforehand , we have to assume that they may use eager, and we need to always register...

cretz: I don't think that's the goal though, and at least for now it's always disabled. Eager workflow start capability is enabled/disabled at a worker level (but we may allow additional overriding to disable it at the start level if enabled at the worker level) and it will be disabled by default (and we're by no means settled on whether that will change in the future).

antlai-temporal: I don't think it is enabled/disabled at worker level in neither java or go.  It is also an unnecessary complexity. Once we decide that we want to enable eager in the server, it should just work if clients use the flag in start_workflow....

cretz: I see, my mistake! So I guess you will register every one as eager-capable.

So how would a worker ever call register on itself twice necessitating an identifier (instead of once and given a thing for it to unregister itself)?

antlai-temporal: With the current implementation it won't happen, if the constructor fails you get a new id. But I was thinking that one day the registration may not be just for single process, but for a pool of local processes that can all benefit from early start, and having  a unique id for a worker in the trait method will hide that implementation. For debugging it also helps to have a unique id per worker, specially if we are imposing the restriction that only one can register.
 
And as I just mentioned, the "given a thing to unregister" forces the client to deal with failed registrations, while unregister with id is just transparent.

cretz: I personally wouldn't be concerned with a potential future of deduplicating, and having an optional "lease" for a successful registration stored on the worker to unregister is normal. The obvious benefits are 1) you don't need a separate trait compared to a single call, and 2) a worker doesn't have to be responsible for its identifier. But if consensus is that we want workers to have this identifier, ok.

client/src/worker_registry/mod.rs
antlai-temporal: Ok, that simplifies a lot of things. I'll just warn with ignored conflict registrations, I don't want to break existing code in the unlikely event they use multiple workers/queue in the same process...

core/src/worker/slot_provider.rs
antlai-temporal: Done

client/src/worker_registry/mod.rs
antlai-temporal: Consuming a Box<Slot> did the trick...

antlai-temporal: The `fn schedule_wft(self: Box<Self>...` did the trick... Thanks!

client/src/raw.rs
antlai-temporal: Thanks for the macro trick! Done...

core/src/worker/slot_provider.rs
```diff
@@ -130,12 +127,13 @@ mod tests {
...
-        if let Some(mut slot) = provider.try_reserve_wft_slot() {
+        if let Some(slot) = provider.try_reserve_wft_slot() {
             let p = slot.schedule_wft(new_validatable_response());
             assert!(p.is_ok());
 
-            let p = slot.schedule_wft(new_validatable_response());
-            assert!(p.is_err());
+            // This is a compiler error, schedule_wft consumes the slot...
+            // let p = slot.schedule_wft(new_validatable_response());
+            // assert!(p.is_err());
```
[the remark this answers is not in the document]
antlai-temporal: Done

client/src/worker_registry/mod.rs
```diff
@@ -58,24 +60,21 @@ impl SlotKey {
...
     /// Maps ids to keys in `providers`.
     index: HashMap<String, SlotKey>,
-}
-
-fn random_permutation(num_entries: usize) -> Vec<usize> {
-    let mut vec: Vec<usize> = (0..num_entries).collect();
-    vec.shuffle(&mut thread_rng());
-    vec
+    /// Do not warn when ignoring worker registrations.
+    quiet: bool,
```
Sushisource, line 68: Why add this? I don't think we need this any more than we need to ability to quiet warnings elsewhere in the codebase. Typically they can be silenced with an appropriate filter string in the logging config anyway.

client/src/worker_registry/mod.rs
```diff
@@ -157,6 +155,7 @@ impl SlotManager {
 
     #[cfg(test)]
     /// Returns (num_providers, num_buckets), where a bucket key is namespace+task_queue.
+    /// There is only one provider per bucket so `num_providers` should be equal to `num_buckets`.
     pub fn num_providers(&self) -> (usize, usize) {
         self.manager.read().num_providers()
     }
```
Sushisource, line 161: I think you can just get rid of this entirely now

core/src/worker/slot_provider.rs
```diff
@@ -130,12 +130,9 @@ mod tests {
...
             let p = slot.schedule_wft(new_validatable_response());
             assert!(p.is_ok());
-
-            // This is a compiler error, schedule_wft consumes the slot...
-            // let p = slot.schedule_wft(new_validatable_response());
-            // assert!(p.is_err());
-
             assert!(external_wft_rx.recv().await.is_some());
+        } else {
+            panic!("failed to reserve slot");
```
Sushisource, line 135: You can just use `.expect()` on the option now

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+
+/// Implements a [WorkerRegistry] and provides a convenient method
+/// to find compatible slots within the collection.
+/// This is an inner class to hide the mutex.
+#[derive(Default, Debug)]
+struct SlotManagerImpl {
+    /// Maps keys, i.e., namespace#task_queue, to provider.
+    providers: HashMap<SlotKey, Box<dyn SlotProvider + Send + Sync>>,
+    /// Maps ids to keys in `providers`.
+    index: HashMap<String, SlotKey>,
```
Sushisource, line 66: We don't need to use UUIDs here, the easiest way to do this would be to use `SlotMap` as seen here: https://github.com/temporalio/sdk-core/blob/ee7f6caaa9ad00bca2e2db8093d444ce563d6016/core/src/worker/workflow/machines/workflow_machines.rs#L129

When registering, workers get back a key, and they can hold onto it to unregister themselves.

Sushisource: I would have the slot map map worker keys to providers, and then have the extra map for `SlotKey -> WorkerKey`

Sushisource (commented):
Looking pretty good to me. Just a few more small changes and I think that's it.

client/src/worker_registry/mod.rs
antlai-temporal: In the unit tests I'm using a static SlotManager, and since workers do not properly unregister it could create many unnecessary warn! messages. So that flag just means "mock" the slot manager...

client/src/worker_registry/mod.rs
antlai-temporal: I was checking the invariant that both maps have the same size in the tests, but yes, not much value...

antlai-temporal: Sorry, or you meant just get rid of the comment?

core/src/worker/slot_provider.rs
antlai-temporal: Much clear, thanks!

client/src/worker_registry/mod.rs
antlai-temporal: This is related to the discussion above of whether we want the registry to identify the workers with a stable id, for debugging, or other future uses... But if both of you think there is no value, I'm happy to get rid of uuids and switch to registry-generated keys...

client/src/worker_registry/mod.rs
Sushisource: I meant the test generally

client/src/worker_registry/mod.rs
Sushisource: It's perfectly fine for unit tests to log warnings. Plus, the logger isn't even turned on for unit tests unless you explicitly add it.

client/src/worker_registry/mod.rs
Sushisource: Yeah, I don't think there's any reason to use uuids.

client/src/worker_registry/mod.rs
antlai-temporal: I use it in all the unit tests in that file, I need to check that things are deallocated/allocated properly, and with the wrapping  I can't get to the maps directly...

client/src/worker_registry/mod.rs
antlai-temporal: Great! Done...

client/src/worker_registry/mod.rs
antlai-temporal: Done, uuids replaced by SlotMap keys.

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+    fn register(&mut self, provider: Box<dyn SlotProvider + Send + Sync>) -> Option<WorkerKey> {
+        let key = SlotKey::new(
+            provider.namespace().to_string(),
+            provider.task_queue().to_string(),
+        );
+        if let Vacant(p) = self.providers.entry(key.clone()) {
+            p.insert(provider);
+            Some(self.index.insert(key))
+        } else {
+            warn!("Ignoring registration for worker in bucket {key:?}.");
```
Sushisource, line 105: "Bucket" isn't going to mean much for a user seeing the warning.
```suggestion
            warn!("Ignoring registration for worker: {key:?}.");
```

client/src/worker_registry/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+
+/// This trait enables local workers to made themselves visible to a shared client instance.
+/// There can only be one worker registered per namespace+queue_name+client, others will get ignored.
+#[cfg_attr(test, mockall::automock)]
+pub trait WorkerRegistry {
+    /// Register a local worker that can provide WFT processing slots.
+    fn register(&self, provider: Box<dyn SlotProvider + Send + Sync>) -> Option<WorkerKey>;
+    /// Unregister a provider, typically when its worker starts shutdown.
+    fn unregister(&self, id: WorkerKey);
+}
```
Sushisource, line 45: I'm inclined still to say this trait can just go away and these can be implemented on the `SlotManager` directly - and in general I try to avoid using mocks except when mocking things that require external resources (ex: the existing mocks are pretty much only for clients / pollers), but this is all internal so typically there's not a big reason to mock anything.

The other two traits make a bit more sense to me since they're implemented outside of the client crate.

core/src/worker/mod.rs
```diff
@@ -75,7 +81,8 @@ use {
 pub struct Worker {
     config: WorkerConfig,
     wf_client: Arc<dyn WorkerClient>,
-
+    /// Registration key for this worker
```
Sushisource, line 84: Worth saying what the registration key is for

Sushisource (approved):
Just a few more minor things, but, LGTM. Thanks!

client/Cargo.toml
```diff
@@ -23,6 +23,8 @@ once_cell = "1.13"
 opentelemetry = { workspace = true, features = ["metrics"] }
 parking_lot = "0.12"
 prost-types = "0.11"
+rand = "0.8.3"
```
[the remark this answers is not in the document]
antlai-temporal: Good catch thanks

client/src/worker_registry/mod.rs
antlai-temporal: Done

client/src/worker_registry/mod.rs
antlai-temporal: I didn't want "core" to see try_reserve_wft_slot, which can only be called by the client, but I'll remove it...

Sushisource: That's what pub(crate) is for

antlai-temporal: Good point, changing the visibility...

PR #623, opened 2023-11-01
Log pushing support

### LINKED ISSUE #618: [Feature Request] Support custom trait for log capture

Add a new enum to `temporal_sdk_core_api::telemetry::Logger` called `Custom` that accepts a `dyn` trait that has a method that is invoked with the `CoreLog` on every log. This is needed by .NET to push logs through the system instead of buffering (and any other SDK that's fast enough to process logs live)

cretz:
## What was changed

Create a new Telemetry option `temporal_sdk_core_api::telemetry::Logger::Push` which accepts a filter like the others, but also accepts an implementation of the `CoreLogConsumer` trait which is a simple single-function trait invoked for each log.

Also exposed a `temporal_sdk_core::telemetry::CoreLogBufferedConsumer` struct which, upon `new`, will return itself (an implementation of `CoreLogConsumer`) and a `temporal_sdk_core::telemetry::CoreLogBuffer` upon which `drain` can be invoked. Also altered log forwarding to leverage the new abstraction which means the foward.

Also expose a `temporal_sdk_core::telemetry::CoreLogStreamConsumer` that is backed by a bounded mpsc channel.

For langs where they need to buffer/throttle on a max-frequency, they are expected to use `CoreLogStreamConsumer::new` and for the stream response (second part of tuple), they can call [tokio_stream::StreamExt::chunks_timeout](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.chunks_timeout) on it.

## Checklist

1. Closes #618

### DIFF
--- a/core/src/telemetry/log_export.rs
+++ b/core/src/telemetry/log_export.rs
@@ -1,3 +1,4 @@
+use futures::channel::mpsc::{channel, Receiver, Sender};
 use parking_lot::Mutex;
 use ringbuf::{Consumer, HeapRb, Producer};
 use std::{collections::HashMap, fmt, sync::Arc, time::SystemTime};
@@ -141,6 +142,32 @@ impl CoreLogBuffer {
     }
 }
 
+/// Core log consumer implementation backed by a mpsc channel.
+pub struct CoreLogStreamConsumer {
+    tx: Sender<CoreLog>,
+}
+
+impl CoreLogStreamConsumer {
+    /// Create a stream consumer and stream of logs.
+    pub fn new(buffer: usize) -> (Self, Receiver<CoreLog>) {
+        let (tx, rx) = channel(buffer);
+        (Self { tx }, rx)
+    }
+}
+
+impl CoreLogConsumer for CoreLogStreamConsumer {
+    fn on_log(&self, log: CoreLog) {
+        // We will drop messages if we can't send
+        let _ = self.tx.clone().try_send(log);
+    }
+}
+
+impl fmt::Debug for CoreLogStreamConsumer {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("<stream consumer>")
+    }
+}
+
 struct JsonVisitor<'a>(&'a mut HashMap<String, serde_json::Value>);
 
 impl<'a> tracing::field::Visit for JsonVisitor<'a> {
@@ -190,7 +217,10 @@ impl<'a> tracing::field::Visit for JsonVisitor<'a> {
 
 #[cfg(test)]
 mod tests {
-    use crate::{telemetry::construct_filter_string, telemetry_init};
+    use crate::{
+        telemetry::construct_filter_string, telemetry::CoreLogStreamConsumer, telemetry_init,
+    };
+    use futures::stream::StreamExt;
     use std::fmt;
     use std::sync::{Arc, Mutex};
     use temporal_sdk_core_api::telemetry::{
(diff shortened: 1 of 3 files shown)


core/src/telemetry/log_export.rs
```diff
@@ -93,10 +95,52 @@ where
...
+        f.write_str("<buffered consumer>")
+    }
+}
+
+/// Buffer of core logs that can be drained.
+pub struct CoreLogBuffer {
+    logs_out: CoreLogsOut,
+}
+
+impl CoreLogBuffer {
```
cretz, line 137: Could expose fuzzy `len()`, but no use case yet

core/src/telemetry/log_export.rs
```diff
@@ -1,33 +1,35 @@
...
+type CoreLogsOut = Consumer<CoreLog, Arc<HeapRb<CoreLog>>>;
 
-pub(super) type CoreLogsOut = Consumer<CoreLog, Arc<HeapRb<CoreLog>>>;
+#[derive(Debug)]
+struct CoreLogFieldStorage(HashMap<String, serde_json::Value>);
 
-pub(super) struct CoreLogExportLayer {
-    logs_in: Mutex<Producer<CoreLog, Arc<HeapRb<CoreLog>>>>,
+pub(super) struct CoreLogConsumerLayer {
+    consumer: Arc<dyn CoreLogConsumer>,
```
Sushisource, line 13: Can this just be a `Box`? It's non-pub and doesn't look like it's cloned.

Sushisource (approved):
Works for me, just one thing to maybe change

core/src/telemetry/log_export.rs
cretz: I had that originally, but I needed `Arc` from a user for the log pushing since since `Logger` enum is cloneable (because all Telemetry options are). But I guess I can accept `Rc` from a user, `.into_inner()` it, and tell users they are not allowed to hold a reference to it. I will try.

Arguably as options contain more and more real implementations, options should not be cloneable.

bergundy (approved):
Makes sense to me.
Wondering if we want to also implement throttling and batching here, I would want this for TS and I would recommend using it in Python too.

cretz:
> Wondering if we want to also implement throttling and batching here, I would want this for TS and I would recommend using it in Python too.

I don't think you can do this properly here because it's dependent on how each lang will schedule a timer for draining. In Python for instance, I think we'll spawn a future call to drain if one is not already waiting. There are some race conditions. I will try out a potential Python implementation before I merge this.

bergundy:
Hmm... can't the timer be scheduled on the tokio runtime? Is that not accessible here?

core/src/telemetry/log_export.rs
Sushisource: Sure, fine to have it stay this way then, that's reasonable.

cretz:
After some discussion, I am going to try a stream-based log buffer without ringbuf

core/src/telemetry/log_export.rs
```diff
@@ -93,10 +96,78 @@ where
...
+}
+
+/// Core log consumer implementation backed by a mpsc channel.
+pub struct CoreLogStreamConsumer {
+    tx: Sender<CoreLog>,
+}
+
+impl CoreLogStreamConsumer {
+    /// Create a stream consumer and stream of logs.
+    pub fn new(buffer: usize) -> (Self, Receiver<CoreLog>) {
```
cretz, line 152: It is clearer here to return `Receiver` than some kind of `Box<dyn Stream<Item = CoreLog>>` because it's not sized which is required by some `StreamExt` calls

Sushisource: You can just return `impl Stream<Item = CoreLog>>`, but this is fine too.

Sushisource (approved):

core/src/telemetry/log_export.rs
cretz: :+1: I wasn't able to cleanly mark the trait as sized which was biting me when using some of the `StreamExt` calls (though can uncleanly do it if I must). Yeah, I usually wouldn't want to leak the specific trait impl I am using to the caller, but that it's sized and this is not some user-facing API settled it for me.

cretz:
Confirmed this works well with Python. Merging to incorporate into Python.

PR #625, opened 2023-11-03
Provide better error message when ephemeral download response is not successful

bergundy:


core/src/ephemeral_server/mod.rs
```diff
@@ -463,15 +463,19 @@ impl EphemeralExe {
...
                     .send()
-                    .await?
-                    .json()
                     .await?;
+                let info: DownloadInfo = match resp.error_for_status() {
+                    Ok(r) => r.json().await?,
+                    Err(e) => {
+                        return Err(anyhow::Error::from(e));
+                    }
+                };
```
josh-berry, line 478: For patterns like:

```rs
match some_result {
  Ok(_) => ...,
  Err(e) => return Err(...),
}
```

I suggest using [.map_err(...)?](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err) instead; I find it a lot cleaner to read.

Quinn-With-Two-Ns (approved):

core/src/ephemeral_server/mod.rs
```diff
@@ -591,10 +595,16 @@ async fn download_and_extract(
...
+    // Only stream on successful responses.
+    let stream = match resp.error_for_status() {
+        // We have to map the error type to an io error
+        Ok(resp) => resp
+            .bytes_stream()
+            .map(|item| item.map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))),
+        Err(e) => {
+            return Err(anyhow::Error::from(e));
+        }
+    };
```
josh-berry, line 607: Same as above

josh-berry (approved):
Non-blocking style comments only

core/src/ephemeral_server/mod.rs
bergundy: Yeah, I also prefer that but the result type is different.

bergundy: I'd need to cast from `Result<Response>` to `anyhow::Result<PathBuf>`.
Same applies below but there's a different return type.

bergundy: Hmm.. I think I figured out a better way to write it.

josh-berry (approved):
Oh nice, yeah, that's even better!

Issue #630, opened 2023-11-14
[Bug] Shutdown worker after replay test


antlai-temporal:
### Describe the bug

The recent changes to support Eager Workflow Start  (EWS) in core #622 introduced a regression in the Python SDK.

During a replay test, it relies on the input stream being closed by the mock client in order to propagate a
shutdown event to the Python worker. Unfortunately, EWS merges an external channel into this stream, and that channel
only closes much later, when the core worker completes shutdown. The result is that the input stream does not close, and 
the Python worker hangs.

PR #631, opened 2023-11-14
Fix worker shutdown during replay test

antlai-temporal:
This fixes issue #630 by closing the external channel early when we are using a mock client associated with the worker. 

This will disable Eager Start Workflow with mock clients, but since ESW is a latency optimization, and we do not use mock clients during its testing, this has no practical consequences.

1. Closes 
#630

### DIFF
--- a/client/src/worker_registry/mod.rs
+++ b/client/src/worker_registry/mod.rs
@@ -113,8 +113,6 @@ impl SlotManagerImpl {
 #[derive(Default, Debug)]
 pub struct SlotManager {
     manager: RwLock<SlotManagerImpl>,
-    /// True if this manager is associated with a mock client.
-    mock: bool,
 }
 
 impl SlotManager {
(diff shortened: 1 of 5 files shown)


client/src/worker_registry/mod.rs
```diff
@@ -113,16 +113,32 @@ impl SlotManagerImpl {
 #[derive(Default, Debug)]
 pub struct SlotManager {
     manager: RwLock<SlotManagerImpl>,
+    /// True if this manager is associated with a mock client.
+    mock: bool,
```
Sushisource, line 117: Rather than adding this here, I think it would be a little cleaner to add `is_mock` to the `WorkerClient` trait and simply not instantiate / merge the external WFT channel if it is.

antlai-temporal: Moved to WorkerClient.

The only reason to merge the closed stream is to satisfy the Rust typing gods... Select<Stream>  is not the same type as Stream, is there a way to cast it into a vanilla Stream?

antlai-temporal: Thanks for your great suggestion of using an Either stream! Channel not merged.

Sushisource (approved):

Issue #578, opened 2023-07-12
[Feature Request] Quickly fail query-only tasks which would overflow task buffer


Sushisource:
It's possible when clients are sending many queries very quickly to end up timing one or more of them out if it would require Core to buffer more than one WFT at a time. Because Core will only buffer 1 WFT at max, any queries dropped from that buffer will time out. This isn't a problem with WFTs with "real" work, since it only makes sense to service the latest one in any case.

When this happens, Core should just auto-fail the query that would have been dropped instead so that at least there is an explicit response to the query rather than a timeout. That response should include some metadata indicating the client should stop spamming so hard.

PR #636, opened 2023-11-17
Better WFT buffer

Sushisource:
## What was changed
Better task buffering.

## Why?
To avoid timing out concurrent queries. There is no reason to buffer more than one WFT, speculative or not, for the reasons described in the second issue. We already definitely will never try to apply more than one WFT at a time if the split brain edge case mentioned there happens.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/578 and closes https://github.com/temporalio/sdk-core/issues/578

3. How was this tested:
New tests

4. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/managed_run.rs
+++ b/core/src/worker/workflow/managed_run.rs
@@ -591,7 +591,6 @@ impl ManagedRun {
     pub(super) fn finish_activation(
         &mut self,
         pred: impl FnOnce(&OutstandingActivation) -> bool,
-        wft_from_complete: Option<PermittedWFT>,
     ) -> (bool, BufferedTasks) {
         let evict = if self.activation().map(pred).unwrap_or_default() {
             let act = self.activation.take();
@@ -599,14 +598,11 @@ impl ManagedRun {
         } else {
             false
         };
-        let mut buffered = if evict {
+        let buffered = if evict {
             mem::take(&mut self.task_buffer)
         } else {
             Default::default()
         };
-        if let Some(wft) = wft_from_complete {
-            buffered.buffer(wft);
-        }
         (evict, buffered)
     }
 
@@ -793,11 +789,10 @@ impl ManagedRun {
         work: PermittedWFT,
     ) -> Option<PermittedWFT> {
         let about_to_issue_evict = self.trying_to_evict.is_some();
-        let has_wft = self.wft().is_some();
         let has_activation = self.activation().is_some();
-        if has_wft || has_activation || about_to_issue_evict || self.more_pending_work() {
+        if has_activation || about_to_issue_evict || self.more_pending_work() {
             debug!(run_id = %self.run_id(),
-                   "Got new WFT for a run with outstanding work, buffering it");
+                   "Got new WFT for a run with outstanding work, buffering it act: {:?} wft: {:?} about to evict: {:?}", &self.activation(), &self.wft, about_to_issue_evict);
             self.task_buffer.buffer(work);
             None
         } else {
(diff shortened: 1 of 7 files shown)


core/src/worker/workflow/workflow_stream.rs
```diff
@@ -31,6 +31,7 @@ use tracing::{Level, Span};
 pub(super) struct WFStream {
     runs: RunCache,
     /// Buffered polls for new runs which need a cache slot to open up before we can handle them
+    /// TODO: Does this even need to exist any more?
```
Sushisource, line 34: Check this

core/src/worker/workflow/workflow_stream.rs
```diff
@@ -339,48 +341,36 @@ impl WFStream {
...
-                res = self.instantiate_or_update(PermittedWFT {
-                    work: wft,
-                    permit: t.permit,
-                    paginator: pag,
-                });
+                .or_else(|| self.buffered_polls_need_cache_slot.pop_front());
+            if let Some(wft) = maybe_buffered_wft {
+                res = self.instantiate_or_update(wft);
+                // We accept that there might be tasks remaining in the buffer if we evicted and
+                // re-instantiated here. It's likely those tasks are now invalidated anyway.
```
Sushisource, line 373: I could theoretically carry over any remaining buffered tasks to the next instantiation, but it adds quite a lot of complexity for little benefit IMO.

core/src/worker/workflow/workflow_stream.rs
Sushisource: It does, in fact, need to exist.

core/src/worker/workflow/managed_run.rs
```diff
@@ -71,11 +73,13 @@ pub(super) struct ManagedRun {
...
-    /// run. This can happen when lang takes too long to complete a task and the task times out, for
-    /// example. Upon next completion, the buffered response will be removed and can be made ready
-    /// to be returned from polling
-    buffered_resp: Option<PermittedWFT>,
+    /// Contains buffered poll responses from the server that apply to this run. This can happen
+    /// when:
+    ///   * Lang takes too long to complete a task and the task times out
+    ///   * Many queries are submitted concurrently and reach this worker (in this case, multiple
+    ///     tasks can be outstanding)
+    ///   * Multiple speculative tasks (ex: for updates) may also exist at once
```
Quinn-With-Two-Ns, line 81: Might be worth clarifying this statement that if multiple are sent only the latest is valid

core/src/worker/workflow/mod.rs
```diff
@@ -1003,6 +1021,38 @@ enum WFTReportStatus {
...
+    fn has_tasks(&self) -> bool {
+        self.wft.is_some() || !self.query_only_tasks.is_empty()
+    }
+
+    /// Remove and return the next WFT from the buffer that should be applied
+    fn get_next_wft(&mut self) -> Option<PermittedWFT> {
+        self.wft
+            .take()
+            .or_else(|| self.query_only_tasks.pop_front())
+    }
```
dandavison, line 1054: A comment about the ordering preference for WFT ahead of Query-only tasks would be good.

core/src/worker/workflow/managed_run.rs
Sushisource: Will add

core/src/worker/workflow/mod.rs
Sushisource: Will add, and this reminds me I meant to test that we won't reply to queries that are "old" if we processed a newer WFT than they targeted.

Sushisource:
It looks like I may *need* to carry over buffered tasks, or else tests that are validating small-cache scenarios have some issues.

PR #639, opened 2023-11-20
Order updates with signals

Sushisource:
## What was changed
Updates are now sorted as equal with signals. IE: Order is preserved relative to signals, but both signals and updates are moved before other jobs.

## Why?
It makes life easier for existing implementations, and specifically for languages that can't exert full control over when routines run (TS, mainly).

## Checklist

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?


sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
```diff
@@ -19,6 +19,32 @@ import "temporal/sdk/core/common/common.proto";
...
 // a cached state.
+//
+// ## Job ordering guarantees and semantics
+//
+// Core will, by default, order jobs within the activation as follows:
+// `patches -> signals/updates -> other -> queries -> evictions`
+//
+// This is because:
+// * Patches are expected to apply to the entire activation
+// * Signal and query handlers should be invoked before workflow routines are iterated
```
dandavison, line 30:
```suggestion
// * Signal and update handlers should be invoked before workflow routines are iterated
```

dandavison: This terminology to "iterate workflow routines" is a bit opaque to me.

sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
```diff
@@ -19,6 +19,32 @@ import "temporal/sdk/core/common/common.proto";
...
+// * Queries always go last (and, in fact, always come in their own activation)
+//
+// The downside of this reordering is that a signal or update handler may not observe that some
+// other event had already happened (ex: an activity completed) when it is first invoked, though it
+// will subsequently when workflow routines are driven. Core only does this reordering to make life
```
dandavison, line 35: >  a signal or update handler may not observe that some other event had already happened (ex: an activity completed) when it is first invoked

OK, this makes sense: the activity completion is in the "other" category.

> though it will subsequently when workflow routines are driven.

Struggling with this. Does "driving workflow routines" happen when jobs in the "other" category are handled? If so, the update handler has stopped executing by that point hasn't it, so in what sense is it able to "observe" something that it didn't when it first executed?

Sushisource: Yes, it does (though, per what I said in that other issue, ideally it happens at the end of processing all the jobs rather than inline with it).

The handler counts as one of the workflow routines, and so is again allowed to advance at the end of job processing. At that point, it would see the activity as completed, if it hadn't already exited.

antlai-temporal:
What happens when update handlers start local activities? Is the order of the local activities preserved, in the sense that they are moved before other jobs, and they respect the signals/update order?

core/src/worker/workflow/mod.rs
```diff
@@ -1441,14 +1445,18 @@ fn prepare_to_ship_activation(wfa: &mut WorkflowActivation) {
...
     #[test]
     fn jobs_sort() {
         let mut act = WorkflowActivation {
             jobs: vec![
                 WorkflowActivationJob {
                     variant: Some(workflow_activation_job::Variant::SignalWorkflow(
-                        Default::default(),
+                        SignalWorkflow {
+                            signal_name: "1".to_string(),
+                            ..Default::default()
```
antlai-temporal, line 1458: What happens if the application code also uses the signal_name "1", do we need to reserve some signal names?

Sushisource: Hmm? This is just test code.

Sushisource:
> What happens when update handlers start local activities? Is the order of the local activities preserved, in the sense that they are moved before other jobs, and they respect the signals/update order?

The two things aren't related. If an update handler starts local activities, the job for that update must have came either in that activation, or some previous one to start the handler. There are no jobs for the local activities until they resolve, and then they come in along with any other jobs, and signals and activities will be present before them.

core/src/worker/workflow/mod.rs
antlai-temporal: Sorry...Duh...

antlai-temporal (approved):

cretz (approved):
Python still has old code that is doing sorting of jobs. I have opened https://github.com/temporalio/sdk-python/issues/433 to remove that code and add an update-and-signal-interleaved test.

PR #645, opened 2023-11-25
Add feature for ephemeral server usage in test utils

h7kanna:
## What was changed
Added ephemeral-server feature to the test-utils similar to SDK and made it default.

## Why?
To optimize compilation for user crate which doesn't not depend on ephemeral server (just replays)

## Checklist

1. Closes 

2. How was this tested:
```
cargo test
cargo integ-test
```

3. Any docs updates needed?
No


Sushisource (approved):
Thanks!

PR #649, opened 2023-11-30
Update WfContext::cancelled to take &self instead of &mut self

rossdylan:
## What was changed
Updated the receiver for `WfContext::cancelled` to take in &self.

## Why?

It's the only method on `WfContext` that takes `&mut self` which makes it tricky to track workflow cancellation in addition to waiting for activities to complete.


Sushisource (approved):

Sushisource:
Thanks @rossdylan !

PR #651, opened 2023-12-04
Fix linter violations

dandavison:
Fix some linter violations introduced in https://github.com/temporalio/sdk-core/pull/649

### DIFF
--- a/tests/integ_tests/workflow_tests/cancel_external.rs
+++ b/tests/integ_tests/workflow_tests/cancel_external.rs
@@ -25,7 +25,7 @@ async fn cancel_sender(ctx: WfContext) -> WorkflowResult<()> {
     Ok(().into())
 }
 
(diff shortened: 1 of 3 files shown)


tests/integ_tests/workflow_tests/cancel_external.rs
```diff
@@ -25,7 +25,7 @@ async fn cancel_sender(ctx: WfContext) -> WorkflowResult<()> {
     Ok(().into())
 }
 
-async fn cancel_receiver(mut ctx: WfContext) -> WorkflowResult<()> {
+async fn cancel_receiver(ctx: WfContext) -> WorkflowResult<()> {
```
cretz, line 28: Can we fail CI on these violations so they don't happen anymore?

cretz (approved):

PR #660, opened 2023-12-14
Fix panic on evict w/ buffered tasks

Sushisource:
This never should have been a panic in the first place. I think I had it that way while testing and forgot to undo it. It's fairly hard to trigger.

### DIFF
--- a/core/src/worker/workflow/workflow_stream.rs
+++ b/core/src/worker/workflow/workflow_stream.rs
@@ -386,7 +386,7 @@ impl WFStream {
                 // We accept that there might be query tasks remaining in the buffer if we evicted
                 // and re-instantiated here. It's likely those tasks are now invalidated anyway.
                 if maybe_buffered.has_tasks() && should_evict {
-                    panic!("There were leftover buffered tasks");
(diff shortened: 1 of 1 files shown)


core/src/worker/workflow/workflow_stream.rs
```diff
@@ -386,7 +386,7 @@ impl WFStream {
                 // We accept that there might be query tasks remaining in the buffer if we evicted
                 // and re-instantiated here. It's likely those tasks are now invalidated anyway.
                 if maybe_buffered.has_tasks() && should_evict {
-                    panic!("There were leftover buffered tasks");
+                    warn!("There were leftover buffered tasks when evicting run");
```
dandavison, line 389: Can you confirm -- you don't need to drain the buffer contents? So these are either going to be treated as valid items, or the buffer is about to be drained anyway, or destroyed?

Sushisource: The tasks will just get dropped. It's not a meaningful issue for the reason the comment mentions. Worst case the task was valid, and it'll time out and get retried.

dandavison (approved):
LGTM!

PR #665, opened 2024-01-08
Make sure current task build id works with queries / caching

Sushisource:
## What was changed
Made the current task build id sent on activations consistent with what I just did in Go / Java.

## Why?
Part of https://github.com/temporalio/features/issues/253

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/machines/workflow_machines.rs
+++ b/core/src/worker/workflow/machines/workflow_machines.rs
@@ -46,8 +46,10 @@ use std::{
     hash::{Hash, Hasher},
     iter::Peekable,
     rc::Rc,
+    sync::Arc,
     time::{Duration, Instant, SystemTime},
 };
+use temporal_sdk_core_api::worker::WorkerConfig;
 use temporal_sdk_core_protos::{
     coresdk::{
         common::{NamespacedWorkflowExecution, VersioningIntent},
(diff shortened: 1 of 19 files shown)


core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -420,9 +418,15 @@ impl WorkflowMachines {
                 Some(workflow_activation_job::Variant::QueryWorkflow(_))
             )
         });
+        let is_replaying = self.replaying || all_query;
+        let build_id_for_current_task = if is_replaying {
+            self.current_wft_build_id.clone().unwrap_or_default()
```
Quinn-With-Two-Ns, line 423: Does core need the same logic as Go to check for an empty build ID?

core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -420,9 +418,15 @@ impl WorkflowMachines {
                 Some(workflow_activation_job::Variant::QueryWorkflow(_))
             )
         });
+        let is_replaying = self.replaying || all_query;
```
Quinn-With-Two-Ns, line 421: Aren't queries considered replaying anyway?

Sushisource: Not until this point, that's where it happens

core/src/worker/workflow/machines/workflow_machines.rs
Sushisource: The lookahead works a bit differently and avoids this problem - I actually ended up changing the Go one to be a bit simpler and more similar in my last fix too.

Quinn-With-Two-Ns (approved):

Issue #672, opened 2024-01-19
[Bug] Queries may still sometimes be dropped when buffering tasks


Sushisource:
### Describe the bug

As uncovered while fuzzing .NET. A query was definitely not delivered to lang although Core did receive it.

The attached log demonstrates we drop query 49. The sequence of events appears to be something like:

1. Get a query (48)
2. Get another query task (49), buffer it
3. Dispatch the first query to lang as an activation
4. Get another task (in this case, an update), buffer it
5. Update is dispatched to lang and somehow query 49 is lost

PR #673, opened 2024-01-19
Query buffering fixes

Sushisource:
## What was changed
Made buffered queries be appropriately handled before a new "real" WFT if one is received while still handling an outstanding one.

## Why?
The previous implementation banked on failing the tasks with the now "stale" queries if a new "real" WFT is received while they are buffered causing server to retry them while the client is still waiting on the query response. For whatever reason this doesn't always work. Instead, the implementation will now try to handle all buffered queries before applying the next real WFT. This potentially slows things down in very high-query-load scenarios, but that's more-or-less what Go and Java do and no one complains. Users shouldn't spam queries that hard anyway.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/672

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/mod.rs
+++ b/core/src/worker/workflow/mod.rs
@@ -50,6 +50,7 @@ use std::{
     collections::VecDeque,
     fmt::Debug,
     future::Future,
+    mem,
     mem::discriminant,
     ops::DerefMut,
     rc::Rc,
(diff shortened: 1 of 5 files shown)


core/src/worker/workflow/mod.rs
```diff
@@ -1023,17 +1024,31 @@ struct BufferedTasks {
...
+            if self.wft.is_none() {
+                self.query_only_tasks.push_back(task);
+            } else {
+                self.query_only_tasks_for_buffered.push_back(task);
+            }
         } else {
+            if self.wft.is_some() {
+                self.query_only_tasks_for_buffered.clear();
+            }
             let _ = self.wft.insert(task);
```
dandavison, line 1052: Why is this OK -- couldn't `self.wft` have signals and updates that we must process?

Sushisource: By definition it is stale and invalid now if there's a new incoming WFT

dandavison (approved):
LGTM!

Issue #477, opened 2023-01-30
[Feature Request] Dynamic client for workers


cretz:
### Is your feature request related to a problem? Please describe.

Users want to be able to change client options on running worker

### Describe the solution you'd like

~Initially we can start with exposing https://docs.rs/tonic/latest/tonic/transport/struct.Channel.html#method.balance_channel somehow via the Core client. Ideally this accepts most of the abstraction we already have in client options for the connection-specific params (would make sense to break up ClientOptions from ConnectOptions since only the latter can be pushed at runtime). Then lang can expose the capacity size and change mechanisms for users to push endpoint changes.~

EDIT: Updated title/description. We just need to be able to replace the client on the running worker for common use cases, e.g. updating endpoint or changing TLS options or something. Needs to do so without interrupting poll. For other non-worker client-change needs, users can just use new clients.

andrey-dubnik:
Hi, is it still a plan to move forward with this approach (we have discussed exactly this use case during the Replay catch-up)?

cretz:
There are discussions at https://github.com/temporalio/sdk-core/issues/565 for custom resolvers. I do not understand how "Replay catch-up" relates to client target host/endpoint. If wanting to discuss, feel free to join us on Slack or the forums.

PR #676, opened 2024-01-25
Dynamic client endpoints

antlai-temporal:
## What was changed

Clients can now dynamically update certain options fields such as the target url, the TLS configuration, the origin, or the keep alive settings, without creating a new client.

## Why?

This will allow, e.g., updating a TLS certificate without restarting workers, or losing state associated with the client, such as, eager workflow start tracking, or metrics...

## Checklist

1. Closes 
#477 
2. How was this tested:

Integration tests and unit tests

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -62,6 +62,10 @@ use temporal_sdk_core_protos::{
     },
     TaskToken,
 };
+use tokio::sync::{
+    mpsc::{error::SendError, Sender},
+    Mutex as TokioMutex,
+};
 use tonic::{
     body::BoxBody,
     client::GrpcService,
@@ -71,7 +75,7 @@ use tonic::{
     transport::{Certificate, Channel, Endpoint, Identity},
     Code, Status,
 };
-use tower::ServiceBuilder;
+use tower::{discover::Change, ServiceBuilder};
 use url::Url;
 use uuid::Uuid;
 
@@ -86,6 +90,9 @@ static LONG_POLL_METHOD_NAMES: [&str; 3] = [
 /// The server times out polls after 60 seconds. Set our timeout to be slightly beyond that.
 const LONG_POLL_TIMEOUT: Duration = Duration::from_secs(70);
 const OTHER_CALL_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Buffer size for the channel that listens to change events
+const DEFAULT_BUFFER_SIZE: usize = 1024;
 
 type Result<T, E = tonic::Status> = std::result::Result<T, E>;
 
@@ -134,8 +141,94 @@ pub struct ClientOptions {
     pub keep_alive: Option<ClientKeepAliveConfig>,
 }
 
+/// Connection options that can be dynamically updated. Construct with [ClientOptionsUpdateBuilder].
+/// All fields are optional, and will be ignored if not set. To disable optional fields in [ClientOptions],
+/// set field to a tombstone specified in the field description.
+#[derive(Clone, Debug, Default, derive_builder::Builder)]
+#[non_exhaustive]
+pub struct ClientOptionsUpdate {
+    /// The URL of the Temporal server to connect to.
+    #[builder(setter(into, strip_option), default)]
+    pub target_url: Option<Url>,
+
+    /// If specified, use TLS as configured by the [TlsConfig] struct. If this is set, core will
+    /// attempt to use TLS when connecting to the Temporal server. Lang SDK is expected to pass any
+    /// certs or keys as bytes, loading them from disk itself if needed.
+    /// If set with all optional fields to `None` in [TlsConfig], it disables TLS.
+    #[builder(setter(strip_option), default)]
+    pub tls_cfg: Option<TlsConfig>,
+
+    /// If set, override the origin used when connecting. May be useful in rare situations where tls
+    /// verification needs to use a different name from what should be set as the `:authority`
+    /// header.
+    /// If set with the URI default value, i.e., "/", it disables the origin override.
+    #[builder(setter(strip_option), default)]
+    pub override_origin: Option<Uri>,
+
+    /// If set (which it is by default), HTTP2 gRPC keep alive will be enabled.
+    /// If specified with all fields in [ClientKeepAliveConfig] set to zero, it disables `keep_alive`.
+    #[builder(setter(strip_option), default)]
+    pub keep_alive: Option<ClientKeepAliveConfig>,
+}
+
+impl ClientOptionsUpdate {
+    /// Override fields that are present in the proposed update.
+    pub fn merge(&mut self, update: ClientOptionsUpdate) {
+        if update.target_url.is_some() {
+            self.target_url = update.target_url;
+        }
+        if update.tls_cfg.is_some() {
+            self.tls_cfg = update.tls_cfg;
+        }
+        if update.override_origin.is_some() {
+            self.override_origin = update.override_origin;
+        }
+        if update.keep_alive.is_some() {
+            self.keep_alive = update.keep_alive;
+        }
+    }
+
+    /// Transform client options using this update.
+    pub fn apply(&self, options: &ClientOptions) -> ClientOptions {
+        let mut result = options.clone();
+        if let Some(target_url) = self.target_url.as_ref() {
+            result.target_url = target_url.clone();
+        }
+
+        if let Some(tls_cfg) = self.tls_cfg.as_ref() {
+            result.tls_cfg = if *tls_cfg == Default::default() {
+                // Default, i.e., all fields `None`, is a tombstone for tls_cfg
+                None
+            } else {
+                Some(tls_cfg.clone())
+            };
+        }
+
+        if let Some(override_origin) = self.override_origin.as_ref() {
+            // Tombstone is "/"
+            result.override_origin = if *override_origin == Uri::default() {
+                None
+            } else {
+                Some(override_origin.clone())
+            }
+        }
+
+        if let Some(keep_alive) = self.keep_alive.as_ref() {
+            // Tombstone sets all durations to zero
+            let zero_duration = Duration::new(0, 0);
+            result.keep_alive =
+                if keep_alive.interval == zero_duration && keep_alive.timeout == zero_duration {
+                    None
+                } else {
+                    Some(keep_alive.clone())
+                }
+        }
+
+        result
+    }
+}
 /// Configuration options for TLS
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, PartialEq)]
 pub struct TlsConfig {
     /// Bytes representing the root CA certificate used by the server. If not set, and the server's
     /// cert is issued by someone the operating system trusts, verification will still work (ex:
@@ -149,7 +242,7 @@ pub struct TlsConfig {
 }
 
 /// If using mTLS, both the client cert and private key must be specified, this contains them.
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
 pub struct ClientTlsConfig {
     /// The certificate for this client
     pub client_cert: Vec<u8>,
@@ -272,6 +365,28 @@ pub enum ClientInitError {
     SystemInfoCallError(tonic::Status),
 }
 
+/// Errors thrown while updating an existing client.
+#[derive(thiserror::Error, Debug)]
+pub enum ClientUpdateError {
+    /// Cannot change the channel configuration.
+    #[error("Cannot send the update: {0:?}")]
+    ControlChannelError(#[from] SendError<Change<i32, Endpoint>>),
+    /// Invalid update, cannot create the new endpoint.
+    #[error("Cannot create an endpoint: {0:?}")]
+    InvalidUpdateError(#[from] ClientInitError),
+}
+
+/// Metadata to enable the dynamic configuration of a client.
+#[derive(Clone, Debug)]
+pub struct DynamicUpdateInfo {
+    /// Version number of the current update, or zero if no updates have been applied.
+    version: i32,
+    /// Control channel sender
+    tx: Sender<Change<i32, Endpoint>>,
+    /// Summary of contributions from all the previous updates.
+    changeset: ClientOptionsUpdate,
+}
+
 /// A client with [ClientOptions] attached, which can be passed to initialize workers,
 /// or can be used directly. Is cheap to clone.
 #[derive(Clone, Debug)]
@@ -282,6 +397,7 @@ pub struct ConfiguredClient<C> {
     /// Capabilities as read from the `get_system_info` RPC call made on client connection
     capabilities: Option<get_system_info_response::Capabilities>,
     workers: Arc<SlotManager>,
+    update: Arc<TokioMutex<DynamicUpdateInfo>>,
 }
 
 impl<C> ConfiguredClient<C> {
@@ -305,6 +421,57 @@ impl<C> ConfiguredClient<C> {
     /// Returns a cloned reference to a registry with workers using this client instance
     pub fn workers(&self) -> Arc<SlotManager> {
         self.workers.clone()
+    }
+
+    /// Patches the client options, and then dynamically updates the endpoint with them.
+    /// If the update fails, the client does not rollback to the previous configuration, and future
+    /// connections will fail.
+    /// Internally it maintains a mutable changeset so that we can apply a sequence of updates without a
+    /// mutable client reference. However, this means it cannot update the `options` field.
+    /// To keep `options` consistent use `update_options_mut` or, when needed, manually `apply()` the
+    /// changeset to the original options.
+    pub async fn update_options(
+        &self,
+        update: ClientOptionsUpdate,
+    ) -> Result<ClientOptions, ClientUpdateError> {
+        // We take a "hard" approach for updating. First remove the previous endpoint, and if successful,
+        // add the updated one. If there is a failure, don't try to recover, and just let the upper layer
+        // handle the failed client.
+        // This approach may leave a short window in which the client cannot open new connections. Again,
+        // we let the upper layer handle this infrequent failure.
+        // The "soft" approach, i.e., remove-last or override-when-ready, cannot guarantee when the update
+        // happens, and also the asynchronous update could fail silently.
+
+        let mut update_meta = self.update.lock().await;
+        let remove_change = Change::Remove(update_meta.version);
+        if let Err(err) = update_meta.tx.send(remove_change).await {
+            return Err(ClientUpdateError::ControlChannelError(err));
+        }
+        update_meta.changeset.merge(update);
+        let new_options = update_meta.changeset.apply(&self.options);
+        update_meta.version += 1;
+        let endpoint = new_options.create_endpoint().await?;
+        let add_change = Change::Insert(update_meta.version, endpoint);
+        if let Err(err) = update_meta.tx.send(add_change).await {
+            return Err(ClientUpdateError::ControlChannelError(err));
+        }
+
+        Ok(new_options)
+    }
+
+    /// Similar to `update_options` but it also updates the `options` field in this `ConfiguredClient`
+    /// to reflect the changes.
+    pub async fn update_options_mut(
+        &mut self,
+        update: ClientOptionsUpdate,
+    ) -> Result<ClientOptions, ClientUpdateError> {
+        match self.update_options(update).await {
+            Ok(new_options) => {
+                *Arc::get_mut(&mut self.options).unwrap() = new_options.clone();
+                Ok(new_options)
+            }
+            Err(error) => Err(error),
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


client/src/lib.rs
```diff
@@ -134,8 +141,94 @@ pub struct ClientOptions {
     pub keep_alive: Option<ClientKeepAliveConfig>,
 }
 
+/// Connection options that can be dynamically updated. Construct with [ClientOptionsUpdateBuilder].
+/// All fields are optional, and will be ignored if not set. To disable optional fields in [ClientOptions],
+/// set field to a tombstone specified in the field description.
+#[derive(Clone, Debug, Default, derive_builder::Builder)]
+#[non_exhaustive]
+pub struct ClientOptionsUpdate {
```
cretz, line 149: Can the entire `ClientOptions` struct be reused instead? Not every field has to be optional. You can require the callers provide the same data they did for original client options if they don't want to change it. The lang side doesn't need help here.

(would also be neat if some of the runtime options like headers was set/updated via a common mechanism and didn't require reconnect, but that doesn't have to be now)

client/src/lib.rs
```diff
@@ -306,6 +422,57 @@ impl<C> ConfiguredClient<C> {
     pub fn workers(&self) -> Arc<SlotManager> {
         self.workers.clone()
     }
+
+    /// Patches the client options, and then dynamically updates the endpoint with them.
+    /// If the update fails, the client does not rollback to the previous configuration, and future
+    /// connections will fail.
```
cretz, line 428: This is rough. Why can't we only apply the update after the new endpoint is connected and works? Can't you do an atomic swap of the endpoint where there is no downtime and it only contains a successfully connected endpoint?

client/src/lib.rs
```diff
@@ -87,6 +91,9 @@ static LONG_POLL_METHOD_NAMES: [&str; 3] = [
 const LONG_POLL_TIMEOUT: Duration = Duration::from_secs(70);
 const OTHER_CALL_TIMEOUT: Duration = Duration::from_secs(30);
 
+/// Buffer size for the channel that listens to change events
+const DEFAULT_BUFFER_SIZE: usize = 1024;
```
Sushisource, line 95: I suspect this can probably be a lot smaller (though I doubt it matters a whole lot either way).

client/src/lib.rs
Sushisource: I think this probably makes decent sense too. I imagine the lang-side API will typically involve the user copying some existing options and changing them, in which case this all works out fine.

antlai-temporal: These are the only fields that I can change dynamically, because they are associated with the endpoint, and I think it makes it more clear what can be changed with a separate struct. Also, fields like `target_url` are not optional in ClientOptions but need to be optional in update, or remember the original value set, which is a pain with concurrent updates...

client/src/lib.rs
```diff
@@ -272,6 +365,28 @@ pub enum ClientInitError {
...
+    #[error("Cannot send the update: {0:?}")]
+    ControlChannelError(#[from] SendError<Change<i32, Endpoint>>),
+    /// Invalid update, cannot create the new endpoint.
+    #[error("Cannot create an endpoint: {0:?}")]
+    InvalidUpdateError(#[from] ClientInitError),
+}
+
+/// Metadata to enable the dynamic configuration of a client.
+#[derive(Clone, Debug)]
+pub struct DynamicUpdateInfo {
```
Sushisource, line 381: Doesn't need to be pub

client/src/lib.rs
Sushisource: Agreed. I would call this `try_update_options` and only swap if the connection succeeds. Also I find leaving the old options to be scary since now things aren't in sync.

It'd be fine to put a mutex around the options as well, so that they can be updated too. This also fits better with the idea of using the `ClientOptions` directly rather than `ClientOptionsUpdate`

client/src/lib.rs
Sushisource: If we go with swapping out the whole options behind a mutex which I mentioned in another comment, then the other fields should be changeable too I think, which could work out nicely.

client/src/lib.rs
antlai-temporal: I've been looking into the atomic swap reusing the id for the endpoint, but the problem is that it s anything but atomic.
 When you reuse the id the endpoint gets in a "preparing" queue, the old endpoint still active, when the new one gets connected the swap happens. The problem is that this is all asynchronous, and if there is a problem with the new endpoint there is no visible error, and the old one continuous forever. It is also impossible to know when this swap will happen, or validate that it happened without sending a carefully crafted request. The endpoints are also created in lazy mode, so they need to be part of channel to validate them.
I think the root problem is that this mechanism was designed for load balancing, not dynamic changes, and if you are adding more (similar) endpoints with different targets, you can be more relaxed about when exactly changes happened.
For this reason, I'm forcing the client to fail when there are problems, otherwise they get silently ignored...

client/src/lib.rs
antlai-temporal: I just copied the constant used internally by tonic, frankly, no idea about the right size... But, yes, being a control channel, that looks large...

client/src/lib.rs
antlai-temporal: They may need to access the changeset to get the current options when there is no mutable ref to the client.

antlai-temporal:
>Agreed. I would call this try_update_options and only swap if the connection succeeds. Also I find leaving the old options to be scary since now things aren't in sync.

This is what I tried first, but as I explained above, it is all asynchronous, with lazy endpoints, and there is no error callback, so it is difficult to  make the update only when it works...
Leaving the old options is really ugly, agreed. I checked that we don't use the possibly outdated fields in any critical manner but there is no guarantee that something will eventually break. It is also unrealistic that we will always have a mutable ref to client unless we add a top level mutex, which would mess up backwards compatibility.
 Adding a lock on ClientOptions does not completely solve the problem, you need to hold both locks (also for update field), to safely do the update, and locks need to be acquired once, and in the correct order, to avoid deadlocks. I thought that could be quite intrusive too...

If we can live with occasional inconsistencies, we could lock them independently... That would improve on what we have when we have no mutable ref... 

>It'd be fine to put a mutex around the options as well, so that they can be updated too. This also fits better with the idea of using 
>the ClientOptions directly rather than ClientOptionsUpdate
See discussion above...
The problem is that you need to hold both locks to keep it consistent

client/src/lib.rs
cretz: > The problem is that this is all asynchronous

Yeah I see that with the balance channel and makes sense. But can we wait to even send the insert until we have connected the endpoint? And can we send the remove after the insert? So logic would be:

```
endpoint = connect_new()
channel_tx.send(insert(endpoint))
channel_tx.send(remove(old_endpoint_seq))
```

Also curious, if you send an insert then a remove, is it guaranteed that the _next_ call will be on the inserted?

client/src/lib.rs
antlai-temporal: Things like `retry_config` are used by the Retry wrapper that I cannot change.  `client_name` or `client_version` or `identity` may be used in metrics, we probably don't want to change them, and I cannot see a use case to change them dynamically... We could fail the update if they don't match. 
I still think that is easier in the API to do deltas with only the changeable attributes, but internally we could use ClientOptions instead of a changeset, if we are ok with a mutex on options. We would still need some state for the updates that does not belong in options, such as the channel, the version id,... but we could probably lock them independently from options...

client/src/lib.rs
antlai-temporal: I don't think that you can easily swap the channel without creating a new client, and the problem of creating new clients is that there is a lot of shared state associated to them, like eager workflow start workers, or metrics,... that need to be moved across atomically.
The semantics of the control channel are really weak. Remove happens immediately, but inserts are queued in a "preparing" queue, so there is no guarantee that after sending insert ok you are getting the new options.
To avoid the gap you can reuse the id in insert, and then you don't need a remove, but there is no guarantee when, or if, the swap will happen.

cretz: Thanks for the detail!

> I don't think that you can easily swap the channel without creating a new client, and the problem of creating new clients is that there is a lot of shared state associated to them, like eager workflow start workers, or metrics,... that need to be moved across atomically.

Yeah I see now that it gets a bit tangled there. I definitely wouldn't want a new top-level client, but I wonder if reasonable to put `TemporalServiceClient` or some of those even lower clients behind a mutex. We of course could do this all the way out on lang side if it wasn't for the need to change a workers' client. Currently this is what our customers do to switch clients, it's just workers that are the problems.

> Remove happens immediately, but inserts are queued in a "preparing" queue

So with the current code there's a potential to have no endpoints (i.e. remove went throw and insert is preparing) or does it read from the stream in that case?

antlai-temporal: Yes, I had this comment in update_options
        // This approach may leave a short window in which the client cannot open new connections. Again,
        // we let the upper layer handle this infrequent failure.
I'm hoping that the retry would take care of it. In practice, the move from "preparing" queue to "ready" seems to be really quick, or triggered by the first request, and I have not been able to see it in the system tests. And changing the endpoint should only affect new requests, so the likelihood is much smaller.
But if all else fails, you could ended up with a no-endpoint client for a short while, and the error will propagate...

About doing surgery on the  low level client, not sure if there is some state that needs to be moved too, @Sushisource  can you comment?

client/src/lib.rs
Sushisource: Why can't they be changed? I agree if they can't be changed it makes sense to accept something that only allows changing what can be. But, I'm not sure I follow why we couldn't change those. They are accessed each time a request is made - seems like changing them would affect subsequent requests and work properly?

client/src/lib.rs
Sushisource: It's not accessible as it stands anyway. It's stored as a private field and there's no accessor.

client/src/lib.rs
antlai-temporal: for retry_config we cache it inside:
impl<SG> RetryClient<SG> {
    /// Use the provided retry config with the provided client
    pub fn new(client: SG, retry_config: RetryConfig) -> Self {
        Self {
            client,
            retry_config: Arc::new(retry_config),
        }
    }
}
we could change that if needed, but not sure if it is that important to dynamically change retry_config...

The others we can, but the question is whether we should, if it is used in metrics, or server side, it may be confusing, and
I don't see much value of changing them dynamically...

client/src/lib.rs
Sushisource: Swapping out `TemporalServiceClient`s would be reasonable, that's just a set of the lowest-level tonic clients. And, in fact, you don't even need a lock necessarily. You could probably use https://docs.rs/arc-swap/latest/arc_swap/ since it only needs to affect new requests.

However, the big problem there is clients wrap other clients generically using traits, so you don't know when you're down to the lowest level where you should be doing the swap. Changing everything to make that work would probably be annoying.

What Chad suggested w/ trying to connect the endpoint, and then sending it down the control channel if it works is what I was imagining. You can sort of arbitrarily make the whole thing sync by doing that, removing the old endpoint, adding the new one, then trying to call some rpc like server capabilities with some tight retry until it either works and you're done, or fails and you restore the old endpoint.

client/src/lib.rs
Sushisource: We don't have to actually expose those to end users, but I take your point. Especially since the `RetryClient` is a level higher. I'm fine with having just the diff struct.

client/src/lib.rs
antlai-temporal: Good point. We need to decide whether to get rid of changeset and update ClientOptions directly (with mutex), or not, and then either add the accessor or make everything private.

client/src/lib.rs
cretz: > // This approach may leave a short window in which the client cannot open new connections. Again,
> // we let the upper layer handle this infrequent failure.

While our worker stuff is often self-recovering, the client stuff is not and some users use low-level pieces directly without explicit retry. I think we're going to have a hard time explaining a series of client failures during endpoint swap. We'll basically tell them to use their own swap if they want zero downtime because ours doesn't have that (and ignore worker warnings if we even log them). If at all possible, would like to avoid downtime just to change out the underlying client.

> However, the big problem there is clients wrap other clients generically using traits, so you don't know when you're down to the lowest level where you should be doing the swap

My naive guess would be to make every trait support updating options and have wrappers' impls call the underlying client until it reaches the true final client.

antlai-temporal: >What Chad suggested w/ trying to connect the endpoint, and then sending it down the control channel if it works is >what I was imagining. You can sort of arbitrarily make the whole thing sync by doing that, removing the old endpoint, >adding the new one, then trying to call some rpc like server capabilities with some tight retry until it either works and >you're done, or fails and you restore the old endpoint.
Yes, I could make it a bit more robust by adding an rpc server check, like passing the ConfiguredClient and calling `get_system_info`, and try to undo the endpoint if it fails. However, other concurrent users of that client could still see a failure during that process, since there is no top level mutex, maybe that it fails for **everybody** gives us better error semantics. 
Trying to test the endpoint beforehand has less value. The api expects an endpoint in lazy mode, and the internal checks for moving from "preparing" to "ready" are repeated anyway. I will probably need to clone the endpoint and use a separate client to test it, and not sure we are gaining much over the "end-to-end" test of the rpc server check above...

antlai-temporal: >While our worker stuff is often self-recovering, the client stuff is not and some users use low-level pieces directly without >explicit retry. I think we're going to have a hard time explaining a series of client failures during endpoint swap.

Do they remove the RetryClient part they get from the bridge, or you are thinking other SDKs like haskell?

I think the biggest value of this is for the worker, clients are easier to update with no need for dynamic stuff. 

But, as I said, the gap is theoretical, have not seen it in system tests that immediately try to use the updated client, and it may be that the request is on hold until the move from "preparing" to  "ready" happens, or is a quick local test...

cretz: > Do they remove the RetryClient part they get from the bridge, or you are thinking other SDKs like haskell?

They make raw gRPC calls to the workflow service.

> But, as I said, the gap is theoretical

At the risk of over-simplifying, can we just do a zero-downtime client swap after guaranteed connection? Ideally this would work for all clients given new client options (or fail trying), but even if it just means a worker getting a `set_client`, that should be enough for most users and they will manage swapping their client uses their own ways.

antlai-temporal: I think multiple endpoints is something else, we could do another PR for that.

Guaranteed connection before we start swapping, and get-system-info check before returning from update sounds good, will make things a bit more robust.

Now, zero-downtime is a trade-off vs predictability on when the change happens. I think this won't be a practical issue, specially when clients retry, and validation is quick, but if we don't delete the previous connection first, we cannot know when the new one is active, so get-system-info doesn't help. I frankly prefer losing a bit of availability to get better consistency.
 
Note that swapping the channel is actually worse for zero-downtime. It would affect existing connections, not only new ones, because returned values from the server may get ignored.

Given your feedback, my preference now is check before, check after, and tell people to retry for raw grpc, or use our RetryClient, or assume that once in a blue moon, you may get a connection error when upgrading...

Things that I am still on the fence:
  -Shall we try to restore the original endpoint if get-system-info fails? The error may already had caused other  users of this client failing. It may be better to have a clean, consistent error rather than a "some fail" type of error. And we check the connection before, so this would be a very rare event...
- We have `options()` and `options_mut()`. If we protect ClientOptions with a mutex, I'm thinking on returning a clone in `options()` and the Arc<Mutex> with `options_mut()`, so that `options()` may be inconsistent but won't block updates.
- With the mutex on ClientOptions get rid of ChangeSet and just update options directly internally.

Any thoughts? @Sushisource @cretz Shall I go ahead with the changes?

cretz: > if we don't delete the previous connection first, we cannot know when the new one is active, so get-system-info doesn't help. I frankly prefer losing a bit of availability to get better consistency.

If this is behind a mutex, can we just assume that all outstanding calls may use old clients but all future calls will use new client? In other words, can all client use obtain the client from the mutex _for every gRPC request_ as needed (e.g. for worker polling) which means if the client is swapped it will be used for the next poll call but won't affect the current?

antlai-temporal: Unfortunately the mutex is not top level, protecting the client, it is just guarding the options field (and  another one internal for update). Trying to change that now will break everything (i.e., all core SDKs bridges, cli,..)... And there are questions on what performance impact, or deadlocks, we will introduce by grabbing a lock, cloning,  on every request. 
 So my goal was to be as little intrusive as I could, even if things are not perfect...

cretz: > Trying to change that now will break everything (i.e., all core SDKs bridges, cli,..)...

It shouldn't break anything, this should be an internal mutex to use the underlying thing. Maybe it is a channel and not a client, I don't know the word, but whatever the "thing that makes RPC calls is and closes itself on drop". If we call that `X`, I'd expect that `Mutex<Arc<X>>` would exist.

>  And there are questions on what performance impact, or deadlocks, we will introduce by grabbing a lock, cloning, on every request.

I don't think either is that much of a concern. The lock is only the life of an `Arc` clone I think. I think this is needed if you want to swap it and at negligible cost.

cretz: Here's what ChatGPT suggested:

```rust
use std::sync::{Arc, RwLock};
use std::thread;

struct Client {
    // Client data
}

impl Client {
    fn new() -> Self {
        Client {
            // Initialization
        }
    }

    fn use_client(&self) {
        // Use the client
    }
}

fn main() {
    let client = Arc::new(RwLock::new(Client::new()));

    // Simulating a thread that uses the client
    let client_clone = Arc::clone(&client);
    thread::spawn(move || {
        let read_client = client_clone.read().unwrap();
        read_client.use_client();
    });

    // Simulating swapping the client
    let new_client = Client::new();
    let mut write_client = client.write().unwrap();
    *write_client = new_client;
}
```

cretz: After chatting with ChatGPT, it ended up with `Arc<RwLock<Arc<Client>>` like so:

```rust
use std::sync::{Arc, RwLock};
use std::thread;

struct Client {
    // Client data
}

impl Client {
    fn new() -> Self {
        Client {
            // Initialization
        }
    }

    fn use_client(&self) {
        // Use the client
    }
}

fn main() {
    let client = Arc::new(RwLock::new(Arc::new(Client::new())));

    // Simulating a thread that uses the client
    let client_clone = Arc::clone(&client);
    thread::spawn(move || {
        let client_arc = client_clone.read().unwrap().clone();
        // The lock is released here, after cloning the Arc
        client_arc.use_client(); // use_client is called outside of the lock
    });

    // Simulating swapping the client
    let new_client = Arc::new(Client::new());
    let mut write_client = client.write().unwrap();
    *write_client = new_client;
}
```

antlai-temporal: If we call `use_client` outside of the lock it doesn't help, right? We need to initiate the request holding the lock...

Interesting that chatGPT gave it a try, scary the number of bugs that it will create with subtle concurrency code...

cretz: > If we call use_client outside of the lock it doesn't help, right? We need to initiate the request holding the lock...

I think it will help. Because that client is still under `Arc` so it is not dropped, it is just no longer on the field once swapped. So if you did a swap, once that `use_client()` call was done, it'd be the last reference and would drop it, but all others accessing the client get the new one.

To me, and again forgive my Rust rustiness, it is a case of atomically getting or swapping the client, but users of the client should not hold onto the lock while it is used. So you can swap a client out on the field that is in use by some function.

cretz: Here's how it would look in Go if it reads better:

```go
type ClientHolder struct {
  // Client is thread safe
  myClient *Client
  myClientLock sync.RWLock
}

func (c *ClientHolder) Swap(newClient *Client) {
  c.myClientLock.Lock()
  defer c.myClientLock.Unlock()
  c.myClient = newClient
}

func (c *ClientHolder) Get() *Client {
  c.myClientLock.RLock()
  defer c.myClientLock.RUnlock()
  return c.myClient
}
```

Basically the `Get` client can be used by the caller regardless of whether it's still the _current_ client on the outer structure. We just expect the caller to use it for one request (again, it may no longer be the _current_ one) and call `Get` again for their next client need.

antlai-temporal: The problem is that it is "lazy". It will create the connection with the first request, and we want to make sure that this connection is not created until we finish the upgrade.

cretz: You should attempt a get-system-info call on the client before swapping it. IMO we want to make sure the connection _is_ created and is a known config to be successful even way before we swap it out. We don't want to swap out lazy or a when given a bad client, the worker will just silently fail for a minute (our default retry iirc) then fatal out.

antlai-temporal: Before the swap we can check the connection, but that doesn't help for lazy. Internally Tonic defers the connection until it gets the first request. And even if we do a get-system-info, there is a race with someone else doing the first call after dropping the lock, if they grabbed it before the update started, and getting the error...

Sushisource: I would probably do this:

```rust
pub struct ConfiguredClient<C> {
    client_and_options: ArcSwap<ClientAndOptions>,
    headers: Arc<RwLock<HashMap<String, String>>>,
    /// Capabilities as read from the `get_system_info` RPC call made on client connection
    capabilities: Option<get_system_info_response::Capabilities>,
    workers: Arc<SlotManager>,
}

pub struct ClientAndOptions<C> {
    client: C,
    options: ClientOptions
}

```

If ArcSwap doesn't work out for whatever reason (it might be hard to use `load()` references in all cases) then you can do the `Arc<RwLock<Arc<ClientAndOptions>>>` approach

Issue #659, opened 2023-12-14
[Bug] Local activity markers might create commands after workflow completion


Sushisource:
This manifests like
```
2023-12-14T17:56:20.923613Z  WARN temporal_sdk_core::worker::workflow: Error while completing workflow activation error=status: InvalidArgument, message: "invalid command sequence: [StartTimer, CompleteWorkflowExecution, RecordMarker], command CompleteWorkflowExecution must be the last command.", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

90% sure LAs are completing after the fact and we're adding the command when we should just be discarding it

* When fixed, note in fuzzy trophy case

PR #678, opened 2024-02-02
Fix possibly sending LA marker after WF Complete command

Sushisource:
## What was changed
Detect terminal commands and avoid doing anything else with LAs other than issuing cancels to outstanding ones.

Also fixed a possible shutdown hang that this surfaced.

## Why?
Bugfix

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/659

2. How was this tested:
Added unit test which repro'd issue consistently, fixed it.

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/mod.rs
+++ b/core/src/worker/workflow/mod.rs
@@ -180,58 +180,60 @@ impl Workflows {
         // We must spawn a task to constantly poll the activation stream, because otherwise
         // activation completions would not cause anything to happen until the next poll.
         let tracing_sub = telem_instance.and_then(|ti| ti.trace_subscriber());
-        let processing_task = thread::spawn(move || {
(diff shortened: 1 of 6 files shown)


core/src/worker/workflow/mod.rs
```diff
@@ -180,58 +180,60 @@ impl Workflows {
...
-            let local = LocalSet::new();
-            local.block_on(&rt, async move {
-                let mut stream = WFStream::build(
-                    basics,
-                    extracted_wft_stream,
-                    locals_stream,
-                    local_activity_request_sink,
-                );
+        let processing_task = thread::Builder::new()
+            .name("workflow-processing".to_string())
```
Sushisource, line 184: Only change here is naming this thread properly. Everything else is whitespace.

cretz (approved):

PR #679, opened 2024-02-02
sdk: allow example code to run

djc:
## What was changed
When running the Rust SDK example, the binary exits with `Error: UninitializedField("worker_build_id")`.

## Why?
Would be nice to have the example work out of the box, even while the Rust SDK is not actually released.

## Checklist
1. How was this tested: ran the code after adding this change, it runs without error
2. Any docs updates needed: probably not at this stage?


djc:
Not sure why there are two comments from the CLAassistant saying different things?

Sushisource (approved):
Thank you!

Issue #594, opened 2023-08-21
[Feature Request] Do not complete shutdown until all activities have returned and results pushed to server


cretz:
### Describe the solution you'd like
Today we are relying on lang (e.g. https://github.com/temporalio/sdk-python/pull/370) to wait for all activities they start to finish completion so we can confirm the values are pushed to server. Ideally core should not complete shutdown with any outstanding activities or unpushed activity completions.

PR #681, opened 2024-02-03
Wait for activity completes to reach server before shutdown

Sushisource:
## What was changed
Make sure we flush activity completions to server before shutting down

## Why?
It's nice. We already wait for outstanding activities to complete. It makes sense to make sure we actually do something with those results.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/594

2. How was this tested:
Added UT

3. Any docs updates needed?


cretz (approved):

bergundy:
@mjameswh we may want to modify TS to not also wait on activity completion to be delivered to the server, need to look into this.

Definitely better done Core side.

PR #682, opened 2024-02-06
client: add retry policy param to start_workflow

Herklos:
## What was changed
Retry policy attribute has been added to client `start_workflow` method to allow the Rust client to specify a retry policy when starting a workflow.

## Why?
It's a useful param for some use cases and it was not available on Rust client while it is available for client of other languages (like go, typescript or python).


Sushisource:
Thanks for this!

Sushisource (approved):

PR #683, opened 2024-02-08
client: clarify encoding of ClientTlsConfig field data

djc:
## What was changed
Add a `_pem` suffix to both `ClientTlsConfig` fields, and remove the `client_` prefix (which is redundant with the `ClientTlsConfig` type name) to try to balance the length increase.

## Why?
While trying to get a Rust SDK worker connected to Temporal Cloud, I got a `TonicTransportError(tonic::transport::Error(Transport, PrivateKeyParseError))`. This turned out to because I was passing DER-encoded private key and certificate to `ClientTlsConfig` instead of PEM. tonic's [`Identity::from_pem()`](https://docs.rs/tonic/latest/tonic/transport/struct.Identity.html#method.from_pem) makes it obvious that they should be PEM, but that context isn't obvious when looking at the rustdoc for `temporal_client::ClientTlsConfig` (and in fact, because it uses a `Vec<u8>` type instead of a `String` type, it suggests using DER over PEM).

## Checklist
1. How was this tested: no logic affected, so relying on the compiler
2. Any doc updates needed: should be reflected in rustdoc automatically

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -151,9 +151,9 @@ pub struct TlsConfig {
 /// If using mTLS, both the client cert and private key must be specified, this contains them.
 #[derive(Clone)]
 pub struct ClientTlsConfig {
-    /// The certificate for this client
+    /// The certificate for this client, encoded as PEM
     pub client_cert: Vec<u8>,
-    /// The private key for this client
+    /// The private key for this client, encoded as PEM
     pub client_private_key: Vec<u8>,
 }
 


client/src/lib.rs
```diff
@@ -151,10 +151,10 @@ pub struct TlsConfig {
 /// If using mTLS, both the client cert and private key must be specified, this contains them.
 #[derive(Clone)]
 pub struct ClientTlsConfig {
-    /// The certificate for this client
-    pub client_cert: Vec<u8>,
-    /// The private key for this client
-    pub client_private_key: Vec<u8>,
```
Sushisource, line 157: I'd prefer not to rename the fields as all the language SDKs are going to have to deal with that change and while it's simple, I don't know that it's totally worth it.

That said, the comment changes are a great way to add clarity. Let's just do that.

djc:
I've rolled back the field rename, and added a commit that cleans up a clippy issue likely new in Rust 1.76.

Sushisource:
@djc This and your other PR need a rebase/merge. Normally I can do that from right in here but the option isn't showing up. I'm guessing maybe because you haven't allowed commits from maintainers: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

djc:
> @djc This and your other PR need a rebase/merge. Normally I can do that from right in here but the option isn't showing up. I'm guessing maybe because you haven't allowed commits from maintainers: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Hmm, I'm not seeing this checkbox show up.

Is there any chance/method by which you can put me on an allowlist to allow CI to run without waiting for explicit approval? What with the timezone differences, these feedback cycles are pretty slow. :(

djc:
> I've changed it to only requiring approval for new contributors.

Thanks! Rebased this on top of current master, I guess that will do the job?

djc:
The remaining test failure looks unrelated to this change?

Sushisource:
> The remaining test failure looks unrelated to this change?

Yeah it flakes occasionally. I'll re-run it and merge it.

Sushisource (approved):

Issue #687, opened 2024-02-16
[Feature Request] Managing dependencies for the Rust SDK


djc:
### Is your feature request related to a problem? Please describe.

It would be great if the Rust SDK dragged in as few dependencies as possible. Dependencies increase build times throughout downstream projects. While some dependencies provide invaluable functionalities, other times there might be a way to do the same work without taking on an additional dependency.

Our monorepo contains about 90k lines of Rust code and pulls in some 800 crates. Adding the temporal-sdk stack of crates added a bunch more. It would be nice to minimize the impact on build times for our repo (and I assume it might also help for your build times in lang SDKs).

### Describe the solution you'd like

Make good trade-offs about which dependencies provide enough value vs being a net drag on compile time.

### Additional context

Here are indirect dependencies that were added to our workspace `Cargo.lock` when adding temporal-sdk:

- arc-swap 1.6.0
- backoff 0.4.0
- convert_case 0.4.0 (via derive_more -- convert_case 0.6.0 was already in our transitive dependencies)
- crossbeam 0.8.4 (but its constituent dependencies were already in our transitive dependencies)
- rustc_version (via derive_more)
- enum-iterator 1.5.0
- enum-iterator-derive 1.3.0
- enum_dispatch 0.3.12
- futures-retry 0.6.0
- lru 0.11.1 [vs lru-cache?]
- matchers 0.1.0 (enabled via tracing-subscriber)
  - regex-automata 0.1.10 (0.4.5 was already in our transitive dependencies)
- mockall 0.11.4
  - downcast 0.11.0
  - fragile 2.0.0
  - mockall_derive 0.11.4
  - predicates 2.1.5
    - difflib 0.4.0
    - float-cmp 0.9.0
    - normalize-line-endings 0.3.0
    - predicates-core 1.0.6
  - predicates-tree 1.0.9
    - termtree 0.4.1
- opentelemetry-prometheus 0.14.1
  - prometheus 0.13.3
  - protobuf 2.28.0
- ringbuf 0.3.3
- rustfsm 0.1.0
  - rustfsm-procmacro 0.1.0
  - rustfsm_trait 0.1.0
- siphasher 1.0.0
- slotmap 1.0.7
- tonic-build 0.9.2 (0.11.0 was already in our transitive dependencies)
  - prettyplease 0.1.25 (0.2.16 was already in our transitive dependencies)
  - prost-build 0.11.9 (0.12.3 was already in our transitive dependencies)
[...]

PR #689, opened 2024-02-16
Use fine-grained features for derive_more

djc:
## What was changed

Changed the derive_more dependency to specify `default-features = "false"` plus a minimal set of derives that are actually used.

## Why?

Helps minimize transitive dependencies brought in by derive_more.

## Checklist

1. Closes: related to #687
2. How was this tested: it still compiles


Sushisource (approved):

djc:
Rebased.

PR #690, opened 2024-02-16
core: make opentelemetry dependencies optional

djc:
## What was changed

Make the telemetry features optional but enabled by default in the core SDK.

This is a first take -- happy to take feedback and iterate.

## Why?

As discussed in #687, it would be nice to limit dependencies. When using the Rust SDK, the application might have its own tracing subscriber setup, and it would be better for the libraries not to interfere with that.

## Checklist

1. See #687
2. How was this tested: it still compiles
3. Any docs updates needed: not sure?

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -13,11 +13,12 @@ categories = ["development-tools"]
 [lib]
 
 [features]
-default = []
+default = ["otel"]
 # Do not enable this feature when building production SDKs. If we ever want a user in the field to
 # record WF input data, we can build them a custom SDK or they can build - it adds significant extra
 # code size in the form of [de]serializers.
 save_wf_inputs = ["rmp-serde", "temporal-sdk-core-protos/serde_serialize"]
+otel = ["dep:opentelemetry", "dep:opentelemetry_sdk", "dep:opentelemetry-otlp", "dep:opentelemetry-prometheus"]
 tokio-console = ["console-subscriber"]
 ephemeral-server = ["dep:flate2", "dep:nix", "dep:reqwest", "dep:tar", "dep:zip"]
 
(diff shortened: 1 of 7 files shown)


core/src/telemetry/metrics.rs
```diff
@@ -1,16 +1,20 @@
...
+};
 use crate::{
     abstractions::dbg_panic,
-    telemetry::{
-        default_resource, metric_temporality_to_selector, prometheus_server::PromServer,
-        TelemetryInstance, TELEM_SERVICE_NAME,
-    },
+    telemetry::{TelemetryInstance, TELEM_SERVICE_NAME},
 };
+#[cfg(feature = "telemetry")]
```
Sushisource, line 9: This is a lot of places to introduce the feature guard. I'd say ideally let's push these down into a submodule which is guarded, and re-export them here behind another guard.

core/src/telemetry/mod.rs
```diff
@@ -3,19 +3,21 @@
...
 mod log_export;
 pub(crate) mod metrics;
+#[cfg(feature = "telemetry")]
 mod prometheus_server;
 
-pub use metrics::{
-    build_otlp_metric_exporter, default_buckets_for, start_prometheus_metric_exporter,
-    MetricsCallBuffer,
-};
+#[cfg(feature = "telemetry")]
```
Sushisource, line 9: Same here about reducing the number of guards. Probably the same module can be used from both places. Likely an `otel` submodule.

core/Cargo.toml
```diff
@@ -13,11 +13,12 @@ categories = ["development-tools"]
 [lib]
 
 [features]
-default = []
+default = ["telemetry"]
 # Do not enable this feature when building production SDKs. If we ever want a user in the field to
 # record WF input data, we can build them a custom SDK or they can build - it adds significant extra
 # code size in the form of [de]serializers.
 save_wf_inputs = ["rmp-serde", "temporal-sdk-core-protos/serde_serialize"]
+telemetry = ["dep:opentelemetry", "dep:opentelemetry_sdk", "dep:opentelemetry-otlp", "dep:opentelemetry-prometheus"]
```
Sushisource, line 21: Given this is really only about otel right now, maybe let's call the feature `otel` rather than `telemetry`

Sushisource (commented):
General idea makes sense, just a few suggestions

djc:
Good suggestion about the module, I think this looks a lot cleaner.

Sushisource (approved):
Thanks!

Sushisource:
@djc This needs to merge from master or rebase. I would do it for you but, same deal I mentioned before: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

There should be a way to turn that on so I can easily do things like that

djc:
I looked at the docs last time and didn't find the option. I can probably fix it up myself in a bit once my daughter is in bed.

djc:
This is what it looks like for me:

Screenshot 2024-03-01 at 20 09 25

Sushisource:
Weird. I don't get why the option isn't showing up for you. That's a bit obnoxious.

PR #693, opened 2024-02-23
Up incoming gRPC message limit / allow override

Sushisource:


bergundy (approved):
Might be worth testing this but this LGTM. Thanks!

client/src/lib.rs
```diff
@@ -512,23 +524,27 @@ where
     }
     /// Get the underlying workflow service client
     pub fn workflow_svc(&self) -> &WorkflowServiceClient<T> {
-        self.workflow_svc_client
-            .get_or_init(|| WorkflowServiceClient::new(self.svc.clone()))
+        self.workflow_svc_client.get_or_init(|| {
+            WorkflowServiceClient::new(self.svc.clone()).max_decoding_message_size(*DECODE_MAX_SIZE)
```
Quinn-With-Two-Ns, line 528: In Go we also apply this limit to to the sending side as well

Quinn-With-Two-Ns (approved):

client/src/lib.rs
Sushisource: Sending side _would_ be limited by server though I believe

client/src/lib.rs
```diff
@@ -493,6 +495,19 @@ pub struct TemporalServiceClient<T> {
     test_svc_client: OnceCell<TestServiceClient<T>>,
     health_svc_client: OnceCell<HealthClient<T>>,
 }
+
+/// We up the limit on incoming messages from server from the 4Mb default to 128Mb. If for
+/// whatever reason this needs to be changed by the user, we support overriding it via env var.
+fn get_decode_max_size() -> usize {
```
bergundy, line 501: I would prefer if we also supported setting this programmatically but this is good enough.

client/Cargo.toml
```diff
@@ -31,6 +31,7 @@ tower = "0.4"
 tracing = "0.1"
 url = "2.2"
 uuid = { version = "1.1", features = ["v4"] }
+lazy_static = "1.4.0"
```
djc: Drive-by comment: it is my understanding that lazy-static is more or less deprecated, and people should be using once_cell instead -- which offers a more idiomatic anyway. In newer versions, you can even rely on `OnceLock` from `std` directly, although that is a little less ergonomic than once_cell's `Lazy`.

djc: https://github.com/temporalio/sdk-core/pull/694

PR #694, opened 2024-02-23
Remove lazy-static in favor of once_cell

djc:
## What was changed

Removed dependencies on lazy-static in favor of once_cell. Use a workspace dependency to manage the once_cell dependency in one place.

## Why?

lazy-static is [effectively deprecated](https://github.com/rust-lang-nursery/lazy-static.rs/issues/214) and the ecosystem has largely moved over to once_cell. once_cell has a more ergonomic API.

## Checklist

1. How was this tested: still compiles
2. Any docs updates needed: don't think so?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ license-file = "LICENSE.txt"
 
 [workspace.dependencies]
 derive_more = { version = "0.99", default-features = false, features = ["constructor", "display", "from", "into"] }
+once_cell = "1.16"
 tonic = "0.9"
 tonic-build = "0.9"
 opentelemetry = "0.21"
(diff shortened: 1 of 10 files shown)


Sushisource (approved):
Cool, thanks for this, just needs formatting.

djc:
Cleaned up the formatting and the missed `lazy_static` in the integration tests.

djc:
Ugh -- what's the reason `cargo check --all-targets` doesn't find the integration tests? Would be nice not to require the use of custom aliases for basic checks like this.

Sushisource:
> Ugh -- what's the reason `cargo check --all-targets` doesn't find the integration tests? Would be nice not to require the use of custom aliases for basic checks like this.

There are a handful of off-by-default features and there was no simple way to get every combo of feature/target to actually test compilation all at once

PR #696, opened 2024-02-27
Slow down retries so we keep retrying for more than 10 seconds.

chronos-tachyon:
## What was changed
* Increase backoff factor to 1.7 (from 1.5) for non-poll RPCs

## Why?
The current backoff factor causes clients using the Core SDK to exhaust their retries within 6.011 to 8.966 seconds, depending on random jitter.  We want to ensure that all clients keep retrying for at least 10 seconds.  With the new backoff factor, retries will not be exhausted until 11.878 to 17.767 seconds after the initial failure.

This advances progress on https://github.com/temporalio/features/issues/27 (aka SDK-118 in internal Jira).


Sushisource (approved):
Thanks!

PR #697, opened 2024-03-01
Update dependencies

djc:
## What was changed

Update outdated dependencies

## Why?

Should help reduce duplicate dependencies in downstream user's repositories. In particular, I noticed that the older derive_builder version pulls in older versions of syn and the darling crates.

## Checklist

1. How was this tested: `cargo lint && cargo test-lint`
2. Any docs updates needed: no

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ license = "MIT"
 license-file = "LICENSE.txt"
 
 [workspace.dependencies]
+derive_builder = "0.12"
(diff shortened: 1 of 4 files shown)


Sushisource (approved):
Thanks! I'd been waiting for otel to update so I can finally update tonic and everything else, but they tend to take forever.

djc:
I think you can do tonic and otel now, actually!

PR #699, opened 2024-03-08
Explicit API key option on client

### LINKED ISSUE #695: [Feature Request] API key client option
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/426

cretz:
## What was changed

* Move `headers` from `connect` to the `ClientOptions` (💥 - intentional compatibility break)
* Add `api_key` to `ClientOptions` and `set_api_key` on client to use as an "Authorization" bearer header if that header not already set

## Checklist

1. Closes #695

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -67,7 +67,7 @@ use tonic::{
     body::BoxBody,
     client::GrpcService,
     codegen::InterceptedService,
-    metadata::{MetadataKey, MetadataValue},
+    metadata::{MetadataKey, MetadataMap, MetadataValue},
     service::Interceptor,
     transport::{Certificate, Channel, Endpoint, Identity},
     Code, Status,
@@ -133,6 +133,15 @@ pub struct ClientOptions {
     /// If set (which it is by default), HTTP2 gRPC keep alive will be enabled.
     #[builder(default = "Some(ClientKeepAliveConfig::default())")]
     pub keep_alive: Option<ClientKeepAliveConfig>,
+
+    /// HTTP headers to include on every RPC call.
+    #[builder(default)]
+    pub headers: Option<HashMap<String, String>>,
+
+    /// API key which is set as the "Authorization" header with "Bearer " prepended. This will only
+    /// be applied if the headers don't already have an "Authorization" header.
+    #[builder(default)]
+    pub api_key: Option<String>,
 }
 
 /// Configuration options for TLS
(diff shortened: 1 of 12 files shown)


client/src/lib.rs
```diff
@@ -309,6 +322,34 @@ impl<C> ConfiguredClient<C> {
...
 }
 
+#[derive(Debug)]
+struct ClientHeaders {
+    user_headers: HashMap<String, String>,
+    api_key: Option<String>,
+}
+
+impl ClientHeaders {
+    fn apply_to_metadata(&self, metadata: &mut MetadataMap) {
```
cretz, line 332: I considered doing some of this work on set rather than apply (md key/val creation, concat w/ "Bearer "), but it didn't seem too burdensome to do it on the call

Sushisource: This makes more sense to me anyway I think. Sure, it's "extra" work but totally irrelevant to performance and this is a bit more flexible

Sushisource (approved):
Sweet, looks good to me. Thanks!

cretz:
Just realized my build/test didn't catch all uses of the call, checking integ_tests and such and will push fixes...

client/src/lib.rs
```diff
@@ -288,8 +297,12 @@ pub struct ConfiguredClient<C> {
...
     pub fn set_headers(&self, headers: HashMap<String, String>) {
-        let mut guard = self.headers.write();
-        *guard = headers;
+        self.headers.write().user_headers = headers;
+    }
+
+    /// Set API key, overwriting previous
+    pub fn set_api_key(&self, api_key: Option<String>) {
+        self.headers.write().api_key = api_key;
     }
```
antlai-temporal, line 306: How do you keep consistent the headers in `options.headers` and the `headers` field in ConfiguredClient? Or are we assuming `options` are just the initial frozen options?

client/src/lib.rs
```diff
@@ -279,7 +288,7 @@ pub enum ClientInitError {
 pub struct ConfiguredClient<C> {
     client: C,
     options: Arc<ClientOptions>,
-    headers: Arc<RwLock<HashMap<String, String>>>,
+    headers: Arc<RwLock<ClientHeaders>>,
```
antlai-temporal, line 291: I'm working on dynamic endpoints, updating the options field and reconnecting. Is the idea that update headers, that do not need reconnect, with always be updated with `set_headers()` and not by modifying `options`?

client/src/lib.rs
cretz: Options are just the initial frozen options as they always have been (but lang can make theirs mutable if they really want, which Python does but .NET doesn't)

client/src/lib.rs
cretz: Yes, that idea has already been the case (`set_headers` has already existed, this just moves things around a bit). I think we need to differentiate things that don't require reconnect (i.e. HTTP headers) from things that do (i.e. hostname and TLS options).

Issue #656, opened 2023-12-11
[Feature Request] Ability for certain task failure types to fail workflow


cretz:
### Describe the solution you'd like

From https://github.com/temporalio/features/issues/322, need:

* `WorkerConfig.workflow_failure_errors: HashSet<WorkflowTaskFailureType>` (only value for `WorkflowTaskFailureType` today is `Nondeterminism`)
* `WorkerConfig.workflow_failure_errors_by_workflow_type: HashMap<String, HashSet<WorkflowTaskFailureType>>`

Sushisource:
I don't think this is the right way to do this in Core. Core doesn't need to try to understand the types of exceptions thrown from lang. Lang can just set an explicit field on the response that says "this should fail the workflow".

cretz:
> Core doesn't need to try to understand the types of exceptions thrown from lang. Lang can just set an explicit field on the response that says "this should fail the workflow".

Completely agree. This is only very specifically for the one task failure type that core _does_ understand - nondeterminism. People want to be able to say "core-based nondeterminism fails workflow". I just figured requesting a collection is more future proof. But all the other cases of task failure lang for sure should handle.

PR #702, opened 2024-03-13
Enable failing WF on nondeterminism errors

Sushisource:
## What was changed
Allow failing the WF instead of the task on nondeterminism errors

## Why?
Useful setting

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/656
2. How was this tested:
Added integ tests

3. Any docs updates needed?

### DIFF
--- a/core-api/src/worker.rs
+++ b/core-api/src/worker.rs
@@ -150,6 +150,20 @@ impl WorkerConfig {
             .saturating_sub(self.max_nonsticky_polls())
             .max(1)
     }
+    /// Returns true if the configuration specifies we should fail a workflow on a certain error
+    /// type rather than failing the workflow task.
+    pub fn should_fail_workflow(
+        &self,
+        workflow_type: &str,
+        error_type: &WorkflowErrorType,
+    ) -> bool {
+        self.workflow_failure_errors.contains(error_type)
+            || self
+                .workflow_types_to_failure_errors
+                .get(workflow_type)
+                .map(|s| s.contains(error_type))
+                .unwrap_or_default()
+    }
 }
 
 impl WorkerConfigBuilder {
--- a/core-api/src/worker.rs
+++ b/core-api/src/worker.rs
@@ -162,7 +162,7 @@ impl WorkerConfig {
                 .workflow_types_to_failure_errors
                 .get(workflow_type)
                 .map(|s| s.contains(error_type))
-                .unwrap_or_default()
+                .unwrap_or(false)
     }
 }
 
--- a/core-api/src/worker.rs
+++ b/core-api/src/worker.rs
@@ -1,4 +1,8 @@
-use std::time::Duration;
+use crate::errors::WorkflowErrorType;
+use std::{
+    collections::{HashMap, HashSet},
+    time::Duration,
+};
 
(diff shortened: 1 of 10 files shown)


core/src/worker/workflow/managed_run.rs
```diff
@@ -570,9 +578,36 @@ impl ManagedRun {
                 )
             }
         } else if should_report {
-            ActivationCompleteOutcome::ReportWFTFail(FailedActivationWFTReport::Report(
-                tt, cause, failure,
-            ))
+            // Check if we should fail the workflow instead of the WFT because of user's preferences
+            if matches!(cause, WorkflowTaskFailedCause::NonDeterministicError)
```
cretz, line 582: There is a `self.metrics.wf_task_failed` above, can we remove this before that or move that into the conditionals? Similarly, will this make sure to increment the workflow failure counter? Any other things about workflow failure that is accounted for by the other workflow failure code path but not this one (e.g. eviction)?

core-api/src/worker.rs
```diff
@@ -136,6 +150,20 @@ impl WorkerConfig {
...
+        &self,
+        workflow_type: &str,
+        error_type: &WorkflowErrorType,
+    ) -> bool {
+        self.workflow_failure_errors.contains(error_type)
+            || self
+                .workflow_types_to_failure_errors
+                .get(workflow_type)
+                .map(|s| s.contains(error_type))
+                .unwrap_or_default()
```
dandavison, line 165: nit: `unwrap_or(false)` seems clearer; does `unwrap_or_default()` have an advantage here?

core/src/worker/workflow/managed_run.rs
Sushisource: Eviction will happen on this path too.

As for the metric, I didn't avoid it since you could in some sense still consider this "failing" the task, it's just reported differently. I suppose we could go either way there. For the workflow failure counter, yeah I'll need to make sure that does get hit, thanks for mentioning.

Sushisource: I have made it go through the same path so we get this stuff without much change.

I've kept it adding to the task fail metric. To my mind this is _both_ failing a task, and the workflow, in some sense. We already increment the task failure counter for WFT fails that aren't reported to server because they've happened >2 times, and this feels similar to that.

cretz: Regarding task vs workflow fail, I think it's either/or and we should increment the counter for which path is taken instead of both. Inability to report a task failure I think may be different (TIL that we don't just send a task failure to the server every time the server sends us a task).

I think it's clearest to think from a user POV. When they opt in to this, e.g. `Workflow(FailureExceptionTypes=new[] { typeof(NondeterminismException), typeof(MyException) })`, they are basically flipping both of those from task fail to wf fail, even if internally we are only sending nondet flip to core and leaving user-exception flip in lang. So from a user POV, they are just two things that can happen that used to increment task fail and now increment wf fail metrics.

Sushisource: Sure, I just think both interpretations are valid, but I'll make it either or since I don't care much

cretz (approved):

Issue #554, opened 2023-05-17
[Feature Request] Possibly add config options for exporting time metrics in seconds w/ unit suffix


Sushisource:
### Is your feature request related to a problem? Please describe.
Core exports time metrics in `ms` where Go / Java use seconds and optionally add a `_seconds` suffix to the metric names.

### Describe the solution you'd like
It's not clear that adding this config option to Core and handling it there makes sense post merge of https://github.com/temporalio/sdk-core/pull/544 

But we should consider how we want that to work once someone's done a lang implementation of that interface

PR #706, opened 2024-03-14
Allow emitting duration metrics in seconds

Sushisource:
## What was changed
Support floating-point histograms to allow emitting metrics in seconds.

:boom:  Breaking changes for lang :boom: 
* The `CoreMeter` interface has changed to support `f64` histograms/gauges and `Duration`-accepting histograms
* `CoreMeter` implementations determine what units are emitted by a `HistogramDuration`
    * The `unit` field on the parameter is expected to be overridden by such impls
* Correspondingly, the `MetricEvent`s used with the buffer for lang-side metrics impls have new variants

## Why?
Some people desire seconds-based latencies. Also will work OOTB with our dashboards for other langs

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/554

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/client/src/metrics.rs
+++ b/client/src/metrics.rs
@@ -6,7 +6,7 @@ use std::{
     time::{Duration, Instant},
 };
 use temporal_sdk_core_api::telemetry::metrics::{
-    CoreMeter, Counter, Histogram, MetricAttributes, MetricKeyValue, MetricParameters,
+    CoreMeter, Counter, HistogramDuration, MetricAttributes, MetricKeyValue, MetricParameters,
     TemporalMeter,
 };
 use tonic::{body::BoxBody, transport::Channel};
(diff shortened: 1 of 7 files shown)


cretz (approved):
Can you confirm from a user Prometheus and OTel output POV there are no breaking changes (in labels or values or anything)?

Also before merging, which will cost a bit of time on core upgrade at least for Python and .NET, can you think of anything coming soon in SDK core that we may want to get in first?

Sushisource:
> Can you confirm from a user Prometheus and OTel output POV there are no breaking changes (in labels or values or anything)?
> 
> Also before merging, which will cost a bit of time on core upgrade at least for Python and .NET, can you think of anything coming soon in SDK core that we may want to get in first?

No there are no changes to any labels or values, unless of course you choose seconds in which case the buckets are necessarily different in numerical terms, but are the same in terms of value.

There's nothing else in progress now that would matter to get in

PR #714, opened 2024-04-05
HTTP CONNECT proxy support

### LINKED ISSUE #309: [Feature Request] Support for HTTP CONNECT proxies

gRPC implementations in JS, Go, and Java all support the `GRPC_PROXY`/`HTTP_PROXY` env vars from https://github.com/grpc/grpc/blob/master/doc/environment_variables.md which allow proxying via `HTTP CONNECT` based proxies, but I can find no such support in https://github.com/hyperium/tonic.

Ideally we could contrib support, but in the absence of that, we may need to call out somewhere that gRPC proxy environment variables are not supported for TypeScript workers and that Envoy or another separate proxy should be used.

This issue is just investigatory to confirm what we do and don't support here.

cretz:
## What was changed

* Added new `client::proxy` module for HTTP CONNECT support
  * Hyper/Tonic don't support this natively, so some workarounds had to be done
  * Based on user need, this also includes authentication
* Added new `http_connect_proxy` client option

Can't really perform a test of this without a more involved integration test that has a full proxy server implementation. I have manually tested this with Python SDK using this version of core and connecting to Temporal cloud through the proxy. Now that proxies are officially supported by Temporal, there's a feature PR at https://github.com/temporalio/features/pull/447 for testing them, and Python will be added once this is merged and the Python SDK is updated

## Checklist

1. Closes #309

### DIFF
--- a/client/src/proxy.rs
+++ b/client/src/proxy.rs
@@ -1,0 +1,85 @@
+use base64::prelude::*;
+use hyper::header;
+use std::future::Future;
+use std::pin::Pin;
+use std::task::Context;
+use std::task::Poll;
+use tokio::net::TcpStream;
+use tonic::transport::Channel;
+use tonic::transport::Endpoint;
+use tower::{service_fn, Service};
+
+/// Options for HTTP CONNECT proxy.
+#[derive(Clone, Debug)]
+pub struct HttpConnectProxyOptions {
+    /// The host:port to proxy through.
+    pub target_addr: String,
+    /// Optional HTTP basic auth for the proxy.
+    pub basic_auth: Option<(String, String)>,
+}
+
+impl HttpConnectProxyOptions {
+    /// Create a channel from the given endpoint that uses the HTTP CONNECT proxy.
+    pub async fn connect_endpoint(
+        &self,
+        endpoint: &Endpoint,
+    ) -> Result<Channel, tonic::transport::Error> {
+        let proxy_options = self.clone();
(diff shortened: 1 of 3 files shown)


client/src/proxy.rs
```diff
@@ -0,0 +1,85 @@
...
+use tonic::transport::Endpoint;
+use tower::{service_fn, Service};
+
+/// Options for HTTP CONNECT proxy.
+#[derive(Clone, Debug)]
+pub struct HttpConnectProxyOptions {
+    /// The host:port to proxy through.
+    pub target_addr: String,
+    /// Optional HTTP basic auth for the proxy.
+    pub basic_auth: Option<(String, String)>,
```
Sushisource, line 18: Definitely ought to say what the two string fields are

Sushisource (approved):
Just one docstring to fix. Looks like this was fairly straightforward, nice.

client/src/proxy.rs
cretz: Fixed

cretz:
> Looks like this was fairly straightforward, nice.

The result was fairly straightforward but this took a lot of time because I had to dig through the internals because Tonic/Hyper don't make this easy. They don't make it easy to change the target TCP stream and they don't make it easy to do a URL call + upgrade after TCP connection. I also had to setup proxies to test with and verify this behavior matches existing Go and Java gRPC behavior. But yes, luckily the amount of code needed is small.

(waiting on external CI fix before merging)

PR #721, opened 2024-04-18
Worker client replacement

### LINKED ISSUE #477: [Feature Request] Dynamic client for workers

cretz:
## What was changed

Added `Worker::replace_client`. Notes:

* Replacement unregisters itself with old client and re-registers with new client for eager workflow start
* Uses lock of arc, so no currently operating poll on existing client will be interrupted (this has been tested)
* This is as close as we can reasonably get to dynamic client endpoints without working through Tonic's connectivity
  * This is actually much cleaner for SDKs since they only really care about this for workers and can't interrupt anything running anyways
* Had to change a few internal things but nothing drastic
* Wrote test that starts two servers and swaps between (had to lower polling timeout on them)

## Checklist

1. Closes #477

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -296,7 +296,7 @@ pub struct ConfiguredClient<C> {
     options: Arc<ClientOptions>,
     headers: Arc<RwLock<ClientHeaders>>,
     /// Capabilities as read from the `get_system_info` RPC call made on client connection
-    capabilities: Option<get_system_info_response::Capabilities>,
+    capabilities: Arc<Option<get_system_info_response::Capabilities>>,
     workers: Arc<SlotManager>,
 }
 
@@ -318,8 +318,8 @@ impl<C> ConfiguredClient<C> {
 
     /// Returns the server capabilities we (may have) learned about when establishing an initial
     /// connection
-    pub fn capabilities(&self) -> Option<&get_system_info_response::Capabilities> {
-        self.capabilities.as_ref()
+    pub fn capabilities(&self) -> Arc<Option<get_system_info_response::Capabilities>> {
+        self.capabilities.clone()
     }
 
     /// Returns a cloned reference to a registry with workers using this client instance
@@ -435,7 +435,7 @@ impl ClientOptions {
             headers,
             client: TemporalServiceClient::new(svc),
             options: Arc::new(self.clone()),
-            capabilities: None,
+            capabilities: Arc::new(None),
             workers: Arc::new(SlotManager::new()),
         };
         match client
@@ -443,7 +443,7 @@ impl ClientOptions {
             .await
         {
             Ok(sysinfo) => {
-                client.capabilities = sysinfo.into_inner().capabilities;
+                client.capabilities = Arc::new(sysinfo.into_inner().capabilities);
             }
             Err(status) => match status.code() {
                 Code::Unimplemented => {}
--- a/core/src/worker/client.rs
+++ b/core/src/worker/client.rs
@@ -2,6 +2,7 @@
 
 pub(crate) mod mocks;
 use std::sync::Arc;
+use std::sync::RwLock;
 use temporal_client::{Client, RetryClient, SlotManager, WorkflowService};
 use temporal_sdk_core_protos::{
     coresdk::workflow_commands::QueryResult,
@@ -26,7 +27,7 @@ type Result<T, E = tonic::Status> = std::result::Result<T, E>;
 
 /// Contains everything a worker needs to interact with the server
 pub(crate) struct WorkerClientBag {
-    client: RetryClient<Client>,
+    replaceable_client: RwLock<RetryClient<Client>>,
     namespace: String,
     identity: String,
     worker_build_id: String,
@@ -42,7 +43,7 @@ impl WorkerClientBag {
         use_versioning: bool,
     ) -> Self {
         Self {
-            client,
+            replaceable_client: RwLock::new(client),
             namespace,
             identity,
             worker_build_id,
@@ -50,8 +51,15 @@ impl WorkerClientBag {
         }
     }
 
+    fn cloned_client(&self) -> RetryClient<Client> {
+        self.replaceable_client
+            .read()
+            .expect("failed client read lock")
+            .clone()
+    }
+
     fn default_capabilities(&self) -> Capabilities {
-        self.capabilities().cloned().unwrap_or_default()
+        self.capabilities().as_ref().clone().unwrap_or_default()
     }
 
     fn binary_checksum(&self) -> String {
@@ -142,8 +150,8 @@ pub(crate) trait WorkerClient: Sync + Send {
         query_result: QueryResult,
     ) -> Result<RespondQueryTaskCompletedResponse>;
 
-    #[allow(clippy::needless_lifetimes)] // Clippy is wrong here
-    fn capabilities<'a>(&'a self) -> Option<&'a get_system_info_response::Capabilities>;
+    fn replace_client(&self, new_client: RetryClient<Client>);
+    fn capabilities(&self) -> Arc<Option<get_system_info_response::Capabilities>>;
     fn workers(&self) -> Arc<SlotManager>;
     fn is_mock(&self) -> bool;
 }
@@ -163,8 +171,7 @@ impl WorkerClient for WorkerClientBag {
         };
 
         Ok(self
-            .client
-            .clone()
+            .cloned_client()
             .poll_workflow_task_queue(request)
             .await?
             .into_inner())
@@ -190,8 +197,7 @@ impl WorkerClient for WorkerClientBag {
         };
 
         Ok(self
-            .client
-            .clone()
+            .cloned_client()
             .poll_activity_task_queue(request)
             .await?
             .into_inner())
@@ -231,8 +237,7 @@ impl WorkerClient for WorkerClientBag {
             metering_metadata: Some(request.metering_metadata),
         };
         Ok(self
-            .client
-            .clone()
+            .cloned_client()
             .respond_workflow_task_completed(request)
             .await?
             .into_inner())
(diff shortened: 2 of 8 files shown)


core/src/worker/client.rs
```diff
@@ -142,8 +150,8 @@ pub(crate) trait WorkerClient: Sync + Send {
         query_result: QueryResult,
     ) -> Result<RespondQueryTaskCompletedResponse>;
 
-    #[allow(clippy::needless_lifetimes)] // Clippy is wrong here
-    fn capabilities<'a>(&'a self) -> Option<&'a get_system_info_response::Capabilities>;
+    fn replace_client(&self, new_client: RetryClient<Client>);
+    fn capabilities(&self) -> Arc<Option<get_system_info_response::Capabilities>>;
```
Sushisource, line 154: This should stay the same. It's generally nicer to return a ref to callers since they can do the clone if they need to, and if they don't the clone is avoided. Esp after getting rid of the Arc

core/src/worker/mod.rs
```diff
@@ -42,7 +42,7 @@ use std::{
     future,
     sync::{
         atomic::{AtomicBool, Ordering},
-        Arc,
+        Arc, Mutex,
```
Sushisource, line 45: Same, prefer the parking lot mutex

core/src/worker/mod.rs
```diff
@@ -215,6 +220,25 @@ impl Worker {
         )
     }
 
+    /// Replace client and return a new client. For eager workflow purposes, this new client will
+    /// now apply and the older client will not.
```
Sushisource, line 224:
```suggestion
    /// now apply to future eager start requests and the older client will not.
```

core/src/worker/mod.rs
```diff
@@ -367,22 +391,23 @@ impl Worker {
...
-        let provider = SlotProvider::new(
+        let slot_provider = SlotProvider::new(
             config.namespace.clone(),
             config.task_queue.clone(),
             wft_semaphore.clone(),
-            external_wft_tx,
+            Arc::new(external_wft_tx),
         );
-        let worker_key = client.workers().register(Box::new(provider));
+        let worker_key = Mutex::new(client.workers().register(Box::new(slot_provider.clone())));
```
Sushisource, line 400: The provider shouldn't need to be cloned?

core/src/worker/slot_provider.rs
```diff
@@ -42,21 +42,21 @@ impl SlotTrait for Slot {
     }
 }
 
-#[derive(derive_more::DebugCustom)]
+#[derive(derive_more::DebugCustom, Clone)]
```
Sushisource, line 45: Not sure why this guy actually needs to be clonable now?

tests/integ_tests/polling_tests.rs
```diff
@@ -88,3 +98,125 @@ async fn out_of_order_completion_doesnt_hang() {
...
+    let client2 = ClientOptionsBuilder::default()
+        .identity("integ_tester".to_owned())
+        .target_url(Url::parse(&format!("http://{}", server2.target)).unwrap())
+        .client_name("temporal-core".to_owned())
+        .client_version("0.1.0".to_owned())
+        .build()
+        .unwrap()
+        .connect("default", None)
+        .await
+        .unwrap();
```
Sushisource, line 139: Can save a lot of lines here by doing the common builder setup, cloning, then adding the specific target.

tests/integ_tests/polling_tests.rs
```diff
@@ -88,3 +98,125 @@ async fn out_of_order_completion_doesnt_hang() {
...
+    info!("Doing initial poll");
+    let act1 = worker.poll_workflow_activation().await.unwrap();
+    assert_eq!(wf1.run_id, act1.run_id);
+    worker
+        .complete_workflow_activation(WorkflowActivationCompletion::from_cmds(
+            act1.run_id,
+            vec![CompleteWorkflowExecution { result: None }.into()],
+        ))
+        .await
+        .unwrap();
```
Sushisource, line 191: I have a shortcut for completing the execution you can use (`complete_execution`), or even just `from_cmd` rather than `from_cmds`

core/src/worker/client.rs
cretz: Will try, I forget the issues I had here keeping the ref. But my whole goal with the Arc was to avoid clone/copy. But in the other comment it sounds like we want clone/copy.

core/src/worker/mod.rs
cretz: Currently it is a non-cloneable struct that is moved into this "register" call. But I need to call this "register" call with it again so now I need to have multiple references to this slot provider, so I had to make it cloneable and store it on the worker. I can wrap it in an Arc and update everywhere its used or I can clone it upon use here (or do more advanced stuff w/ references and lifetimes).

client/src/lib.rs
```diff
@@ -296,7 +296,7 @@ pub struct ConfiguredClient<C> {
     options: Arc<ClientOptions>,
     headers: Arc<RwLock<ClientHeaders>>,
     /// Capabilities as read from the `get_system_info` RPC call made on client connection
-    capabilities: Option<get_system_info_response::Capabilities>,
+    capabilities: Arc<Option<get_system_info_response::Capabilities>>,
```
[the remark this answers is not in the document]
cretz: I remember the problem now. The issue is that `WorkerClient::capabilities()` wants a reference, but that is now on a replaceable client. Should I clone every single `WorkflowClient::capabilites()` call?

core/src/worker/client.rs
```diff
@@ -26,7 +27,7 @@ type Result<T, E = tonic::Status> = std::result::Result<T, E>;
 
 /// Contains everything a worker needs to interact with the server
 pub(crate) struct WorkerClientBag {
-    client: RetryClient<Client>,
+    replaceable_client: RwLock<RetryClient<Client>>,
```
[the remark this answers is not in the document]
cretz: Fixed

core/src/worker/mod.rs
Sushisource: Ah, yeah, I see since it is getting put into `Self` below, I missed that initially.

I think the alternative here would be to not instantiate the worker key here any longer, since you can create it from within self now that it owns the provider. It probably ends up being the same issue though.

Imo it probably does make sense to wrap the provider in an arc rather than the channel inside it.

core/src/worker/client.rs
cretz: Commented above, this was to avoid `WorkerClient::capabilities()` call from having to copy the capabilities on every invocation (can't be a reference anymore since client is replaced and lifetime no longer tied to self).

core/src/worker/mod.rs
cretz: Fixed

core/src/worker/mod.rs
cretz: Fixed

tests/integ_tests/polling_tests.rs
cretz: Fixed

core/src/worker/mod.rs
cretz: I was hoping to get away with as little of changes to the existing worker registration/eager code as I could, but I fear we're basically saying I may need to do a bit more to it.

client/src/lib.rs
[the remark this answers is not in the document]
cretz: :+1: Will make this call return a clone of the proto on each invocation and change back everywhere else to continue doing the same.

core/src/worker/mod.rs
Sushisource: It's not that much change, and it's nice to maintain some good stewardship where then structure makes sense rather than a minimum possible change, not like this is in a huge rush or anything.

cretz: :+1: Done. You're right that it wasn't much change to switch from `Box` to `Arc`: https://github.com/temporalio/sdk-core/pull/721/commits/dfe58a56751267b8dab9b4fd28abea4622e26b29

core/src/worker/slot_provider.rs
cretz: Replaced cloning with Arc

core/src/worker/client.rs
cretz: Changed this to clone on every call now

core/src/worker/mod.rs
cretz: Switched from Box to Arc, you were correct, not a big change (https://github.com/temporalio/sdk-core/commit/417b70959bc872aa8c5f451002b1057d5a32026a).

Sushisource (approved):

core/src/worker/mod.rs
```diff
@@ -78,8 +79,9 @@ use {
 pub struct Worker {
     config: WorkerConfig,
     wf_client: Arc<dyn WorkerClient>,
+    slot_provider: Arc<SlotProvider>,
     /// Registration key to enable eager workflow start for this worker
```
antlai-temporal, line 83: @cretz Sorry for being so late in the game (already merged...) I wonder if holding a reference to slot_provider will hang the worker shutdown in some cases. slot_provider contains a reference to `external_wft_tx` and I think we need to drop all refs to it, so that the rx part closes, and that unblocks the worker shutdown...

cretz: Wouldn't this have failed tests that rely on worker shutdown to complete (or are there none)? Is the `SlotProvider` drop needed in time for `shutdown` or `finalize_shutdown`. The latter consumes this worker and therefore this slot provider should be dropped.

Can we write a test for this? Do you have alternative suggestions on how to handle this?

cretz: Ok, I have opened #725. I think that is a cleaner approach anyways and I wish I thought of it before.

PR #725, opened 2024-04-23
Do not hold eager workflow slot provider on worker

cretz:
## What was changed

Based on comment at https://github.com/temporalio/sdk-core/pull/721#discussion_r1575409008, I have basically changed the approach that PR took where it stored slot provider on the worker, and just put it back the way it was, but also returned the slot provider from the `unregister` call so it could be used again for the next `register` call.

### DIFF
--- a/core/src/worker/mod.rs
+++ b/core/src/worker/mod.rs
@@ -79,7 +79,6 @@ use {
 pub struct Worker {
     config: WorkerConfig,
     wf_client: Arc<dyn WorkerClient>,
-    slot_provider: Arc<SlotProvider>,
     /// Registration key to enable eager workflow start for this worker
     worker_key: Mutex<Option<WorkerKey>>,
     /// Manages all workflows and WFT processing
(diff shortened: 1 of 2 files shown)


Sushisource (approved):
Nice, better for sure.

core/src/worker/mod.rs
```diff
@@ -222,15 +221,14 @@ impl Worker {
     pub fn replace_client(&self, new_client: ConfiguredClient<TemporalServiceClientWithMetrics>) {
         // Unregister worker from current client, register in new client at the end
         let mut worker_key = self.worker_key.lock();
+        let slot_provider = (*worker_key).and_then(|k| self.wf_client.workers().unregister(k));
         if let Some(key) = *worker_key {
             self.wf_client.workers().unregister(key);
         }
```
antlai-temporal, line 227: Do you need this? Calling unregister twice may not do harm but...

antlai-temporal (commented):
This is a great, very clever approach to get rid of Arc.

Just one comment,  you may want to delete the old `unregister()` call

core/src/worker/mod.rs
cretz: Ah thanks! Gonna have to be a new PR sadly.

cretz: #727 opened

PR #728, opened 2024-04-24
Normalize workspace

djc:
## What was changed

Move the integration tests around so that `cargo c --all-targets --all-features` works, and integration test compilation failures become visible in Rust-Analyzer-based IDEs.

## Why?

I again got frustrated after finding compilation failures via `cargo lint` and `cargo test-lint` in an upcoming PR. Move the integration test code for the core crate into more conventional places so that Cargo / Rust-Analyzer can find all the code, while minimizing interface changes to running it.

## Checklist

1. Closes: no issue
2. How was this tested: ran `cargo lint` and `cargo test-lint` and exercised the `integ_runner`


Sushisource (commented):
Hmm, I'd prefer the integ tests to not have ignored tags by default. It's potentially easy to forget when adding a new one, and makes for slightly confusing output with a bunch of ignored tests when running normally.

If there's a different way we can get the visibility to rust-analyzer / linting both with one command without ignoring, that'd be great. Maybe a feature flag? I think I had tried that in the past without success though.

djc:
Would the feature flag be harder to forget when adding a new one? Could make the slightly confusing output better by employing better module/file naming? IMO this use case is really what `#[ignore]` was intended for -- it seems like a pretty good fit to me.

Sushisource:
Sorry for the late replay - had some other urgent things and was on vacation for a while.

So I think the feature flag forgetting problem can probably be handled by just having the flag go around whole modules at once? Then you don't need to think about it most of the time, only when adding whole new modules. So seems like that'd work fairly nicely to me

> using --ignored instead of --include-ignored in the integ_runner

AFAIK the only place I've used `--include-ignored` is when running normal tests to run the one weird state machine transition coverage test.

dandavison:
FWIW I can confirm this improves the experience of using VSCode for `tests/integ_tests` -- e.g. rust-analyzer working and "Run/debug tests" code lenses active.

dandavison (commented):
@Sushisource what do you think about just letting `cargo test` run all tests, add adding a command to run unit tests only?

Sushisource:
> @Sushisource what do you think about just letting `cargo test` run all tests, add adding a command to run unit tests only?

Yeah, we can do that. Bonus points if the command has an option for running with https://nexte.st/ (would be nice to add that to the integ test runner too), but not a requirement.

dandavison:
@djc This PR is a really valuable improvement. (I also work on `sdk-core` using rust-analyzer.) Sorry we've let it pick up merge conflicts. Do you have any appetite for getting these changes mergable again?

djc:
@dandavison happy to rebase this if there's a path to getting it merged. But,

> add adding a command to run unit tests only?

I'm not sure how you're envisioning this should work.

dandavison:
OK so to recap our requirements, I'm thinking they are:

- Be able to run unit tests, and integration tests, separately
- Not have to remember to add annotations
- Be `rust-analyzer` compatible
- Not do anything that makes nextest adoption harder

So what I'm thinking is
- move code as you did to be `rust-analyzer` compatible
- Have a command `cargo unit-test` that runs all tests except integration tests, and `cargo integ-test` that runs only integration tests (perhaps via filepath-based logic?).
- `cargo test` would run all tests.
- And I think we'd want `cargo heavy-tests` also.

djc:
I'm leaving the company where I was working with Temporal, so I probably won't be pushing this forward anymore. Feel free to reuse any of the code here for future changes.

dandavison:
OK, thanks for letting us know.

PR #732, opened 2024-04-26
Remove temporalite ephemeral server and expose child process ID

### LINKED ISSUE #731: [Feature Request] Expose ephemeral server PID and confirm shutdown destroys process

Need to see ephemeral server PID for lang to do some validations against it (and users may want to know about it too). Also need to confirm the process is gone in CI tests. See https://github.com/temporalio/features/issues/470.

cretz:
## What was changed

* Remove Temporalite as an ephemeral server option
* Expose child process ID for ephemeral servers
* Add assertion for CLI dev server that process is gone after shutdown

Done as part of https://github.com/temporalio/features/issues/470

## Checklist

1. Closes #731

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -93,6 +93,7 @@ bimap = "0.6.1"
 clap = { version = "4.0", features = ["derive"] }
 criterion = "0.5"
 rstest = "0.18"
+sysinfo = "0.30"
 temporal-sdk-core-test-utils = { path = "../test-utils" }
 temporal-sdk = { path = "../sdk" }
 
--- a/core/src/ephemeral_server/mod.rs
+++ b/core/src/ephemeral_server/mod.rs
@@ -22,99 +22,6 @@ use zip::read::read_zipfile_from_stream;
 use std::os::unix::fs::OpenOptionsExt;
 use std::process::Stdio;
 
-/// Configuration for Temporalite.
-/// Will be removed eventually as its successor, Temporal CLI matures.
-/// We don't care for the duplication between this struct and [TemporalDevServerConfig] and prefer that over another
-/// abstraction since the existence of this struct is temporary.
-#[derive(Debug, Clone, derive_builder::Builder)]
-pub struct TemporaliteConfig {
-    /// Required path to executable or download info.
-    pub exe: EphemeralExe,
-    /// Namespace to use.
-    #[builder(default = "\"default\".to_owned()")]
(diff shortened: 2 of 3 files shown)


core/src/ephemeral_server/mod.rs
```diff
@@ -326,6 +233,12 @@ impl EphemeralServer {
             Ok(())
         }
     }
+
+    /// Get the process ID of the child. This will be None if the process is
+    /// considered to be complete.
```
cretz, line 238: > considered to be complete

Note, this is vague on purpose, ref https://docs.rs/tokio/latest/tokio/process/index.html#unix-processes

tests/integ_tests/ephemeral_server_tests.rs
```diff
@@ -30,45 +35,6 @@ async fn temporal_cli_fixed() {
...
-        .exe(fixed_cached_download("v0.2.0"))
-        .build()
-        .unwrap();
-    let mut server = config.start_server().await.unwrap();
-    assert_ephemeral_server(&server).await;
-    server.shutdown().await.unwrap();
-}
-
-#[tokio::test]
-async fn temporalite_shutdown_port_reuse() {
```
mjameswh, line 56: Looks like we don't have an this specific test for cli. Would it be pertinent to repurpose it rather than delete it?

mjameswh (approved):
Left one comment, but LGTM overall.

core/Cargo.toml
```diff
@@ -93,6 +93,7 @@ bimap = "0.6.1"
 clap = { version = "4.0", features = ["derive"] }
 criterion = "0.5"
 rstest = "0.18"
+sysinfo = "0.30"
```
Sushisource, line 96: Heh, I need this crate for the resource slot controller anyway, so it'll get promoted to non-dev dep

Sushisource (approved):

tests/integ_tests/ephemeral_server_tests.rs
cretz: We never have had it, so for CLI we've never confirmed port reuse, only Temporalite. I was hoping not to have to address new requirements (e.g. confirming CLI allows port reuse post shutdown) as part of this smaller issue, but alas I suppose I can.

cretz: Fixed

PR #739, opened 2024-05-14
Fix request failure metric & add error code labels

Sushisource:
## What was changed
The request failure metrics weren't actually being incremented when they should've been due to the interceptor only caring about http errors and not inspecting grpc statuses. That is fixed, and error code labels are attached w/ an option to disable.

## Why?
Part of https://github.com/temporalio/features/issues/144

## Checklist

1. Closes 

2. How was this tested:
Added IT

3. Any docs updates needed?

### DIFF
--- a/client/src/metrics.rs
+++ b/client/src/metrics.rs
@@ -9,7 +9,7 @@ use temporal_sdk_core_api::telemetry::metrics::{
     CoreMeter, Counter, HistogramDuration, MetricAttributes, MetricKeyValue, MetricParameters,
     TemporalMeter,
 };
-use tonic::{body::BoxBody, transport::Channel};
+use tonic::{body::BoxBody, transport::Channel, Code};
 use tower::Service;
 
 /// Used to track context associated with metrics, and record/update them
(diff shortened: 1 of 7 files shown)


client/src/metrics.rs
```diff
@@ -126,12 +136,40 @@ pub(crate) fn svc_operation(op: String) -> MetricKeyValue {
     MetricKeyValue::new(KEY_SVC_METHOD, op)
 }
 
+pub(crate) fn status_code_kv(code: Code) -> MetricKeyValue {
+    MetricKeyValue::new(KEY_STATUS_CODE, code_as_screaming_snake(&code))
+}
+
+/// This is done to match the way Java sdk labels these codes (and also matches gRPC spec)
+fn code_as_screaming_snake(code: &Code) -> &'static str {
```
cretz, line 144: Pretty annoying https://docs.rs/tonic/latest/tonic/enum.Code.html doesn't give you the raw enum from https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto so that you can use the proper enum name

client/src/metrics.rs
```diff
@@ -113,6 +122,7 @@ impl MetricsContext {
 const KEY_NAMESPACE: &str = "namespace";
 const KEY_SVC_METHOD: &str = "operation";
 const KEY_TASK_QUEUE: &str = "task_queue";
+const KEY_STATUS_CODE: &str = "status_code";
```
cretz, line 125: Can you update the original features issue with details on the exact key name we're using, that we're using the proper screaming case enum from https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto, and which metrics we're adding it on? This can help other SDKs make sure they conform.

client/src/metrics.rs
```diff
@@ -167,13 +205,29 @@ impl Service<http::Request<BoxBody>> for GrpcMetricSvc {
...
             if let Some(metrics) = metrics {
                 metrics.record_svc_req_latency(started.elapsed());
-                if res.is_err() {
-                    metrics.svc_request_failed();
+                if let Ok(ref ok_res) = res {
+                    if let Some(number) = ok_res
+                        .headers()
+                        .get("grpc-status")
+                        .and_then(|s| s.to_str().ok())
+                        .and_then(|s| s.parse::<i32>().ok())
```
cretz, line 219: Ug, I wonder what the Tonic recommended way is to access the full gRPC status in an interceptor-like pattern or if it's just not supported.

cretz (approved):

client/src/metrics.rs
Sushisource: It's not really, at least not directly in any simple way.

client/src/metrics.rs
Sushisource: Sure, this is all exactly what Java is doing now

Issue #716, opened 2024-04-12
[Bug] Legacy queries may get lost


cretz:
### Describe the bug

There is a rare-but-acknowledged server case where legacy query is delivered for a run ID to a worker but before it is answered a new task can be delivered. After lots of debugging, it appears https://github.com/temporalio/sdk-core/blob/409e74ec8e80ae4c1f9043e8b413b1371b65f946/core/src/worker/workflow/workflow_stream.rs#L468 overwrites it.

### Minimal Reproduction

So this is kinda hard to replicate. I have seen the following Python script replicate it after running for a short bit:

```python
import asyncio
import logging
from datetime import timedelta
from uuid import uuid4

from temporalio import activity, workflow
from temporalio.client import Client
from temporalio.worker import Worker


@activity.defn
async def wait_forever_activity() -> None:
    await asyncio.Future()


@workflow.defn
class WaitForeverWorkflow:
    @workflow.run
    async def run(self) -> None:
        await asyncio.Future()


@workflow.defn
class WeirdWorkflow:
    def __init__(self) -> None:
        self._signal_count = 0

    @workflow.run
    async def run(self) -> None:
        # Start several things in background
        tasks = [
            asyncio.create_task(
                workflow.execute_activity(
                    wait_forever_activity, start_to_close_timeout=timedelta(hours=1)
                )
            ),
            asyncio.create_task(
                workflow.execute_child_workflow(WaitForeverWorkflow.run)
            ),
            asyncio.create_task(asyncio.sleep(1000)),
            asyncio.shield(
                workflow.execute_activity(
                    wait_forever_activity, start_to_close_timeout=timedelta(hours=1)
                )
            ),
            asyncio.create_task(workflow.wait_condition(lambda: False)),
        ]
        asyncio.gather(*tasks, return_exceptions=True)
        try:
            # Wait for signal count to reach 2
            await workflow.wait_condition(lambda: self._signal_count > 1)
        finally:
            # Wait for signal count to reach 3
            await workflow.wait_condition(lambda: self._signal_count > 2)

    @workflow.signal
    async def signal(self) -> None:
        self._signal_count += 1

    @workflow.query
    def signal_count(self) -> int:
        return self._signal_count


task_queue = f"tq-{uuid4()}"


async def run_workflow(client: Client) -> None:
    id = f"wf-{uuid4()}"
    print("Running workflow", id)
    handle = await client.start_workflow(
        WeirdWorkflow.run,
        id=id,
[...]

PR #740, opened 2024-05-15
Allow buffering more than one WFT per run when at cache limit

Sushisource:
## What was changed
Allows the buffer for WFTs that are waiting for a cache slot to hold more than one task per run id

## Why?
Solves situation where multiple tasks can arrive containing legacy queries

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/716

2. How was this tested:
Expanded unit test

3. Any docs updates needed?

### DIFF
--- a/core/src/core_tests/queries.rs
+++ b/core/src/core_tests/queries.rs
@@ -893,8 +893,9 @@ async fn build_id_set_properly_on_query_on_first_task() {
     core.drain_pollers_and_shutdown().await;
 }
 
+#[rstest::rstest]
 #[tokio::test]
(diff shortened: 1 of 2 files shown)


core/src/core_tests/queries.rs
cretz, line 1: Can you confirm this test failed before your fix?

cretz (approved):
I admit to not understanding the details here but if tests pass, works for me

core/src/core_tests/queries.rs
Sushisource: Yeah, it did

PR #742, opened 2024-05-17
get-free-port: prevent duplicate ports on Linux

mjameswh:
## What changed

- When assigning ephemeral ports for use by Ephemeral Servers, on all platforms except macOS and Windows, immediately open a connection to that port, and force the connection into `TIME_WAIT` state. This avoids various race conditions that could happen due to the port being reallocated between the moment the port is obtained and the moment the server actually binds to that port.

- Added a test to confirm that it is possible to start multiple dev server concurrently; however, that test is currently skipped, as it will fail on Linux until temporalio/cli#564 gets released (presumably in 0.12.1). That can still be used locally by building CLI manually.

### DIFF
--- a/core/src/ephemeral_server/mod.rs
+++ b/core/src/ephemeral_server/mod.rs
@@ -372,14 +372,54 @@ impl EphemeralExe {
     }
 }
 
+/// Returns a TCP port that is available to listen on for the given local host.
+///
+/// This works by binding a new TCP socket on port 0, which requests the OS to
+/// allocate a free port. There is no strict guarantee that the port will remain
+/// available after this function returns, but it should be safe to assume that
+/// a given port will not be allocated again to any process on this machine
+/// within a few seconds.
+///
+/// On Unix-based systems, binding to the port returned by this function
+/// requires setting the `SO_REUSEADDR` socket option (Rust already does that by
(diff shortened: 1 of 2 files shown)


cretz (approved):

core/src/ephemeral_server/mod.rs
```diff
@@ -372,14 +372,54 @@ impl EphemeralExe {
...
+#[allow(clippy::let_and_return)]
 fn get_free_port(bind_ip: &str) -> u16 {
-    // Can just ask OS to give us a port then close socket. OS's don't give that
-    // port back to anyone else anytime soon.
-    std::net::TcpListener::bind(format!("{bind_ip}:0"))
-        .unwrap()
-        .local_addr()
-        .unwrap()
-        .port()
+    let listen = std::net::TcpListener::bind((bind_ip, 0)).unwrap();
```
Sushisource, line 392: IMO no reason not to make this return a result and unwrap it in places that don't care about erroring.

Sushisource (approved):

core/src/ephemeral_server/mod.rs
mjameswh: Sure, will fix.

Issue #667, opened 2024-01-10
[Bug] Worker.finalize_shutdown seems to hang when poll never succeeded due to server permission failure


cretz:
### Describe the bug

Originally reported at https://github.com/temporalio/sdk-python/issues/459.

To replicate, first run a local Temporal server, then alter https://github.com/temporalio/samples-go/blob/main/grpc-proxy/proxy-server/main.go to change the gRPC dialing to:

```go
	grpcClient, err := grpc.Dial(
		upstreamFlag,
		grpc.WithTransportCredentials(insecure.NewCredentials()),
		// grpc.WithUnaryInterceptor(clientInterceptor),
		grpc.WithUnaryInterceptor(func(
			ctx context.Context, method string,
			req, reply any,
			cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
			if strings.HasPrefix(method, "/temporal.api.workflowservice.v1.WorkflowService/Poll") {
				return status.Error(codes.PermissionDenied, "DENIED!")
			}
			return invoker(ctx, method, req, reply, cc, opts...)
		}),
	)
```

This will simulate permission denied upstream. Then run it, e.g. `go run ./grpc-proxy/proxy-server` and it'll run a proxy at `127.0.0.1:8081` that proxies to `127.0.0.1:7233` by default. Then run the Python script given in that issue pointing to this proxy:

```python
import asyncio
from temporalio import activity, client, workflow, worker

@activity.defn
async def a() -> None:
    pass


@workflow.defn
class Workflow:
    @workflow.run
    async def run(self) -> None:
        pass


async def main():
    c = await client.Client.connect(
        "127.0.0.1:8081",
        rpc_metadata={"authorization": "wrong_token"},
        # tls=True,
    )
    w = worker.Worker(
        c,
        task_queue="default",
        activities=[a],
        workflows=[Workflow],
    )
    await w.run()


if __name__ == "__main__":
    asyncio.run(main())
```

After about a minute or so, the worker will fatally error due to too many permission denials. But the worker process gets to await finalize_shutdown and hangs. Feel free to add logs inside `.venv/Lib/site-packages/temporalio/worker/_worker.py`'s `run` call to see how far it gets.

pwalessi-dell:
[...]

PR #746, opened 2024-05-22
Fix possible shutdown hang if pollers keep getting gRPC errors

Sushisource:
## What was changed
Fixed a scenario where Core wouldn't properly shut itself down if it only gets poll error responses

## Why?
Hang on shutdown bad

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/667

2. How was this tested:
Added reproing unit test

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/workflow_stream.rs
+++ b/core/src/worker/workflow/workflow_stream.rs
@@ -165,7 +165,6 @@ impl WFStream {
                         None
                     }
                     WFStreamInput::PollerError(e) => {
-                        warn!("WFT poller errored, shutting down");
                         return Err(PollWfError::TonicError(e));
                     }
                 };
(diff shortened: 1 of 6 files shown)


core/src/worker/workflow/workflow_stream.rs
```diff
@@ -165,7 +165,6 @@ impl WFStream {
                         None
                     }
                     WFStreamInput::PollerError(e) => {
-                        warn!("WFT poller errored, shutting down");
```
cretz, line 168: Is there a consequence to removing this warning? Can/should we warn somewhere in general when a fatal worker error occurs, or is the fact that it is returned as an error from poll for lang to warn and start shutdown enough?

cretz (approved):

core/src/worker/workflow/workflow_stream.rs
Sushisource: It's actually just redundant. I removed it because there's another line that will always log for the same reason.

Issue #735, opened 2024-05-06
[Feature Request] Local activity support for next-retry-delay


cretz:
### Describe the solution you'd like

Re https://github.com/temporalio/features/issues/419, it was decided we need to support next retry delay in application failures from local activity too. Implementation details are still pending.

PR #747, opened 2024-05-22
Explicit activity retry delays for local activites

Sushisource:
## What was changed
Allow explicit configuration of next retry backoff for LAs

* Also made breaking changes to the unreleased Rust SDK to make testing this work better. Activity functions may need to call `into()` on returned errors.

## Why?
Parity with normal activities

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/735

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/tests/integ_tests/workflow_tests/local_activities.rs
+++ b/tests/integ_tests/workflow_tests/local_activities.rs
@@ -7,7 +7,7 @@ use std::{
 };
 use temporal_client::WorkflowOptions;
 use temporal_sdk::{
-    interceptors::WorkerInterceptor, ActContext, ActivityCancelledError, CancellableFuture,
+    interceptors::WorkerInterceptor, ActContext, ActivityError, CancellableFuture,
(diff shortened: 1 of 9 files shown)


tests/integ_tests/workflow_tests/local_activities.rs
cretz, line 1: Do we need a test for next retry delay?

cretz (approved):

tests/integ_tests/workflow_tests/local_activities.rs
Sushisource: There are tests? https://github.com/temporalio/sdk-core/pull/747/files#diff-6201d121c4591b6c569356a6cb108abad27ef1c6a921f53f6b86f6b8bdea656eR1431

cretz: :+1: I missed it because I was looking at something testing it came through the application error failure proto. But I think that can be added on lang side when lang side adds their part of this.

PR #748, opened 2024-05-24
Allow handing of speculative wft with command events

Sushisource:
## What was changed
Core will now be OK with speculative WFTs which also contain command events

## Why?
Allows server to make a WFT speculative in more situations

## Checklist

1. Closes 

2. How was this tested:
Added UT

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/machines/workflow_machines.rs
+++ b/core/src/worker/workflow/machines/workflow_machines.rs
@@ -474,17 +474,19 @@ impl WorkflowMachines {
     pub(crate) fn reset_last_started_id(&mut self, id: i64) {
         debug!("Resetting back to event id {} due to speculative WFT", id);
         self.current_started_event_id = id;
-        // This is pretty nasty to just + 1 like this, but, we know WFT complete always follows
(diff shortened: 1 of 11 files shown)


core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -474,17 +474,19 @@ impl WorkflowMachines {
...
+        // schedule event just before it.
+        self.last_processed_event -= 2;
         // Then, we have to drop any state machines (which should only be one workflow task machine)
         // we may have created when servicing the speculative task.
         // Remove when https://github.com/rust-lang/rust/issues/59618 is stable
         let remove_these: Vec<_> = self
             .machines_by_event_id
             .iter()
-            .filter(|(mid, _)| **mid > id)
+            .filter(|(mid, _)| **mid > self.last_processed_event)
```
Sushisource, line 489: This is the whole actual fix

cretz:
I don't understand this one enough on a casual glance, would like to defer to others more familiar here, or will come back and review if there is nobody.

Quinn-With-Two-Ns (approved):

Issue #738, opened 2024-05-10
[Feature Request] Call describe namespace on worker start


cretz:
### Describe the solution you'd like

See https://github.com/temporalio/features/issues/477

PR #750, opened 2024-05-24
Ensure namespace exists on worker startup

Sushisource:
## What was changed
Added a new worker validation function that lang SDKs should call before starting polling

## Why?
Eager validation that namespace exists, future room for other validations

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/738

2. How was this tested:
Integ test

3. Any docs updates needed?

### DIFF
--- a/core-api/src/errors.rs
+++ b/core-api/src/errors.rs
@@ -2,6 +2,17 @@
 
 use temporal_sdk_core_protos::coresdk::activity_result::ActivityExecutionResult;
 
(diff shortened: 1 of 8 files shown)


cretz (approved):
I have opened https://github.com/temporalio/sdk-typescript/issues/1430, https://github.com/temporalio/sdk-python/issues/532, and https://github.com/temporalio/sdk-dotnet/issues/253 to ensure SDKs call this

PR #757, opened 2024-06-07
Update dependencies

djc:
Just noticed these. There are some outdated hyper/http dependencies in the client crate, but those interact with tonic which is still on 0.11 (see https://github.com/hyperium/tonic/pull/1670).


Sushisource (approved):
Thanks!

Issue #704, opened 2024-03-13
[Feature Request]  poll_workflow_task_queue retried X times needs to be an ERROR log, not WARN


danthegoodman1:
```
2024-03-13T14:27:20.049568Z  WARN temporal_client::retry: gRPC call poll_workflow_task_queue retried 16 times error=Status { code: Cancelled,
message: "Timeout expired", source: Some(tonic::transport::Error(Transport, TimeoutExpired(()))) }
```

This log absolutely needs to be an `ERROR`. Maybe after some threshold, but I've seen it get to >100 when the client was having connectivity issues to the client.

I understand metrics might solve this, but client metrics should not be required to run temporal clients reliably.

If a client is unable to get work, after a **very long time** of trying and failing, that should be considered an error.

jhecking:
We have recently observed a case where a client had retired nearly 100,000 times and still no error was reported:

    temporal_client::retry: gRPC call poll_activity_task_queue retried 94372 times

Even querying the worker's state didn't indicate any problem, so the service's heartbeat (which checks worker state) also didn't fail:

```
{"runState":"RUNNING","numHeartbeatingActivities":0,"workflowPollerState":"SHUTDOWN","activityPollerState":"POLLING","hasOutstandingWorkflowPoll":false,"hasOutstandingActivityPoll":true,"numCachedWorkflows":0,"numInFlightWorkflowActivations":0,"numInFlightActivities":0}
```

PR #758, opened 2024-06-07
Make grpc retries eventually turn into error messages

Sushisource:
## What was changed
For calls that can retry forever (polling), eventually start logging errors if retrying has been going on for a while

## Why?
More obvious for users who have alerting set up on error lines

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/704

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/client/src/retry.rs
+++ b/client/src/retry.rs
@@ -146,16 +146,26 @@ where
             throttle_backoff: throttle_cfg.into_exp_backoff(throttle_clock),
         }
     }
-    const fn should_log_retry_warning(&self, cur_attempt: usize) -> bool {
+
+    fn maybe_log_retry(&self, cur_attempt: usize, err: &tonic::Status) {
+        let mut do_log = false;
         // Warn on more than 5 retries for unlimited retrying
         if self.max_retries == 0 && cur_attempt > 5 {
-            return true;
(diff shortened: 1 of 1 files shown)


client/src/retry.rs
```diff
@@ -146,16 +146,26 @@ where
...
-            return true;
+            do_log = true;
         }
         // Warn if the attempts are more than 50% of max retries
         if self.max_retries > 0 && cur_attempt * 2 >= self.max_retries {
-            return true;
+            do_log = true;
+        }
+
+        if do_log {
```
mjameswh, line 161: At what speed can this repeat? Should we be worried about filling disk?

mjameswh (approved):
LGTM. Left an open question for a potential risk, but that's far fetch. Feel free to dismiss if you don't think that would be a problem.

client/src/retry.rs
Sushisource: It'll only go forever for long polling, which will hit the max retry interval of 10s quickly.

For others, they might retry faster but will eventually hit max retries.

In principle someone could specify a custom retry policy that just goes ham but, then they'd be absolutely blowing up their server too so I'm not too stressed.

PR #759, opened 2024-06-07
Fix issue with lookahead not properly seeing flags in multi-wft sequences

Sushisource:
## What was changed
WFT look-ahead was not properly handling the fact that in a "logical WFT" which spans multiple individual WFTS we want to merge flags from more than just the immediate-next WFT. Fix that.

## Why?
Bug

## Checklist

1. Closes 

2. How was this tested:
Added UT

3. Any docs updates needed?

### DIFF
--- a/core/src/core_tests/workflow_tasks.rs
+++ b/core/src/core_tests/workflow_tasks.rs
@@ -2900,7 +2900,7 @@ async fn sets_build_id_from_wft_complete() {
     });
     let timer_started_event_id = t.add_by_type(EventType::TimerStarted);
     t.add_timer_fired(timer_started_event_id, "2".to_string());
-    t.add_full_wf_task();
+    t.add_workflow_task_scheduled_and_started();
 
     let mock = mock_workflow_client();
     let mut worker = mock_sdk_cfg(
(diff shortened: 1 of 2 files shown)


core/src/core_tests/workflow_tasks.rs
```diff
@@ -2856,13 +2913,14 @@ async fn sets_build_id_from_wft_complete() {
...
         // First task, it should be empty, since replaying and nothing in first WFT completed
-        ctx.timer(Duration::from_secs(1)).await;
         assert_eq!(ctx.current_build_id(), None);
         ctx.timer(Duration::from_secs(1)).await;
         assert_eq!(ctx.current_build_id(), Some("enchi-cat".to_string()));
         ctx.timer(Duration::from_secs(1)).await;
         // Not replaying at this point, so we should see the worker's build id
         assert_eq!(ctx.current_build_id(), Some("fierce-predator".to_string()));
+        ctx.timer(Duration::from_secs(1)).await;
+        assert_eq!(ctx.current_build_id(), Some("fierce-predator".to_string()));
```
Sushisource, line 2923: This test was just actually wrong before. All the assertions should've been before rather than after the timer.

cretz (approved):
Approving with the understanding that this does not change any activation ordering/count or job ordering/count, it just affects lookahead to populate SDK flags.

PR #760, opened 2024-06-10
sdk: create spans for RunWorkflow and RunActivity

valkum:
## What was changed
This PR expands OpenTelemetry support in the Rust SDK. The workflow and activity futures get instrumented using tracing (similar to other places in the code).
To follow other SDKs such as the python one, the spans use "otel.name" (special keyword from `tracing-opentelemtry` to create a span name with the type name of the workflow/activity.

## Why?
Currently, these futures get spawned and thus don't have a root span. To get a root span boilerplate code is required in all workflow and activity functions. To have less boilerplate and to follow the naming conventions more strictly first party support is desired.
I plan on expanding this with support for propagation.

## Discussion
This currently relies on internals of the `tracing-opentelemetry` crate (`otel.name` and `otel.kind` resource names). Also, this is currently not configurable. Other crates in the workspace already use the `tracing` crate and/or the OpenTelemetry crates. I think this is a far better state than relying on copied boilerplate especially when adding propagation support.

## Checklist

1. Closes n/a

2. How was this tested:
This was tested using a local Jaeger instance. The generated span based on one of our workflows looks like this in Jaeger UI.
image
 

3. Any docs updates needed?
 Unsure if needed due to the non-official nature of the Rust SDK.

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -53,6 +53,7 @@ mod workflow_future;
 
 pub use activity_context::ActContext;
 pub use temporal_client::Namespace;
+use tracing::{field, Instrument, Span};
 pub use workflow_context::{
     ActivityOptions, CancellableFuture, ChildWorkflow, ChildWorkflowOptions, LocalActivityOptions,
     PendingChildWorkflow, Signal, SignalData, SignalWorkflowOptions, StartedChildWorkflow,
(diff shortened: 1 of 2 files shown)


Sushisource (approved):
Thanks for the contribution! And double thanks for taking the time to make the attributes & names match the other SDKs.

PR #766, opened 2024-06-18
Cloud ops API protos/client

### LINKED ISSUE #737: [Feature Request] Cloud Operations API Client

See https://github.com/temporalio/features/issues/440 and https://github.com/temporalio/sdk-go/pull/1426. We need a separate cloud client using the cloud protos (they are separate today). The client should be marked experimental.

cretz:
## What was changed

* Created new subtree at `sdk-core-protos/protos/api_cloud_upstream` sourced from https://github.com/temporalio/api-cloud
  * Note, we expect this to be in the API repo someday
* Added necessary client calls and confirmed it can be used in Python

Note for core developers using this, some calls will have to be qualified between workflow/operator service and cloud service.

## Checklist

1. Closes #737


Sushisource (approved):
Love to see these macros doin' so much work

PR #768, opened 2024-06-25
Use Resource::default in order to respect common OTEL env vars

invokermain:
## What was changed
Changed the `otel::default_resource` function use the Resource's default trait instead of providing a single custom default attributes, which is just a fixed service name. 

## Why?
This makes the OTLP Exporter respect common env vars such as `OTEL_RESOURCE_ATTRIBUTES`, [see here](https://opentelemetry.io/docs/languages/sdk-configuration/general/#otel_resource_attributes).

### DIFF
--- a/core/src/telemetry/otel.rs
+++ b/core/src/telemetry/otel.rs
@@ -12,7 +12,7 @@ use crate::{abstractions::dbg_panic, telemetry::metrics::DEFAULT_S_BUCKETS};
 use opentelemetry::{
     self,
     metrics::{Meter, MeterProvider as MeterProviderT, Unit},
-    KeyValue,
+    Key, KeyValue, Value,
 };
 use opentelemetry_otlp::WithExportConfig;
 use opentelemetry_sdk::{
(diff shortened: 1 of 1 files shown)


core/src/telemetry/otel.rs
```diff
@@ -330,18 +330,11 @@ impl GaugeF64 for MemoryGauge<f64> {
...
-    static INSTANCE: OnceCell<[KeyValue; 1]> = OnceCell::new();
-    INSTANCE.get_or_init(|| [KeyValue::new("service.name", TELEM_SERVICE_NAME)])
-}
-
 fn default_resource(override_values: &HashMap<String, String>) -> Resource {
     let override_kvs = override_values
         .iter()
         .map(|(k, v)| KeyValue::new(k.clone(), v.clone()));
-    Resource::new(default_resource_kvs().iter().cloned()).merge(&Resource::new(override_kvs))
+    Resource::default().merge(&Resource::new(override_kvs))
```
cretz, line 337: I think we still want to default `service.name` to `TELEM_SERVICE_NAME` if not already set

Sushisource: Yeah, we can't eliminate that - but we can have the `default_resource_kvs` function start from using `Resource::default()` and then merge in the `service.name` value.

invokermain: I've updated the PR to not change the behaviour around unknown_service

core/src/telemetry/otel.rs
```diff
@@ -359,3 +372,26 @@ fn metric_temporality_to_selector(t: MetricTemporality) -> impl TemporalitySelec
...
 }
+
+#[cfg(test)]
+pub(crate) mod tests {
+    use super::*;
+    use opentelemetry::Key;
+    use std::env;
+
+    #[test]
+    pub(crate) fn default_resource_instance_respects_otel_env_vars() {
```
cretz, line 383: I don't know enough about Rust, do these tests run concurrently? If so, we need to turn it off during this test since this has global side effects (I would also be ok not having this test if others decide that)

invokermain: Yeah I've never used env vars in rust tests before, I'll just remove it as it's basically testing `opentelemetry-sdk` functionality anyway. Was just nice to see it pass as it was the whole point of this PR :)

Sushisource (approved):
Thank you!

PR #769, opened 2024-06-27
sdk: expose WfContext:task_queue()

djc:
## What was changed

Add a `WfContext::task_queue()` member allowing the workflow code to access the task queue from which it is being run.

## Why?

Would like to schedule child workflows or activities on the same task queue as the current workflow.


Sushisource (approved):
This is a nice add regardless, but FYI for your specific use case if you simply leave TQ blank it will inherit the TQ of the initiating workflow for both children and activities.

PR #776, opened 2024-07-11
Honor all non-terminal commands

### LINKED ISSUE #778: Honor all non-terminal commands
See https://github.com/temporalio/features/issues/481

Today, core truncates the command sequence sent by lang at the first "terminal" (i.e. workflow-completing) command.

We will change this behavior to instead honor all non-terminal commands, i.e. even when in the command sequence sent by lang they are preceded by a terminal command. In practice, this means moving any terminal commands to the end of the sequence. If there are multiple terminal commands, core will truncate at the first, i.e. respecting the order sent by lang.

The primary motivation for this change is to honor update completions when they occur in the same workflow task as workflow completions.

dandavison:
Fixes #778 

- Previously, core was truncating commands received from lang at the first terminal command.
- With this PR, if there are terminal commands, we remove them all and replace the first one at the end of the command sequence.

### Evidence this is correct
- unit tests added for new and old logic
- TODO: add tests of replay / flag logic to `core`
- https://github.com/temporalio/sdk-python/pull/569 uses these changes and contains tests demonstrating that update completions are now honored, even when after workflow completion.

### DIFF
--- a/core/src/core_tests/workflow_tasks.rs
+++ b/core/src/core_tests/workflow_tasks.rs
@@ -45,7 +45,7 @@ use temporal_sdk_core_protos::{
             StartWorkflow, UpdateRandomSeed, WorkflowActivationJob,
         },
         workflow_commands::{
-            update_response::Response, ActivityCancellationType, CancelTimer,
+            update_response::Response, workflow_command, ActivityCancellationType, CancelTimer,
             CompleteWorkflowExecution, ContinueAsNewWorkflowExecution, FailWorkflowExecution,
             RequestCancelActivity, ScheduleActivity, SetPatchMarker, StartChildWorkflowExecution,
             UpdateResponse,
@@ -56,7 +56,7 @@ use temporal_sdk_core_protos::{
     temporal::api::{
         command::v1::command::Attributes,
         common::v1::{Payload, RetryPolicy, WorkerVersionStamp},
-        enums::v1::{EventType, WorkflowTaskFailedCause},
+        enums::v1::{CommandType, EventType, WorkflowTaskFailedCause},
         failure::v1::Failure,
         history::v1::{
             history_event, TimerFiredEventAttributes,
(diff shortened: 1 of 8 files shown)


core/src/internal_flags.rs
```diff
@@ -29,6 +29,15 @@ pub(crate) enum CoreInternalFlags {
...
     TooHigh = u32::MAX,
+    /// Prior to this flag, we truncated commands received from lang at the
+    /// first terminal (i.e. workflow-terminating) command. With this flag, we
+    /// reorder commands such that all non-terminal commands come first,
+    /// followed by the first terminal command, if any (it's possible that
+    /// multiple workflow coroutines generated a terminal command). This has the
+    /// consequence that all non-terminal commands are sent to the server, even
+    /// if in the sequence delivered by lang they came after a terminal command.
+    /// See https://github.com/temporalio/features/issues/481.
+    ReorderCommands = 3,
```
Sushisource, line 40: Can we move this above `TooHigh`? I might also call it `ReorderTerminalCommands` just to be a bit more explicit

core/src/core_tests/workflow_tasks.rs
```diff
@@ -2518,8 +2518,9 @@ async fn post_terminal_commands_are_discarded() {
         mock_workflow_client(),
     );
     mh.completion_mock_fn = Some(Box::new(|c| {
-        // Only the complete execution command should actually be sent
-        assert_eq!(c.commands.len(), 1);
+        // The start timer and complete execution commands should be sent, in
+        // that order.
+        assert_eq!(c.commands.len(), 2);
```
Sushisource, line 2522: This should probably assert the actual command types as well while we're at it

core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -470,6 +470,12 @@ impl WorkflowMachines {
             .add_lang_used(flags);
     }
 
+    pub(crate) fn try_use(&self, flag: CoreInternalFlags, should_record: bool) -> bool {
```
Sushisource, line 473: Maybe `try_use_flag` since it's now slightly less obvious w/o the field access

core/src/worker/workflow/managed_run.rs
```diff
@@ -1169,6 +1197,67 @@ impl ManagedRun {
...
+// Note that multiple coroutines may have generated a terminal command, leading
+// to multiple terminal commands in the input to this function.
+fn preprocess_command_sequence(
+    mut commands: Vec<WFCommand>,
+) -> (Vec<WFCommand>, Vec<QueryResult>, bool) {
+    let mut query_results = vec![];
+    let mut terminals = vec![];
+    let (mut seen_terminal, mut any_moved) = (false, false);
+
+    commands = std::mem::take(&mut commands)
```
Sushisource, line 1214: `take` here and in the next fn no longer needed. (I guess it wasn't really before either, at some point that was a ref rather than being passed by value)

Sushisource (commented):
Looking good to me. Just needs the unit tests to test dealing with flag being present or not in old history, and integ test like you said.

core/src/internal_flags.rs
```diff
@@ -114,9 +123,9 @@ impl InternalFlags {
         }
     }
 
-    /// Wipes the recorded flags used during the current WFT and returns a partially filled
-    /// sdk metadata message that can be combined with any existing data before sending the WFT
-    /// complete
+    /// Return a partially filled sdk metadata message containing core and lang flags added since
+    /// the last WFT complete. The returned value can be combined with other data before sending the
+    /// WFT complete.
```
Sushisource, line 128: I think it's important to keep saying that this wipes the since last complete fields.

core/src/core_tests/workflow_tasks.rs
dandavison: Done

core/src/core_tests/workflow_tasks.rs
```diff
@@ -2505,43 +2505,121 @@ async fn core_internal_flags() {
...
+        expected_command_types_emitted,
+        t,
+    )
+    .await
+}
+
+async fn _do_post_terminal_commands_test(
+    commands_sent_by_lang: Vec<workflow_command::Variant>,
+    response_types: impl IntoIterator<Item = impl Into<ResponseType>>,
+    expected_command_types: Option<Vec<CommandType>>,
```
Sushisource, line 2593: nit: Doesn't need to be an option, can just be empty vec when `None`

Sushisource (approved):
Thanks Dan!

core/src/core_tests/workflow_tasks.rs
dandavison: I think option is a reasonable choice: `None` means the concept of expected commands is irrelevant, since the worker is purely replaying all history and will never emit a completion; that's as opposed to `Some(empty_vec)` which would mean SDK did a non-replay WFT but replied without any commands (e.g. all coroutines are blocked on a workflow condition; also WFT fail might be represented like that).

Issue #774, opened 2024-07-09
[Bug] Respect lang-set Tonic request timeouts and designate some more calls with long-polling timeout defaults


cretz:
### Describe the bug

Two issues:

* When `set_timeout` is used by lang on non-long-poll call (retry client or not), it seems to get overridden by the 30s default timeout despite the interceptor checking for the `grpc-timeout` header first.
* `poll_workflow_task_queue` and `poll_activity_task_queue` are the only calls that get the 70s long poll timeout today. This means things like waiting for workflow result get the default 30s which actually returns empty response in 20s. We need to also set that long poll timeout for `update_workflow_execution`, `poll_workflow_execution_update` and `get_workflow_execution_history` (when `wait_new_event` is true).

Sushisource:
Timeouts at least seem to be working properly: https://github.com/temporalio/sdk-core/commit/e64b05b8f26f132deba7c32eab7c682eeff861d0

cretz:
I can confirm that timeout does work as expected on update start. I am unsure why/how I came to the conclusion it didn't. Changing title/description to reflect this as a long-poll timeout issue only.

PR #780, opened 2024-07-17
Ensure some rpcs use long-poll timeout

Sushisource:
## What was changed
Added tests for grpc-timeout header (which was already working properly turns out)
Made certain rpcs that do/can act like long-polls use the long poll timeout.
One unrelated bit of cleanup around slot permit reservation I had pending.

## Why?
Better timeouts

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/774

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/client/src/raw.rs
+++ b/client/src/raw.rs
@@ -567,7 +567,7 @@ proxier! {
         |r| {
             let labels = AttachMetricLabels::namespace(r.get_ref().namespace.clone());
             r.extensions_mut().insert(labels);
-            if r.get_ref().wait_new_event == true {
+            if r.get_ref().wait_new_event {
                 r.set_timeout(LONG_POLL_TIMEOUT);
             }
         }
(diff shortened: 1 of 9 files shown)


cretz (approved):

client/src/raw.rs
```diff
@@ -980,6 +983,7 @@ proxier! {
         |r| {
             let labels = AttachMetricLabels::namespace(r.get_ref().namespace.clone());
             r.extensions_mut().insert(labels);
+            r.set_timeout(LONG_POLL_TIMEOUT);
```
dandavison, line 986: These changes cause the Python test `tests/worker/test_workflow.py::test_workflow_update_timeout_or_cancel` to fail, because the `rpc_timeout` argument in client calls is ignored. See https://github.com/temporalio/sdk-core/pull/781

Issue #736, opened 2024-05-09
[Feature Request] Set Temporal-Namespace header on every namespace-specific gRPC request


cretz:
### Describe the solution you'd like

See https://github.com/temporalio/features/issues/475. It needs to be an interceptor surely, but unsure how the implementation will do this generically. Maybe macros or code gen can help here. This is left up to the implementer. All that matters is that new RPCs that come in with namespaces automatically have the header set.

PR #783, opened 2024-07-30
Attach `temporal-namespace` header to namespace requests

Sushisource:
## What was changed
Attach `temporal-namespace` header to namespaced calls

## Why?
Requested by cloud

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/736

2. How was this tested:
Integ test

3. Any docs updates needed?

### DIFF
--- a/client/src/raw.rs
+++ b/client/src/raw.rs
@@ -7,7 +7,7 @@ use crate::{
     raw::sealed::RawClientLike,
     worker_registry::{Slot, SlotManager},
     Client, ConfiguredClient, InterceptedMetricsSvc, RequestExt, RetryClient,
-    TemporalServiceClient, LONG_POLL_TIMEOUT,
+    TemporalServiceClient, LONG_POLL_TIMEOUT, TEMPORAL_NAMESPACE_HEADER_KEY,
 };
 use futures::{future::BoxFuture, FutureExt, TryFutureExt};
 use std::sync::Arc;
@@ -23,7 +23,10 @@ use temporal_sdk_core_protos::{
     },
 };
 use tonic::{
-    body::BoxBody, client::GrpcService, metadata::KeyAndValueRef, Request, Response, Status,
+    body::BoxBody,
+    client::GrpcService,
+    metadata::{AsciiMetadataValue, KeyAndValueRef},
+    Request, Response, Status,
 };
 
 pub(super) mod sealed {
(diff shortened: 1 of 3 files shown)


client/src/raw.rs
```diff
@@ -473,6 +476,21 @@ macro_rules! proxier {
...
+        $req.metadata_mut().insert(
+            TEMPORAL_NAMESPACE_HEADER_KEY,
+            ns_str
+                .parse()
+                .unwrap_or_else(|_| AsciiMetadataValue::from_static("")),
+        );
+        // Init metric labels
+        AttachMetricLabels::namespace(ns_str)
+    }};
+}
```
dandavison, line 492: Does this need be a macro?

Sushisource: Yes. There's no trait for `.namespace` and adding one would involve writing a whole new procmacro crate. This is the simplest type safe way to do it. That's the same reason there are a lot of macros in general in here, because I have to work with generated code I can't easily control.

dandavison: OK, I was just curious because it seemed structurally similar to the `r.extensions_mut().insert(labels);` call nearby (and wasn't doing anything with symbol information available at compile-time only).

client/src/raw.rs
```diff
@@ -473,6 +476,21 @@ macro_rules! proxier {
...
 
+macro_rules! namespaced_request {
+    ($req:ident) => {{
+        let ns_str = $req.get_ref().namespace.clone();
+        // Attach namespace header
+        $req.metadata_mut().insert(
+            TEMPORAL_NAMESPACE_HEADER_KEY,
+            ns_str
+                .parse()
+                .unwrap_or_else(|_| AsciiMetadataValue::from_static("")),
```
dandavison, line 487: Can you confirm that it's correct / what the reason is to insert an empty value in this case rather than omitting the header?

Sushisource: I dunno that's it's well defined what would make sense to do here. The call _is supposed to be_ namespace specific, but, the name of it cannot be set as a header. So I suppose it makes sense to me to indicate it was trying to do a namespace call but couldn't fill the value.

dandavison: I think there should be a warning containing the value that could not be parsed.

Sushisource: Sure, I can add that. But, I don't think it's actually possible since I doubt we would allow registering namespaces that have invalid chars.

dandavison: > I don't think it's actually possible
Yeah, a subset of impossible that maps to warn rather than panic, I'm thinking? Even if we knew we were talking to a server that wants this header, failing ATs / WFTs / WFs would be too strict I think. But I kind of don't think any subset of impossible maps to silent, except maybe some security situations.

PR #784, opened 2024-07-30
rust-sdk: prevent panic on non-existing workflow

### LINKED ISSUE #558: [Bug] Worker running which encounters an unknown workflow type will exit
I'm trying to run a worker using the (alpha) rust SDK so I'm not sure if this is a bug or is just not supported. But currently if a worker is running and an unknown workflow type is encountered on the same task queue, then the worker will exit with the error `Workflow type {workflow_type} not found`

### Describe the bug
When a worker is running and an unknown workflow type is encountered on the same task queue, then the worker will exit with the error `Workflow type {workflow_type} not found`

This happens in this line of code:

https://github.com/temporalio/sdk-core/blob/master/sdk/src/lib.rs#L405

Is this intended behavior and if so, what is the suggested solution? Or is this simply unimplemented at this stage.

### Minimal Reproduction
1. Start a worker using the rust lib.
1. Create a workflow that has a different workflow type then what you defined.
1. The worker will exit.

valkum:
## What was changed
This changes the check to warn about the workflow and allows the worker to continue polling from the queue similar to the other SDKs.

Currently, there is no good way of writing a test to make sure this isn't introduced by a refactor in the future.
Maybe with a custom tracing subscriber to check for the warn output, but that would be something for it's own PR.

## Why?
Previously, the worker would panic due to the error return value of the activation handler.
Other SDKs do not panic/exit upon an unknown workflow. 

## Checklist

1. Closes #558

2. How was this tested:
Start an empty worker with a local dev instance and start a workflow with any name. The worker should not exit.

3. Any docs updates needed?
Not really.

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -403,9 +403,10 @@ impl WorkflowHalf {
         }) {
             let workflow_type = &sw.workflow_type;
             let wf_fns_borrow = self.workflow_fns.borrow();
-            let wf_function = wf_fns_borrow
-                .get(workflow_type)
-                .ok_or_else(|| anyhow!("Workflow type {workflow_type} not found"))?;
+            let Some(wf_function) = wf_fns_borrow.get(workflow_type) else {
+                warn!("Workflow type {workflow_type} not found");
+                return Ok(None);
+            };
 
             let (wff, activations) = wf_function.start_workflow(
                 common.worker.get_config().namespace.clone(),


Sushisource (approved):
Makes perfect sense, thank you!

Sushisource:
@valkum As far as testing goes, you could test this by making sure that the history in an integration test contains a workflow task failure. But, we don't have to block the PR on that so I've set it to merge. You're welcome to add a test in another PR if you like.

valkum:
> @valkum As far as testing goes, you could test this by making sure that the history in an integration test contains a workflow task failure. But, we don't have to block the PR on that so I've set it to merge. You're welcome to add a test in another PR if you like.

Hmm. Are you sure that a workflow failure is recorded in this case? I had a look through the go-sdk but couldn't find anything in that regard.

Sushisource:
> > @valkum As far as testing goes, you could test this by making sure that the history in an integration test contains a workflow task failure. But, we don't have to block the PR on that so I've set it to merge. You're welcome to add a test in another PR if you like.
> 
> Hmm. Are you sure that a workflow failure is recorded in this case? I had a look through the go-sdk but couldn't find anything in that regard.

Yep, that's definitely what should happen. In Go that error eventually bubbles up from here https://github.com/temporalio/sdk-go/blob/2baa60eb75c4e35cdc027d17552846807f9b9f50/internal/internal_event_handlers.go#L1426.

valkum:
Following that code path up, it should be this one, right? https://github.com/temporalio/sdk-go/blob/2baa60eb75c4e35cdc027d17552846807f9b9f50/internal/internal_task_pollers.go#L402

Not sure, how much of that machinery is already implemented in the Rust SDK. Will have a look and maybe come back with a PR.

Sushisource:
> Following that code path up, it should be this one, right? https://github.com/temporalio/sdk-go/blob/2baa60eb75c4e35cdc027d17552846807f9b9f50/internal/internal_task_pollers.go#L402
> 
> Not sure, how much of that machinery is already implemented in the Rust SDK. Will have a look and maybe come back with a PR.

The machinery for failing workflow tasks is there, it's just not quite hooked up to this particular spot. Probably it would mostly be a matter of just sending a failure down the `completions_tx` channel in the same spot where you made the change with a detail that the workflow function wasn't registered.

PR #787, opened 2024-08-06
Rust-SDK: Handle unknown workflow

valkum:
## What was changed
This makes sure unknown workflows are handled on-par with other SDKs. This is done by emitting a `WorkflowActivationComplete` message with state set to failure.

This is based on feedback from https://github.com/temporalio/sdk-core/pull/784

## Why?
The recent changed avoided panics caused by unknown workflows. The workflow activation was never marked as failed.
This made it hard to add a test for this behavior.

## Checklist

1. How was this tested:
Test is added to check for the correct behavior.

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -97,6 +97,7 @@ use temporal_sdk_core_protos::{
     },
     temporal::api::{
         common::v1::Payload,
+        enums::v1::WorkflowTaskFailedCause,
         failure::v1::{failure, Failure},
     },
     TaskToken,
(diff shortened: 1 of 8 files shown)


valkum:
The test currently does not shutdown properly. The `temporal_sdk_core::worker: Initiated shutdown` message is logged, so it hangs somewhere after that point. I am not sure if this is caused by emitting the `WorkflowActiviationComplete` message, or by the test structure itself.

sdk/src/lib.rs
```diff
@@ -445,6 +454,23 @@ impl WorkflowHalf {
...
         } else {
+            // When we failed to start a workflow, we never inserted it into the cache.
+            // But core sends us a `RemoveFromCache` job when we mark the StartWorkflow workflow activation
+            // as a failure, which we need to complete.
+            // Other SDKs add the workflow to the cache even when the workflow type is unknown/not found.
+            // To circumvent this, we simply mark any RemoveFromCache job for workflows that are not in the cache as complete.
+            if activation
+                .jobs
+                .iter()
+                .any(|j| matches!(j.variant, Some(Variant::RemoveFromCache(_))))
```
Sushisource, line 465: This should check that the eviction is the *only* job, which should really be the only possible situation here.

Sushisource (commented):
Awesome! Thanks! This looks good to me, just one minor thing.

Sushisource (approved):
Thanks again!

PR #789, opened 2024-08-09
Job ordering updates

Sushisource:
## What was changed
Make `UpdateRandomSeed` jobs come after start workflow, but before everything else.

## Why?
These could technically have consequences for later actions, and should go first. It's intuitive.

The more delicate change here is actually sorting `StartWorkflow` before other stuff, which also seems intuitive, but a number of tests at least in core had some expectations around updates being able to come first. Frankly, that doesn't make a ton of sense, and I don't really expect this move to be any kind of real issue, but it might require some changes to langs that are using `StartWorkflow` as a synonym for "do first iteration of the event loop"? If such changes are needed, they should be the ones described here: https://github.com/temporalio/sdk-python/issues/606 (.NET may need an equivalent item) as none of these problems exist if all jobs are all read and applied to state _first_, before doing any iterating of the event loop.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/790

2. How was this tested:
Updated existing tests

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/mod.rs
+++ b/core/src/worker/workflow/mod.rs
@@ -1331,7 +1331,7 @@ impl LocalActivityRequestSink for LAReqSink {
 /// activations must uphold.
 ///
 /// ## Ordering
-/// `patches -> start workflow -> random-seed-updates -> signals/updates -> other -> queries -> evictions`
+/// `init workflow -> patches -> random-seed-updates -> signals/updates -> other -> queries -> evictions`
 ///
 /// ## Invariants:
 /// * Queries always go in their own activation
@@ -1363,8 +1363,8 @@ fn prepare_to_ship_activation(wfa: &mut WorkflowActivation) {
         }
         fn variant_ordinal(v: &workflow_activation_job::Variant) -> u8 {
             match v {
-                workflow_activation_job::Variant::NotifyHasPatch(_) => 0,
-                workflow_activation_job::Variant::InitializeWorkflow(_) => 1,
+                workflow_activation_job::Variant::InitializeWorkflow(_) => 0,
+                workflow_activation_job::Variant::NotifyHasPatch(_) => 1,
                 workflow_activation_job::Variant::UpdateRandomSeed(_) => 2,
                 workflow_activation_job::Variant::SignalWorkflow(_) => 3,
                 workflow_activation_job::Variant::DoUpdate(_) => 3,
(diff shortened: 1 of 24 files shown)


core/src/worker/workflow/mod.rs
```diff
@@ -1363,17 +1363,19 @@ fn prepare_to_ship_activation(wfa: &mut WorkflowActivation) {
...
         fn variant_ordinal(v: &workflow_activation_job::Variant) -> u8 {
             match v {
-                workflow_activation_job::Variant::NotifyHasPatch(_) => 1,
-                workflow_activation_job::Variant::SignalWorkflow(_) => 2,
-                workflow_activation_job::Variant::DoUpdate(_) => 2,
+                workflow_activation_job::Variant::NotifyHasPatch(_) => 0,
+                workflow_activation_job::Variant::StartWorkflow(_) => 1,
+                workflow_activation_job::Variant::UpdateRandomSeed(_) => 2,
+                workflow_activation_job::Variant::SignalWorkflow(_) => 3,
+                workflow_activation_job::Variant::DoUpdate(_) => 3,
```
cretz, line 1370: I don't think this works. I think signal/update in the same task as start always need to come before start. It's important for data loss reason that signal/update handlers run before the primary method.

Sushisource: Lengthy discussion outside of the PR summarized:

This change's impetus is to make the sending of _information_ be appropriately ordered, and lang can always assume that a main routine ought to be started at the end of the first activation.

core/src/worker/workflow/mod.rs
```diff
@@ -1363,17 +1363,19 @@ fn prepare_to_ship_activation(wfa: &mut WorkflowActivation) {
         }
         fn variant_ordinal(v: &workflow_activation_job::Variant) -> u8 {
             match v {
-                workflow_activation_job::Variant::NotifyHasPatch(_) => 1,
-                workflow_activation_job::Variant::SignalWorkflow(_) => 2,
-                workflow_activation_job::Variant::DoUpdate(_) => 2,
+                workflow_activation_job::Variant::NotifyHasPatch(_) => 0,
+                workflow_activation_job::Variant::InitializeWorkflow(_) => 1,
```
cretz, line 1367: I was under the impression the contract now was that this should always be the first job on the activation, so I wonder if this should be 0 and notify has patch should be 1

Sushisource: Yeah, I suppose there's really no reason not to do that. Will change.

cretz (approved):

Issue #794, opened 2024-08-14
[Feature Request] Add task validation to activity tasks


Sushisource:
Workflow tasks get some basic sanity checks like "have a task token", etc. Activity tasks don't, but probably should, since as we've discovered some users stand up proxies and muck with returned poll tasks and this confuses the worker into thinking it's got a real task instead of an empty poll response.

PR #795, opened 2024-08-14
Basic validation of activity tasks

Sushisource:
## What was changed
At least check that activity tasks have a task token, to avoid user proxies tricking the SDK into thinking an empty poll response wasn't empty.

## Why?
Custom proxy implementations breaking things.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/794

2. How was this tested:
Existing tests

3. Any docs updates needed?

### DIFF
--- a/sdk-core-protos/src/lib.rs
+++ b/sdk-core-protos/src/lib.rs
@@ -436,8 +436,10 @@ pub mod coresdk {
     pub mod workflow_activation {
         use crate::{
             coresdk::{
+                activity_result::{activity_resolution, ActivityResolution},
                 common::NamespacedWorkflowExecution,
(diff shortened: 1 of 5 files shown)


sdk-core-protos/src/lib.rs
```diff
@@ -604,6 +613,28 @@ pub mod coresdk {
             }
         }
 
+        impl Display for ActivityResolution {
+            fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
```
Sushisource, line 617: I added this stuff while debugging something else, it's useful.

cretz (approved):

PR #799, opened 2024-08-19
Add a basic cloud TLS test

Sushisource:
Just for some basic sanity checking on connecting to cloud


Sushisource:
Need to open from not-fork

PR #800, opened 2024-08-19
Add a basic cloud TLS test

Sushisource:


.github/workflows/per-pr.yml
```diff
@@ -61,8 +61,13 @@ jobs:
           name: state-machine-coverage
           path: machine_coverage/
 
-  fmt:
+  integ-tests:
     name: Integ tests
+    env:
+      TEMPORAL_CLOUD_ADDRESS: https://sdk-ci.a2dd6.tmprl.cloud:7233
+      TEMPORAL_CLOUD_NAMESPACE: sdk-ci.a2dd6
```
cretz, line 68:
```suggestion
      TEMPORAL_CLOUD_ADDRESS: https://${{ vars.TEMPORAL_CLIENT_NAMESPACE }}.tmprl.cloud:7233
      TEMPORAL_CLOUD_NAMESPACE: ${{ vars.TEMPORAL_CLIENT_NAMESPACE }}
```

If you wanted, but not required (we hardcode this namespace in some other SDKs because I/we forgot we had this var)

cretz (approved):

.github/workflows/per-pr.yml
Sushisource: Ah, cool yeah that works

PR #804, opened 2024-08-26
Debugger Client in core

twin-drill:
Adding methods for using our VSCode extension into core, so the debugger can be extended into other languages without having to reimplement these basic methods.

Drafted while I'm still working on it.

### DIFF
--- a/core/src/debug_client.rs
+++ b/core/src/debug_client.rs
@@ -37,7 +37,8 @@ impl DebugClient {
 
     async fn post_wft_started(&self, event_id: &i64) -> Result<reqwest::Response, anyhow::Error> {
         let url = self.debugger_url.clone() + "/current-wft-started";
-        Ok(self.client
+        Ok(self
+            .client
             .get(url)
             .header("Temporal-Client-Name", CLIENT_NAME)
             .header("Temporal-Client-Version", CLIENT_VERSION)
--- a/core/src/debug_client.rs
+++ b/core/src/debug_client.rs
@@ -63,10 +63,11 @@ impl DebugClient {
     }
 }
 
-#[ignore]
 #[cfg(test)]
 mod tests {
+    use super::*;
     // this test wont work without a local instance of the debugger running.
+    #[ignore]
     #[tokio::test]
     async fn test_debug_client_delete() {
         let dc = DebugClient::new("http://127.0.0.1:51563".into());
--- a/core/src/debug_client.rs
+++ b/core/src/debug_client.rs
@@ -27,7 +27,7 @@ impl DebugClient {
         }
     }
 
-    /// Get the history from the instance of the debug plugin server, then return the protobinary if successful.
+    /// Get the history from the instance of the debug plugin server
     pub async fn get_history(&self) -> Result<History, anyhow::Error> {
         let url = self.debugger_url.clone() + "/history";
         let resp = self
--- a/core/src/debug_client.rs
+++ b/core/src/debug_client.rs
@@ -8,20 +8,27 @@ use temporal_sdk_core_protos::temporal::api::history::v1::History;
 const CLIENT_NAME: &str = "temporal-core";
 const CLIENT_VERSION: &str = "0.1.0";
 
-struct DebugClient {
+/// Struct representing the internal debug client
+#[derive(Clone)]
+pub struct DebugClient {
+    /// URL for the local instance of the debugger server
     debugger_url: String,
+
+    /// reqwest::Client to avoid re-creating a client for every request.
     client: reqwest::Client,
 }
 
 impl DebugClient {
-    fn new(url: String) -> DebugClient {
+    /// Create a new instance of a DebugClient with the specified url.
+    pub fn new(url: String) -> DebugClient {
         DebugClient {
             debugger_url: url,
             client: reqwest::Client::new(),
         }
     }
 
-    async fn get_history(&self) -> Result<History, anyhow::Error> {
+    /// Get the history from the instance of the debug plugin server, then return the protobinary if successful.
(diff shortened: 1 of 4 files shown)


core/src/debug_client.rs
```diff
@@ -0,0 +1,65 @@
...
+
+        let bytes = resp.bytes().await?;
+        Ok(History::decode(bytes)?) // decode_length_delimited() does not work
+
+    }
+
+    async fn post_wft_started(&self, event_id: &i64) -> bool {
+        let url = self.debugger_url.clone() + "/current-wft-started";
+        let resp = self._client.get(url)
+            .header("Temporal-Client-Name", "temporal-debug-core")
```
Sushisource, line 36: It doesn't tremendously matter what these are but they should at least be the same for these two methods. You can lift both of these up into constants.

core/src/debug_client.rs
```diff
@@ -0,0 +1,65 @@
...
+            .header("Temporal-Client-Version", "0.1.0")
+            .timeout(Duration::from_secs(5))
+            .json(event_id)
+            .send()
+            .await;
+
+        match resp {
+            Ok(r) => { r.status() == reqwest::StatusCode::OK },
+            Err(_) => false
+        }
```
Sushisource, line 46: This hides the error in the event something went wrong - I think logging a warning here makes sense, and probably also just returning the result to the caller makes sense.

core/src/debug_client.rs
```diff
@@ -0,0 +1,65 @@
+use reqwest;
+use prost::Message;
+use temporal_sdk_core_protos::temporal::api::history::v1::History;
+use std::time::Duration;
+
+struct DebugClient {
+    debugger_url: String,
+    _client: reqwest::Client
```
Sushisource, line 8: There's no need for the `_` prefix

core/src/debug_client.rs
```diff
@@ -0,0 +1,71 @@
...
+impl DebugClient {
+    /// Create a new instance of a DebugClient with the specified url.
+    pub fn new(url: String) -> DebugClient {
+        DebugClient {
+            debugger_url: url,
+            client: reqwest::Client::new(),
+        }
+    }
+
+    /// Get the history from the instance of the debug plugin server, then return the protobinary if successful.
```
Sushisource, line 30: This isn't actually encoded yet, so you can just lop that bit off
```suggestion
    /// Get the history from the instance of the debug plugin server
```

core/src/debug_client.rs
```diff
@@ -0,0 +1,71 @@
...
+
+use prost::Message;
+use reqwest;
+use std::time::Duration;
+use temporal_sdk_core_protos::temporal::api::history::v1::History;
+
+const CLIENT_NAME: &str = "temporal-core";
+const CLIENT_VERSION: &str = "0.1.0";
+
+/// Struct representing the internal debug client
```
Sushisource, line 11: This sort of comment isn't telling you much. Prefer instead something like this
```suggestion
/// A client for interacting with the VSCode debug plugin
```

core/src/debug_client.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    // this test wont work without a local instance of the debugger running.
+    // #[tokio::test]
+    // async fn test_debug_client_delete() {
+    //     let dc = DebugClient::new("http://127.0.0.1:51563".into());
+    //     dc.get_history().await.unwrap();
+    // }
```
Sushisource, line 70: Rather than commenting it out, just use `#[ignore]`

core/src/debug_client.rs
```diff
@@ -0,0 +1,71 @@
...
+        let resp = self
+            .client
+            .get(url)
+            .header("Temporal-Client-Name", CLIENT_NAME)
+            .header("Temporal-Client-Version", CLIENT_VERSION)
+            .send()
+            .await?;
+
+        let bytes = resp.bytes().await?;
+        Ok(History::decode(bytes)?) // decode_length_delimited() does not work
```
Sushisource, line 42:
```suggestion
        Ok(History::decode(bytes)?)
```

core/src/debug_client.rs
```diff
@@ -0,0 +1,71 @@
...
+const CLIENT_NAME: &str = "temporal-core";
+const CLIENT_VERSION: &str = "0.1.0";
+
+/// Struct representing the internal debug client
+#[derive(Clone)]
+pub struct DebugClient {
+    /// URL for the local instance of the debugger server
+    debugger_url: String,
+
+    /// reqwest::Client to avoid re-creating a client for every request.
```
Sushisource, line 17: You never need to repeat type names in comments because the doc generation and any good IDE is gonna show you the type anyway
```suggestion
    /// Underlying HTTP client
```

core/src/debug_client.rs
```diff
@@ -0,0 +1,71 @@
...
+use temporal_sdk_core_protos::temporal::api::history::v1::History;
+
+const CLIENT_NAME: &str = "temporal-core";
+const CLIENT_VERSION: &str = "0.1.0";
+
+/// Struct representing the internal debug client
+#[derive(Clone)]
+pub struct DebugClient {
+    /// URL for the local instance of the debugger server
+    debugger_url: String,
```
Sushisource, line 15: This can be an actual `Url` type which we already depend on https://docs.rs/url/latest/url/struct.Url.html (which also has safe ways to construct paths like adding `/history`)

Sushisource (commented):
Looking pretty good, mostly just comment changes here

Sushisource:
Closed in favor of https://github.com/temporalio/sdk-core/pull/829

PR #805, opened 2024-08-27
Move Payload pretty-print from Display to Debug

yuandrew:
## What was changed

Spencer previously implemented a pretty-print for `Display` for `Payload`. `Debug` was not touched because prost automatically generated the `Debug` impl for `Payload`. Prost and tonic-build now support skipping the `Debug` impl, https://github.com/tokio-rs/prost/pull/797, so we can leverage this to provide our own `Debug` impl.

Also added support to skip this truncation by specifying the `TEMPORAL_PRINT_FULL_PAYLOADS` env variable

## Why?

A nice-to-have feature to make debug logs more readable.

## Checklist

1. Closes 
#806

3. How was this tested:

Generated temporal.api.common.v1.rs now generates 
```
#[cfg_attr(
    feature = "serde_serialize",
    derive(::serde::Serialize, ::serde::Deserialize)
)]
#[allow(clippy::derive_partial_eq_without_eq)]
#[derive(Clone, PartialEq, ::prost::Message)]
#[prost(skip_debug)]
pub struct Payload {
    #[prost(map = "string, bytes", tag = "1")]
    pub metadata: ::std::collections::HashMap<
        ::prost::alloc::string::String,
        ::prost::alloc::vec::Vec<u8>,
    >,
    #[prost(bytes = "vec", tag = "2")]
    pub data: ::prost::alloc::vec::Vec<u8>,
}

impl std::fmt::Debug for Payload {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        if self.data.len() > 64 {
            let mut windows = self.data.as_slice().windows(32);
            write!(
                f,
                "[{}..{}]",
                BASE64_STANDARD.encode(windows.next().unwrap_or_default()),
                BASE64_STANDARD.encode(windows.next_back().unwrap_or_default())
            )
        } else {
            write!(f, "[{}]", BASE64_STANDARD.encode(&self.data))
        }
    }
}
```
4. Any docs updates needed?

### DIFF
--- a/sdk-core-protos/build.rs
+++ b/sdk-core-protos/build.rs
@@ -1,9 +1,4 @@
-use std::{
-    env,
-    fs::File,
-    io::{BufRead, BufReader},
-    path::PathBuf,
-};
+use std::{env, path::PathBuf};
 
 static ALWAYS_SERDE: &str = "#[cfg_attr(not(feature = \"serde_serialize\"), \
                                derive(::serde::Serialize, ::serde::Deserialize))]";
@@ -12,6 +7,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("cargo:rerun-if-changed=./protos");
     let out = PathBuf::from(env::var("OUT_DIR").unwrap());
     let descriptor_file = out.join("descriptors.bin");
+
+    let mut config = prost_build::Config::default();
+    config.skip_debug(["temporal.api.common.v1.Payload"]);
+
     tonic_build::configure()
         // We don't actually want to build the grpc definitions - we don't need them (for now).
         // Just build the message structs.
@@ -118,7 +117,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             "::prost_wkt_types::Value"
         )
         .file_descriptor_set_path(descriptor_file)
-        .compile(
+        .compile_with_config(
+            config,
             &[
                 "./protos/local/temporal/sdk/core/core_interface.proto",
                 "./protos/api_upstream/temporal/api/workflowservice/v1/service.proto",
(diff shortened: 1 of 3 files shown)


sdk-core-protos/build.rs
```diff
@@ -131,5 +136,58 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             ],
         )?;
 
+    let file_path = out.join("temporal.api.common.v1.rs");
+    post_process_payload_file(file_path)?;
```
Sushisource, line 140: So I just saw in the docs there's a built-in thing on the configurator: https://docs.rs/prost-build/0.13.1/prost_build/struct.Config.html#method.skip_debug

I think we might not need to do any manual replacement.

sdk-core-protos/build.rs
```diff
@@ -131,5 +136,58 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        }
+
+        // Write the current line to the new content
+        new_file_contents.push_str(&line);
+        new_file_contents.push('\n');
+
+        if found_payload && line == "}" && !written {
+            let debug_impl = r#"
+impl std::fmt::Debug for Payload {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
```
Sushisource, line 165: Even if we do need to keep manual replacement, there's no need to inject this here. The debug impl can live in the same place the current Display impl is.

yuandrew: Without Debug being in the generated files, we get an error with `derive(::prost::Message)`

```
`temporal::api::common::v1::Payload` doesn't implement `derive_more::Debug`
the trait `derive_more::Debug` is not implemented for `temporal::api::common::v1::Payload`, which is required by `&std::vec::Vec<temporal::api::common::v1::Payload>: derive_more::Debug`
add `#[derive(Debug)]` to `temporal::api::common::v1::Payload` or manually `impl derive_more::Debug for temporal::api::common::v1::Payload`
the trait `derive_more::Debug` is implemented for `std::vec::Vec<T, A>`
required for `std::vec::Vec<temporal::api::common::v1::Payload>` to implement `derive_more::Debug`
1 redundant requirement hidden
required for `&std::vec::Vec<temporal::api::common::v1::Payload>` to implement `derive_more::Debug`
required for the cast from `&&std::vec::Vec<temporal::api::common::v1::Payload>` to `&dyn derive_more::Debug`
```

sdk-core-protos/build.rs
Sushisource: Aaaah, I see. That would be awesome, yes. In the meantime looks like you could potentially use this https://docs.rs/tonic-build/latest/tonic_build/struct.Builder.html#method.compile_with_config to interact directly with the prost config

sdk-core-protos/build.rs
Sushisource: Ugh, that's annoying. I wonder exactly where that's coming from. It still does this even after you have the `Debug` impl in  `sdk-core-protos/lib.rs`? I'm really surprised by that because the generated code is literally inlined into that file, so it should be identical to having it written into the generated code

yuandrew: I didn't try putting the `Debug` impl in `sdk-core-protos/lib.rs`, you're right, that resolves the error

sdk-core-protos/build.rs
yuandrew: https://github.com/hyperium/tonic/pull/1900 created, using `compile_with_config` in the meantime, thanks for the suggestion!

Sushisource (commented):
Yaaaaaaay! I'm jazzed for this.

My only remaining thought is that sometimes it may be pretty important to get the full payload dumped to the screen, even if it is longer than 64 bytes. I can't really think of any reasonable way to configure than other than an environment variable.

Let's add that so we at least have the option when debugging stuff or if a customer needs to show us some info. We can call it something like `TEMPORAL_PRINT_FULL_PAYLOADS` and if that's set in the environment, then don't do the >64 byte truncation thing

Sushisource (approved):
Noice!

PR #809, opened 2024-09-04
Fix legacy query failure replying

Sushisource:
## What was changed
Make sure we still reply to legacy queries in specific failure scenarios

## Why?
Could cause timeouts where they were not necessary

## Checklist

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?


Quinn-With-Two-Ns:
Is this fixing a regression in core or has it always been broken?

Sushisource:
> Is this fixing a regression in core or has it always been broken?

No, this is making a refinement to a recent change introduced here: https://github.com/temporalio/sdk-core/pull/792

Now rather than always dumping out of replying to queries if history application failed in an earlier activation, now we'll still reply if the only thing the task was doing was issuing a legacy query.

Quinn-With-Two-Ns (approved):

Issue #807, opened 2024-08-30
[Bug] Recent reports of activity completion timing


cretz:
### Describe the bug

In latest .NET (1.3.0) and Python SDKs (1.7.0), users are reporting seeing things like:

> 2024-08-30T15:15:19.610801Z  WARN temporal_sdk_core::worker::activities: Network error while completing activity error=Status { code: Cancelled, message: "operation was canceled", source: Some(tonic::transport::Error(Transport, hyper::Error(Canceled, "connection closed"))) }

Or

> 2024-08-23T18:12:54.800672Z  WARN temporal_sdk_core::worker::activities::activity_heartbeat_manager: Error when recording heartbeat: Status { code: Cancelled, message: "operation was canceled", source: Some(tonic::transport::Error(Transport, hyper::Error(Canceled, "connection closed"))) }

This happens intermittently (i.e. not all the time) and has been confirmed not to happen when the users downgrade (.NET 1.2.0 and Python 1.6.0). There are not much more details to go on. It is technically possible these are unrelated and lang-specific things. I have no replication from any user.

My initial suspicion was #780 (I saw something at #781 that fixed something but unfamiliar), but I see nothing obvious there. Maybe a Tonic upgrade at #782 messed things up here. I think the first thing is to confirm what the timeout is set to for these activity completion calls. I wonder if we can simulate a tad slow respond activity complete or activity heartbeat server side and see if it replicates.

talmawash:
We are experiencing the same issue in Python SDK (1.7.1)

tetrakatech:
In case it's helpful for future debugging (either by project maintainers, or by anyone else), we ran into the same error in production using version 1.11.1 of the Typescript SDK, and it seems to have been resolved by upgrading to 1.11.3 (which includes #811).

Our log lines:
```json
{
  "jsonPayload": {
    "metadata": {
      "error": "Status { code: Cancelled, message: \"operation was canceled\", source: Some(tonic::transport::Error(Transport, hyper::Error(Canceled, \"connection closed\"))) }",
      "sdkComponent": "core",
      "label": "worker"
    },
    "message": "Network error while completing activity"
  },
  "timestamp": "2024-11-06T03:29:36.765000104Z",
  "severity": "WARNING",
}
```

[...]

PR #811, opened 2024-09-05
Attempt to address unusal rpc cancellation error

Sushisource:
## What was changed
* Fixed get workflow history not being considered a long-poll call (discovered while trying to repro)
* Retry at most once the specific error users have seen occasionally during activity completions

## Why?
Avoid spurious rpc failures due to goaways

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/807

2. How was this tested:
Could not repro the exact case, but, by analysis this should allow that situation to retry at most once.

3. Any docs updates needed?

### DIFF
--- a/client/src/raw.rs
+++ b/client/src/raw.rs
@@ -149,12 +149,12 @@ where
         F: FnMut(&mut Self, Request<Req>) -> BoxFuture<'static, Result<Response<Resp>, Status>>,
         F: Send + Sync + Unpin + 'static,
     {
-        let rtc = self.get_retry_config(call_name);
+        let info = self.get_call_info(call_name, Some(&req));
         let fact = || {
             let req_clone = req_cloner(&req);
             callfn(self, req_clone)
         };
-        let res = Self::make_future_retry(rtc, fact, call_name);
+        let res = Self::make_future_retry(info, fact);
         res.map_err(|(e, _attempt)| e).map_ok(|x| x.0).await
     }
 }
@@ -318,7 +318,6 @@ impl RawClientLike for Client {
 }
 
 /// Helper for cloning a tonic request as long as the inner message may be cloned.
-/// We drop extensions, so, lang bridges can't pass those in :shrug:
 fn req_cloner<T: Clone>(cloneme: &Request<T>) -> Request<T> {
     let msg = cloneme.get_ref().clone();
     let mut new_req = Request::new(msg);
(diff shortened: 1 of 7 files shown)


client/src/retry.rs
```diff
@@ -201,7 +229,26 @@ where
         let long_poll_allowed =
             is_long_poll && [Code::Cancelled, Code::DeadlineExceeded].contains(&e.code());
 
-        if RETRYABLE_ERROR_CODES.contains(&e.code()) || long_poll_allowed {
+        // Sometimes we can get a GOAWAY that, for whatever reason, isn't quite properly handled
```
Sushisource, line 232: This is the relevant fix

client/src/retry.rs
```diff
@@ -201,7 +229,26 @@ where
...
+        // bug and it is fixed.
+        let mut goaway_retry_allowed = false;
+        if !self.have_retried_goaway_cancel && e.code() == Code::Cancelled {
+            if let Some(e) = e
+                .source()
+                .and_then(|e| e.downcast_ref::<tonic::transport::Error>())
+                .and_then(|te| te.source())
+                .and_then(|tec| tec.downcast_ref::<hyper::Error>())
+            {
+                if format!("{e:?}").contains("connection closed") {
```
Sushisource, line 244: There is no other good way to do this without again down-casting _this_ to a &'static str

client/src/raw.rs
```diff
@@ -586,6 +609,9 @@ proxier! {
         |r| {
             let labels = namespaced_request!(r);
             r.extensions_mut().insert(labels);
+            if r.get_ref().wait_new_event {
+                r.extensions_mut().insert(ForceConsiderLongPoll);
+            }
```
cretz, line 614: I wonder if this should also occur on `UpdateWorkflowExecution` and `PollWorkflowExecutionUpdate`

cretz (approved):

client/src/retry.rs
cretz: I think the bigger concern is if they change the string on us since we can't really test this

client/src/raw.rs
Sushisource: I considered it, but update pretty sure you wouldn't want to work exactly the same way where it retries forever. Poll could be ok to do that with potentially.

cretz: Hrmm, arguably we wouldn't want that for get history either. Users may wait for next history event and wait for update response the same way.

PR #815, opened 2024-09-16
La metrics

Sushisource:
## What was changed
Added local activity metrics to core. No flag needed as these were not being recorded under existing activity metrics anyway.

## Why?
Parity with other SDKs

## Checklist

1. Closes https://github.com/temporalio/features/issues/522

2. How was this tested:
Added integ test against Prometheus metrics

3. Any docs updates needed?
Yes, docs site needs tables etc on metrics page updated


cretz (approved):
Probably need another issue somewhere to update https://docs.temporal.io/references/sdk-metrics to include Core on these now

PR #823, opened 2024-09-24
Upgrade quinn-proto version to fix security vulnerability

yuandrew:
## What was changed

Added a `quinn-proto` version requirement of 0.11.7. This ends up adding v0.11.8 to Cargo.lock.

## Why?

This security issue was flagged from the python SDK, https://github.com/temporalio/sdk-python/issues/642,


Sushisource:
We don't need to do this here -Core doesn't depend directly on this and don't need to. All that needs to happen is to run `cargo update` in the Python SDK so that the lock there gets the new version.

PR #824, opened 2024-09-27
Use the system-native TLS CAs instead of rustls's built-in CAs

### LINKED ISSUE #818: [Bug] Broken download for ephemeral_server due to wrong CA

Trying to run tests in CICD using auto downloaded ephemeral_server.

### Describe the bug

The pipeline where we try to run the test can access outside internet only via proxy. So downloading anything via SSL requires adding our own CA either to CA root ot by appending it via some env variable. We automatically do both and it works everywhere except in pipeline that is downloading ephemeral_server to execute unit tests. 

The problem is this: https://github.com/temporalio/sdk-core/blob/c27b739e1a4c3f42c694b203ae3a269f5a8ec63f/core/Cargo.toml#L56

rustls-tls feature == rustls-tls-webpki-roots (https://github.com/rustls/webpki-roots)

That means that downloading ephemeral_server  will be validated only via Hardcoded Mozilla's trusted root certificates and nothing else. System root CA is not included and it is not possible to append any other CA via ENV variable.

I suggest switching to rustls-tls-native-roots. This feature also allows to append custom CA for validation.

### Minimal Reproduction

running test via

```
from temporalio.testing import WorkflowEnvironment

env = await WorkflowEnvironment.start_local()
```

in an environment where access to outside is provided via self sign proxy

### Environment/Versions

- OS and processor: Linux x86_64/CICD/Jenkins
- Temporal Version: latest
- Are you using Docker or Kubernetes or building Temporal from source? Using provided temporal pkgs

josh-berry:
Closes #818.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -53,7 +53,7 @@ prometheus = "0.13"
 prost = { workspace = true }
 prost-types = { version = "0.6", package = "prost-wkt-types" }
 rand = "0.8.3"
-reqwest = { version = "0.12", features = ["json", "stream", "rustls-tls"], default-features = false, optional = true }
+reqwest = { version = "0.12", features = ["json", "stream", "rustls-tls-native-roots"], default-features = false, optional = true }
 ringbuf = "0.4"
 serde = "1.0"
 serde_json = "1.0"
--- a/sdk-core-protos/build.rs
+++ b/sdk-core-protos/build.rs
@@ -113,7 +113,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             "::prost_wkt_types::Value"
         )
         .file_descriptor_set_path(descriptor_file)
-        .compile(
+        .compile_protos(
             &[
                 "./protos/local/temporal/sdk/core/core_interface.proto",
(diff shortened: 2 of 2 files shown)


josh-berry:
Note: This still needs to be manually tested, but I'm posting here anyway to get a CI run and because I don't think that testing has to block the (trivial) review.

cretz (approved):
LGTM. Deprecation error in CI is due to brand new tonic change release yesterday in 0.12.3 and we don't use lock files in this repo so it downloads the latest on each CI. Will have to change the called method and commit to get CI to pass.

josh-berry:
Did some manual testing:

1. Created an nginx server to serve the temporal CLI (/dev-server) locally
2. Created a custom CA for nginx
3. Verified that `cargo test --test=integ_tests temporal_cli_default` failed with an SSL error
4. Added the CA to my system's trusted CAs
5. Verified that `cargo test --test=integ_tests temporal_cli_default` succeded (well, actually, failed with a different error since it was expecting JSON and I gave it a binary by mistake)

So I can confirm we are actually paying attention to the system CA roots with this feature flag changed. Merging accordingly.

PR #825, opened 2024-10-06
Replace futures with future-util

negezor:
## What was changed
Replaced `futures` crate with its direct dependencies

## Why?
It's much easier to control small dependencies instead of one big one. Most of the functions from `futures` crate are not used.

## Checklist

1. How was this tested:
Run `cargo test`

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -20,7 +20,7 @@ backoff = "0.4"
 base64 = "0.22"
 derive_builder = { workspace = true }
 derive_more = { workspace = true }
-futures = "0.3"
+futures-util = { version = "0.3", default-features = false }
 futures-retry = "0.6.0"
 http = "1.1.0"
 http-body-util = "0.1"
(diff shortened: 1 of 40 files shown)


Sushisource:
Looks like the futures::channel usage isn't actual in futures_util (see failing integ test compile)

In general I'm OK with this sort of thing, but I am curious about the motivation here beyond the obvious "reduce dependency size". Is there something in particular that motivated these PRs?

negezor:
Actually, the reason is quite simple - compilation time. When I installed the sdk for my Rust project, I noticed how much my `Cargo.lock` got fat, and the compile time dramatically increased. Before that I had no dependency on `futures`, only on its little crates. 

I also sent PR to other repositories that temporal sdk depends on to reduce direct dependencies on `futures`. I hope step by step it will be possible to reduce other dependencies as well.

Sushisource (approved):
Cool, thanks!

PR #826, opened 2024-10-06
Use mpsc instead crossbeam-channel

negezor:
## What was changed
Replaced `crossbeam-channel` crate with `std::sync::mpsc`

## Why?
Noticed that `crossbeam-channel` is being used unnecessarily.

## Checklist

1. How was this tested:
Run `cargo test`


Sushisource (approved):
Makes sense, thanks!

Sushisource (approved):

PR #827, opened 2024-10-06
Replace once_cell with std::sync::{OnceLock,LazyLock}

negezor:
## What was changed
Replaced `once_cell` with `std::sync`

## Why?
Since you can already directly use `OnceLock` and `LazyLock` from std without an additional crate. However, this would require raising MSVR to 1.80 because of `LazyLock`, but `OnceLock` can be used from 1.70 onwards.

Considering that the current approximate MSVR is `1.77` (based on GitHub Workflow) I think it won't be a big deal to raise it to `1.80`.

## Checklist

1. How was this tested:
Run `cargo test`

### DIFF
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -21,7 +21,7 @@ jobs:
           submodules: recursive
       - uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: 1.77.0
+          toolchain: 1.80.0
       - name: Install protoc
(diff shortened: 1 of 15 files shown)


Sushisource (approved):
Thanks! I have no problem bumping the minimum ver, but the docker image version also needs to get bumped for tests to pass looks like.

Also needs a cargo fmt to pass CI, I'll do it myself later today if you don't do it first.

Sushisource (approved):

PR #828, opened 2024-10-09
add id_conflict_policy to client::WorkflowOptions

EddieWhi:
## What was changed

Added the WorkflowOptions::id_conflict_policy property.

## Why?

I have a need to use this feature of temporal when (unofficially) using the Rust client library.

## Checklist

1. Closes 
No issue

3. How was this tested:

Tested this locally (i.e. used this feature from code on my machine).
Happy to add automated tests if you think there should be some... If you're open to this change and think so, could point me to a reasonable place to add a test and/or another similar test to follow?

4. Any docs updates needed?

Don't think so.


Sushisource (approved):
Thanks!

EddieWhi:
Many thanks @Sushisource :)

PR #829, opened 2024-10-14
Client for VSCode Debug Plugin

Sushisource:
Simple client to help languages interact with the vscode debug plugin.

Arguably this isn't really doing much, but, it's at least vaguely useful.

### DIFF
--- a/core/src/debug_client.rs
+++ b/core/src/debug_client.rs
@@ -37,7 +37,8 @@ impl DebugClient {
 
     async fn post_wft_started(&self, event_id: &i64) -> Result<reqwest::Response, anyhow::Error> {
         let url = self.debugger_url.clone() + "/current-wft-started";
-        Ok(self.client
+        Ok(self
+            .client
             .get(url)
             .header("Temporal-Client-Name", CLIENT_NAME)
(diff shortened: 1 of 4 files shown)


core/src/debug_client.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /// Get the history from the instance of the debug plugin server
+    pub async fn get_history(&self) -> Result<History, anyhow::Error> {
+        let url = self.debugger_url.join("history")?;
+        let resp = self
+            .client
+            .get(url)
+            .header("Temporal-Client-Name", CLIENT_NAME)
+            .header("Temporal-Client-Version", CLIENT_VERSION)
```
bergundy, line 45: You'll want to let Lang set this.

bergundy, line 45: Consider using https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.default_headers instead of setting this for every request (yes there's only two now).

bergundy (approved):
Not a codeowner but approving FTR.

cretz (approved):

PR #833, opened 2024-10-23
Fix possible replay bug with empty WFT sequences and updates

Sushisource:
## What was changed
Fixed a bug where an empty WFT sequence followed by an update could cause the update job to be sent before things that happened preceding the empty sequence.

## Why?
Bugfix

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/673
2. How was this tested:
UT / IT

3. Any docs updates needed?


cretz (approved):

dandavison (approved):
Nice, I've spent a bit of time with this. The fix LGTM and I've convinced myself that I understand what at least some of the tests are doing, although I haven't looked at `replay_with_signal_and_update_same_task`.

PR #838, opened 2024-11-01
Changes needed for lang-implementable slot suppliers

Sushisource:
Unlocks https://github.com/temporalio/sdk-typescript/pull/1553 and other langs


mjameswh (approved):
Everything looks ok here...

Issue #822, opened 2024-09-24
[Feature Request] Call ShutdownWorker API on worker shutdown


cretz:
### Describe the solution you'd like

See https://github.com/temporalio/features/issues/542

PR #839, opened 2024-11-04
Update proto submodule in prep for ShutdownWorker API call

yuandrew:
## What was changed

Ran `git pull --squash --rebase=false -s subtree ssh://git@github.com/temporalio/api.git master --allow-unrelated-histories` from the README to update the submodule and fix up the new rust errors from the update.

## Why?

Need to update the proto submodule to access ShutdownWorker API for #822


client/src/raw.rs
```diff
@@ -1124,6 +1124,52 @@ proxier! {
             r.extensions_mut().insert(labels);
         }
     );
+    (
+        shutdown_worker,
+        ShutdownWorkerRequest,
+        ShutdownWorkerResponse,
+        // TODO: need the namespaced_request label?
+        |r| {
```
yuandrew, line 1132: @Sushisource I added these in by default to match the majority, but lmk if this should be removed

Sushisource: Yeah it should be kept. All namespaced calls should have this label.

Sushisource (approved):
Sweet, thanks!

client/src/raw.rs
yuandrew: oops, forgot to remove this comment, will remove in the follow-up PR

PR #841, opened 2024-11-15
Allow langs to specify custom on-thread-start functions

Sushisource:
Needed for Python custom slot suppliers


cretz (approved):
Hrmm, I admit I'm curious here how this will be used in Python

PR #842, opened 2024-11-18
Ensure all permits are released before shutdown resolves

Sushisource:
Needed for custom slot supplier work

### DIFF
--- a/core/src/worker/activities/local_activities.rs
+++ b/core/src/worker/activities/local_activities.rs
@@ -1,11 +1,8 @@
 use crate::{
-    abstractions::{
-        dbg_panic, MeteredPermitDealer, OwnedMeteredSemPermit, PermitDealerContextData,
-        UsedMeteredSemPermit,
-    },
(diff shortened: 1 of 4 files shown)


core/src/worker/mod.rs
```diff
@@ -484,6 +515,8 @@ impl Worker {
         if let Some(acts) = self.at_task_mgr.as_ref() {
             acts.shutdown().await;
         }
+        // Wait for all permits to be released
+        self.all_permits_tracker.lock().await.all_done().await;
```
Sushisource, line 519: This is the real fix

core/src/worker/activities/local_activities.rs
```diff
@@ -740,6 +734,8 @@ impl LocalActivityManager {
         while !self.set_shutdown_complete_if_ready(&mut self.dat.lock()) {
             self.complete_notify.notified().await;
         }
+        // This makes sure we drop any permits that might be held inside the stream
+        self.rcvs.lock().await.inner = stream::empty().boxed();
```
Sushisource, line 738: This is the real fix

cretz (approved):

PR #843, opened 2024-11-18
Call ShutdownWorker API on worker shutdown

### LINKED ISSUE #822: [Feature Request] Call ShutdownWorker API on worker shutdown

yuandrew:
## What was changed

Added a new call to ShutdownWorker that was recently implemented in the server.

## Why?

New server feature already implemented in Go and Java, implementing in core to match.

## Checklist

1. Closes #822

2. How was this tested:

Add tests to verify when API is called


Sushisource (commented):
Just a few minor things

Sushisource (approved):
Sweet, thanks!

core/src/worker/mod.rs
```diff
@@ -502,6 +502,10 @@ impl Worker {
     /// completed
     async fn shutdown(&self) {
         self.initiate_shutdown();
+        if let Some(name) = self.workflows.get_sticky_queue_name() {
+            // This is a best effort call and we can still shutdown the worker if it fails
+            let _ = self.client.shutdown_worker(name).await;
```
cretz, line 507: I think we should warn if this is any `Err` except `Unavailable`

Issue #459, opened 2023-01-03
[Feature Request] Emit larger `workflow_task_schedule_to_start_latency` bucket sizes


lorensr:
### Is your feature request related to a problem? Please describe.

> We noticed the metric caps out at 10 seconds. If we start a large batch workflows we are generally fine waiting for them to process for 10-15 minutes. Given the 10 second maximum it’s pretty hard to tell how busy the workflow queue is.

(small number of workflow task slots)

cretz:
I have changed the title to reflect the general desire to customize bucket sizes. Unsure if this can be done on a per metric basis or just in general for all durations/histograms.

punit-kulal:
Also, if we are to record our own metrics, even those use the default buckets, which are not always ideal

cbrewster:
It would be nice to override this for the activity execution latency metrics. Right now it caps out at 1 minute, but we have some activities that take longer than that.

PR #844, opened 2024-11-20
Custom metric buckets

Sushisource:
## What was changed
Allow configuring custom buckets for histogram metrics

## Why?
Oft-requested feature

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/459

2. How was this tested:
Added tests

3. Any docs updates needed?

I'm looking at passing buckets in with user-created histograms, but my initial inclination is that that may not actually make sense since the otel API separates how aggregation happens from the instrument definition, so we would need to do some odd contortions to take the buckets from the histogram definition and somehow apply that to the overall meter provider definition.

Ultimately, it's pretty easy for the user to just use a constant for their histogram name, and then register the buckets for that in the telemetry options.

### DIFF
--- a/core-api/src/telemetry.rs
+++ b/core-api/src/telemetry.rs
@@ -81,11 +81,11 @@ pub struct PrometheusExporterOptions {
     #[builder(default)]
     pub global_tags: HashMap<String, String>,
     /// If set true, all counters will include a "_total" suffix
-    #[builder(default = "false")]
+    #[builder(default)]
     pub counters_total_suffix: bool,
     /// If set true, all histograms will include the unit in their name as a suffix.
     /// Ex: "_milliseconds".
-    #[builder(default = "false")]
+    #[builder(default)]
     pub unit_suffix: bool,
     /// If set to true, use f64 seconds for durations instead of u64 milliseconds
     #[builder(default)]
--- a/core-api/src/telemetry.rs
+++ b/core-api/src/telemetry.rs
@@ -68,6 +68,9 @@ pub struct OtelCollectorOptions {
     /// If set to true, use f64 seconds for durations instead of u64 milliseconds
     #[builder(default)]
     pub use_seconds_for_durations: bool,
+    /// Overrides for histogram buckets. Units depend on the value of `use_seconds_for_durations`.
+    #[builder(default)]
+    pub histogram_bucket_overrides: HistogramBucketOverrides,
 }
 
 /// Options for exporting metrics to Prometheus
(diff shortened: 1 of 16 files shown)


core-api/src/telemetry.rs
```diff
@@ -78,15 +81,33 @@ pub struct PrometheusExporterOptions {
...
+    /// Overrides where the key is the metric name and the value is the list of bucket boundaries.
+    /// The metric name will apply regardless of name prefixing, if any. IE: the name acts like
+    /// `*metric_name`.
+    ///
+    /// The string names of core's built-in histogram metrics are publicly available on the
+    /// `core::telemetry` module and the `client` crate.
+    ///
+    /// See [here](https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/metrics/enum.Aggregation.html#variant.ExplicitBucketHistogram.field.boundaries)
+    /// for the exact meaning of boundaries.
+    pub overrides: HashMap<&'static str, Vec<f64>>,
```
cretz, line 110: I am not sure this can be `&'static` here since it comes from users on lang side. It may have to be `String`.

core-api/src/telemetry.rs
```diff
@@ -78,15 +81,33 @@ pub struct PrometheusExporterOptions {
...
+    #[builder(default)]
+    pub histogram_bucket_overrides: HistogramBucketOverrides,
+}
+
+/// Allows overriding the buckets used by histogram metrics
+#[derive(Debug, Clone, Default)]
+pub struct HistogramBucketOverrides {
+    /// Overrides where the key is the metric name and the value is the list of bucket boundaries.
+    /// The metric name will apply regardless of name prefixing, if any. IE: the name acts like
+    /// `*metric_name`.
```
cretz, line 103: Does this wildcard thing only apply to OTel or does it apply to Prometheus too?

core-api/src/telemetry.rs
cretz, line 110: I was going to suggest we also provide the configured buckets as part of buffered/custom metric to the user, but that may not be needed.

cretz (commented):
I'm ok with requiring named bucket configuration to be done at metric init time instead of histogram creation time (at least at first and can wait if/until it becomes a real issue for users).

core-api/src/telemetry.rs
Sushisource: Yeah that's what the PR description is about. It's just extremely annoying to do

core-api/src/telemetry.rs
Sushisource: Both

core-api/src/telemetry.rs
Sushisource: Duh, good point, thanks

cretz: To clarify, here I am saying it may be neat to have the `MetricEvent::Create` buffered metric event include the buckets for those writing generic buffered metric impls, not necessarily concerned about the creation. But I understand that is a bit difficult too.

cretz (approved):

core-api/src/telemetry.rs
Sushisource: Yeah, it's the same thing. The problem is you define these `Views` at meter creation time, and instrument (metric) creation time is too late. So, not just annoying really but AFAICT may not even be possible.

Issue #836, opened 2024-10-29
[Feature Request] Evict cached workflows on workflow complete


cretz:
### Describe the solution you'd like

Like Go and Java, Core should evict cached workflows when server accepts workflow completion. Since post-close queries don't use the sticky queue anyways, Core was saving lang some steps, but full history was being downloaded anyways and it was just luck if the query landed on the same worker as the cache anyways. It was decided that we might as well save users memory in the common case. After discussion, there was not enough concern for reliance on existing behavior to need to have an option.

This also means Core needs to support the `temporal_sticky_cache_total_forced_eviction` metric now like Go and JAva.

aezomz:
Hi Cretz, just want to understand. For current behavior, eviction only happens when the memory runs out? or after hitting MAX_CACHED_WORKFLOWS ?

Also issues like this in this repo, is it normally for Temporal maintainers to keep track and pick up?

cretz:
With current behavior, eviction only happens if the spot is needed (so the cache size never decreases). We have this tracked on our near-term backlog.

PR #847, opened 2024-11-22
Evict workflows when they complete

Sushisource:
## What was changed
Workflows now are evicted if they emit a terminal command and there is no more work to be done (no already pending queries, for example).

## Why?
Same behavior as Go/Java. Reduce memory usage for most scenarios, since not everyone queries closed workflows often.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/836

2. How was this tested:
Existing tests & new

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/workflow/workflow_stream.rs
+++ b/core/src/worker/workflow/workflow_stream.rs
@@ -505,7 +505,7 @@ impl WFStream {
         let num_existing_evictions = self
             .runs
             .runs_lru_order()
-            .filter(|(_, h)| h.is_trying_to_evict())
+            .filter(|(_, h)| h.trying_to_evict().is_some())
             .count();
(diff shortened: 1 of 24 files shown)


core/src/worker/workflow/workflow_stream.rs
```diff
@@ -378,6 +377,20 @@ impl WFStream {
...
+                {
+                    res = rh
+                        .request_eviction(RequestEvictMsg {
+                            run_id: run_id.to_string(),
+                            message: "Workflow completed".to_string(),
+                            reason: EvictionReason::WorkflowExecutionEnding,
+                            auto_reply_fail_tt: None,
+                        })
+                        .into_run_update_resp()
+                }
```
Sushisource, line 393: This is the actual implementation. Everything else is pretty much just test fixing or cleanup / refactor.

cretz (approved):

PR #852, opened 2024-12-10
Have Opentelemetry error handler to use tracing::error

### LINKED ISSUE #810: [Feature Request] Have OpenTelemetry failure logs go to logger instead of stderr

Unfortunately at https://github.com/open-telemetry/opentelemetry-rust/blob/976bc54dba564afdc8fd7de51a8feb123f44ecdb/opentelemetry/src/global/error_handler.rs#L62, OTel just dumps errors to stderr unless there is a global error handler set. We should set it and use `tracing::error!` or similar.

yuandrew:
## What was changed

Set OpenTelemetry global setting to log to `tracing::error`

## Checklist

1. Closes #810

2. How was this tested:

3. Any docs updates needed?


Sushisource (approved):

core/src/telemetry/otel.rs
```diff
@@ -120,6 +120,9 @@ pub(super) fn augment_meter_provider_with_defaults(
 pub fn build_otlp_metric_exporter(
     opts: OtelCollectorOptions,
 ) -> Result<CoreOtelMeter, anyhow::Error> {
+    global::set_error_handler(|err| {
+        tracing::error!("{}", err);
+    })?;
```
cretz, line 125: I wonder if there is a better more global place to set this instead of on each `build_otlp_metric_exporter` call? Unsure if we have global metric initialization code.

core/src/telemetry/otel.rs
```diff
@@ -120,6 +120,9 @@ pub(super) fn augment_meter_provider_with_defaults(
 pub fn build_otlp_metric_exporter(
     opts: OtelCollectorOptions,
 ) -> Result<CoreOtelMeter, anyhow::Error> {
+    global::set_error_handler(|err| {
```
cretz, line 123: Note this was removed recently in 0.27 (ref https://github.com/open-telemetry/opentelemetry-rust/pull/2260), so we either need to say in the dep that we are 0.26 (not sure if you can say `< 0.27`) or look at that migration guide and see if just upgrading to 0.27 is enough to get proper logs here.

yuandrew: I believe the change only changes the behavior on how you can override their default logging behavior, we would still need to override their behavior, just in a different way now. I'm not sure it's worth upgrading to 0.27 over this.

cretz: I think my concern is whether _other_ people want to upgrade to 0.27+. I am not sure what Cargo's auto-upgrade logic is for zero-ver. Obviously _one day_ we will need to as well even if that day is not today (can't stay on 0.26 forever).

yuandrew: Who are the "other" people mentioned here? customers? 

>  I am not sure what Cargo's auto-upgrade logic is for zero-ver. 

I assume cargo wouldn't auto-upgrade from 0.26 to 0.27, since there are a number of breaking changes with that update.

> Obviously one day we will need to as well even if that day is not today (can't stay on 0.26 forever).

Agreed, I'm just reluctant to upgrade today 😅 but now that I've written a test, that future upgrade should be easier

cretz: > Who are the "other" people mentioned here? customers?

Nope just us internally

> Agreed, I'm just reluctant to upgrade today

Works for me

cretz (approved):

PR #853, opened 2024-12-10
Add OTel HTTP support

### LINKED ISSUE #820: [Feature Request] OpenTelemetry HTTP API support

Currently we only support OTLP over gRPC, but users like https://github.com/temporalio/sdk-python/issues/647 use systems that (surprisingly) don't support OTLP gRPC.

yuandrew:
## What was changed

1. Added HTTP support for OTel.
2. Added new testing support in CI to test otel/prometheus running in a separate process.
    - must specify `DOCKER_PROMETHEUS_RUNNING` env variable to indicate we expect docker/prometheus to be running
    - test must start with `docker_` so we can easily filter out which tests have this dependency
4. Updated otel-collector-*.yaml to stop using deprecated field, see #854.

## Why?

Users have requested HTTP support, on top of existing GRPC support

## Checklist

1. Closes #820

2. How was this tested:

Added test to validate prometheus running in a separate docker container can pick up metrics via HTTP and GRPC

5. Any docs updates needed?


cretz (commented):
Looks great from my POV, but saving approval for @Sushisource

tests/integ_tests/metrics_tests.rs
```diff
@@ -651,6 +654,81 @@ async fn request_fail_codes_otel() {
...
+    let client = starter.get_client().await;
+    client.list_namespaces().await.unwrap();
+
+    // Give Prometheus time to scrape metrics
+    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+
+    // Query Prometheus API for metrics
+    let client = reqwest::Client::new();
+    let query = format!("temporal_sdk_{}num_pollers", test_uid.clone());
+    println!("[query]: {:?}", query);
```
Sushisource, line 711: More leftover

tests/integ_tests/metrics_tests.rs
```diff
@@ -651,6 +654,81 @@ async fn request_fail_codes_otel() {
...
+            CompleteWorkflowExecution { result: None }.into(),
+        ))
+        .await
+        .unwrap();
+
+    let client = starter.get_client().await;
+    client.list_namespaces().await.unwrap();
+
+    // Give Prometheus time to scrape metrics
+    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
Sushisource, line 706: I wonder if you can hit some endpoint to force a scrape, but, since this is only in CI it's fine if it takes a bit.

tests/integ_tests/metrics_tests.rs
```diff
@@ -651,6 +654,81 @@ async fn request_fail_codes_otel() {
...
+        .unwrap()
+        .json::<serde_json::Value>()
+        .await
+        .unwrap();
+
+    println!("[response]: {:?}", response["data"]);
+
+    // Validate the Prometheus response
+    if let Some(data) = response["data"]["result"].as_array() {
+        assert!(!data.is_empty(), "No metrics found for query: {test_uid}");
```
Sushisource, line 726: Probably a good idea to assert some specific metrics exist too

.github/workflows/per-pr.yml
```diff
@@ -84,3 +84,31 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: 1.80.0
+      - name: Install protoc
+        uses: arduino/setup-protoc@v3
+        with:
+          # TODO: Upgrade proto once https://github.com/arduino/setup-protoc/issues/99 is fixed
+          version: '23.x'
```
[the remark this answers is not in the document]
cretz: I think it does. That action I think is basically abandoned, we need to find a new/better way to install protoc (not just here but in many repos).

tests/integ_tests/metrics_tests.rs
yuandrew: Since this is only CI, is there any reason I shouldn't just set the scrape interval to 1s, and then this sleep can go down to 2s?

docker/docker-compose-ci.yaml
```diff
@@ -0,0 +1,25 @@
+version: '3.5'
+
+services:
+  otel-collector:
```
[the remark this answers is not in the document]
yuandrew: Looks like docker supports importing env variables, but i don't think env variables are able to change the flow of what containers get stood up

core-api/src/telemetry.rs
```diff
@@ -72,8 +72,8 @@ pub struct OtelCollectorOptions {
     #[builder(default)]
     pub histogram_bucket_overrides: HistogramBucketOverrides,
     /// Protocol to use for communication with the collector
-    #[builder(default = "OtlpProtocl::Grpc")]
-    pub protocol: OtlpProtocl,
```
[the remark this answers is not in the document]
yuandrew: Good to know! 

Looks like I made the same typo mistake in my own test trying to spell the same word 😆

tests/integ_tests/metrics_tests.rs
Sushisource: Yeah, you can do that too, but I thought that was defined in the same config file that gets used by the non-ci version of the docker compose. If not, then great.

Sushisource (approved):
Sweet, thanks.

This `queries_tests::query_after_execution_complete::case_2_with_eviction` is flaking a lot lately. Looks like I need to fix that.

PR #854, opened 2024-12-11
opentelemetry-collector: Switch from deprecated logging to debug

yuandrew:
## What was changed

`logging` is no longer supported, need to switch to `debug`

## Why?

https://github.com/open-telemetry/opentelemetry-collector/issues/11337

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Sushisource (approved):
This can also just go with your other PR, but, fine with merging it separately if you want too

PR #857, opened 2024-12-19
Nexus WF Machines & Lang Support

Sushisource:
> [!NOTE]
> Note that this isn't going into main, but a nexus branch 

## What was changed
Add ability for langs to perform nexus operations from workflows

## Why?
Nexus is cool yo

## Checklist

1. Closes 

2. How was this tested:
Added integration tests

3. Any docs updates needed?

### DIFF
--- a/sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
+++ b/sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
@@ -125,6 +125,8 @@ message WorkflowActivationJob {
         ResolveNexusOperationStart resolve_nexus_operation_start = 15;
         // A nexus operation resolved.
         ResolveNexusOperation resolve_nexus_operation = 16;
+        // A request to cancel a nexus operation resolved.
+        ResolveCancelNexusOperation resolve_cancel_nexus_operation = 17;
         // Remove the workflow identified by the [WorkflowActivation] containing this job from the
         // cache after performing the activation. It is guaranteed that this will be the only job
         // in the activation if present.
@@ -327,12 +329,15 @@ message ResolveNexusOperationStart {
     uint32 seq = 1;
     oneof status {
         // The operation started asynchronously. Contains an ID that can be used to perform
-        // operations on the started operation by, ex, clients.
+        // operations on the started operation by, ex, clients. A `ResolveNexusOperation` job will
+        // follow at some point.
         string operation_id = 2;
         // The operation "started" but only because it's also already resolved. A
         // `ResolveNexusOperation` job will be in the same activation.
         google.protobuf.Empty started_sync = 3;
         // The operation was cancelled before it was ever sent to server (same WFT).
+        // Note that core will still send a `ResolveNexusOperation` job in the same activation, so
+        // there does not need to be an exceptional case for this in lang.
         temporal.api.failure.v1.Failure cancelled_before_start = 4;
     }
 }
@@ -341,6 +346,11 @@ message ResolveNexusOperation {
     // Sequence number as provided by lang in the corresponding ScheduleNexusOperation command
     uint32 seq = 1;
     nexus.NexusOperationResult result = 2;
+}
+
+message ResolveCancelNexusOperation {
+    // Sequence number as provided by lang in the corresponding RequestCancelNexusOperation command
+    uint32 seq = 1;
 }
 
 message RemoveFromCache {
--- a/sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
+++ b/sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
@@ -125,8 +125,6 @@ message WorkflowActivationJob {
         ResolveNexusOperationStart resolve_nexus_operation_start = 15;
         // A nexus operation resolved.
         ResolveNexusOperation resolve_nexus_operation = 16;
-        // A request to cancel a nexus operation resolved.
-        ResolveCancelNexusOperation resolve_cancel_nexus_operation = 17;
         // Remove the workflow identified by the [WorkflowActivation] containing this job from the
         // cache after performing the activation. It is guaranteed that this will be the only job
         // in the activation if present.
@@ -332,9 +330,9 @@ message ResolveNexusOperationStart {
         // operations on the started operation by, ex, clients. A `ResolveNexusOperation` job will
         // follow at some point.
         string operation_id = 2;
-        // The operation "started" but only because it's also already resolved. A
+        // If true the operation "started" but only because it's also already resolved. A
         // `ResolveNexusOperation` job will be in the same activation.
-        google.protobuf.Empty started_sync = 3;
+        bool started_sync = 3;
         // The operation was cancelled before it was ever sent to server (same WFT).
         // Note that core will still send a `ResolveNexusOperation` job in the same activation, so
         // there does not need to be an exceptional case for this in lang.
@@ -346,11 +344,6 @@ message ResolveNexusOperation {
     // Sequence number as provided by lang in the corresponding ScheduleNexusOperation command
     uint32 seq = 1;
     nexus.NexusOperationResult result = 2;
-}
-
-message ResolveCancelNexusOperation {
-    // Sequence number as provided by lang in the corresponding RequestCancelNexusOperation command
-    uint32 seq = 1;
 }
 
 message RemoveFromCache {
--- a/sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
+++ b/sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
@@ -9,6 +9,7 @@ option ruby_package = "Temporalio::Internal::Bridge::Api::WorkflowActivation";
 
 import "google/protobuf/timestamp.proto";
 import "google/protobuf/duration.proto";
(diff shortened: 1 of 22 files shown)


sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
```diff
@@ -316,6 +324,35 @@ message DoUpdate {
...
+    // Sequence number as provided by lang in the corresponding ScheduleNexusOperation command
+    uint32 seq = 1;
+    oneof status {
+        // The operation started asynchronously. Contains an ID that can be used to perform
+        // operations on the started operation by, ex, clients. A `ResolveNexusOperation` job will
+        // follow at some point.
+        string operation_id = 2;
+        // The operation "started" but only because it's also already resolved. A
+        // `ResolveNexusOperation` job will be in the same activation.
+        google.protobuf.Empty started_sync = 3;
```
cretz, line 337: Can we make this a bool so we don't have to bring in "empty"? Since we have source compat with lang on protos, we are safe changing this to a full message later, we don't need to be concerned with wire compat IMO.

Arguably we could just remove this. I can't see the situation where lang cares whether start and complete are in the same activation or not (same for child workflows). Probably same for the field below, but I guess too many fields doesn't technically hurt.

sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
```diff
@@ -119,6 +121,12 @@ message WorkflowActivationJob {
         ResolveRequestCancelExternalWorkflow resolve_request_cancel_external_workflow = 13;
         // A request to handle a workflow update.
         DoUpdate do_update = 14;
+        // A nexus operation started.
+        ResolveNexusOperationStart resolve_nexus_operation_start = 15;
+        // A nexus operation resolved.
+        ResolveNexusOperation resolve_nexus_operation = 16;
+        // A request to cancel a nexus operation resolved.
+        ResolveCancelNexusOperation resolve_cancel_nexus_operation = 17;
```
cretz, line 129: Hrmm, I guess this doesn't mean a Nexus operation was cancelled, but rather that the cancel was sent? I do wonder what situation we need this as an awaitable thing that can come back. I would similarly not expect a `ResolveCancelChildWorkflowExecution` since it effectively is always expected to succeed. But I do see the `NexusOperationCancelRequestedEventAttributes` event. Hrmm.

The issue from a lang POV is that cancellation of things you start often does not have a path for "resolving" of the cancellation alone. Because, unlike external workflow cancellation, we often use cancellation token like things when starting things, we don't expose a fallible/awaitable cancellation. If we supported cancelling Nexus operations you didn't start, definitely would be worth the extra wait.

sdk-core-protos/protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto
```diff
@@ -337,4 +340,46 @@ message UpdateResponse {
...
+    uint32 seq = 1;
+}
+
+// Request cancellation of a nexus operation that has already started
+message RequestCancelStartedNexusOperation {
+    // A sequence number specifically for this cancel request
+    uint32 seq = 1;
+    // The sequence number of the `ScheduleNexusOperation` command being cancelled
+    uint32 schedule_seq = 2;
+}
```
cretz, line 385: Can Core's internals handle the difference between these two and just expose a single cancel command instead of having lang differentiate the status before picking which of these commands to use?

sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
Sushisource: Sure, it can be a bool

sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
Sushisource: They need to be awaitable. Support will be added later for ensuring cancel was received, so we need to have them be awaitable. This is a lot like how child workflows get cancelled via the external workflow cancel command.

And yeah, cancelling ones you didn't start likely will be supported at some point too.

sdk-core-protos/protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto
Sushisource: Not really because you need a different sequence number for the cancel resolution, which is needed b/c of the reasons I mentioned in the other comment.

sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
Sushisource: As for the cancel token thing, yeah, in this case you're just going to end up internally seeing the resolution and maybe not showing it to the user. But, likely the type that represents a started nexus operation will also have an explicit awaitable cancel method.

cretz: > They need to be awaitable

I don't see how this can be done with cancellation via something you pass in to start often using language-native cancellation mechanisms that are not awaitable.

> This is a lot like how child workflows get cancelled via the external workflow cancel command.

But there is no separate activation for awaiting a child cancel received right? Not sure we have ever had to wait for a cancel to be received by something we've started, be it activities, child workflows, etc.

> cancelling ones you didn't start likely will be supported at some point too.

I think, just as lang would not use cancellation token approach but instead an explicit awaitable cancel call, it would deserve a separate command/activation pair.

Sushisource: > I don't see how this can be done with cancellation via something you pass in to start often using language-native cancellation mechanisms that are not awaitable.

They are conceptually awaitable. That doesn't mean the user literally _has_ to wait on them. Like I said you can just track the sequence number and resolution internally for cancel-token-like cases.

> it would deserve a separate command/activation pair.

That is exactly what's happening. You send the `RequestCancelStartedNexusOperation` command, and then you wait for that to be resolved.

> I wonder why we even care about the resolution from lang side then. Not sure it has any lang value. I'd have to see a lang sample that shows where it is even able to be used I think.

This is present in the integ tests - you can see how the Rust SDK is doing this.

@bergundy Suggested I make the cancel explicitly awaitable because of what I mentioned re:
> Support will be added later for ensuring cancel was received

Personally, I don't hugely care. So I'll let him make the case for that if he feels it's important to deliver cancel resolution separately from operation resolution. I would be OK internalizing the after-started cancel, but probably the command will always need to exist (or will have to be reintroduced) when we support cancelling operations you didn't start.

bergundy: To implement `WAIT_CANCELLATION_REQUESTED`, you'll need to notify lang when the cancelation was requested.

cretz: > To implement WAIT_CANCELLATION_REQUESTED, you'll need to notify lang when the cancelation was requested.

How is this implemented with child workflows today? If we get to external Nexus cancel operation, I agree that has to be a request/response, but for start-in-this-workflow items (children, activities, and Nexus operations), cancellation is a one-way operation. Languages don't even have a way to wait for result.

Can either of y'all show what waiting on a Nexus operation cancel looks like in any of our SDK languages (not to be confused with what waiting on a result looks like). And if 0 SDK languages can support this activation anyways, why would a language ever even handle it? And if 0 SDK languages handle this resulting activation job, why have it effectively as a dead message in the proto?

bergundy: Ah I see what you mean, you don't wait for the cancelation itself in lang, we're expecting core to resolve the nexus operation as canceled with `WAIT_CANCELLATION_REQUESTED`, it's not the result of the cancel command.
I definitely see external nexus operations being a thing and allowing waiting on its cancelation, that may be a separate thing though.

cretz: So maybe we do want a "Nexus cancellation type"? How does one wait for a Nexus cancellation in Go today? Go child and activity options both have `WaitForCancellation` booleans, should we do the same for Nexus operation options? And for Core-based SDKs, should we have a cancellation type like child and activity have?

bergundy: Yes we need to add a cancellation type for Nexus too. To fully support it, we need to add a history event to notify the SDK that a cancelation request has been delivered (or permanently failed).

Sushisource: OK, I'll have operations you started have a non-awaitable cancel operation then I suppose. That's not quite what you mentioned @bergundy when I brought it up in the Nexus channel earlier:

> Make it an awaitable, we are going to add support for waiting for cancellation to be delivered in the future.
It’s basically like child workflow cancellation

But, I can change it back. I might just keep the external cancel command around if we're going to bring that back later.

cretz (approved):

Sushisource:
@Quinn-With-Two-Ns I'm merging this so I can work on the next part that will go into this branch, but feel free to leave any comments here if you like.

PR #859, opened 2025-01-08
Fix DB filename argument for ephemeral server

iand675:
## What was changed
This is a relatively simple fix: the ephemeral server code isn't using the correct option name for setting the SQLite filename argument.

## Why?
I didn't check old versions of the ephemeral server, but I am at least confident that for `temporal version 1.0.0 (Server 1.24.2, UI 2.28.0)`, the current flag is incorrect.

## Checklist

1. Closes 

2. How was this tested:
Ran the updated SDK against our internal codebase and validated that the flag worked. I can add a regression test if y'all like, but wasn't sure if it was worth it here.

3. Any docs updates needed?


turnage (approved):

Sushisource (approved):
Thanks!

PR #860, opened 2025-01-10
Nexus Tasks Core->Lang

Sushisource:
> [!WARNING]
> Can't merge this until CLI gets updated. Depends on changes in server 1.26 

## What was changed
Add nexus task delivery and tracking

## Why?
Part of Nexus functionality

## Checklist

1. Closes 

2. How was this tested:
Integ tests

3. Any docs updates needed?


core-api/src/errors.rs
```diff
@@ -67,6 +55,21 @@ pub enum CompleteActivityError {
     },
 }
 
+/// Errors thrown by [crate::Worker::complete_nexus_task]
+#[derive(thiserror::Error, Debug)]
+pub enum CompleteNexusError {
+    /// Lang SDK sent us a malformed nexus completion. This likely means a bug in the lang sdk.
+    #[error("Lang SDK sent us a malformed nexus completion: {reason}")]
+    MalformeNexusCompletion {
```
cretz, line 63:
```suggestion
    MalformedNexusCompletion {
```

cretz (approved):
Admittedly only reviewed the two new exposed worker calls and the local core protos. Both LGTM.

core-api/src/errors.rs
Sushisource: Woopsie. Thanks!

Quinn-With-Two-Ns (approved):

PR #864, opened 2025-01-22
Write sdk name & version to WFT completes

Sushisource:
## What was changed
Title. Automatically derived from client settings, no lang-visible change needed.

## Why?
Parity, useful for investigations

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/588

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/internal_flags.rs
+++ b/core/src/internal_flags.rs
@@ -175,12 +175,16 @@ impl InternalFlags {
                     .collect();
                 core.extend(core_since_last_complete.iter());
                 lang.extend(lang_since_last_complete.iter());
-                let (sdk_name, sdk_version) =
-                    if last_sdk_name != sdk_name || last_sdk_version != sdk_version {
-                        (sdk_name.clone(), sdk_version.clone())
-                    } else {
-                        ("".to_owned(), "".to_owned())
-                    };
+                let sdk_name = if last_sdk_name != sdk_name {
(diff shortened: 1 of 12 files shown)


core/src/internal_flags.rs
```diff
@@ -146,11 +175,17 @@ impl InternalFlags {
                     .collect();
                 core.extend(core_since_last_complete.iter());
                 lang.extend(lang_since_last_complete.iter());
+                let (sdk_name, sdk_version) =
+                    if last_sdk_name != sdk_name || last_sdk_version != sdk_version {
+                        (sdk_name.clone(), sdk_version.clone())
+                    } else {
+                        ("".to_owned(), "".to_owned())
+                    };
```
cretz, line 183: I think other SDKs (and mentioned at https://github.com/temporalio/features/issues/321) compare name and version independently, meaning if the version upgraded but not the SDK, only the version is written.

Sushisource: Well, that wouldn't be totally independent, since it wouldn't make sense the other way around, but, yeah can do

cretz: Thanks! I think _technically_ the other SDK impls are independent meaning in some strange world you are able to switch from Go to Java but keep the version number, the version is unchanged.

cretz (approved):

PR #870, opened 2025-02-01
Client refactoring

Sushisource:
## What was changed
Some client cleanup I needed to do while working on poller scaling

## Why?
Paying down some debt & better organization

## Checklist

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -47,6 +47,7 @@ path = "../core-api"
 [dev-dependencies]
 assert_matches = "1"
 mockall = "0.13"
+rstest = "0.24"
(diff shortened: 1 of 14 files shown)


cretz (approved):
Does this have any external effect on lang? If so, a quick note on what that is could help

Sushisource:
> Does this have any external effect on lang? If so, a quick note on what that is could help

Some publicly visible types changed, but not ones that lang would be using, and they changed in ways that should be backwards compat or very simple to deal with (import another trait)

PR #874, opened 2025-02-08
Poller scaling

Sushisource:
I intend to add a few unit tests too, and some simpler integ tests once server is actually released with this. This PR will have to sit and wait for server to release with the changes https://github.com/temporalio/temporal/pull/7300

## What was changed
Read and handle poller scaling decisions from server

## Why?
 Part of the worker management effort to simplify configuration of workers for users.

## Checklist

1. Closes 

2. How was this tested:
Big manual tests + integ tests (to come)

3. Any docs updates needed?
Doc updates will come as part of adding to all SDKs


core/src/pollers/poll_buffer.rs
```diff
@@ -186,12 +379,218 @@ where
...
+                    tokio::select! {
+                        _ = interval.tick() => {}
+                        _ = shutdown.cancelled() => { break; }
+                    }
+                    let ingested = rhc.ingested_this_period.swap(0, Ordering::Relaxed);
+                    let ingested_last = rhc.ingested_last_period.swap(ingested, Ordering::Relaxed);
+                    rhc.scale_up_allowed
+                        .store(ingested_last >= ingested, Ordering::Relaxed);
+                }
+            }))
```
cretz, line 474: So to confirm, the algorithm is:

Server returns amount of pollers to scale up (or down if negative) on each poll response, and SDK respects that decision so long as it's bounded by min/max, and in the case of scale up, there were at least as many polls accepted this 100ms period as last 100ms?

cretz (approved):
LGTM but didn't get into the details. I think an English description of what's happening could help. I tried to guess at one in on the comments. But since the default behavior is unchanged for users, no problem.

I do think we should consider holding off (or at least holding off lang side) until a server is released that can even take advantage of this. We will want _some_ end to end _automated_ test somewhere that just proves poller auto scaling works (often we can just do this in lang in a smoke-test kinda integration test just to confirm).

Sushisource:
> LGTM but didn't get into the details. I think an English description of what's happening could help. I tried to guess at one in on the comments. But since the default behavior is unchanged for users, no problem.
> 
> I do think we should consider holding off (or at least holding off lang side) until a server is released that can even take advantage of this. We will want _some_ end to end _automated_ test somewhere that just proves poller auto scaling works (often we can just do this in lang in a smoke-test kinda integration test just to confirm).

Yeah, I'm happy to wait for a server release so that we have an integ test.

core/src/pollers/poll_buffer.rs
Sushisource: I'm glad you looked at this again because it sort of obviously makes no sense. I was wondering why my testing wasn't showing good restriction of overshooting like I know it did at one point, and I had tried like a bazillion different methods but I knew this worked and it was simple, so I went back to it -- but results were inconsistent.

Somehow this comparison got flipped at some point. I've changed it back and the results are much better again.

Sushisource: Honestly though I'm tempted to get rid of this entirely. My tests show it can still be "defeated" semi often by rapid scale ups, and ingestion is indeed going up so there's not really any reason to say no, and then you end up with a bunch of polls sitting there once the backlog clears.

Smoothing out the calculation so that the shorter timeout gets set more reliably might have more value

cretz: Agreed. I am always a bit wary when I see hardcoded time expectations like 100ms unless that's a good number arrived by testing or something (which maybe it is). Maybe a more naive "do not make poller count changes more frequently than X interval" can help, but I don't understand the details and haven't run the tests.

Issue #708, opened 2024-03-20
[Bug] Core doesn't expose the `cause` from `WorkflowExecutionCancelRequestedEventAttributes`


bergundy:
Event has a reason:

https://github.com/temporalio/api/blob/6606687fb5ca8d899fe3e93313d53d9c7ed2a379/temporal/api/history/v1/message.proto#L403

Activation Job has details:

https://github.com/temporalio/sdk-core/blob/4840a9c570fa3388e190e20f01f61065d0b7e965/sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto#L242

Must have been a mix up with `WorkflowExecutionCanceledEventAttributes`:

https://github.com/temporalio/api/blob/6606687fb5ca8d899fe3e93313d53d9c7ed2a379/temporal/api/history/v1/message.proto#L414

PR #875, opened 2025-02-11
Expose workflow cancel cause

Sushisource:
## What was changed
Expose cancellation cause from user requests

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/708

2. How was this tested:
Upgraded integ test

3. Any docs updates needed?


sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
cretz, line 1: I think this may be needed on the child/external cancel commands outbound from workflow too. And we should decide if we want to call this "cancel" or "reason" (API is inconsistent but we don't have to be)

Sushisource: Added

cretz (approved):

PR #877, opened 2025-02-12
disable tracing-log feature for tracing-subscriber

konradsz:
## What was changed
Disable the `tracing-log` feature for `tracing-subscriber` by turning off [default features](https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/Cargo.toml#L27). The `ansi` feature was explicitly listed because it's used in the code.

## Why?
I am trying to use `core` crate from Temporal SDK in a Rust application that already uses `tracing-subscriber`. However, it fails to initialize the subscriber because `core` depends on `tracing-subscriber` with `tracing-log` feature enabled. This behavior is documented [here](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/util/trait.SubscriberInitExt.html#method.try_init):

> This method panics if a global default subscriber has already been set, or if a log logger has already been set (when the “tracing-log” feature is enabled).

In a perfect world, `core` would not depend on `tracing-subscribers` since it is a [library](https://github.com/tokio-rs/tracing?tab=readme-ov-file#in-libraries), but I propose this as an immediate fix.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -67,7 +67,7 @@ tokio-util = { version = "0.7", features = ["io", "io-util"] }
 tokio-stream = "0.1"
 tonic = { workspace = true, features = ["tls", "tls-roots"] }
 tracing = "0.1"
-tracing-subscriber = { version = "0.3", features = ["parking_lot", "env-filter", "registry"] }
+tracing-subscriber = { version = "0.3", default-features = false, features = ["parking_lot", "env-filter", "registry", "ansi"] }
 url = "2.2"
 uuid = { version = "1.1", features = ["v4"] }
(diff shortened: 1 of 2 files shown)


Sushisource:
Thanks for this @konradsz . I think this is left-over from when `log` was used internally too. 

And yes, I agree about
>In a perfect world, core would not depend on tracing-subscribers 

In principle, but Core is in a bit of an odd position where most consumers are using FFI and so a lot of the time logging needs to be enabled by the library itself too. Potentially we could introduce another crate that thinly layers over core by adding the tracing subscriber dep and maybe a handful of other things meant for non-Rust consumers, but that's for another day.

Thanks!

Sushisource (approved):

Issue #879, opened 2025-02-20
[Feature Request] Fix recent clippy warnings


cretz:
### Describe the solution you'd like

From a recent Ruby build:

```
error: useless conversion to the same type: `temporal::api::common::v1::RetryPolicy`
    --> ext/sdk-core/sdk-core-protos/src/lib.rs:1266:53
     |
1266 |                         retry_policy: r.retry_policy.map(Into::into),
     |                                                     ^^^^^^^^^^^^^^^^ help: consider removing
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
     = note: `-D clippy::useless-conversion` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(clippy::useless_conversion)]`

error: useless conversion to the same type: `temporal::api::common::v1::Payload`
    --> ext/sdk-core/sdk-core-protos/src/lib.rs:1424:50
     |
1424 |                 Some(p) => p.payloads.into_iter().map(Into::into).collect(),
     |                                                  ^^^^^^^^^^^^^^^^ help: consider removing
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion

error: useless conversion to the same type: `temporal::api::common::v1::Payload`
    --> ext/sdk-core/sdk-core-protos/src/lib.rs:1442:36
     |
1442 |                     payloads: iterd.map(Into::into).collect(),
     |                                    ^^^^^^^^^^^^^^^^ help: consider removing
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
```

PR #880, opened 2025-02-21
Rust / Packages upgrade

Sushisource:
## What was changed
Upgrade rust, and crates, ~and Otel (maybe)~

Problem is Otel has managed to break something. In this case, error logging (they also are logging some uninteresting junk at INFO level). AFAICT errors simply aren't logged while trying to export metrics. This happens in a new thread now and after wasting a bunch of time on this I gave up trying to figure out exactly why. I opened an [issue](https://github.com/open-telemetry/opentelemetry-rust/issues/2697) for them.

We can simply not upgrade otel for now, but we'll be stuck on that git revision of the prom exporter.

## Why?
Updates good.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/879

2. How was this tested:
Existing tests

3. Any docs updates needed?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ license-file = "LICENSE.txt"
 
 [workspace.dependencies]
 derive_builder = "0.20"
-derive_more = { version = "1.0", features = ["constructor", "display", "from", "into", "debug"] }
+derive_more = { version = "2.0", features = ["constructor", "display", "from", "into", "debug"] }
 thiserror = "2"
 tonic = "0.12"
 tonic-build = "0.12"
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -21,7 +21,7 @@ jobs:
           submodules: recursive
       - uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: 1.84.0
+          toolchain: 1.85.0
       - name: Install protoc
         uses: arduino/setup-protoc@v3
         with:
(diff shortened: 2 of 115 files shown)


Cargo.toml
```diff
@@ -8,11 +8,11 @@ license-file = "LICENSE.txt"
 
 [workspace.dependencies]
 derive_builder = "0.20"
-derive_more = { version = "1.0", features = ["constructor", "display", "from", "into", "debug"] }
+derive_more = { version = "2.0", features = ["constructor", "display", "from", "into", "debug"] }
 thiserror = "2"
 tonic = "0.12"
 tonic-build = "0.12"
-opentelemetry = { version = "0.26", features = ["metrics"] }
+opentelemetry = { version = "0.28", features = ["metrics"] }
```
cretz, line 15: So the options are: 1) don't upgrade `opentelemetry` and therefore stay on old `opentelemetry-prometheus`, 2) upgrade `opentelemetry` and start getting a lot of info-level log spam (assuming langs have enabled info level which I don't think they do by default but will by annoying for users that do)

Tough choice, but assuming https://github.com/open-telemetry/opentelemetry-rust/issues/2697 moves quickly and 0.29 is not too far off, maybe option 1 is best for a bit longer.

Sushisource:
Dropped the otel upgrades and left them in https://github.com/temporalio/sdk-core/tree/otel-upgrade

cretz (approved):

Issue #871, opened 2025-02-03
[Feature Request] Support priority annotations


Sushisource:
**Is your feature request related to a problem? Please describe.**
Add support for attaching priority to workflows/activities

**Describe the solution you'd like**
See https://github.com/temporalio/features/issues/593 for details

This is really just proto-passthrough from lang->server

PR #892, opened 2025-03-25
Priority field passthrough

Sushisource:
## What was changed
Add passing priorities to server

## Why?
Priorities! Yay!

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/871

2. How was this tested:
Added integ test

3. Any docs updates needed?


Sushisource:
Need to update the version of CLI CI is using so this will work, but, the code itself is working properly against that version locally.

yuandrew (approved):

Quinn-With-Two-Ns (approved):

PR #901, opened 2025-04-15
Fix sync started Nexus ops not coming across as such

Sushisource:


dandavison (approved):
Thanks! Confirmed this works as expected from Python.

PR #905, opened 2025-04-19
Handle metrics for benign application failures

### LINKED ISSUE #897: [Feature Request] Plumb `ApplicationErrorCategory` to lang SDKs
See corresponding features issue:
https://github.com/temporalio/features/issues/614

Core SDK needs to plumb `ApplicationErrorCategory` to lang SDKs and apply metrics behaviour based on the given `ApplicationErrorCategory` from lang (core handles metrics, lang handles logging)

THardy98:
## What was changed
Updated temporal api to get application error category, use the category to filter out benign application failures from failure metrics.

## Why?
Part of benign exceptions work

Closes #897

### DIFF
--- a/core/src/telemetry/metrics.rs
+++ b/core/src/telemetry/metrics.rs
@@ -594,12 +594,10 @@ pub(super) const TASK_SLOTS_USED_NAME: &str = "worker_task_slots_used";
 pub(super) const STICKY_CACHE_SIZE_NAME: &str = "sticky_cache_size";
 
 /// Track a failure metric if the failure is not a benign application failure.
-pub(crate) fn should_record_failure_metric(
-    failure: &Option<Failure>,
-) -> bool {
+pub(crate) fn should_record_failure_metric(failure: &Option<Failure>) -> bool {
     !failure
         .as_ref()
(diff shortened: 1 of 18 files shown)


cretz (approved):
LGTM, though may want to wait for @Sushisource approval

core/src/telemetry/metrics.rs
```diff
@@ -592,6 +593,19 @@ pub(super) const TASK_SLOTS_AVAILABLE_NAME: &str = "worker_task_slots_available"
...
+    metric_fn: impl FnOnce(),
+) {
+    let is_benign = failure
+        .as_ref()
+        .map_or(false, |f| f.is_benign_application_failure());
+    if !is_benign {
+        metric_fn();
+    }
+}
+
```
Sushisource, line 608: I don't think we really need this function - the closure passing is a bit performative. I think it'd be easier to call this at the callsite, or, if you want to save a few lines, you can make a trait that handles the option stuff an implement it for `Option<Failure>` and that way you can call `should_record_metric()` directly on the `Option<Failure>`

Sushisource (approved):
Nothing blocking but I think we can ditch the closure-passing

PR #910, opened 2025-04-29
Fix possibly incorrect `is_replaying` values when processing empty WFTs

Sushisource:
Currently, everything passes except some UTs that specifically test WFT boundaries involving histories with empty WFTs, which this fix changes.

I tried this change with Python, and everything passes. My intuition is that this is probably a compatible change, but we could always flag it to be sure. If anyone can come up with a counterexample please do. It's possible that this change can potentially move jobs from one activation to another (without changing ordering), but I don't think that matters in any way that actually affects anything in terms of how user code is woken up (besides making is_replay be correct in places where it previously wasn't, which would've led to incorrect behavior like in the bug, or NDEs anyway)


core/src/worker/workflow/history_update.rs
```diff
@@ -728,13 +732,14 @@ fn find_end_index_of_next_wft_seq(
...
-                }
-                // If we've never seen an interesting event and the next two events are a completion
-                // followed immediately again by scheduled, then this is a WFT heartbeat and also
-                // doesn't conclude the sequence.
-                else if next_event_type == EventType::WorkflowTaskCompleted {
+                } else if next_event_type == EventType::WorkflowTaskCompleted {
                     if let Some(next_next_event) = events.get(ix + 2) {
-                        if next_next_event.event_type() == EventType::WorkflowTaskScheduled {
+                        if !saw_command
+                            && next_next_event.event_type() == EventType::WorkflowTaskScheduled
```
Sushisource, line 738: This is the fix

core/src/worker/workflow/history_update.rs
```diff
@@ -921,7 +926,9 @@ mod tests {
         let seq = next_check_peek(&mut update, 0);
         assert_eq!(seq.len(), 6);
         let seq = next_check_peek(&mut update, 6);
-        assert_eq!(seq.len(), 13);
+        assert_eq!(seq.len(), 4);
```
Sushisource, line 929: This is the test that is updated to work with the new boundaries

cretz (approved):
LGTM. I can't think of an obvious way to break existing workflows that may inadvertently rely on the existing task-end-boundary expectation. We may want to consider an environment variable to be able to flip back for a release or two just in case.

Sushisource:
Added emergency env var

PR #916, opened 2025-05-14
Don't retry on grpc message too large errors

### LINKED ISSUE #462: [Bug] Do not auto-retry gRPC-message-size-too-large errors
### Describe the bug
Not _really_ a bug on this repo's part, but needs to be treated as one nonetheless.

When server gets > 4MB message, it returns ResourceExhausted error with something like:

> grpc: received message larger than max (10248653 vs. 4194304)

Which is retried by our client as are all other `ResourceExhausted`. We need to stop retrying this one in particular but still retry other `ResourceExhausted`. If we can prevent it from going over the wire without manually pre-calculating the size that would be best, but otherwise we may need to hack something to change the error type for this particular error before it gets to the retry interceptor (or inside the retry interceptor if you can). Yes that may mean a message string check which is admittedly gross.

Also, add an integration test with a workflow whose activity payload will trigger this and confirm that it's a workflow task failure (can be in sdk-features).

tconley1428:
## What was changed
Don't retry on grpc message too large errors. Forward the error and fail the workflow task

## Why?
Retries will not succeed, as they don't change the size of the message

## Checklist

1. Closes #462 

2. How was this tested:
New worker tests

3. Any docs updates needed?

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -5,6 +5,7 @@ CLI_VERSION_OVERRIDE = "v1.3.1-priority.0"
 
 [alias]
 integ-test = ["run", "--package", "temporal-sdk-core", "--example", "integ_runner", "--"]
+cloud-test = ["run", "--package", "temporal-sdk-core", "--example", "integ_runner", "--", "--test-name", "cloud_tests"]
 lint = ["clippy", "--workspace", "--examples", "--all-features",
     "--test", "integ_tests", "--test", "heavy_tests", "--test", "manual_tests",
     "--", "--D", "warnings"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -5,7 +5,6 @@ CLI_VERSION_OVERRIDE = "v1.3.1-priority.0"
 
 [alias]
 integ-test = ["run", "--package", "temporal-sdk-core", "--example", "integ_runner", "--"]
-cloud-test = ["run", "--package", "temporal-sdk-core", "--example", "integ_runner", "--", "--test-name", "cloud_tests"]
 lint = ["clippy", "--workspace", "--examples", "--all-features",
     "--test", "integ_tests", "--test", "heavy_tests", "--test", "manual_tests",
     "--", "--D", "warnings"]
--- a/client/src/retry.rs
+++ b/client/src/retry.rs
@@ -201,7 +201,11 @@ where
 {
     type OutError = tonic::Status;
 
-    fn handle(&mut self, current_attempt: usize, e: tonic::Status) -> RetryPolicy<tonic::Status> {
+    fn handle(
+        &mut self,
+        current_attempt: usize,
+        mut e: tonic::Status,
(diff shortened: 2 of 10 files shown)


client/src/retry.rs
```diff
@@ -213,6 +217,17 @@ where
...
         }
 
+        // Short circuit if message is too large - this is not retryable
+        if e.code() == Code::ResourceExhausted
+            && e.message()
+                .contains("grpc: received message larger than max")
+        {
+            // Leave a marker so we don't have duplicate detection logic in the workflow
+            e.metadata_mut()
+                .insert("message-too-large", tonic::metadata::MetadataValue::from(0));
```
Sushisource, line 227: Probably extract this out to a static that can be re-used from the worker side

tests/integ_tests/worker_tests.rs
```diff
@@ -153,3 +158,35 @@ async fn resource_based_few_pollers_guarantees_non_sticky_poll() {
     }
     worker.run_until_done().await.unwrap();
 }
+
+static OVERSIZE_GRPC_MESSAGE_RUN: AtomicBool = AtomicBool::new(false);
```
Sushisource, line 162: I typically move these statics inside the function to avoid accidental re-use

Sushisource (commented):
Overall idea looks good to me. Will want to add cloud test as discussed on Slack.

core/src/worker/workflow/mod.rs
```diff
@@ -317,6 +318,180 @@ impl Workflows {
...
+                            }
+                            self.handle_eager_activities(
+                                reserved_act_permits,
+                                response.activity_tasks,
+                            );
+                        }
+                        Err(e) if e.metadata().contains_key("message-too-large") => {
+                            let failure = Failure {
+                                failure: Some(
+                                    temporal_sdk_core_protos::temporal::api::failure::v1::Failure {
```
cretz, line 394: Arguably we should put an application error info in here with an error type string of `GrpcMessageTooLarge` or something. When we build rules to react to task failure (e.g. ones that pause or notify), users will need something to uniquely identify the error. The task failed cause enum will do this too, but I suspect most users will inspect the failure. Not required of course.

cretz (commented):
No comment on the general Rust logic, just a couple of notes. Leaving code review to @Sushisource.

.cargo/config.toml
```diff
@@ -5,6 +5,7 @@ CLI_VERSION_OVERRIDE = "v1.3.1-priority.0"
 
 [alias]
 integ-test = ["run", "--package", "temporal-sdk-core", "--example", "integ_runner", "--"]
+cloud-test = ["run", "--package", "temporal-sdk-core", "--example", "integ_runner", "--", "--test-name", "cloud_tests"]
```
Sushisource, line 8: You can bypass the integ runner here (and it saves some compile time) since all that's really doing is starting the local server, which ofc you don't need for cloud.

Sushisource (approved):
Sweet! Looking good

PR #919, opened 2025-05-16
Fix missing timeout marker for local activities

### LINKED ISSUE #891: [Bug] Missing timeout marker when local activities timeout during worker shutdown
I am trying to stop a worker in the middle of a local activity, with no `graceful_shutdown_period` specified.

### Describe the bug

Currently, history ends with `WorkflowTaskFailed` and `WorkflowExecutionTimedOut`, but we should be getting a timeout marker for the local activity.

### Minimal Reproduction

The local_activity_stop_follow_up test in 
https://github.com/yuandrew/sdk-core/blob/missing-timeout-marker-bug/tests/integ_tests/workflow_tests/local_activities.rs#L149

### Environment/Versions

- OS and processor: [e.g. M1 Mac, x86 Windows, Linux]
- Temporal Version: [e.g. 1.14.0?] and/or SDK version
- Are you using Docker or Kubernetes or building Temporal from source?

yuandrew:
## What was changed

Handle the edge case where the `schedule_to_close` timeout occurs before we've added the `TaskToken` to `outstanding_activity_tasks`.

## Why?

Antz > Bug's Life

## Checklist

1. Closes #891

2. How was this tested:

Added test that found the bug

3. Any docs updates needed?


yuandrew:
Fixed up the test setup, now seeing the expected timeout Marker in history

PR #925, opened 2025-05-27
Fix bad balacing between sticky & nonsticky WFT polling with low slot/poller counts

Sushisource:
Fixes an issue that could cause task timeouts when using very small (<2) numbers of WFT slots or pollers.

### DIFF
--- a/core-api/src/worker.rs
+++ b/core-api/src/worker.rs
@@ -521,7 +521,7 @@ pub enum PollerBehavior {
     /// requires a slot to be available before beginning polling.
     Autoscaling {
         /// At least this many poll calls will always be attempted (assuming slots are available).
-        /// Cannot be less than two for workflow tasks, or one for other tasks.
+        /// Cannot be zero.
         minimum: usize,
         /// At most this many poll calls will ever be open at once. Must be >= `minimum`.
         maximum: usize,
--- a/core-api/src/worker.rs
+++ b/core-api/src/worker.rs
@@ -225,6 +225,19 @@ impl WorkerConfigBuilder {
             }
         }
 
+        if matches!(self.max_outstanding_workflow_tasks.as_ref(), Some(Some(v)) if *v == 0) {
+            return Err("`max_outstanding_workflow_tasks` must be > 0".to_owned());
+        }
+        if matches!(self.max_outstanding_activities.as_ref(), Some(Some(v)) if *v == 0) {
+            return Err("`max_outstanding_activities` must be > 0".to_owned());
+        }
+        if matches!(self.max_outstanding_local_activities.as_ref(), Some(Some(v)) if *v == 0) {
+            return Err("`max_outstanding_local_activities` must be > 0".to_owned());
+        }
+        if matches!(self.max_outstanding_nexus_tasks.as_ref(), Some(Some(v)) if *v == 0) {
+            return Err("`max_outstanding_nexus_tasks` must be > 0".to_owned());
+        }
+
         if let Some(cache) = self.max_cached_workflows.as_ref() {
             if *cache > 0 {
                 if let Some(Some(max_wft)) = self.max_outstanding_workflow_tasks.as_ref() {
@@ -288,8 +301,8 @@ impl WorkerConfigBuilder {
 /// For more, see the docstrings of the traits in the return types of its functions.
 pub trait WorkerTuner {
     /// Return a [SlotSupplier] for workflow tasks. Note that workflow task slot suppliers must be
-    /// willing to hand out a minimum of two slots if workflow caching is enabled, otherwise the
-    /// worker may fail to process new tasks.
+    /// willing to hand out a minimum of one non-sticky slot and one sticky slot if workflow caching
+    /// is enabled, otherwise the worker may fail to process new tasks.
     fn workflow_task_slot_supplier(
         &self,
     ) -> Arc<dyn SlotSupplier<SlotKind = WorkflowSlotKind> + Send + Sync>;
@@ -413,11 +426,6 @@ impl SlotSupplierPermit {
     /// Returns none if there is no data or the data is not of the appropriate type.
     pub fn user_data_mut<T: Any + Send + Sync>(&mut self) -> Option<&mut T> {
         self.user_data.as_mut().and_then(|b| b.downcast_mut())
-    }
-
-    /// Deconstruct this permit and return the inner data
-    pub fn into_user_data(self) -> Option<Box<dyn Any + Send + Sync>> {
-        self.user_data
     }
 }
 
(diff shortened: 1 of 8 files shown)


core/src/worker/workflow/wft_poller.rs
```diff
@@ -93,15 +100,19 @@ impl WFTPollerShared {
...
         if !is_sticky {
             let backlog = *self.last_seen_sticky_backlog.0.borrow();
-            if backlog > 1 {
+            if backlog >= 1 {
                 let _ = self
                     .last_seen_sticky_backlog
                     .0
                     .clone()
-                    .wait_for(|v| *v <= 1)
+                    .wait_for(|v| *v < 1)
```
dandavison, line 115:
```suggestion
                    .wait_for(|v| *v == 0)
```

(I have no idea currently what's going on in this code, but I see `v` is `usize`.)

core/src/worker/workflow/wft_poller.rs
```diff
@@ -119,7 +130,10 @@ impl WFTPollerShared {
                 } else {
                     let _ = non_sticky_active
                         .clone()
-                        .wait_for(|v| *v <= *sticky_active.borrow())
+                        .wait_for(|v| {
+                            *v < *sticky_active.borrow()
+                                || !self.have_done_first_poll.load(Ordering::Relaxed)
+                        })
```
dandavison, line 136: Is there an argument for a comment here to help future readers?

core/src/worker/workflow/wft_poller.rs
```diff
@@ -74,6 +77,8 @@ pub(crate) struct WFTPollerShared {
     last_seen_sticky_backlog: (watch::Receiver<usize>, watch::Sender<usize>),
     sticky_active: OnceLock<watch::Receiver<usize>>,
     non_sticky_active: OnceLock<watch::Receiver<usize>>,
+    wait_for_first_nonsticky_poll: Notify,
```
mjameswh, line 80: Neither of these two new fields are ever changed/notified. Uncompleted work on an optimization?

core/src/worker/workflow/wft_poller.rs
mjameswh: For now, it looks to me like this condition will always eval to `true`, since `have_done_first_poll` is set to false on creation, and never changed afterward.

mjameswh: Another question - Assuming `have_done_first_poll` would work as its name suggest, your change would prioritize sticky pollers over non-sticky except on the first polls. So assuming the case where max wft poller is one, it would forcibly be a sticky one from that point on, no? No more polling on the non-sticky tq, ever? Is that really the intended behavior?

Also, what if workflow cache is disabled (`max_cached_workflows = 0`), so that we don't even have a sticky task queue? I really don't know how this is handled internally, but it seems like everything would still go through `WFTPollerShared`, and so I assume that `sticky_active` would remain 0 forever in that case. And so we'd found ourselves in the situation of being incapable of doing anymore poll past the very first one.

Sushisource: Yeah, you're right that those somehow just weren't used. They were before and then I refactored to simplify things and somehow this weird broken version actually does the right thing (probably accidentally by just not using sticky). Will fix.

Sushisource: Added a comment as well

Sushisource:
> Can you describe what this "fix" is? and why is it only relevant if you have a low number of pollers/slots?

The fix is enforcing the 2-value minimums now, as well as dealing with the problem James mentioned where the balancing was waiting on changes to one channel when it needed to respect changes to both channels. This caused a problem where there was some random chance of getting unbalanced and "stuck" because only one channel was changing. This is more obvious at low poller counts.

tests/integ_tests/workflow_tests.rs
```diff
@@ -85,56 +83,6 @@ async fn parallel_workflows_same_queue() {
...
-static RUN_CT: AtomicUsize = AtomicUsize::new(0);
-
-pub(crate) async fn cache_evictions_wf(command_sink: WfContext) -> WorkflowResult<()> {
-    RUN_CT.fetch_add(1, Ordering::SeqCst);
-    command_sink.timer(Duration::from_secs(1)).await;
-    Ok(().into())
-}
-
-#[tokio::test]
-async fn workflow_lru_cache_evictions() {
```
Sushisource, line 97: This test is removed now since it depended on having only 1 poller to work. It's covered by unit tests anyway.

core-api/src/worker.rs
```diff
@@ -224,6 +225,28 @@ impl WorkerConfigBuilder {
             }
         }
 
+        if let Some(cache) = self.max_cached_workflows.as_ref() {
```
mjameswh, line 228: Should we validate `max_outstanding_workflow_tasks` and `workflow_task_poller_behavior` are both at least 1 if workflow cache is disabled, or is this already asserted somewhere else?

core/src/worker/workflow/wft_poller.rs
```diff
@@ -96,31 +100,39 @@ impl WFTPollerShared {
         // If there's a sticky backlog, prioritize it.
```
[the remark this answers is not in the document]
Sushisource: Yes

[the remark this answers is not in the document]
Sushisource: What I mean is this code runs before permit acquisition and the scaler. So, we balance attempts to acquire permits, and then scale. IE: It's possible in autoscale mode to, for example, have both the sticky and nonsticky proceed past this balance, acquire permits, allow scale, but from then on the scaler might hold up nonsticky indefinitely until the sticky backlog is clear.

However... it is making me realize the whole "balance" part does not need to apply to autoscale at all. Only the backlog clearing part. I will fix that. I was just running some tests with the autoscale on and since the sticky backlog constantly bounces off of 0, it ties them together more than necessary.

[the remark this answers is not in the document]
Quinn-With-Two-Ns: Reading through the comments I am sorry I still don't understand why this logic exists for the autoscaling case at all. 

In the fixed pollers we need to have some algorithm to balance the number of pollers , but in the autoscaling case the SDK is picking the number of pollers based off of feedback from the server that should be all the balancing we need. I understand there is some argument for balancing with autoscaling if we are low on slots, but that logic is not being applied if we are autoscaling.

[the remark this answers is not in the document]
Sushisource: Yeah, I feel like I really tied myself in knots on this one. I've changed it to just be way more obvious about what's the real problem here, which is tying up all the slots with one kind.

PR #927, opened 2025-06-02
Nexus cancellation types

### LINKED ISSUE #911: [Feature Request] Support Nexus operation cancellation types

We would like to expose cancellation types for Nexus operations to give users control over what happens to their Nexus operations when the parent context is cancelled. This behavior should follow the pattern set by child workflows. There should be 4 new cancellation types where are specified as part of the input options when starting an operation.

1. `ABANDON` -> Do not send a RequestCancelNexusOperation command
2. `TRY_CANCEL` -> Send a RequestCancelNexusOperation command and immediately report cancellation to the caller without waiting for confirmation that the operation was cancelled.
3. `WAIT_REQUESTED` -> Request cancellation of the operation and wait for confirmation that the cancel request was received by the operation handler.
4. `WAIT_COMPLETED` -> Request cancellation of the operation and wait for the operation to complete. It may or may not complete as cancelled. This is the current behavior and therefore should be the default for backwards compatibility.

To support `WAIT_REQUESTED`, two new history events were added in https://github.com/temporalio/api/pull/564 for `NexusOperationCancelRequestCompleted` and `NexusOperationCancelRequestFailed`. When either of these events are seen, that means that the cancel request was successfully delivered to the operation handler and futures for operations started with `WAIT_REQUESTED` can be unblocked.

tconley1428:
## What was changed
Added nexus cancellation types to the core API and state machine

## Checklist

1. Closes #911 

2. How was this tested:
New integration tests

3. Any docs updates needed?
No

### DIFF
--- a/tests/integ_tests/workflow_tests/nexus.rs
+++ b/tests/integ_tests/workflow_tests/nexus.rs
@@ -25,7 +25,7 @@ use temporal_sdk_core_protos::{
         },
     },
 };
-use temporal_sdk_core_test_utils::{CoreWfStarter, rand_6_chars};
+use temporal_sdk_core_test_utils::{CoreWfStarter, WorkflowHandleExt, rand_6_chars};
 use tokio::sync::watch;
 use tokio::{join, sync::mpsc};
 
@@ -69,7 +69,7 @@ async fn nexus_basic(
             Ok(res.into())
         }
     });
-    starter.start_with_worker(wf_name, &mut worker).await;
+    let wf_handle = starter.start_with_worker(wf_name, &mut worker).await;
 
     let client = starter.get_client().await.get_client().clone();
     let nexus_task_handle = async {
@@ -165,6 +165,10 @@ async fn nexus_basic(
         }
         _ => unreachable!(),
     }
+    wf_handle
+        .fetch_history_and_replay(worker.inner_mut())
+        .await
+        .unwrap();
 }
 
 #[rstest::rstest]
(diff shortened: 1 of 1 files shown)


core/src/worker/workflow/machines/nexus_operation_state_machine.rs
```diff
@@ -65,10 +68,17 @@ fsm! {
...
+    Cancelled --(NexusOperationCancelRequested)--> Cancelled;
+    Cancelled --(NexusOperationCompleted(NexusOperationCompletedEventAttributes), shared on_completed)--> Cancelled;
+    Cancelled --(NexusOperationFailed(NexusOperationFailedEventAttributes), shared on_failed)--> Cancelled;
+    Cancelled --(NexusOperationTimedOut(NexusOperationTimedOutEventAttributes), shared on_timed_out)--> Cancelled;
+    Cancelled --(NexusOperationCanceled(NexusOperationCanceledEventAttributes))--> Cancelled;
+
     // Ignore cancels in all terminal states
     Completed --(Cancel)--> Completed;
     Failed --(Cancel)--> Failed;
-    Cancelled --(Cancel)--> Cancelled;
```
Sushisource, line 71: I'm a little worried removing this could cause errors in SDKs that allow invoking a cancel on an old handle or something. Probably no harm in leaving it in?

Sushisource (commented):
Generally looking good, just some small things

core/src/worker/workflow/machines/nexus_operation_state_machine.rs
Sushisource: This guy is still gone

Issue #933, opened 2025-06-09
[Bug] cur_mem_usage being overrided with cgroup limits


antlai-temporal:
### Describe the bug
https://github.com/temporalio/sdk-core/blob/master/core/src/worker/tuner/resource_based.rs#L514-L522

 It seems that we carefully calculate cur_mem_usage with cgroup limits but then immediately override it with the node used memory...

Issue #903, opened 2025-04-17
[Feature Request] Support CPU Cgroup Limits in Resource-based tuner


Sushisource:
### Is your feature request related to a problem? Please describe.

A few users have expressed a desire to be able to target CPU as a % of CPU limits/quota in containerized environments.

### Describe the solution you'd like

Will need to use cgroup functionality to determine CPU usage as a fraction of the container's used CPU vs it's limit/quota.

Sushisource:
There is an impl in Go that can be referenced: https://github.com/temporalio/sdk-go/blob/aecdca8be43def99f6237a07385f58612869cd65/contrib/resourcetuner/cgroups.go#L46

I'm not 100% sure if this is what we want or not, but, it's one option. We probably want to be able to configure how the tuner should calculate usage. IE: `Host | CGroups | Whatever`

PR #934, opened 2025-06-09
Fix incorrect cgroups memory usage

Sushisource:
Fix this obviously wrong calculation.

No specific new tests here b/c this area needs more love in general and should get more tests w/ https://github.com/temporalio/sdk-core/issues/903

Fixes https://github.com/temporalio/sdk-core/issues/933

### DIFF
--- a/core/src/worker/tuner/resource_based.rs
+++ b/core/src/worker/tuner/resource_based.rs
@@ -509,7 +509,6 @@ impl RealSysInfo {
         let mut lock = self.sys.lock();
         lock.refresh_memory();
         lock.refresh_cpu_usage();
-        let mem = lock.used_memory();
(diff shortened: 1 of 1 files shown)


core/src/worker/tuner/resource_based.rs
```diff
@@ -518,8 +518,9 @@ impl RealSysInfo {
                 cgroup_limits.total_memory - cgroup_limits.free_memory,
                 Ordering::Release,
             );
+        } else {
+            self.cur_mem_usage.store(mem, Ordering::Release);
```
cretz, line 522: Might as well move `let mem = lock.used_memory();` into this conditional too

cretz (approved):
Nothing blocking

PR #941, opened 2025-06-20
Fix wft poll balancing with maxpoll=2

mjameswh:
## What was changed

Fix a bug that prevented `PollWorkflowTaskQueue` to the sticky task queue if `workflow_task_poller_behavior = PollerBehavior::SimpleMaximum(2)`. This was resulting in slower performance, and would cause "sticky worker unavailable, please use original task queue." error messages on the server.

### DIFF
--- a/core/src/worker/workflow/wft_poller.rs
+++ b/core/src/worker/workflow/wft_poller.rs
@@ -102,16 +102,6 @@ impl WFTPollerShared {
     /// Makes either the sticky or non-sticky poller wait pre-permit-acquisition so that we can
     /// balance which kind of queue we poll appropriately.
     pub(crate) async fn wait_if_needed(&self, is_sticky: bool) {
-        // If there's a sticky backlog, prioritize it.
-        if !is_sticky {
-            let _ = self
-                .last_seen_sticky_backlog
-                .0
-                .clone()
-                .wait_for(|v| *v == 0)
-                .await;
-        }
-
         // We need to make sure there's at least one poller of both kinds available. So, we check
         // that we won't end up using every available permit with one kind of poller. In practice
         // this is only ever likely to be an issue with very small numbers of slots.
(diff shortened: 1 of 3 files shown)


core/src/worker/workflow/wft_poller.rs
```diff
@@ -121,26 +111,63 @@ impl WFTPollerShared {
...
+                let mut sticky_backlog = self.last_seen_sticky_backlog.0.clone();
+
                 loop {
                     let num_sticky_active = *sticky_active.borrow_and_update();
                     let num_non_sticky_active = *non_sticky_active.borrow_and_update();
-                    let both_are_zero = num_sticky_active == 0 && num_non_sticky_active == 0;
-                    if both_are_zero {
+                    let num_sticky_backlog = *sticky_backlog.borrow_and_update();
+
+                    let allow = || {
```
Sushisource, line 121: This could avoid being a closure by using else-ifs... but I don't have a firm position on if that's more readable or not.

core/src/worker/workflow/wft_poller.rs
```diff
@@ -121,26 +111,63 @@ impl WFTPollerShared {
...
-                    } else {
-                        let would_exceed_max_slots =
-                            (num_sticky_active + num_non_sticky_active + 1) >= max_slots;
-                        let must_wait = would_exceed_max_slots
-                            && (num_sticky_active == 0 || num_non_sticky_active == 0);
-                        if !must_wait {
-                            break;
+
+                        // Just balance the two poller types.
+                        // FIXME: Do we need anything more here, to ensure proper balancing?
```
Sushisource, line 155: Go effectively just forces there to be an even amount of both kinds of pending requests, except always favoring sticky during a backlog, and when counts are even: https://github.com/temporalio/sdk-go/blob/9682a692145f503a02b73f32435cb04bdc6f57c1/internal/internal_task_pollers.go#L782

We already are favoring it with a backlog. It's not obvious to me that just making an even amount is the best move either, but, it is what Core used to do and clearly it at least works, so we could go with that.

Sushisource (approved):

PR #951, opened 2025-07-01
Moved C Bridge from dotnet-sdk to core-sdk repo

maciejdudko:
## What was changed

Moved C Bridge from dotnet-sdk to core-sdk repo.

The PR has been squashed into 2 commits to make it easier to review:
- The first commit is a straight copy of what's in dotnet-sdk repo.
- The second commit contains all the other changes.

**💥 There are breaking changes to the C bridge:**
- Renamed crate to `temporal_sdk_core_c_bridge`
- Renamed header file to `temporal-sdk-core-c-bridge.h`
- Added `temporal_core_` prefix to exported C functions and `TemporalCore` prefix to exported types
- Changed `TestServerOptions::download_ttl_ms` to `download_ttl_seconds`

Other changes:
- Upgraded to edition 2024, cleaned up dependencies
- Made exported structs and their fields public (so that Rust tests can use them)
- Added missing RPC calls
- Added some smoke tests for the bridge
- Moved LTO profile to workspace root
- Various lints and formatting

There's also one unrelated change to integration tests that fixes unused import warning on Linux/ARM.

## Why?

It makes the C bridge available for other languages that prefer to use C interface to interact with Rust. See https://github.com/temporalio/sdk-dotnet/issues/423

## Checklist

1. How was this tested:

- The bridge crate has some basic smoke tests.
- A matching PR will be opened for dotnet-sdk repo. Before merging this PR we should make sure dotnet-sdk builds and works using this version of the bridge.

2. Any docs updates needed?

Need to update README to mention the new crate.


core-c-bridge/src/client.rs
```diff
@@ -0,0 +1,677 @@
...
+    ($client:ident, $call:ident, $trait:tt, $call_name:ident) => {
+        if $call.retry {
+            rpc_resp($trait::$call_name(&mut $client, rpc_req($call)?).await)
+        } else {
+            rpc_resp($trait::$call_name(&mut $client.into_inner(), rpc_req($call)?).await)
+        }
+    };
+}
+
+async fn call_workflow_service(
```
cretz, line 299: I think we should either change this to be code generated (w/ CI check confirming code gen is accurate), or something else to confirm that every string is represented here. In .NET we do have a test IIRC to confirm every gRPC call is available. In Ruby we ended up writing a code gen to generate these big match statements (though if the match statement gets too big it can have stack problems in debug build).

If we'd like though, we can defer this to another PR. But we do need something at some point to confirm this doesn't get out of sync with the proto RPC names.

core-c-bridge/include/temporal-sdk-core-c-bridge.h
cretz, line 1: I think we should have a CI step that confirms that there is no git diff after build (thereby confirming that the checked in header is the latest)

cretz (commented):
Looks great, thanks for all the work here! Only a minor CI suggestion and code-gen/validation suggestion (that can be deferred).

Also, looking forward to seeing maybe a .NET draft PR that changes submodule to this branch and just to confirm we don't need any more change on this side to be usable.

core-c-bridge/src/client.rs
cretz: Thanks! This is good enough for me. I do think we may want code gen one day, but so long as the test is there to make sure we have to add them as they come up, sounds good.

core-c-bridge/include/temporal-sdk-core-c-bridge.h
maciejdudko: Added. Here's an example failure: https://github.com/temporalio/sdk-core/actions/runs/16008506207/job/45160619527?pr=951#step:12:1

cretz (approved):
Looking at https://github.com/temporalio/sdk-dotnet/pull/494, I think we're good here

core-c-bridge/src/tests/context.rs
```diff
@@ -0,0 +1,649 @@
+#![allow(dead_code)]
+#![allow(unused_imports)]
```
Sushisource, line 2: Would be good to make these specific to the area they're needed in

maciejdudko: Oops. Fixed in new PR #954

PR #953, opened 2025-07-02
Worker heartbeat

yuandrew:
## What was changed

New worker heartbeat functionality. There are a few calls that also send this heartbeat data, and there is a timer that runs in the background to ensure the worker heartbeats within `heartbeat_interval` if there are no calls that send this heartbeat data.

misc changes:
- Renamed `mock_workflow_client` to `mock_worker_client`
- fixed spelling for `la_permit_dealer`

## Why?

new feature

## Checklist

1. Closes 

2. How was this tested:
 
new unit test

3. Any docs updates needed?


core/src/worker/heartbeat.rs
```diff
@@ -0,0 +1,255 @@
...
+            .unwrap();
+
+        let heartbeat_info = Arc::new(Mutex::new(WorkerHeartbeatInfo::new(config.clone())));
+        let client = Arc::new(mock);
+        heartbeat_info.lock().add_client(client.clone());
+        let worker = worker::Worker::new(config, None, client, None, Some(heartbeat_info.clone()));
+        let _ = heartbeat_info.lock().capture_heartbeat();
+
+        // heartbeat timer fires once
+        tokio::time::sleep(Duration::from_millis(150)).await;
```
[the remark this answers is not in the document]
yuandrew: I actually want to test specific sleep times to make sure I try heartbeating in the right spot, new iteration of the test should be more obvious for this

core/src/worker/heartbeat.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub(crate) sdk_version: String,
+    /// Worker start time
+    pub(crate) start_time: SystemTime,
+}
+
+impl WorkerHeartbeatData {
+    fn new(worker_config: WorkerConfig) -> Self {
+        Self {
+            // TODO: Is this right for worker_identity?
+            worker_identity: worker_config
```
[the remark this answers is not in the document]
yuandrew: added a `bail!` statement in `init_worker` to ensure either lang or client sets this value

core/src/worker/heartbeat.rs
```diff
@@ -141,8 +68,15 @@ impl WorkerHeartbeatData {
         };
 
         self.heartbeat_time = Some(now);
+        if let Some(reset_tx) = &self.reset_tx {
+            let _ = reset_tx.send(());
```
Sushisource: This state should be unrepresentable. A fun type puzzle!

I think the way to fix this is to separate out data from some kind of handle for calling this function. The handle is created early on (including the reset channel inside it) and the data is created later, down inside `Worker::new`. Initially the handle has no capability but once it's passed into the constructor for data it becomes useful.

This sort of "uninitialized" state still sort of exists between those two calls, but, at least it exists in a way that works fairly naturally with the return type of `Option<WorkerHeartbeat>`, or in a way that can block until everything is initialized.

One option is something like having the handle contain a `req_resp: Sender<Receiver<Option<WorkerHeartbeat>` which can be used inside the timing task to decide if the caller should get a heartbeat or not (also obviates then need for the dedicated reset channel). Dunno if that's overkill or not lol.

The more obvious option is to just have a `OnceLock<Box<Fn() -> Optional<WorkerHeartbeat>>>` or similar, which can be `wait`ed on to do semantically the same thing as the channel-of-channel. It's probably a bit cleaner than nested channels.

I dunno, fun one, I thought about this for a while and had to stop myself from just trying to do it. Feel free to play around. I maybe spend too much time trying to eliminate states like these. If you think it's not worth it, which is legit, change this to a `dbg_panic!`

core/src/worker/mod.rs
```diff
@@ -484,6 +497,25 @@ impl Worker {
...
+                data.sdk_name = sdk_name_and_ver.0.clone();
+                data.sdk_version = sdk_name_and_ver.1.clone();
+                data.start_time = SystemTime::now();
+                data.set_reset_tx(reset_tx);
+            }
+            Some(create_worker_heartbeat_process(
+                heartbeat_data.clone(),
+                client.clone(),
+                reset_rx,
+            ))
```
Sushisource, line 514: This can all move inside `WorkerHeartbeatData::new` after my other comment (which can internalize the join handle and expose a `shutdown` method or something).

(And probably should regardless of which direction you go)

Sushisource (commented):
Type system puzzle totally nerd sniped me

core/src/worker/mod.rs
yuandrew: makes sense

core/src/worker/heartbeat.rs
```diff
@@ -1,50 +1,128 @@
...
+        self.heartbeat_handle.abort()
+    }
+}
+
 #[derive(Debug, Clone)]
-pub struct WorkerHeartbeatData {
+struct WorkerHeartbeatData {
     worker_instance_key: String,
-    pub(crate) worker_identity: String,
+    worker_identity: String,
```
Sushisource: Ahh, de-pub :feelsgoodman:

core/src/worker/heartbeat.rs
```diff
@@ -1,50 +1,128 @@
...
+
+impl WorkerHeartbeatManager {
+    pub(crate) fn new(
+        config: WorkerConfig,
+        identity: String,
+        heartbeat_fn: Arc<OnceLock<Box<dyn Fn() -> Option<WorkerHeartbeat> + Send + Sync>>>,
+        client: Arc<dyn WorkerClient>,
+    ) -> Self {
+        let sdk_name_and_ver = client.sdk_name_and_version();
+        let (reset_tx, reset_rx) = watch::channel(());
```
Sushisource, line 29: I'm realizing this should probably just be a https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html

core/src/worker/heartbeat.rs
```diff
@@ -0,0 +1,224 @@
...
+            loop {
+                tokio::select! {
+                    _ = ticker.tick() => {
+                        let heartbeat = if let Some(heartbeat) = data_clone.lock().capture_heartbeat_if_needed() {
+                            heartbeat
+                        } else {
+                            continue
+                        };
+                        if let Err(e) = client.clone().record_worker_heartbeat(heartbeat).await {
+                            warn!(error=?e, "Network error while sending worker heartbeat");
```
Sushisource, line 50: I think probably move this warning after the return-if-unimplemented

core/src/worker/heartbeat.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    fn capture_heartbeat_if_needed(&mut self) -> Option<WorkerHeartbeat> {
+        let now = SystemTime::now();
+        let elapsed_since_last_heartbeat = if let Some(heartbeat_time) = self.heartbeat_time {
+            let dur = now.duration_since(heartbeat_time).unwrap_or(Duration::ZERO);
+
+            // Only send poll data if it's nearly been a full interval since this data has been sent
+            // In this case, "nearly" is 90% of the interval
+            if dur.as_secs_f64() < 0.9 * self.heartbeat_interval.as_secs_f64() {
+                println!("Heartbeat interval not yet elapsed, not sending poll data");
```
Sushisource, line 137: Leftover print

core/src/worker/heartbeat.rs
```diff
@@ -0,0 +1,224 @@
...
+            .build()
+            .unwrap();
+
+        let heartbeat_fn = Arc::new(OnceLock::new());
+        let client = Arc::new(mock);
+        let worker = worker::Worker::new(config, None, client, None, Some(heartbeat_fn.clone()));
+        heartbeat_fn.get().unwrap()();
+
+        // heartbeat timer fires once
+        tokio::time::sleep(Duration::from_millis(300)).await;
```
Sushisource, line 216: Looks like we can avoid the sleeps by using `advance` and family which are behind `test-utils` flags for tokio: https://docs.rs/tokio/latest/tokio/time/index.html

Which is very nice.

Sushisource (commented):
Looking good to me. Just a few things to clean up.

core/src/worker/heartbeat.rs
yuandrew: ahh I was originally thinking we'd want a warning either way, but I guess it makes sense not to bother informing users that don't even have this feature that it's being skipped

yuandrew: i guess an argument for logging it would be to try to entice folks to upgrade their server versions for this new feature?

Sushisource: Eh, if they read about it and decide they want it, that's how I'd expect they'd learn anyway.

Sushisource (approved):
We'll want more tests for sure, but, those can be added after you've done the second part of the work to move this over to the runtime and have it be shared across workers.

PR #954, opened 2025-07-02
Removed #![allow(unused_imports)] and #![allow(dead_code)] from C bridge tests

maciejdudko:
## What was changed

Recently merged PR #951 had `#![allow(unused_imports)]` and `#![allow(dead_code)]` accidentally left in in the tests. This PR fixes unused imports and puts `#[allow(dead_code)]` to the relevant functions directly (there are a few getters that are unused but may be useful in the future).

## Why?

The warnings were turned off for development and it was supposed to be removed before merge.


Sushisource (approved):
Thanks!

Issue #883, opened 2025-03-07
[Feature Request] Expose way for worker clients to use their own client


cretz:
### Describe the solution you'd like

Today the `worker::client::WorkerClient` trait is internal for our use, but users want to replace the client implementation for workers. Whether this is simply replacing calls the worker client makes or exposing a way to replace the underlying RPC calls of the whole client is up for discussion.

PR #955, opened 2025-07-09
Introduce a new callback based bridge worker

FranzBusch:
# Motivation

The bridge code allows other languages to interop with the Core Rust SDK through the C-FFI layer. Sometimes those languages also want to bring their own custom gRPC implementation from their languages ecosystem.

# Modification

This PR introduces a new callback based worker client that is also exposed through the bridge header. This allows other languages to use their gPRC client to serve the requests from the Core SDK.

### DIFF
--- a/core-c-bridge/include/temporal-sdk-core-c-bridge.h
+++ b/core-c-bridge/include/temporal-sdk-core-c-bridge.h
@@ -5,6 +5,29 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+/**
+ * Interop mirror of the Rust tonic [tonic::Code] enum
+ */
+typedef enum TemporalCoreCallbackBasedWorkerClientGRPCCode {
+  Ok = 0,
+  Cancelled = 1,
+  Unknown = 2,
+  InvalidArgument = 3,
+  DeadlineExceeded = 4,
+  NotFound = 5,
+  AlreadyExists = 6,
+  PermissionDenied = 7,
+  ResourceExhausted = 8,
+  FailedPrecondition = 9,
+  Aborted = 10,
+  OutOfRange = 11,
+  Unimplemented = 12,
+  Internal = 13,
+  Unavailable = 14,
+  DataLoss = 15,
+  Unauthenticated = 16,
+} TemporalCoreCallbackBasedWorkerClientGRPCCode;
+
 typedef enum TemporalCoreForwardedLogLevel {
   Trace = 0,
   Debug,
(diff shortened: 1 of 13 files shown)


core-c-bridge/src/callback_based_worker_client/interop.rs
```diff
@@ -0,0 +1,264 @@
...
+/// - `client_config`: pointer to [CallbackBasedWorkerClientConfig] (client name, version, identity).
+/// - `system_info_ref`: zero-copy [GetSystemInfoResponse] bytes or empty.
+/// - `runtime`:        pointer to [Runtime].
+///
+/// # Returns
+/// A [WorkerOrFail] struct containing:
+/// - `worker`: on success, a heap-allocated [Worker] pointer; null on error.
+/// - `fail`: on error, a C string (UTF-8) describing the failure; null on success.
+#[unsafe(no_mangle)]
+pub extern "C" fn temporal_core_new_callback_based_worker_client(
```
cretz, line 168:
```suggestion
pub extern "C" fn temporal_core_worker_callback_based_new(
```

or

```suggestion
pub extern "C" fn temporal_core_callback_based_worker_new(
```

Pedantic, but just clearer that you're creating a worker, not a client

core-c-bridge/include/temporal-sdk-core-c-bridge.h
```diff
@@ -629,6 +652,130 @@ typedef struct TemporalCoreWorkerReplayPushResult {
...
+   * timeout in milliseconds
+   */
+  uint64_t timeout_ms;
+} TemporalCoreCallbackBasedWorkerClientPollOptions;
+
+/**
+ * Function pointer for RPC callbacks delivering a `(success, failure)`
+ * pair as non-owning `ByteArrayRef`s to Rust.
+ */
+typedef void (*TemporalCoreWorkerClientCallbackTrampoline)(void *user_data,
```
cretz, line 695:
```suggestion
typedef void (*TemporalCoreWorkerClientResultCallback)(void *user_data,
```

Pedantic, but we have not traditionally called this a trampoline in this C header, can it just be called a callback?

core-c-bridge/include/temporal-sdk-core-c-bridge.h
```diff
@@ -629,6 +652,130 @@ typedef struct TemporalCoreWorkerReplayPushResult {
...
+                                                           struct TemporalCoreByteArrayRef failure);
+
+/**
+ * Interop function pointer table implementing the RPC logic of the [temporal_sdk_core::WorkerClient] trait via callbacks.
+ */
+typedef struct TemporalCoreCallbackBasedWorkerClientRPCs {
+  void *user_data;
+  void (*poll_workflow_task)(void *user_data,
+                             struct TemporalCoreByteArrayRef request,
+                             const struct TemporalCoreCallbackBasedWorkerClientPollOptions *poll_opts,
```
cretz, line 706: How are poll options expected to be used?

core-c-bridge/include/temporal-sdk-core-c-bridge.h
```diff
@@ -629,6 +652,130 @@ typedef struct TemporalCoreWorkerReplayPushResult {
...
+ * pair as non-owning `ByteArrayRef`s to Rust.
+ */
+typedef void (*TemporalCoreWorkerClientCallbackTrampoline)(void *user_data,
+                                                           struct TemporalCoreByteArrayRef success,
+                                                           struct TemporalCoreByteArrayRef failure);
+
+/**
+ * Interop function pointer table implementing the RPC logic of the [temporal_sdk_core::WorkerClient] trait via callbacks.
+ */
+typedef struct TemporalCoreCallbackBasedWorkerClientRPCs {
```
cretz, line 702: So, thinking on this, I wonder if we should just have a general lower-level callback-based client and nothing specific to workers. So basically a C abstraction for gRPC client.

So basically have a `temporal_core_client_from_callbacks_new` that accepts some options including the single function called back for every gRPC request (so RPC name is one of the params), and it returns a pointer to TemporalCoreClient (or rather a struct that has that or fail as mutually exclusive). Now that client can be used for `temporal_core_worker_new` can be passed to `temporal_core_client_free`, etc. The API would be cleaner/simpler and the implementer does not have to be aware of retry or worker details or any of that.

I know we had mentioned this approach at #883, but instead at #924 we'er exposing the single worker client trait. I think it'd be wiser to make a C bridge for gRPC client in general. This would allow our SDKs to let people just replace the gRPC implementation at client creation time instead of worker creation time. It would also allow our SDKs to provide gRPC interception if they wanted (by delegating to another client), though we could make a bit of sugar to make delegating to the default implementation easier at that time.

Thoughts @FranzBusch and @Sushisource?

core-c-bridge/include/temporal-sdk-core-c-bridge.h
```diff
@@ -629,6 +652,130 @@ typedef struct TemporalCoreWorkerReplayPushResult {
...
+  void *user_data;
+  void (*poll_workflow_task)(void *user_data,
+                             struct TemporalCoreByteArrayRef request,
+                             const struct TemporalCoreCallbackBasedWorkerClientPollOptions *poll_opts,
+                             TemporalCoreWorkerClientCallbackTrampoline callback,
+                             void *callback_ud);
```
cretz, line 708:
```suggestion
                             const struct TemporalCoreCallbackBasedWorkerClientPollOptions *poll_options,
                             TemporalCoreWorkerClientCallbackTrampoline callback,
                             void *callback_user_data);
```

Pedantic, but would like not to abbreviate things that are exposed in the header

core-c-bridge/include/temporal-sdk-core-c-bridge.h
```diff
@@ -812,6 +959,32 @@ void temporal_core_complete_async_reserve(void *sender, uintptr_t permit_id);
...
+ * - `client_config`: pointer to [CallbackBasedWorkerClientConfig] (client name, version, identity).
+ * - `system_info_ref`: zero-copy [GetSystemInfoResponse] bytes or empty.
+ * - `runtime`:        pointer to [Runtime].
+ *
+ * # Returns
+ * A [WorkerOrFail] struct containing:
+ * - `worker`: on success, a heap-allocated [Worker] pointer; null on error.
+ * - `fail`: on error, a C string (UTF-8) describing the failure; null on success.
+ */
+struct TemporalCoreWorkerOrFail temporal_core_new_callback_based_worker_client(const struct TemporalCoreCallbackBasedWorkerClientRPCs *client_rpcs,
```
cretz, line 982:
```suggestion
struct TemporalCoreWorkerOrFail temporal_core_callback_based_worker_new(const struct TemporalCoreCallbackBasedWorkerClientRPCs *client_rpcs,
```

or

```suggestion
struct TemporalCoreWorkerOrFail temporal_core_worker_callback_based_new(const struct TemporalCoreCallbackBasedWorkerClientRPCs *client_rpcs,
```

Pedantic, but worth being clear this is creating a _worker_ not a _client_

core-c-bridge/include/temporal-sdk-core-c-bridge.h
```diff
@@ -812,6 +959,32 @@ void temporal_core_complete_async_reserve(void *sender, uintptr_t permit_id);
...
+                                                                               const struct TemporalCoreWorkerOptions *options,
+                                                                               const struct TemporalCoreCallbackBasedWorkerClientConfig *client_config,
```
cretz, line 984: Pedantic, but would be ok putting all other non-runtime parameters in this structure instead of as separate parameters. And call it `TemporalCoreCallbackBasedWorkerOptions` instead of `Config`. This is consistent with what we've done elsewhere I think. But not a big deal.

cretz (commented):
Just reviewed API, didn't get to implementation yet. But I am starting to think we need to be able to replace the gRPC implementation, not the client implementation. There are other benefits to our other SDKs to be able to replace the low-level gRPC calls only.

core-c-bridge/include/temporal-sdk-core-c-bridge.h
FranzBusch: I spent some time looking to see if we can drop the callback based substitution one layer down to general `Client` level. Overall it seems like a good alternative approach but it looks like it would require a significant amount of refactoring to get us there. The approach I took here with just allowing the worker client to be substituted was significantly more straight forward since the protocol already existed. It was also enough for our use-case since I only cared about replacing the worker client and not the overall client.

I would love to proceed with what @Sushisource suggested here and proceed with the strategy of this PR and we can track the alternative approach in an issue for further down the road. If you all are fine with this I will tackle the inline feedback on this PR.

Sushisource:
Closing since we merged the other approach

PR #957, opened 2025-07-10
Fallback to system env vars if none are provided

THardy98:
## What was changed
Update env config to fallback to system env vars if none are explicitly provided.

## Why?
This should be the default behaviour

1. How was this tested:
Small test


cretz (approved):
Looks great!

PR #962, opened 2025-07-23
[DRAFT] Process wide worker heartbeat

yuandrew:
No where ready for in depth review, still need to implement the actual nexus call handling, looking for some high level feedback. Added a few comments for different discussion points

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -53,9 +53,7 @@ use crate::abstractions::dbg_panic;
 pub use crate::worker::client::{
     PollActivityOptions, PollOptions, PollWorkflowOptions, WorkerClient, WorkflowTaskCompletion,
 };
-use crate::worker::heartbeat::{
-    ClientIdentity, HeartbeatFn, HeartbeatMap, HeartbeatNamespaceManager,
-};
+use crate::worker::heartbeat::{ClientIdentity, HeartbeatFn, HeartbeatMap, HeartbeatNamespaceManager, WorkerHeartbeatDetails};
 use crate::{
     replay::{HistoryForReplay, ReplayWorkerInput},
     telemetry::{
@@ -130,42 +128,13 @@ where
         sticky_q,
         client_bag.clone(),
         Some(&runtime.telemetry),
-        // Worker sets heartbeat, and client bag AND runtime both use to heartbeat
-        Some(heartbeat_fn.clone()), // TODO: combine with heartbeat_map into an enum
-        None,
+        Some(WorkerHeartbeatDetails::Fn(heartbeat_fn.clone())),
     );
 
     if runtime.heartbeat_worker.get().is_none() {
         let process_key = Uuid::new_v4();
         // TODO: set max_concurrent_nexus_polls to 1?
         // let nexus_config = WorkerConfig {
-        //     namespace: "".to_string(),
-        //     task_queue: "".to_string(),
-        //     client_identity_override: None,
-        //     max_cached_workflows: 0,
-        //     tuner: None,
-        //     workflow_task_poller_behavior: (),
-        //     nonsticky_to_sticky_poll_ratio: 0.0,
-        //     activity_task_poller_behavior: (),
-        //     nexus_task_poller_behavior: (),
-        //     no_remote_activities: false,
-        //     sticky_queue_schedule_to_start_timeout: Default::default(),
-        //     max_heartbeat_throttle_interval: Default::default(),
-        //     default_heartbeat_throttle_interval: Default::default(),
-        //     max_task_queue_activities_per_second: None,
-        //     max_worker_activities_per_second: None,
-        //     ignore_evicts_on_shutdown: false,
-        //     fetching_concurrency: 0,
-        //     graceful_shutdown_period: None,
-        //     local_timeout_buffer_for_activities: Default::default(),
-        //     workflow_failure_errors: Default::default(),
-        //     workflow_types_to_failure_errors: Default::default(),
-        //     max_outstanding_workflow_tasks: None,
-        //     max_outstanding_activities: None,
-        //     max_outstanding_local_activities: None,
-        //     max_outstanding_nexus_tasks: None,
-        //     versioning_strategy: Default::default(),
-        //     heartbeat_interval: Default::default(),
         // };
         // construct nexus worker
         let nexus_worker = Worker::new(
@@ -173,8 +142,7 @@ where
             None,
             client_bag,
             None,
-            None,
-            Some(runtime.get_heartbeat_map()),
+            Some(WorkerHeartbeatDetails::Map(runtime.get_heartbeat_map())),
         );
         runtime.add_heartbeat_worker(nexus_worker);
     }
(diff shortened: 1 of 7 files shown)


core/src/lib.rs
```diff
@@ -113,13 +123,48 @@ where
...
         Some(&runtime.telemetry),
-        Some(heartbeat_fn),
-    ))
+        Some(WorkerHeartbeatDetails::Fn(heartbeat_fn.clone())),
+    );
+
+    if runtime.heartbeat_worker.get().is_none() {
+        let process_key = Uuid::new_v4();
+        // TODO: set max_concurrent_nexus_polls to 1?
+        // let nexus_config = WorkerConfig {
```
yuandrew, line 137: Seems like we want to have a separate config for this worker, or do we want to group them with the existing `WorkerConfig`? But what if different workers on the same namespace have 2 different configs?

Also, not sure what config options need to be unique here, I'm thinking potentially `max_concurrent_nexus_polls`, and also `heartbeat_interval` at least?

core/src/worker/heartbeat.rs
```diff
@@ -1,90 +1,189 @@
...
+                                warn!(error=?e, "Network error while sending worker heartbeat");
                             }
-                            warn!(error=?e, "Network error while sending worker heartbeat");
-                        }
                     }
                     _ = reset_notify.notified() => {
                         ticker.reset();
                     }
+                    // TODO: handle nexus tasks
+                    res = manager.next_nexus_task() => {
```
yuandrew, line 103: Currently working with Yuri on this

core/src/lib.rs
```diff
@@ -89,7 +97,9 @@ pub fn init_worker<CT>(
 where
     CT: Into<sealed::AnyClient>,
 {
-    let client = init_worker_client(&worker_config, *client.into().into_inner());
+    let client_inner = *client.into().into_inner();
+    let endpoint = client_inner.options().clone().target_url;
```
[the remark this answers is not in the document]
Sushisource: Yep that's right

core/src/lib.rs
Sushisource: So, yeah, the config for these workers should just be entirely in our control, and not derived from some of the other worker's configs.

For example, if you look at how we initialize replay workers, we set a lot of values to 1 or disable things, same story here really.

Off the top of my head we probably want:
* Disable workflow and activity polling - this will happen by just not calling the poll APIs, but we can also explicitly set `no_remote_activities: true`.
* Probably use poller autoscaling for the nexus poller with minimum/initial set to 1 and max set fairly low like 10.
* Fixed size nexus slots. Honestly not sure what the value should be here, but we can start with something low to begin with.

Pretty much everything else can just be default or off or something, whatever makes sense.

The only option that needs to be derived from an existing worker here is the newly-added `heartbeat_interval`. The problem is it could be conflicting. So, we can just always pick the smallest value among workers sharing the same heartbeat. That should be added to the docstring.

core-api/src/worker.rs
```diff
@@ -33,8 +34,8 @@ pub struct WorkerConfig {
     /// Workflows are evicted according to a least-recently-used policy one the cache maximum is
     /// reached. Workflows may also be explicitly evicted at any time, or as a result of errors
     /// or failures.
-    #[builder(default = "0")]
-    pub max_cached_workflows: usize,
+    #[builder(default = "Arc::new(AtomicUsize::new(0))")]
+    pub max_cached_workflows: Arc<AtomicUsize>,
```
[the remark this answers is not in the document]
yuandrew: `AtomicUsize` doesn't implement Clone, which `WorkerConfig` derives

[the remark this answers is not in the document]
Sushisource: Sorry, I mean the config itself shouldn't change at all. It should stay a normal usize. Then you create the arc'd atomic based on it in init.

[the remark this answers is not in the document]
yuandrew: ahhh, yeah that makes sense. ty

core/src/test_help/mod.rs
```diff
@@ -180,10 +183,36 @@ pub(crate) fn mock_worker(mocks: MocksHolder) -> Worker {
                 .unwrap_or_else(|| mock_poller_from_resps([])),
         },
         None,
-        None,
+        false,
     )
 }
 
+pub(crate) fn mock_worker_with_heartbeat(mock: MockWorkerClient, config: WorkerConfig) -> Worker {
```
yuandrew, line 190: not used rn, will use for the test i need to fix up, feel free to ignore for now

core/src/worker/mod.rs
```diff
@@ -121,8 +126,206 @@ pub struct Worker {
     local_activities_complete: Arc<AtomicBool>,
     /// Used to track all permits have been released
     all_permits_tracker: tokio::sync::Mutex<AllPermitsTracker>,
-    /// Used to shutdown the worker heartbeat task
-    worker_heartbeat: Option<WorkerHeartbeatManager>,
+    worker_heartbeat_data: Option<Arc<Mutex<WorkerHeartbeatData>>>,
```
Sushisource, line 129: Similar issue here - `WorkerHeartbeatData` is mixing a lot of concerns. Leave data as just data, and separate out the behavior. My comments on `capture_heartbeat` / `fetch_config` get at that.

Like I had in my comment on the last review round - we can just have `get_worker_heartbeat_data()`, which you do have, but we don't need to actually store data - you can just construct a new one every time it's called.

core/src/worker/heartbeat.rs
```diff
@@ -1,90 +1,270 @@
...
+                    "Attempted to remove key {} that was already removed",
+                    worker_instance_key
+                );
+            }
+        })
     }
 }
-
+// TODO: rename
+// TODO: impl trait so entire struct doesn't need to be passed to Worker and SharedNamespaceWorker
```
Sushisource, line 265: Shouldn't be necessary anyway, per my other comments.

core/src/worker/mod.rs
```diff
@@ -404,13 +621,15 @@ impl Worker {
                 };
 
                 let np_metrics = metrics.with_new_attrs([nexus_poller()]);
+                // This starts the poller thread.
```
Sushisource, line 624: Not really a thread - tasks. But, kind of an uninteresting comment anyway

core/src/abstractions.rs
```diff
@@ -36,8 +36,8 @@ pub(crate) struct MeteredPermitDealer<SK: SlotKind> {
     /// is at this number, no more permits will be requested from the supplier until one is freed.
     /// This avoids requesting slots when we are at the workflow cache size limit. If and when
     /// we add user-defined cache sizing, that logic will need to live with the supplier and
-    /// there will need to be some associated refactoring.
-    max_permits: Option<usize>,
+    /// there will need to be some associated refactoring. // TODO: sounds like we'll need to do this
+    max_permits: Option<Arc<AtomicUsize>>,
```
Sushisource, line 40: Potentially, yes. Let's leave proper implementations of the commands for later though, since this PR is already quite large, and they won't be working server side for a while anyway.

Issue #965, opened 2025-07-24
[Bug] Support relaying sync Nexus operation failures in Nexus start resolution


cretz:
### Describe the bug

Today unlike Go and Java, Core's Nexus resolving of Nexus operation start _always_ succeeds. To match Go and Java, and general expectations, we need a way for start to resolve with failure if `EVENT_TYPE_NEXUS_OPERATION_FAILED` comes before `EVENT_TYPE_NEXUS_OPERATION_STARTED` and not have to do a check-sync-started-plus-lookahead to know whether start actually failed.

PR #967, opened 2025-07-25
Fail Nexus start operations on sync failure

Sushisource:
## What was changed
Changed core to fail nexus start ops when failure happens in sync operations, rather than asking lang to look ahead to the operation resolution.

## Why?
Easier for lang. Not really a bug, this was working as intended, but it can be easier this way.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/965

2. How was this tested:
Existing tests.

3. Any docs updates needed?

### DIFF
--- a/sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
+++ b/sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
@@ -350,10 +350,9 @@ message ResolveNexusOperationStart {
         // If true the operation "started" but only because it's also already resolved. A
         // `ResolveNexusOperation` job will be in the same activation.
         bool started_sync = 3;
-        // The operation was cancelled before it was ever sent to server (same WFT).
-        // Note that core will still send a `ResolveNexusOperation` job in the same activation, so
-        // there does not need to be an exceptional case for this in lang.
-        temporal.api.failure.v1.Failure cancelled_before_start = 4;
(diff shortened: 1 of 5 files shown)


sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
```diff
@@ -350,10 +350,9 @@ message ResolveNexusOperationStart {
         // If true the operation "started" but only because it's also already resolved. A
         // `ResolveNexusOperation` job will be in the same activation.
         bool started_sync = 3;
-        // The operation was cancelled before it was ever sent to server (same WFT).
-        // Note that core will still send a `ResolveNexusOperation` job in the same activation, so
-        // there does not need to be an exceptional case for this in lang.
-        temporal.api.failure.v1.Failure cancelled_before_start = 4;
+        // The operation either failed to start, was cancelled before it started, timed out, or
+        // failed synchronously. Details are included inside the message.
+        temporal.api.failure.v1.Failure failed = 4;
```
cretz, line 355: Can you confirm (maybe in comment too) that this means the resolve job will never be sent?

cretz (approved):

sdk-core-protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
Sushisource: Yes

PR #971, opened 2025-08-04
Fairness protos

Sushisource:
> [!WARNING]
> DRAFT until we have a CLI RC with this in it that can be used for tests 

## What was changed
Updated API to 1.51.0 & add fairness keys/weights

## Why?
New feature complete

## Checklist

1. Closes 

2. How was this tested:
Updated tests

3. Any docs updates needed?


tests/cloud_tests.rs
```diff
@@ -56,8 +20,9 @@ async fn tls_test() {
 #[tokio::test]
 async fn grpc_message_too_large_test() {
     let wf_name = "oversize_grpc_message";
-    let mut starter =
-        CoreWfStarter::new_with_client(wf_name, get_client("grpc_message_too_large").await);
+    let mut starter = CoreWfStarter::new_cloud_or_local(wf_name, "1.26")
```
yuandrew, line 23: Should this test live in `shared_tests/` if we're using `new_cloud_or_local` here?

yuandrew (commented):
big fan of the test change, feels like an objective improvement with minimal extra effort

tests/cloud_tests.rs
Sushisource: It can, was gonna move it.

PR #986, opened 2025-08-23
feat: Expose envconfig functionality through C bridge

THardy98:
## What was changed
C bridge code for .NET env config

1. Part of https://github.com/temporalio/sdk-dotnet/issues/490
2. How was this tested:
Integration tests lang-side

3. Any docs updates needed?
No

### DIFF
--- a/core-c-bridge/src/client.rs
+++ b/core-c-bridge/src/client.rs
@@ -685,7 +685,7 @@ async fn call_workflow_service(
         "UpdateWorkerBuildIdCompatibility" => {
             rpc_call!(client, call, update_worker_build_id_compatibility)
         }
-        rpc => Err(anyhow::anyhow!("Unknown RPC call {}", rpc)),
+        rpc => Err(anyhow::anyhow!("Unknown RPC call {rpc}")),
     }
 }
 
@@ -715,7 +715,7 @@ async fn call_operator_service(
         "UpdateNexusEndpoint" => {
             rpc_call_on_trait!(client, call, OperatorService, update_nexus_endpoint)
         }
-        rpc => Err(anyhow::anyhow!("Unknown RPC call {}", rpc)),
+        rpc => Err(anyhow::anyhow!("Unknown RPC call {rpc}")),
     }
 }
 
@@ -785,7 +785,7 @@ async fn call_cloud_service(client: &CoreClient, call: &RpcCallOptions) -> anyho
         "GetConnectivityRule" => rpc_call!(client, call, get_connectivity_rule),
         "GetConnectivityRules" => rpc_call!(client, call, get_connectivity_rules),
         "DeleteConnectivityRule" => rpc_call!(client, call, delete_connectivity_rule),
-        rpc => Err(anyhow::anyhow!("Unknown RPC call {}", rpc)),
+        rpc => Err(anyhow::anyhow!("Unknown RPC call {rpc}")),
     }
 }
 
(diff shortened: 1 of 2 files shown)


core-c-bridge/src/envconfig.rs
```diff
@@ -0,0 +1,251 @@
...
+
+        Ok(core_config.into())
+    };
+
+    send_result(result(), user_data, callback);
+}
+
+/// Load a single client profile from given sources with env overrides
+#[unsafe(no_mangle)]
+pub extern "C" fn temporal_core_client_config_profile_load(
```
cretz, line 212: Can you be a bit more specific in docs on what the callback receives on success and make it clear that the user must free callback success/fail?

cretz, line 212: There are several parameters here. I think it'd be ok if we accepted a `ClientConfigProfileLoadOptions` struct with all of this in it instead. Similarly, would have a `ClientConfigLoadOptions` for the other exposed function.

cretz (commented):
Minor suggestions. Also, can we submit use of this (which I am guessing is at https://github.com/temporalio/sdk-dotnet/pull/509) for PR review before merging this just to make sure we're good there from a review POV? We can basically merge both at once when both approved (obviously this first and submodule update over there).

core-c-bridge/src/envconfig.rs
```diff
@@ -0,0 +1,251 @@
...
+        Err(e) => {
+            let err = ByteArray::from_utf8(e);
+            unsafe { callback(user_data, std::ptr::null(), err.into_raw()) };
+        }
+    }
+}
+
+/// Load all client profiles from given sources
+#[unsafe(no_mangle)]
+pub extern "C" fn temporal_core_client_config_load(
```
cretz, line 183: Was reviewing .NET and noticed - this doesn't need to be async w/ a callback. Can just return the result or failure. Can see calls like `temporal_core_runtime_new` or `temporal_core_worker_new` to see how they return success or fail.

cretz:
Let me know if/when the PR is ready for re-review (if you want, you can respond to comments to make it easier to know which are agreed with and/or handled)

core-c-bridge/src/envconfig.rs
THardy98: Returns `ClientConfigOrFail` now

core-c-bridge/src/envconfig.rs
```diff
@@ -0,0 +1,321 @@
...
+) -> ClientConfigOrFail {
+    if options.is_null() {
+        let err = ByteArray::from_utf8("Options cannot be null".to_string());
+        return ClientConfigOrFail {
+            success: std::ptr::null(),
+            fail: err.into_raw(),
+        };
+    }
+
+    let result = || -> Result<ClientConfig, String> {
```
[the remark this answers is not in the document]
THardy98: Returning a `Result` allows us to use the `?` operator, which is a bit nicer than writing out the `match` cases for each call that returns `Result` (`parse_env_vars`, `parse_config_source`, `load_client_config`)

core-c-bridge/include/temporal-sdk-core-c-bridge.h
```diff
@@ -240,6 +240,53 @@ typedef void (*TemporalCoreClientRpcCallCallback)(void *user_data,
...
                                                   const struct TemporalCoreByteArray *failure_details);
 
+/**
+ * OrFail result for client config loading operations.
+ * Either success or fail will be null, but never both.
+ * If success is not null, it contains JSON-serialized client configuration data.
+ * If fail is not null, it contains UTF-8 encoded error message.
+ * The returned ByteArrays must be freed by the caller.
+ */
+typedef struct TemporalCoreClientConfigOrFail {
```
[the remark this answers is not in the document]
THardy98: Not at all - renamed all structs to have `Env` before `Config`. Renamed the function names `config` -> `env_config`

core-api/src/envconfig.rs
```diff
@@ -1592,6 +1592,9 @@ address = "some-address"
 
     #[test]
     fn test_load_client_config_profile_from_system_env() {
+        // WARNING: This test modifies system environment variables which can cause
+        // test pollution if tests run in parallel.
+
```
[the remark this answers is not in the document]
THardy98: Hmm the change I made to this test is working locally, but I guess CI may have more restrictions around setting env variables, given that the thread panicked:
```
thread 'envconfig::tests::test_load_client_config_profile_from
  _system_env_impl' panicked at 
  core-api/src/envconfig.rs:1627:45
```
the alternative I can think of is to revert the test change and just it serially

cretz (approved):

PR #994, opened 2025-08-28
Enable slot supplier for nexus pollers

dandavison:
Enable slot supplier for nexus pollers

See https://github.com/temporalio/sdk-python/pull/1066 for additional test coverage.


Sushisource (approved):
Nice. Thanks!

Issue #991, opened 2025-08-27
[Feature Request] Ensure gRPC binary metadata headers are supported in C bridge


cretz:
### Describe the solution you'd like

See https://github.com/temporalio/features/issues/671. Probably only have to update C bridge to send different tonic headers if key is suffixed with `-bin`. But using binary means users may want `\n` in the body which we don't support today, unsure if we want to handle that now or maybe even require users base64 `-bin`-suffixed key values it before it reaches us.

PR #997, opened 2025-08-30
Add support for configuring client-wide gRPC binary metadata values, validate metadata earlier

jazev-stripe:
## What was changed
This PR allows configuring [gRPC binary metadata key/values](https://chromium.googlesource.com/external/github.com/grpc/grpc/+/HEAD/doc/PROTOCOL-HTTP2.md#:~:text=Applications%20define%20binary%20headers%20by%20having%20their%20names%20end%20with%20%E2%80%9C%2Dbin%E2%80%9D.%20Runtime%20libraries%20use%20this%20suffix%20to%20detect%20binary%20headers%20and%20properly%20apply%20base64%20encoding%20%26%20decoding%20as%20headers%20are%20sent%20and%20received.) (where the key ends in `-bin`) on the Temporal client struct. It's already possible for Rust users of the Temporal client struct to set per-request gRPC binary metadata key/values; this allows setting client-wide headers.

The main API changes are:
- Added: `ClientOptions.binary_headers: Option<HashMap<String, Vec<u8>>>`
- Added: `ConfiguredClient::set_binary_headers(&self, binary_headers: HashMap<String, Vec<u8>>) -> Result<(), InvalidHeaderError>`
- Added: `InvalidHeaderError`
- Modified: `ConfiguredClient::set_headers` (for setting ASCII headers) now returns `Err(InvalidHeaderError)` if the keys or values are invalid, **and no headers are modified**. Before, it would just silently discard/skip invalid keys or values.
- Modified: `ClientOptions::connect` (and the related methods) now returns an `Err(ClientInitError::InvalidHeaders(InvalidHeaderError))` if any of the keys or values passed in `ClientOptions` are invalid

The approach taken is to parse the metadata keys/values early, and store an already-parsed representation (that can be infallibly applied to the outbound metadata, in `ClientHeaders ::apply_to_metadata`).

This PR is depended on by https://github.com/temporalio/sdk-python/pull/1070

## Why?
https://github.com/temporalio/sdk-core/issues/991

(for the specific parts in Core SDK that are needed for the Python SDK; this doesn't cover all Core-SDK changes that are needed (for e.g. the Rust SDK or the C bridge)).

## Checklist
1. ~~Closes~~
   - This is a partial implementation of https://github.com/temporalio/sdk-core/issues/991, and is a prereq to closing https://github.com/temporalio/sdk-python/issues/1063

2. How was this tested:
    - See added/modified unit tests in `client/src/lib.rs`
    - Existing tests pass (so the behavior for ASCII headers shouldn't have regressed)
    - In https://github.com/temporalio/sdk-python/pull/1070, I modified the client tests to cover both per-request gRPC binary metadata and client-wide gRPC binary metadata

3. Any docs updates needed?
    - I don't think so? Afaik there are no docs on the Core SDK from a "Rust integrators" perspective. The updates to docs.rs should be sufficient IMO.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -150,9 +150,9 @@ pub struct ClientOptions {
 
     /// HTTP headers to include on every RPC call.
     ///
-    /// These must be valid gRPC metadata keys, and must not end with a `-bin` suffix (to set binary
-    /// headers, see [ClientOptions::binary_headers]). Invalid header keys will cause an error to be
-    /// returned when connecting.
+    /// These must be valid gRPC metadata keys, and must not be binary metadata keys (ending in
+    /// `-bin). To set binary headers, use [ClientOptions::binary_headers]. Invalid header keys or
+    /// values will cause an error to be returned when connecting.
     #[builder(default)]
     pub headers: Option<HashMap<String, String>>,
 
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -156,8 +156,7 @@ pub struct ClientOptions {
     #[builder(default)]
     pub headers: Option<HashMap<String, String>>,
 
-    /// HTTP headers to include on every RPC call as binary gRPC metadata (typically encoded as
-    /// base64).
+    /// HTTP headers to include on every RPC call as binary gRPC metadata (encoded as base64).
     ///
     /// These must be valid binary gRPC metadata keys (and end with a `-bin` suffix). Invalid
     /// header keys will cause an error to be returned when connecting.
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -338,7 +338,7 @@ pub enum ClientInitError {
     #[error("Invalid URI: {0:?}")]
     InvalidUri(#[from] InvalidUri),
     /// Invalid gRPC metadata headers. Configuration error.
-    #[error("Invalid headers: {0:?}")]
+    #[error("Invalid headers: {0}")]
     InvalidHeaders(#[from] InvalidHeaderError),
     /// Server connection error. Crashing and restarting the worker is likely best.
     #[error("Server connection error: {0:?}")]
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -32,6 +32,9 @@ pub use temporal_sdk_core_protos::temporal::api::{
     },
 };
 pub use tonic;
+use tonic::metadata::{
+    AsciiMetadataKey, AsciiMetadataValue, BinaryMetadataKey, BinaryMetadataValue,
+};
 pub use worker_registry::{Slot, SlotManager, SlotProvider, WorkerKey};
 pub use workflow_handle::{
     GetWorkflowResultOpts, WorkflowExecutionInfo, WorkflowExecutionResult, WorkflowHandle,
@@ -148,6 +151,13 @@ pub struct ClientOptions {
     /// HTTP headers to include on every RPC call.
     #[builder(default)]
     pub headers: Option<HashMap<String, String>>,
+
+    /// HTTP headers to include on every RPC call as binary gRPC metadata (typically encoded as
+    /// base64).
+    ///
+    /// These must be valid binary gRPC metadata keys (and end with a `-bin` suffix).
+    #[builder(default)]
+    pub binary_headers: Option<HashMap<String, Vec<u8>>>,
 
     /// API key which is set as the "Authorization" header with "Bearer " prepended. This will only
     /// be applied if the headers don't already have an "Authorization" header.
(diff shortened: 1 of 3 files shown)


Sushisource:
@jazev-stripe I haven't reviewed this since it's still DRAFT. Thanks for contributing though! Anything you need to un-draft? Happy to review it once it is.

jazev-stripe:
@Sushisource sorry about the lack of details; I've been meaning to finish this PR (and the corresponding Python one).

I have a thread in the Stripe support channel in the Temporal Slack about this with @cretz where we've discussed the shape of the change, and I think we're mostly aligned [0]. I'm aiming to get this PR in a more reviewable state by the end of the week (once I get a spare couple hours at work🤞).

[0] One change I think I'll make based on our discussion is to make `ClientHeaders.user_binary_headers` and `ClientHeaders.user_headers` store pre-validated `tonic` metadata keys/values, so that applying them to the metadata is infallible (and instead return errors earlier, when the client is constructed or when `set_headers`/`set_binary_headers` is called.

client/src/lib.rs
```diff
@@ -558,6 +638,57 @@ impl ClientOptions {
...
 
+fn parse_ascii_headers(
+    headers: HashMap<String, String>,
+) -> Result<HashMap<AsciiMetadataKey, AsciiMetadataValue>, InvalidHeaderError> {
+    let mut parsed_headers = HashMap::with_capacity(headers.len());
+    for (k, v) in headers.into_iter() {
+        let key = match AsciiMetadataKey::from_str(&k) {
+            Ok(key) => key,
+            Err(err) => {
+                return Err(InvalidHeaderError::InvalidAsciiHeaderKey {
```
jazev-stripe, line 649: This is somewhat verbose, but I used it instead of `.map_err` to avoid cloning the `key`, in the case that we need to return an error

client/src/lib.rs
```diff
@@ -558,6 +638,57 @@ impl ClientOptions {
...
+                    key: k,
+                    source: err,
+                });
+            }
+        };
+        let value = match MetadataValue::from_str(&v) {
+            Ok(value) => value,
```
jazev-stripe, line 656: Same here (for the `value`)

client/src/lib.rs
```diff
@@ -331,6 +349,37 @@ pub enum ClientInitError {
...
+    /// An ASCII header key was invalid
+    #[error("Invalid ASCII header key '{key}': {source}")]
+    InvalidAsciiHeaderKey {
+        /// The invalid key
+        key: String,
+        /// The source error from tonic
+        source: tonic::metadata::errors::InvalidMetadataKey,
+    },
+    /// An ASCII header value was invalid
+    #[error("Invalid ASCII header value for key '{key}': {source}")]
```
jazev-stripe, line 372: I didn't include the invalid value here because I was a bit worried that including it in error messages may lead to a credentials logging risk for users (if headers are used for security values, for example).

Let me know if that makes sense, or if I should make any changes

Sushisource (approved):
Looking good to me. Thanks for the contribution! Just one potential docstring update.

client/src/lib.rs
```diff
@@ -146,9 +149,21 @@ pub struct ClientOptions {
...
     /// HTTP headers to include on every RPC call.
+    ///
+    /// These must be valid gRPC metadata keys, and must not be binary metadata keys (ending in
+    /// `-bin). To set binary headers, use [ClientOptions::binary_headers]. Invalid header keys or
+    /// values will cause an error to be returned when connecting.
     #[builder(default)]
     pub headers: Option<HashMap<String, String>>,
 
+    /// HTTP headers to include on every RPC call as binary gRPC metadata (typically encoded as
+    /// base64).
```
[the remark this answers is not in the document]
jazev-stripe: When I was doing research, I saw https://github.com/grpc/proposal/blob/HEAD/G1-true-binary-metadata.md, which involves sometimes **not** using base64 for binary headers, if negotiated. However, it's unclear to me if that proposal is actually used in practice (or what the state of it is). It doesn't _seem_ accepted based on the header of the doc, but I also see [references to that symbol in the gRPC C++ codebase](https://github.com/search?q=repo%3Agrpc%2Fgrpc%20GRPC_ALLOW_TRUE_BINARY_METADATA&type=code) (possibly in a partially-implemented state).

Either way, I think it's almost always encoded as base64, and whatever the encoding is, it's internal to the gRPC transport mechanism. I'll just remove "typically"; this distinction probably doesn't matter much for consumers. The doc-comments on tonic seem to reflect that it only uses base64 too: https://docs.rs/tonic/latest/tonic/metadata/struct.MetadataValue.html

Issue #993, opened 2025-08-28
[Feature Request] Nexus should deduplicate and lowercase inbound headers


cretz:
### Describe the solution you'd like

Today, server returns `Request-Timeout` _and_ `request-timeout` headers apparently for backwards compatibility reasons with some other GA Nexus SDKs. We believe Core should lowercase all headers and ignore duplicates (regardless of case). This is being done in some languages manually at the moment (like .NET) but deserves to be done in Core.

This should probably also apply to callback headers, but confirm with Nexus team first.

PR #1006, opened 2025-09-09
Lowercaseify Nexus request headers

Sushisource:
## What was changed
Lowercaseify Nexus start request headers

## Why?
So lang doesn't see weird duplicated camel and kebab case of the same header.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/993

2. How was this tested:
Existing integ test with modification

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/nexus.rs
+++ b/core/src/worker/nexus.rs
@@ -42,7 +42,7 @@ use tokio::{
 use tokio_stream::wrappers::UnboundedReceiverStream;
 use tokio_util::sync::CancellationToken;
 
-static REQUEST_TIMEOUT_HEADER: &str = "Request-Timeout";
+static REQUEST_TIMEOUT_HEADER: &str = "request-timeout";
 
 /// Centralizes all state related to received nexus tasks
 pub(super) struct NexusManager {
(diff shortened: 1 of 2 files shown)


sdk-core-protos/src/utilities.rs
```diff
@@ -26,3 +28,13 @@ pub fn pack_any<T: Message>(
     Message::encode(msg, &mut value)?;
     Ok(prost_wkt_types::Any { type_url, value })
 }
+
+/// Given a header map, lowercase all the keys and return it as a new map.
+/// Any keys that are duplicated after lowercasing will clobber each other in underfined ordering.
```
yuandrew, line 33:
```suggestion
/// Any keys that are duplicated after lowercasing will clobber each other in undefined ordering.
```

core/src/worker/nexus.rs
```diff
@@ -245,11 +249,18 @@ where
...
                 let res = match t {
                     TaskStreamInput::Poll(t) => match *t {
-                        Ok(t) => {
+                        Ok(mut t) => {
                             if let Some(dur) = t.resp.sched_to_start() {
                                 self.metrics.nexus_task_sched_to_start_latency(dur);
                             };
 
+                            if let Some(ref mut req) = t.resp.request {
+                                req.header = normalize_http_headers(std::mem::take(&mut req.header));
```
yuandrew, line 258: why do we `std::mem::take` here? Just for safety, to make sure that memory is cleared out?

yuandrew (approved):

sdk-core-protos/src/utilities.rs
Sushisource: Hahah derfined

cretz (commented):
(sorry, forgot to publish this review, too late now)

Issue #1004, opened 2025-09-08
[Bug] Native certs used for TLS client even when server root CA override present


cretz:
### Describe the bug

We are using native roots even when a server CA cert is provided, see https://github.com/temporalio/sdk-core/blob/3a6e5ffc18a717be7a3670d1880882d70c92af80/client/src/lib.rs#L531-L536. Tonic fails if this setting is set and it doesn't find native roots, see https://github.com/hyperium/tonic/blob/d02ee2d1dba2f77570097cbbd87a611399ea98c1/tonic/src/transport/channel/service/tls.rs#L65-L74.

Users should be allowed to provide a CA override on a system that doesn't have native roots. However, we may have to factor in backwards compatibility of removing native roots for those that may (incorrectly) be relying on them _and_ passing in a root override.

cretz:
One option pointed out by @maciejdudko is to try to build with native roots, and if it fails with the `TlsError::NativeCertsNotFound` and has a CA override set, try again without. Arguably though, when you want a CA override, you don't want native roots to be in the config at all, you want client connections to fail if they aren't using your expected CA override.

PR #1007, opened 2025-09-10
:boom: Only enable native roots if CA override not specified

Sushisource:
## What was changed
Don't enable native TLS roots if a CA override is specified. This is technically breaking, but only if you were doing a somewhat odd  thing in the first place.

## Why?
People using CA overrides often want to do it in an environment where there are no native roots, which this change enables.

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/1004

2. How was this tested:
Could not come up with any kind of reasonable test that wasn't dramatically more LOC than it feels worth. Open to ideas.

3. Any docs updates needed?


Sushisource:
> SDKs need to call out this breaking change in their release notes when they update Core. Not sure how we can make sure they do.

That's what the :boom: is for, and I hope when people update they see it in the log, but, yeah, we might want a script for that or something.

PR #1012, opened 2025-09-17
Improved retry policy handling

maciejdudko:
## What was changed

- For local activities, the default values are applied to retry policy fields on activity task initialization. This is encapsulated with new `ValidatedRetryPolicy` type. `RetryPolicyExt` trait is removed and `should_retry` function is moved to `ValidatedRetryPolicy`. The lang side will now see the modified retry policy with defaults applied.
- For non-local activities and for workflow activations, a safeguard was added to ensure `initial_interval` is always present in retry policy even if it was missing in server message. Note that this is different from the whole retry policy being missing (which can happen on old server versions).

## Why?

It ensures the retry policy stored in local activity info is the same as the one actually used for retries. It also allows lang side to assume the invariant that a retry policy always has some initial interval set.

## Checklist

1. How was this tested: modified tests in `core/src/retry_logic.rs`.


core/src/worker/activities/local_activities.rs
```diff
@@ -570,7 +570,7 @@ impl LocalActivityManager {
             macro_rules! calc_backoff {
                 ($fail: ident) => {
                     info.la_info.schedule_cmd.retry_policy.should_retry(
-                        info.attempt as usize,
+                        info.attempt.try_into().unwrap_or(NonZero::<u32>::MIN),
```
Sushisource, line 573: Hahah this is a very elaborate way of saying `1`

core/src/retry_logic.rs
```diff
@@ -1,177 +1,323 @@
...
-            Some(tried_mul.unwrap_or(max_iv).min(max_iv))
-        } else {
-            // No retries if initial interval is not specified
-            None
-        }
+        let delay = i32::try_from(attempt_number.get())
+            .ok()
+            .and_then(|attempt| {
+                let factor = self.backoff_coefficient.powi(attempt - 1);
+                Duration::try_from_secs_f64(factor * self.initial_interval.as_secs_f64()).ok()
```
Sushisource, line 109: Oh, this is non-nightly now! Nice!

Sushisource (approved):
Looking good!

PR #1025, opened 2025-09-30
💥 C Bridge: Custom slot supplier rework

maciejdudko:
## What was changed

**💥 BREAKING CHANGE:** Redesigned the code flow around custom slot supplier in C bridge to ensure memory safety in case of cancellation. From Lang's point of view it works mostly the same as before, with the following differences:
- All callbacks take an extra `user_data` argument (not used by .Net implementation but very useful for other languages).
- In `reserve`, the `ctx` argument is short-lived, and `completion_ctx` replaces `sender`.
- `token_source` is removed from `SlotReserveCtx` and function `temporal_core_set_reserve_cancel_target` is removed.
- `cancel_reserve` takes `completion_ctx` argument instead of `token_source`.
- `temporal_core_complete_async_reserve` is safe to call after cancellation, making it possible to avoid race condition. This function now returns bool indicating whether the reservation was completed or cancelled.
- Added separate `temporal_core_complete_async_cancel_reserve` that has to be called after cancellation to clean up resources.

Additionally, implemented `available_slots` functionality, and renamed the callback type aliases to make their relation to `CustomSlotSupplierCallbacks` more obvious.

There is a matching PR in .Net SDK that implements the new API: https://github.com/temporalio/sdk-dotnet/pull/532

## Why?

The previous implementation had memory corruption bugs, see https://github.com/temporalio/sdk-dotnet/issues/458

## Checklist

1. Part of https://github.com/temporalio/sdk-dotnet/issues/458

2. How was this tested:
**There is no test for the original bug.** Due to the nature of this bug, it's impossible to write a test that reliably triggers the problematic conditions. The fix can be verified over time by the lack of transient CI failures in .Net SDK caused by this bug.

General tests of the custom slot supplier exercising these APIs are done inside .Net SDK, see https://github.com/temporalio/sdk-dotnet/pull/532

### DIFF
--- a/core-c-bridge/src/worker.rs
+++ b/core-c-bridge/src/worker.rs
@@ -207,15 +207,15 @@ pub struct CustomSlotSupplierCallbacks {
     /// to reserve a slot until it eventually succeeds, or until reservation is cancelled by Core.
     pub reserve: CustomSlotSupplierReserveCallback,
     /// Called to cancel slot reservation. `completion_ctx` specifies which reservation is being
-    /// cancelled; the matching [`reserve`] call was made with the same `completion_ctx`. After
-    /// cancellation, the implementation should call [`temporal_core_complete_async_cancel_reserve`]
+    /// cancelled; the matching [`reserve`](Self::reserve) call was made with the same `completion_ctx`.
+    /// After cancellation, the implementation should call [`temporal_core_complete_async_cancel_reserve`]
     /// with the same `completion_ctx`. Calling [`temporal_core_complete_async_reserve`] is not
     /// needed after cancellation.
     pub cancel_reserve: CustomSlotSupplierCancelReserveCallback,
     /// Called to try an immediate slot reservation. The callback should return 0 if immediate
     /// reservation is not currently possible, or permit ID if reservation was successful. Permit ID
-    /// is arbitrary, but must be unique among live reservations as it's later used for `mark_used`
-    /// and `release` callbacks.
+    /// is arbitrary, but must be unique among live reservations as it's later used for [`mark_used`](Self::mark_used)
+    /// and [`release`](Self::release) callbacks.
     pub try_reserve: CustomSlotSupplierTryReserveCallback,
     /// Called after successful reservation to mark slot as used. See [`SlotSupplier`](temporal_sdk_core_api::worker::SlotSupplier)
     /// trait for details.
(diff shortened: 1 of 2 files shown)


core-c-bridge/src/worker.rs
```diff
@@ -249,31 +293,64 @@ pub enum SlotInfo {
...
+                        inner.user_data,
+                    ),
+                }
             }
         }
     }
 }
-unsafe impl Send for CancelReserveGuard {}
+
+//unsafe impl<'a, SK: SlotKind + Send + Sync> Send for CancelReserveGuard<'a, SK> {}
```
Sushisource, line 353: Leftover comment

core-c-bridge/src/worker.rs
```diff
@@ -249,31 +293,64 @@ pub enum SlotInfo {
...
+        if !self.completed {
+            let state = self
+                .completion_ctx
+                .state
+                .swap(SlotReserveOperationState::Cancelled);
             unsafe {
-                (self.callback)(self.token_src);
+                let inner = &*self.slot_supplier.inner.0;
+                match state {
+                    SlotReserveOperationState::Cancelled => (), // should not be possible but it's safe
```
Sushisource, line 336: Might be worth adding a debug-only panic here

core-c-bridge/src/worker.rs
```diff
@@ -282,36 +359,53 @@ impl<SK: SlotKind + Send + Sync> temporal_sdk_core_api::worker::SlotSupplier
...
 
     fn mark_slot_used(&self, ctx: &dyn SlotMarkUsedContext<SlotKind = Self::SlotKind>) {
         let ctx = SlotMarkUsedCtx {
             slot_info: Self::convert_slot_info(ctx.info().downcast()),
-            slot_permit: ctx.permit().user_data::<usize>().copied().unwrap_or(0),
+            slot_permit: ctx
+                .permit()
+                .user_data::<usize>()
+                .copied()
+                .expect("Failed to read custom slot supplier permit ID"),
```
Sushisource, line 404: Small stylistic note, but I think it's better when `expect` calls read as what should be expected. In this case, something like "Custom slot supplier permit ID should always be a usize"

core-c-bridge/src/worker.rs
```diff
@@ -939,29 +1043,84 @@ pub extern "C" fn temporal_core_worker_replay_push(
...
+    if completion_ctx.is_null() {
+        panic!("completion_ctx is null");
+    }
+    let state = unsafe { (*completion_ctx).state.load() };
+    match state {
+        SlotReserveOperationState::Cancelled => {
+            drop(unsafe { Arc::from_raw(completion_ctx) });
+            true
+        }
+        SlotReserveOperationState::Pending => false,
```
Sushisource, line 1118: According to the docstring seems like this should also be a panic? It does seem wrong to call this if the cancel didn't happen

maciejdudko: While it could panic, there is no reason to panic as no invariants have been broken. Safely returning false here may make it easier to make the implementation safe. The other branch panics because it potentially reads from freed memory.

maciejdudko: I revised the docstring to make the intention clearer.

PR #1029, opened 2025-10-08
Cgroup cpu usage

### LINKED ISSUE #903: [Feature Request] Support CPU Cgroup Limits in Resource-based tuner

VegetarianOrc:
## What was changed

Update RealSysInfo to read cgroupv2 CPU limits when the sysinfo crate has cgroup memory limits available.

## Why?

When running in a container it's more appropriate to report the CPU usage as a factor of it's limit rather than the host system.

## Checklist

1. Closes #903 

2. How was this tested:

- Added unit tests for reading the cgroupv2 CPU files and calculating usage based on their contents.
- Added unit tests that confirm RealSysInfo respects cgroup limits. These tests skip when not running in a cgroup. A small helper script is included at `docker-cgroup-tests.sh` to cross-compile tests and execute only cgroup tests via docker.

3. Any docs updates needed?

### DIFF
--- a/core/src/worker/tuner/resource_based.rs
+++ b/core/src/worker/tuner/resource_based.rs
@@ -513,16 +513,16 @@ impl RealSysInfoInner {
             );
 
             let cpu = self.cgroup_cpu_info.calc_cpu_percent().unwrap_or_else(|| {
-                //there won't be a cgroup cpu usage if there is no limit applied to the cgroup
-                //or if an error is encountered when reading cpu.stat or cpu.max
-                //in these cases, fallback to global cpu usage
+                // There won't be a cgroup cpu usage if there is no limit applied to the cgroup
+                // or if an error is encountered when reading cpu.stat or cpu.max.
+                // In these cases, fallback to global cpu usage.
                 lock.refresh_cpu_usage();
                 lock.global_cpu_usage() as f64 / 100.
             });
             self.cur_cpu_usage.store(cpu.to_bits(), Ordering::Release);
         } else {
-            //always update the total_mem b/c we could transiently fall to host if
-            // lock.cgroup_limits() returns None
+            // Always update the total_mem b/c we could transiently fall to host if
+            // lock.cgroup_limits() returns None.
             self.total_mem.store(lock.total_memory(), Ordering::Release);
             self.cur_mem_usage
                 .store(lock.used_memory(), Ordering::Release);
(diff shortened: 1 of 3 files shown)


core/src/worker/tuner/resource_based.rs
```diff
@@ -495,26 +497,40 @@ struct RealSysInfoInner {
...
                 .store(cgroup_limits.total_memory, Ordering::Release);
             self.cur_mem_usage.store(
                 cgroup_limits.total_memory - cgroup_limits.free_memory,
                 Ordering::Release,
             );
+
+            let cpu = self.cgroup_cpu_info.calc_cpu_percent().unwrap_or_else(|| {
+                //there won't be a cgroup cpu usage if there is no limit applied to the cgroup
+                //or if an error is encountered when reading cpu.stat or cpu.max
+                //in these cases, fallback to global cpu usage
```
Sushisource: Nit: I'm vaguely surprised this isn't a cargo fmt thing, but, the lack a space after your `//` freaks me out

core/src/worker/tuner/resource_based.rs
```diff
@@ -768,4 +915,204 @@ mod tests {
...
+        let cpu_percent = cgroup_info
+            .calc_cpu_percent()
+            .expect("expected usage value");
+        assert!(cpu_percent > 0., "expected cpu percent > 0");
+    }
+
+    #[test]
+    fn cgroup_realsysinfo_uses_cgroup_limits_cpu() {
+        let sys_info = RealSysInfo::new();
+        let cgroup_info = CGroupCpuInfo::new(CgroupV2CpuFileSystem);
```
Sushisource: Honestly this is one of the few spots where I don't like Rust - really this should be an integ test. It could be its own binary and only run in CI. Having to have the skips in a unit test feels a little awkward, but having to `pub` all this or put it in its own crate is even stupider.

I don't have any actual suggestion here, just whining.

... Actually maybe I do - let's protect these with explicit env vars saying to run the tests. That way they also won't silently succeed when they're _supposed_ to be able to read stuff but can't

Sushisource (approved):
Awesome! Thanks for the work on this

PR #1042, opened 2025-10-23
Use deployment_build_id as part of key for eager workflow start

yuandrew:
## What was changed

Added `deployment_build_id` as key for eager workflow slot

## Why?

Valid use case that today causes an error

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> Eager slot management now keys by namespace/task_queue/build_id, allows multiple local workers per queue with randomized selection/retry, updates duplicate-registration rules, and adds `rand` dependency.
> 
> - **Client worker registry (eager workflow start)**:
>   - Key slot providers by `namespace` + `task_queue` + `deployment_build_id` (store `Vec<(Uuid, Option<String>)>`), allowing multiple providers per queue when build IDs differ.
>   - Selection retries across providers if a worker has no slot; randomized order in prod, deterministic in tests.
>   - Registration now rejects duplicates only when same `namespace`/`task_queue`/`build_id`; error message updated. Unregister removes only the specific worker and cleans up when empty.
>   - `num_providers` now counts total providers across keys.
>   - Extensive tests added/updated for retry behavior, task-queue boundaries, duplicate handling, and build ID cases.
> - **API/docs**:
>   - `WorkflowOptions.enable_eager_workflow_start` doc note updated to remove BuildID incompatibility warning (still experimental).
> - **Dependencies**:
>   - Add `rand` for provider selection shuffling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 427dd961c273ca0b0f19c8cce8b254d64f52ca74. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/common/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
+++ b/crates/common/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
@@ -92,6 +92,8 @@ message WorkflowActivation {
     // build id, if this worker was using the deprecated Build ID-only
     // feature(s).
     common.WorkerDeploymentVersion deployment_version_for_current_task = 9;
+    // The last seen SDK version from the most recent WFT completed event
+    string last_sdk_version = 10;
 }
 
 message WorkflowActivationJob {
--- a/crates/common/src/protos/mod.rs
+++ b/crates/common/src/protos/mod.rs
@@ -519,6 +519,7 @@ pub mod coresdk {
                 history_size_bytes: 0,
                 continue_as_new_suggested: false,
                 deployment_version_for_current_task: None,
+                last_sdk_version: String::new(),
             }
         }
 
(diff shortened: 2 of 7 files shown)


crates/client/src/worker_registry/mod.rs
```diff
@@ -78,15 +79,18 @@ impl ClientWorkerSetImpl {
...
-            let deployment_options = worker.deployment_options();
-            return Some(SlotReservation {
-                slot,
-                deployment_options,
-            });
+        if let Some(worker_list) = self.slot_providers.get(&key) {
+            let worker_id = worker_list.choose(&mut rand::rng());
+            if let Some(worker_entry) = worker_id
+                && let Some(worker) = self.all_workers.get(&worker_entry.0)
+                && let Some(slot) = worker.try_reserve_wft_slot()
```
Sushisource: In principle we should then try the next worker in the list if this reservation fails. In practice I doubt this is going to come up much. Probably not hard to add, but maybe annoying to test.

Sushisource (approved):

PR #1049, opened 2025-10-30
Antithesis instrumentation

THardy98:
## What was changed
Add `antithesis_assertions` feature flag to core SDK. Minimal assertions included, but can be added with more usage.

## Why?
antithesis testing

2. How was this tested:
It's not really, just compiles

### DIFF
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -43,7 +43,7 @@ jobs:
 
   test:
     name: Unit Tests
-    timeout-minutes: 10
+    timeout-minutes: 15
     strategy:
       fail-fast: false
       matrix:
@@ -123,7 +123,7 @@ jobs:
 
   integ-tests:
     name: Integ tests
-    timeout-minutes: 20
+    timeout-minutes: 25
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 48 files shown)


chris-olszewski (approved):

crates/sdk-core/src/antithesis.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub(crate) fn ensure_init() {}
+
+    /// No-op assertion when feature is disabled.
+    macro_rules! assert_always {
+        ($condition:expr, $message:literal, $details:expr) => {{
+            let _ = ($condition, $message);
+            let _ = $details;
+        }};
+        ($condition:expr, $message:literal) => {{
+            let _ = ($condition, $message);
```
Sushisource: I think assigning things like this is actually going to cause the compiler not to optimize some stuff away.

Honestly, I'm not sure we want to do this technique at all. It's slightly more wordy, but guarding every call to `assert_always` with the flag at the callsite is safer in terms of actually making sure everything gets compiled out

THardy98:
@Sushisource just FYI, I've been holding off on this given that the Antithesis experiment finishes next week and we're blocked on upgrading core due to worker heartbeating anyways (also doesn't seem worth to cherry-pick this commit into SDKs). It's more useful to investigate what we're getting already with the Antithesis + omes runs as of right now.

Depending on the success of the experiment, i'll either address + merge or just close the PR.

Sushisource (approved):

PR #1075, opened 2025-12-09
fix: resolve local activities in order of history events

### LINKED ISSUE #1744: [Bug] NDE replaying nested promises
### Describe the bug

The following code throws an NDE during replay 

```


    await Promise.all(
        [1, 3].map(async (offset) => {
            await Promise.all(
                ["a", "b", "c", "d", "e"].map(async (c) => {

                    console.log(`sleepRandomLocal start ${offset} ${c}`);


                    printActivities();
                    //schedule local activity
                    await sleepRandomLocal(1, 100, `${offset} ${c}`);
                    

                    console.log(`sleepRandom start ${offset} ${c}`);
                    printActivities();
                    //schedule activity
                    await sleepRandom(offset * 1000, offset * 1000, `${offset} ${c}`);
                })
            );

            console.log(`greet start ${offset}`);
            printActivities();
            //schedule activity
            await greet(offset.toString());
        })
    );


    printActivities();

    // Throw an error to force replay
    throw new Error("Something went wrong");

```

```
2025-07-15T08:54:54.526968Z  WARN temporal_sdk_core::worker::workflow: Failing workflow task run_id=01980d4b-3cd8-7271-abfb-0c98f28d3e91 failure=Failure { failure: Some(Failure { message: "[TMPRL1100] Nondeterminism error: No command scheduled for event HistoryEvent(id: 38, ActivityTaskScheduled)", source: "", stack_trace: "", encoded_attributes: None, cause: None, failure_info: 
Some(ApplicationFailureInfo(ApplicationFailureInfo { r#type: "", non_retryable: false, details: None, next_retry_delay: None, category: Unspecified })) }), force_cause: NonDeterministicError }

```

I added a log to the interceptor and activities seems to be scheduled in the same order (considering only activity type) 

**Runtime**

`sleepRandomLocal-1 a, sleepRandomLocal-1 b, sleepRandomLocal-1 c, sleepRandomLocal-1 d, sleepRandomLocal-1 e, sleepRandomLocal-3 a, sleepRandomLocal-3 b, sleepRandomLocal-3 c, sleepRandomLocal-3 d, sleepRandomLocal-3 e, sleepRandom-1 c, sleepRandom-1 a, sleepRandom-3 e, sleepRandom-3 d, sleepRandom-1 b, sleepRandom-3 a, sleepRandom-1 e, sleepRandom-3 c, sleepRandom-3 b, sleepRandom-1 d, greet-undefined, greet-undefined`

**Replay**

`sleepRandomLocal-1 a, sleepRandomLocal-1 b, sleepRandomLocal-1 c, sleepRandomLocal-1 d, sleepRandomLocal-1 e, sleepRandomLocal-3 a, sleepRandomLocal-3 b, sleepRandomLocal-3 c, sleepRandomLocal-3 d, sleepRandomLocal-3 e, sleepRandom-1 a, sleepRandom-1 b, sleepRandom-1 c, sleepRandom-1 d, sleepRandom-1 e, sleepRandom-3 a, sleepRandom-3 b, sleepRandom-3 c, sleepRandom-3 d`

### Minimal Reproduction

here is the code shared by the customer to reproduce the issue: 

[Archive.zip](https://github.com/user-attachments/files/21229315/Archive.zip)

### Environment/Versions

version 1.12.1

chris-olszewski:
## What was changed
On replay, local activities will now be resolved in the order that their markers appear in history as opposed to their schedule order.

In order to achieve this a few things had to change:

### Peekahead Preresolutions

Previously we stored these as a map, switch to a vec so we can preserve the order we peek them in the history.

### LA machine changes
Before
image

After
Screenshot 2025-10-15 at 15 46 21

Important change here is the addition of a state between scheduling and getting resolved. This allows us to still create the LA machines as they come in, but delays resolving them until later so we can resolve them in the order we peeked the resolutions.

### Removal of `FakeMarker`
This was causing failures with the new state machine as we would end up applying workflow completion events to LA machines.

I did a quick test off of `master` and removing the `FakeMarker` that gets emitted on transitioning to `WaitingMarkerEventPreResolved` doesn't result in any test failure. I can do this change in a separate PR if desired.

## Why?
This could cause NDE errors if there were additional commands produced on each LA completion and the first scheduled LA didn't complete first e.g.
```
Promise.all([a().then(() => b()), c().then(() => d())])
```
would trigger NDE on replay if in history `c` finished before `a`.

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/1744 once TS is updated to include this commit

2. How was this tested:
Existing test suite. Added test where LAs are resolved in different order than they are scheduled.
Updated TS to this branch and the failing [nested promise replay test](https://github.com/temporalio/sdk-typescript/tree/olszewski/fix_nested_promise) now passes without the NDE

3. Any docs updates needed?
No

### DIFF
--- a/crates/sdk-core/src/worker/workflow/machines/workflow_machines.rs
+++ b/crates/sdk-core/src/worker/workflow/machines/workflow_machines.rs
@@ -836,12 +836,16 @@ impl WorkflowMachines {
                 DelayedAction::WakeLa(mk, la_dat) => {
                     let mach = self.machine_mut(mk);
                     if let Machines::LocalActivityMachine(ref mut lam) = *mach {
-                        // self.local_activity_data.insert_peeked_marker(*la_dat);
                         if lam.will_accept_resolve_marker() {
                             let resps = lam.try_resolve_with_dat((*la_dat).into())?;
                             self.process_machine_responses(mk, resps)?;
                         } else {
-                            self.local_activity_data.insert_peeked_marker(*la_dat);
+                            // Since the LA machine exists, we have encountered the LA WF command.
+                            // But since it will not accept a resolve marker, we have no use saving
+                            // this marker.
+                            // Previously, peeked markers were stored in a map, but since we are
+                            // now storing in an ordered collection we no longer can store this
+                            // unnecessary data as `apply_local_action_peeked_resolutions` will attempt to apply it.
                         }
                     }
                 }
--- a/crates/sdk-core/src/worker/workflow/machines/workflow_machines.rs
+++ b/crates/sdk-core/src/worker/workflow/machines/workflow_machines.rs
@@ -540,11 +540,31 @@ impl WorkflowMachines {
     pub(crate) fn iterate_machines(&mut self) -> Result<()> {
         let results = self.drive_me.fetch_workflow_iteration_output();
         self.handle_driven_results(results)?;
+        self.apply_local_action_peeked_resolutions()?;
         self.prepare_commands()?;
         if self.workflow_is_finished()
             && let Some(rt) = self.total_runtime()
         {
             self.metrics.wf_e2e_latency(rt);
+        }
+        Ok(())
+    }
+
+    fn apply_local_action_peeked_resolutions(&mut self) -> Result<()> {
+        while let Some(seq) = self.local_activity_data.peek_preresolution_seq() {
+            let Ok(mk) = self.get_machine_key(CommandID::LocalActivity(seq)) else {
+                // If we haven't encountered the LA schedule yet, stop processing the
+                // preresolutions.
+                break;
+            };
+            // Look to make this "safe"
+            let dat = self.local_activity_data.take_preresolution(seq).unwrap();
+            if let Machines::LocalActivityMachine(lam) = self.machine_mut(mk) {
+                let resps = lam.try_resolve_with_dat(dat)?;
+                self.process_machine_responses(mk, resps)?;
+            } else {
+                panic!("Found non-LAM for LA command");
+            }
         }
         Ok(())
     }
@@ -816,6 +836,7 @@ impl WorkflowMachines {
                 DelayedAction::WakeLa(mk, la_dat) => {
                     let mach = self.machine_mut(mk);
                     if let Machines::LocalActivityMachine(ref mut lam) = *mach {
+                        // self.local_activity_data.insert_peeked_marker(*la_dat);
                         if lam.will_accept_resolve_marker() {
                             let resps = lam.try_resolve_with_dat((*la_dat).into())?;
                             self.process_machine_responses(mk, resps)?;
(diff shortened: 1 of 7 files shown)


Sushisource (commented):
Really good fix. Love how you simplified things here.

The only thing left to do here is ensure that we're not breaking any existing histories (that wouldn't have already been broken by this bug). The way to do that is to take your new tests, and run them without your fix (multiple times since they will have different LA ordering?), save the histories (crates/sdk-core/tests/histories), and then use those in the replayer too. 

I think, with this fix, it'll be the case that _either_ the history triggers an NDE or not depending on if the original run happened to line up with the now-deterministic ordering of the LAs or not. I think the potentially worrying scenario here is that users possibly could've had workflows that, under replay, would sporadically throw NDEs but eventually pass after a few retries because they happened to hit the right sequence. Now, those workflows may fail/pass 100% of the time. I think that's probably acceptable, considering how rare this bug was to trigger anyway, but the moral of the story is we want to prove to ourselves the circumstances where the behavior change does or does not cause old histories to fail to replay. @mjameswh probably has some good advice about what other tests can be added based on his original research.

crates/sdk-core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -811,7 +800,12 @@ impl WorkflowMachines {
                             let resps = lam.try_resolve_with_dat((*la_dat).into())?;
                             self.process_machine_responses(mk, resps)?;
                         } else {
-                            self.local_activity_data.insert_peeked_marker(*la_dat);
+                            // Since the LA machine exists, we have encountered the LA WF command.
```
mjameswh: Is this case even possible? If it is, it feels wrong, and possibly dangerous, to simply drop the marker.

Unless we can prove this case exists, or otherwise prove that dropping the marker would be safe, I think we should just remove the call to `will_accept_resolve_marker` above, and let `try_resolve_with_dat` propagate an error should we ever reached that from an unexpected state.

mjameswh:
Great! Thanks for completing that work. And really pleased to see this ended up removing the need for fake markers, that's a welcome simplification.

> I think, with this fix, it'll be the case that either the history triggers an NDE or not depending on if the original run happened to line up with the now-deterministic ordering of the LAs or not. I think the potentially worrying scenario here is that users possibly could've had workflows that, under replay, would sporadically throw NDEs but eventually pass after a few retries because they happened to hit the right sequence.

I'm quite sure there's no risk of "sporadic" repro on this issue. Legacy behavior was actually deterministic (according to the traditional, non-temporal definition of that word) in both the non-replay and replay cases, though possibly _inconsistent_ between these two (hence the non-determinism issue according to Temporal's definition of that word). That is, in the non-replay case, LAs would always notify lang in the order that the LA execution completed, while in replay, they would always complete in the order that they got scheduled, assuming they originally completed within the same WFT. It's true that some workflows were replaying successfully despite technically hitting this LA bug, but that's still fully deterministic (traditional definition) for a given workflow execution. There would be no case of "retry a few times, and it may eventually pass".

Given the test plan, I indeed have specific concerns that I want us to cover. I'll detail that tomorrow.

Sushisource:
Replaced by https://github.com/temporalio/sdk-rust/pull/1442

PR #1076, opened 2025-12-10
Support gRPC binary metadata headers in C bridge

### LINKED ISSUE #991: [Feature Request] Ensure gRPC binary metadata headers are supported in C bridge

jmaeagle99:
## What was changed
- Added support for binary metadata headers for gRPC calls in the C bridge
- Unified the data format used for ASCII headers and binary headers to use `GrpcMetadataRef` across the C bridge.
- Added a test to validate that ascii and binary headers are properly passed to the RPC handler.

## Why?

Tonic will return errors if binary headers are passed in as ASCII headers. Add support for passing binary headers and ensure they are combined together appropriately for gRPC requests.

## Checklist

1. Closes #991

2. How was this tested: Add tests to verify header passing

3. Any docs updates needed? Not sure

### DIFF
--- a/crates/sdk-core-c-bridge/src/runtime.rs
+++ b/crates/sdk-core-c-bridge/src/runtime.rs
@@ -239,11 +239,10 @@ impl Runtime {
 
         let core_runtime_options = CoreRuntimeOptions::builder()
             .telemetry_options(telemetry_options)
-            .maybe_heartbeat_interval((options.worker_heartbeat_interval_millis != 0).then(|| {
-                Some(Duration::from_millis(
-                    options.worker_heartbeat_interval_millis,
-                ))
-            }))
+            .heartbeat_interval(
+                (options.worker_heartbeat_interval_millis != 0)
+                    .then(|| Duration::from_millis(options.worker_heartbeat_interval_millis)),
+            )
             .build()
             .map_err(|e| anyhow::anyhow!(e))?;
 
--- a/crates/sdk-core/tests/shared_tests/mod.rs
+++ b/crates/sdk-core/tests/shared_tests/mod.rs
@@ -58,7 +58,6 @@ pub(crate) async fn grpc_message_too_large() {
                     false
                 }
         }),
-        "Expected workflow task failure or termination b/c grpc message too large: {:?}",
-        events
+        "Expected workflow task failure or termination b/c grpc message too large: {events:?}",
     );
 }


crates/sdk-core-c-bridge/include/temporal-sdk-core-c-bridge.h
cretz: Arguably could just get rid of the concept of newline-delimited metadata ref altogether. It is a bit confusing to differentiate `TemporalCoreGrpcMetadataRef` and `TemporalCoreMetadataRef` I think

crates/sdk-core-c-bridge/src/client.rs
```diff
@@ -249,6 +252,18 @@ pub extern "C" fn temporal_core_client_update_metadata(
         .set_headers(metadata.to_string_map_on_newlines());
 }
 
+#[unsafe(no_mangle)]
+pub extern "C" fn temporal_core_client_update_binary_metadata(
```
cretz: I think, like github.com/temporalio/sdk-ruby/pull/348 and others, we should just differentiate plain vs binary metadata based on whether the key suffixed with `-bin`. This is what the gRPC spec does and most gRPC clients do. From a user POV, they only need to provide one metadata map.

crates/sdk-core-c-bridge/src/client.rs
```diff
@@ -337,19 +358,24 @@ pub extern "C" fn temporal_core_client_grpc_override_request_rpc(
 #[unsafe(no_mangle)]
 pub extern "C" fn temporal_core_client_grpc_override_request_headers(
     req: *const ClientGrpcOverrideRequest,
-) -> MetadataRef {
+) -> GrpcMetadataRef {
     let req = unsafe { &*req };
     // Lazily create the headers on first access
     let headers = req.built_headers.get_or_init(|| {
+        // CONSIDER: Should binary headers be decoded here before they are returned?
```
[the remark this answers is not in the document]
cretz: We should continue to provide the headers as is like this IMO. gRPC clients are supposed to be smart enough to understand what `-bin` keys represent. Same for accepting the headers/metadata IMO, it should just be one map.

crates/sdk-core-c-bridge/include/temporal-sdk-core-c-bridge.h
jmaeagle99: I performed the rename suggested in the latter part of your comment. Converting the remaining usages of the former `MetadataRef` (namely metrics tags, OTEL headers, and histogram buckets) feels more out of scope to me.

cretz (approved):
Looks great to me, but may want @Sushisource's approval before merging. Also, could develop the .NET SDK aspect with your branch as the submodule to confirm this does what is needed before merging.

Sushisource (approved):
Looking good to me!

PR #1079, opened 2025-12-15
Client Crate Refactorings

Sushisource:
Update client crate to simplify types / match proposal

Highlights:
* Added `Connection` type, which is a lot like it is in other Core SDKs (tempted to call this `Client`, since it really is)
* `Client` is what `NamespacedClient` was before (maybe go back to that? Maybe call `Connection` `BaseClient`? Open to ideas)
* Options structs for both (`ConnectionOptions` is exactly like `ClientOptions` was before, minus namespace)
* `RetryClient` eliminated, lives inside `Connection`
* Added `core-based-sdk` feature to gate options that should be hidden from Rust SDK users

Overall things are a bit simpler now which is great

Next up:
* Move runtime/telemetry stuff into common, as right now you can't attach metrics to a client without depending on Core
* Make `WorkflowClientTrait` nice
* Go back to finishing activity definitions & use in SDK

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -107,7 +107,7 @@ const VERSION: &str = env!("CARGO_PKG_VERSION");
 
 type Result<T, E = tonic::Status> = std::result::Result<T, E>;
 
-/// Options for [Connection::connect] and [Client::connect].
+/// Options for [Connection::connect].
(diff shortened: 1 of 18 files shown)


chris-olszewski (approved):
Removal of `RetryClient` is very nice

PR #1086, opened 2025-12-30
SDK Worker Executes activities

Sushisource:
* Hook up new activity definitions to execution
* Activity invocation now looks like: `ctx.activity(MyActivities::my_activity, input, ActivityOptions::default());`
* Worker now takes `Client` not `Connection`
* Hash current binary for default build id

### DIFF
--- a/crates/sdk-core/tests/common/mod.rs
+++ b/crates/sdk-core/tests/common/mod.rs
@@ -484,14 +484,15 @@ impl CoreWfStarter {
                     let client = Client::new(connection.clone(), client_opts);
                     (connection, client)
                 };
-                let worker = init_worker(
-                    rt,
-                    self.sdk_config
-                        .to_core_options(client.namespace())
-                        .expect("sdk config converts to core config"),
-                    connection,
-                )
-                .expect("Worker inits cleanly");
+                let mut core_config = self
+                    .sdk_config
+                    .to_core_options(client.namespace())
+                    .expect("sdk config converts to core config");
+                if let Some(ref ccm) = self.core_config_mutator {
+                    ccm(&mut core_config);
+                }
+                let worker =
+                    init_worker(rt, core_config, connection).expect("Worker inits cleanly");
                 InitializedWorker {
                     worker: Arc::new(worker),
                     client,
--- a/crates/sdk-core/tests/heavy_tests/fuzzy_workflow.rs
+++ b/crates/sdk-core/tests/heavy_tests/fuzzy_workflow.rs
@@ -82,8 +82,7 @@ async fn fuzzy_workflow() {
     starter.sdk_config.tuner = Arc::new(TunerHolder::fixed_size(25, 25, 100, 100));
     let mut worker = starter.worker().await;
     worker.register_wf(wf_name.to_owned(), fuzzy_wf_def);
-
-    starter.sdk_config.register_activities(StdActivities);
+    worker.register_activities(StdActivities);
 
     let client = starter.get_client().await;
 
--- a/crates/sdk-core/tests/integ_tests/worker_heartbeat_tests.rs
+++ b/crates/sdk-core/tests/integ_tests/worker_heartbeat_tests.rs
@@ -135,7 +135,7 @@ async fn docker_worker_heartbeat_basic(#[values("otel", "prom", "no_metrics")] b
     let wf_name = format!("worker_heartbeat_basic_{backing}");
     let mut starter = CoreWfStarter::new_with_runtime(&wf_name, rt);
     starter.sdk_config.max_cached_workflows = 5_usize;
-    starter.sdk_config.tuner = Arc::new(TunerHolder::fixed_size(5, 5, 1, 1));
+    starter.sdk_config.tuner = Arc::new(TunerHolder::fixed_size(5, 5, 100, 0));
     starter.set_core_cfg_mutator(|c| {
         c.plugins = vec![
             PluginInfo {


crates/sdk/src/lib.rs
```diff
@@ -250,32 +374,53 @@ struct WorkflowFutureHandle<F: Future<Output = Result<WorkflowResult<Payload>, J
...
     /// Maps activity type to the function for executing activities of that type
-    activity_fns: HashMap<String, ActivityFunction>,
+    activities: ActivityDefinitions,
     task_tokens_to_cancels: HashMap<TaskToken, CancellationToken>,
 }
 
 impl Worker {
-    // /// Create a new worker from an existing connection, and options.
-    // pub fn new(connection: Connection, options: WorkerOptions) -> Self {}
+    // TODO [rust-sdk-branch]: Not 100% sure I like passing runtime here
```
Sushisource: This is probably my main open question on this PR. This is really only used for telemetry options, and worker heartbeat interval.

We could have worker heartbeat on connection (inside client) as discussed way back, and then only pass in telemetry explicitly... but not sure how worth it that is or not

crates/common/src/activity_definition.rs
cretz: Arguably this should live alongside the rest of the activity code IMO (i.e. in `temporal_sdk::activities`). Same for workflow definition (and workflow signal definition, workflow update definition, and workflow query definition).

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -58,25 +59,44 @@ use temporalio_sdk_core::{
...
+impl SleepyActivities {
+    /// Activity that echoes input after sleeping for 2 seconds
+    #[activity]
+    async fn sleepy_echo(_ctx: ActivityContext, echo_me: String) -> Result<String, ActivityError> {
+        sleep(Duration::from_secs(2)).await;
+        Ok(echo_me)
+    }
+}
+
+async fn one_activity_wf(ctx: WfContext) -> WorkflowResult<Payload> {
```
cretz: We'll have to come back and fix names like `WfContext` I think at some point

crates/sdk/src/workflow_context.rs
```diff
@@ -213,18 +217,33 @@ impl WfContext {
...
+        let pc = PayloadConverter::serde_json();
+        let payload = pc.to_payload(&input, &SerializationContext::Workflow)?;
+        Ok(self.activity_untyped(AD::name().to_string(), payload, opts))
+    }
+
+    /// Request to run an activity with an explicit name and payload
+    pub fn activity_untyped(
         &self,
+        activity_type: String,
+        input: Payload,
```
cretz: As unfortunate as it is, I think all SDKs have to have _some_ way to both implement and invoke multi-param activities, even if it's not the suggested or typed approach.

cretz: I wonder from a Temporal user POV if we should not be exposing literal prost-generated `Payload` but instead offer a specific `Payload` type of our own with any helpers we might want and such. This will also come in handy for people that may want to accept literal `Payload` which can be preferred in cases of dynamic activities/workflows or just lazy/deferred deserialization.

crates/sdk/src/lib.rs
cretz: In other SDKs, we have expected the connection to carry the runtime, the client to carry a connection, and the worker to therefore access the runtime that way upon instantiation. I admit I haven't dug into how one creates a client/connection in this SDK with a runtime, but that's where it should come from IMO.

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -58,25 +59,44 @@ use temporalio_sdk_core::{
...
-        activity_type: "echo_activity".to_string(),
-        start_to_close_timeout: Some(Duration::from_secs(5)),
-        input: "hi!".as_json_payload().expect("serializes fine"),
-        ..Default::default()
-    })
-    .await;
-    Ok(().into())
+pub(crate) struct SleepyActivities {}
+
+#[activities]
```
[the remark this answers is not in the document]
Sushisource: you can use the macros on a function whose body is just unimplemented!(), or implement activity definition by hand on your own struct

I can add an example

crates/sdk/src/activities.rs
```diff
@@ -327,3 +328,84 @@ pub trait ExecutableActivity: ActivityDefinition {
...
+
+/// Contains activity registrations in a form ready for execution by workers.
+#[derive(Default, Clone)]
+pub struct ActivityDefinitions {
+    activities: HashMap<&'static str, ActivityInvocation>,
+}
+
+impl ActivityDefinitions {
+    /// Registers all activities on an activity implementer that don't take a receiver.
+    pub fn register_activities_static<AI>(&mut self) -> &mut Self
```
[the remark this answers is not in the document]
Sushisource: Yeah fair point, initializing it would typically be a no-op in that case

crates/common/src/activity_definition.rs
Sushisource: Sure, I can see naming it worker. I don't think it maters all that much but it's a reasonable rename.

The definition makes sense in common because workflow definitions will have to live here so they can be used by the client. Having the kinds of definitions together makes sense.

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -44,9 +45,9 @@ use temporalio_common::{
         test_utils::schedule_activity_cmd,
     },
 };
+use temporalio_macros::activities;
```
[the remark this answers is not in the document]
Sushisource: No. Procmacros must be in their own crate.

crates/sdk/src/workflow_context.rs
Sushisource: I think that makes sense but I'm not really going to deal with that until I flesh out the data conversion stuff

crates/common/src/activity_definition.rs
```diff
@@ -9,7 +9,7 @@ use crate::data_converters::{TemporalDeserializable, TemporalSerializable};
 /// Implement on a marker struct to define an activity.
 pub trait ActivityDefinition {
```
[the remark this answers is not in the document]
cretz: Would definitely be interested in a sample/snippet showing how to impl and register activity definition. May be worth a helper struct/fn, though hopefully people don't use that approach to create activities much. I will say one place where we see this approach used a lot in .NET is for DI where they want to control lifetimes and inject deps and such to the invocation.

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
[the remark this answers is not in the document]
cretz: Right, but maybe the crates are split out and a common one exposes them together somehow? For instance, how do https://docs.rs/pyo3/latest/pyo3/#attributes and https://docs.rs/magnus/latest/magnus/#attributes have the nice appearance of having a common library without asking users to depend on independent crates?

crates/sdk/src/lib.rs
Sushisource: Currently the connection only needs the metric meter. The whole runtime would have to move into common to have things work that way - but, that's not necessarily an issue. More the problem is that the idea of the runtime in general maybe feels a bit odd in Rust... but maybe that's not a huge issue. It holding onto a tokio runtime/handle is the main odd part. The other thing it holds onto is a tracing subscriber, which actually does make sense.

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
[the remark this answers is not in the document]
Sushisource: Oh, yeah, they can be re-exported. That's already mentioned in the proposal. I'm not going to bother with that until further down the line when stuff is more settled.

crates/sdk/src/workflow_context.rs
```diff
@@ -213,18 +217,33 @@ impl WfContext {
     }
 
     /// Request to run an activity
-    pub fn activity(
+    pub fn activity<AD: ActivityDefinition>(
+        &self,
+        _activity: AD,
+        input: AD::Input,
```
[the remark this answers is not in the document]
cretz: I think it can matter. There is a big difference between multiple scalar args and a single array of multiple values. I don't think you can even resolve the array ambiguity in some deserialization cases. I think you have to clearly define the difference between single param array and multiple params, and with the latter being rare (only for interop needs IMO), it can probably just be a specialized type.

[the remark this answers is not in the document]
Sushisource: I'm actually not sure if it's ambiguous or not... but regardless yeah, we can just use a special type.

PR #1090, opened 2026-01-08
Update api_upstream to v1.60.0

jmaeagle99:
## What was changed

- Updated api_upstream to the [1.60.0](https://github.com/temporalio/api/releases/tag/v1.60.0) release from temporal/api.
- Adjusted docs to correct location of the subtree.
- Adjusted sdk-core code to changes in the protos.

## Why?

Getting latest protos from latest release to leverage them in new features.

## Checklist
1. How was this tested: Unit and integration tests

### DIFF
--- a/crates/sdk-core/tests/heavy_tests.rs
+++ b/crates/sdk-core/tests/heavy_tests.rs
@@ -173,6 +173,7 @@ async fn chunky_activities_resource_based() {
                     wf_type.to_owned(),
                     vec![],
                     WorkflowOptions {
+                        #[allow(deprecated)]
                         id_reuse_policy: WorkflowIdReusePolicy::TerminateIfRunning,
                         ..Default::default()
                     },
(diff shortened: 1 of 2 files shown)


crates/common/src/protos/mod.rs
```diff
@@ -2352,6 +2354,8 @@ pub mod temporal {
                                 Attributes::WorkflowExecutionOptionsUpdatedEventAttributes(_) => true,
                                 Attributes::NexusOperationCancelRequestCompletedEventAttributes(_) => false,
                                 Attributes::NexusOperationCancelRequestFailedEventAttributes(_) => false,
+                                Attributes::WorkflowExecutionPausedEventAttributes(_) => false,
```
Sushisource: I'm actually wondering if maybe these _should_ be ignorable? I dunno if these events actually have any impact on replay or if they're really just there for server. Worth double checking.

gow: At this point there is no intention to use the `WorkflowExecutionPausedEvent` & `WorkflowExecutionUnpausedEvent` in the SDKs. So it's safe to ignore them.
Just for future reference - Should the events (and attributes) be explicitly listed here if we are adding a new ignorable events?

PR #1101, opened 2026-01-22
Rust SDK Prerelease

Sushisource:
Introduce the prerelease version of the Rust SDK

Follow-up list:
* ~~Get rid of WfExitValue and use special errors for CaN, etc.~~
* ~~WorkflowContextView needs to be populated~~
* ~~Prevent signal handlers from using `WorkflowContext::state_mut`~~

Probably post-this-PR / initial release:
* Getting typed workflow handles from not-start currently broken because WorkflowDefinition is implemented for workflow struct _and_ run method but handler types only point at run method.  Passing just the type/struct to start workflow doesn't work, because it's a type, but referring to `::run` in get handle is slightly awkward and has disambiguation problems sometimes. Not sure what the solution is. Current workaround is to use `snake_case_module_name::Run`.
* Hook up / impl Failure converters. 
* User friendly replay worker in SDK crate
* More clearly define interceptor APIs
* `RawValue::empty()` should validly serialize to `()` but doesn't

### DIFF
--- a/crates/common/src/telemetry.rs
+++ b/crates/common/src/telemetry.rs
@@ -1,6 +1,6 @@
 //! Contains tracing/logging and metrics related functionality
 
-/// Metric instrument types and the [`CoreMeter`](metrics::CoreMeter) trait.
+/// Metric instrument types and the [`CoreMeter`] trait.
 pub mod metrics;
 
 #[cfg(feature = "core-based-sdk")]


crates/sdk/src/workflows.rs
```diff
@@ -0,0 +1,638 @@
...
+    /// Dispatch the signal with payload deserialization.
+    fn dispatch(
+        ctx: WorkflowContext<Self>,
+        payloads: Payloads,
+        converter: &PayloadConverter,
+    ) -> LocalBoxFuture<'static, Result<(), WorkflowError>> {
+        match deserialize_input::<S::Input>(payloads.payloads, converter) {
+            Ok(input) => {
+                let mut ctx_clone = ctx.clone();
+                ctx.state_mut(|wf| Self::handle(wf, &mut ctx_clone, input));
```
chris-olszewski: Need to provide clear warning for implementers (or better yet prevent compilation) from writing a signal handler that mutates state through `ctx.state_mut(|s| {...})` inside the handler instead of directly mutate `self`.

At the moment this causes the signal handler to panic due to multiple mutable borrows caused by this `ctx.state_mut`

crates/client/src/workflow_handle.rs
```diff
@@ -0,0 +1,826 @@
...
+                UpdateWorkflowExecutionRequest,
+            },
+        },
+    },
+};
+use tonic::IntoRequest;
+use uuid::Uuid;
+
+/// Enumerates terminal states for a particular workflow execution
+// TODO: Add non-proto failure types, flesh out details, etc.
```
Sushisource: I think this todo can go away now

crates/sdk/src/workflow_context/options.rs
```diff
@@ -99,7 +100,8 @@ impl IntoWorkflowCommand for ActivityOptions {
                         .and_then(|d| d.try_into().ok()),
                     heartbeat_timeout: self.heartbeat_timeout.and_then(|d| d.try_into().ok()),
                     cancellation_type: self.cancellation_type as i32,
-                    arguments: vec![self.input],
+                    // TODO [rust-sdk-branch]: Handle multi-args
+                    arguments: vec![input],
```
Sushisource: Probably should deal with this one in this PR

PR #1102, opened 2026-01-22
Update Worker::validate to allow SDKs to get namespace limit information

jmaeagle99:
## What was changed

Update Worker::validate method to allow SDKs to still validate the server connection but get namespace information in response.

## Why?

Allows SDKs to get namespace limit information.

## Checklist

1. How was this tested: Tests and locally updated sdk-python to run tests.

### DIFF
--- a/crates/common/protos/api_upstream/openapi/openapiv2.json
+++ b/crates/common/protos/api_upstream/openapi/openapiv2.json
@@ -8195,6 +8195,17 @@
         }
       }
     },
+    "ExecuteMultiOperationResponseResponse": {
+      "type": "object",
+      "properties": {
+        "startWorkflow": {
+          "$ref": "#/definitions/v1StartWorkflowExecutionResponse"
+        },
+        "updateWorkflow": {
+          "$ref": "#/definitions/v1UpdateWorkflowExecutionResponse"
+        }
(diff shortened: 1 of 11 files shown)


crates/common/src/lib.rs
```diff
@@ -35,8 +35,14 @@ pub trait Worker: Send + Sync {
     /// Validate that the worker can properly connect to server, plus any other validation that
     /// needs to be done asynchronously. Lang SDKs should call this function once before calling
     /// any others.
+    /// TODO: Move all SDKs over to validate_namespace and then remove
     async fn validate(&self) -> Result<(), WorkerValidationError>;
 
+    /// Validate that the worker can properly connect to server, plus any other validation that
+    /// needs to be done asynchronously. Lang SDKs should call this function once before calling
+    /// any others.
+    async fn validate_namespace(&self) -> Result<NamespaceInfo, WorkerValidationError>;
```
cretz: IMO you can just change the existing validate to return the stuff you need. There are no compatibility concerns for this core code. (this should not be public to users of Rust SDK IMO, cc @Sushisource)

jmaeagle99: Ripped the bandaid.

PR #1106, opened 2026-02-02
Fix TemporalPatchVersion search attribute type to use KeywordList format

### LINKED ISSUE #1105: [Bug] Improper "type" used for TemporalPatchVersion search attribute
### Describe the bug

The metadata `type` added as part of #1032 is not in the expected format, but server doesn't validate it properly. That PR set it as `INDEXED_VALUE_TYPE_KEYWORD_LIST`, but all SDKs set it (and expect it) as `KeywordList`. The reason the server accepts this improperly is likely due to our special generated code like https://pkg.go.dev/go.temporal.io/api/enums/v1#IndexedValueTypeFromString that accepts both forms of enum string for when we switch from gogoproto and had to write these helpers.

cretz:
## What was changed

#1032 used the wrong string for SA `type` for keyword list, causing some issues in parsing in some SDKs

## Checklist

1. Closes #1105

### DIFF
--- a/crates/sdk-core/src/worker/workflow/machines/patch_state_machine.rs
+++ b/crates/sdk-core/src/worker/workflow/machines/patch_state_machine.rs
@@ -141,10 +141,10 @@ pub(super) fn has_change<'a>(
             );
             vec![]
         } else {
-            serialized.metadata.insert(
-                "type".to_string(),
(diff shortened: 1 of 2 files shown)


chris-olszewski:
(Failures are from https://www.githubstatus.com/incidents/xwn6hjps36ty)

PR #1109, opened 2026-02-10
Nexus error types

VegetarianOrc:
## What was changed

- **Nexus error types**: Updated Nexus task completion to support the new Temporal `Failure`-based error format alongside the legacy `HandlerError` format. The SDK now uses a `NexusTaskFailure` enum (`Legacy`/`Temporal` variants) and negotiates the format via a new `temporal_failure_responses` capability flag on the Nexus request.
- **Start operation failure variant**: Added handling for the new `StartOperationResponse::Failure` variant, which carries a full Temporal `Failure` with `ApplicationFailureInfo` or `CanceledFailureInfo`. The deprecated `operation_error` path is still supported.
- **Handler failure completion**: Added a new `Status::Failure` completion path for Nexus tasks. When the handler returns a failure with `NexusHandlerFailureInfo`, it is sent as a Temporal `Failure` to modern servers or converted to a legacy `HandlerError` for older servers.
- **pbjson serialization**: Added `pbjson`/`pbjson-build` dependencies and integrated them into the build to generate JSON serialization for `temporal.api.failure`, `temporal.api.common`, and `temporal.api.enums` proto packages. This is used to serialize Temporal failures into Nexus failure detail payloads.
- **Worker instance key**: Added a `worker_instance_key` (UUID) to `WorkerClientBag`, included in all poll requests (workflow, activity, nexus) and shutdown requests. The shutdown method now also sends `task_queue` and `task_queue_types` to support server-side poll cancellation.
- **Proto API update to 1.62.1**: Pulled in upstream proto changes including new `CountSchedules` API, `PauseWorkflowExecution`/`UnpauseWorkflowExecution` APIs, Nexus failure fields (`stack_trace`, `cause`), and `Request.Capabilities`.

## Why?

The Temporal server is adding support for richer Nexus error propagation using native Temporal `Failure` objects instead of the limited `HandlerError` format. This lets Nexus handlers return structured failure information (application errors, cancellation, cause chains) that can be properly surfaced to workflow callers. The capability negotiation ensures backward compatibility with servers that haven't been updated yet. The worker instance key and enhanced shutdown fields support upcoming server-side worker lifecycle improvements.

## Checklist

1. Closes 

2. How was this tested:
   - Unit tests for all new Nexus completion paths:
     - `nexus_task_completion_with_failure_status` — handler failure with Temporal failure support
     - `nexus_task_completion_with_failure_converts_to_legacy_for_old_server` — backward compat conversion
     - `nexus_task_completion_with_failure_status_missing_handler_info_fails` — validation of required NexusHandlerFailureInfo
     - `nexus_start_operation_failure_with_application_failure_info` — StartOperationResponse failure with ApplicationFailureInfo
     - `nexus_start_operation_failure_with_canceled_failure_info` — StartOperationResponse failure with CanceledFailureInfo
     - `nexus_convert_temporal_to_nexus_failure_no_temporal_failure_support` — Temporal-to-Nexus failure conversion for legacy servers
   - Existing integration tests updated

3. Any docs updates needed?


crates/sdk-core/tests/integ_tests/polling_tests.rs
```diff
@@ -122,6 +122,7 @@ async fn out_of_order_completion_doesnt_hang() {
 }
 
 #[tokio::test]
+#[ignore]
```
VegetarianOrc: These tests suffer from a wire incompatibility. This will be resolved once the api is updated.

crates/common/build.rs
```diff
@@ -123,5 +163,14 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
         )?;
 
+    let descriptors = std::fs::read(&descriptor_file)?;
+    pbjson_build::Builder::new()
+        .register_descriptors(&descriptors)?
+        .build(&[
+            ".temporal.api.failure",
+            ".temporal.api.common",
+            ".temporal.api.enums",
+        ])?;
```
Sushisource: Do we have to add this new dep? Is it because the internally serialized JSON stuff needs to be proto compat? I don't love the idea of having multiple different serde serialization implementations depending on the type.

Quinn-With-Two-Ns: We used to do this in Lang right, but now we are moving it to core since core needs to convert the new Failure response format to the old Error format right?

Sushisource (approved):

PR #1122, opened 2026-02-20
no longer cancel pending polls on shutdown

yuandrew:
## What was changed

If `worker_poll_complete_on_shutdown` namespace capability exists, then don't cancel outstanding poll requests on shutdown. Instead expect server to return an empty result.

## Why?

During worker shutdown, the SDK cancels outstanding poll requests client-side by racing them against the shutdown token. This
  creates a race condition: the server may dispatch a task to a poll at the same instant the SDK kills the connection, orphaning the
  task until start_to_close timeout.

## Checklist

1. Closes 

2. How was this tested:

Added new test verifying graceful shutdown behavior

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Touches generated/public OpenAPI surface area (path reshapes and endpoint additions/removals), which can break client integrations; the new CI guard is low risk but may block builds if it flags previously unnoticed conflicts.
> 
> **Overview**
> Updates the generated OpenAPI v2 spec to **restructure activity-related REST endpoints**, introducing `.../activities/{activityId}/...` routes (e.g. `cancel`, `terminate`, `outcome`, `complete/fail/heartbeat`) while moving older pause/reset/unpause/update-options routes under an `activities-deprecated` prefix and adding top-level `activity-*` worker-style endpoints.
> 
> Removes the experimental `workflows/execute-multi-operation` API from the OpenAPI spec and adjusts several schema definitions/fields (including new namespace capability `workerPollCompleteOnShutdown` and related enum/fields changes).
> 
> Adds a Go-based CI step (`cmd/check-path-conflicts`) that fails the build when the OpenAPI spec contains ambiguous path patterns (literal vs parameter segment collisions).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 41246e8dc6a66ac24e80d77e1858bebcbf14c7c2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/client/src/workflow_handle.rs
+++ b/crates/client/src/workflow_handle.rs
@@ -104,6 +104,7 @@ impl WorkflowHistory {
 
 /// A workflow handle which can refer to a specific workflow run, or a chain of workflow runs with
 /// the same workflow id.
+#[derive(Clone)]
 pub struct WorkflowHandle<ClientT, W> {
     client: ClientT,
     info: WorkflowExecutionInfo,
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -9,17 +9,30 @@ homepage = "https://temporal.io/"
 repository = "https://github.com/temporalio/sdk-core"
 keywords = ["temporal", "workflow"]
 categories = ["development-tools"]
+exclude = ["protos/*/.github/*"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
 history_builders = ["rand"]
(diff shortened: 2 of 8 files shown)


chris-olszewski (changes requested):
Makes sense to me, but I think we want to keep respecting `TEMPORAL_POLL_SHUTDOWN_INTERRUPT_WAIT_MS` set by users in the case that `worker_poll_complete_on_shutdown` isn't supported. Not sure how widely this was advertised, but it did make it into [Python release notes](https://github.com/temporalio/sdk-python/releases/tag/1.21.0)

PR #1129, opened 2026-02-23
fix: encode ActivityOptions.summary as json/plain payload

thejens:
## What was changed

`ActivityOptions::into_command` now encodes the `summary` field using `as_json_payload()` (`json/plain` encoding) instead of `.into()` (`binary/plain` encoding).

## Why?

`LocalActivityOptions::into_command` (same file, line ~203) already encodes its summary via `as_json_payload()`. The `ActivityOptions` path used `.into()` which produces a `binary/plain` payload — the Temporal UI cannot decode this and shows "Decoding failed" instead of the summary text.

## Checklist

1. Closes — no existing issue

2. How was this tested: built and ran locally, see screenshots:
### Before:
image

### After:
image

### DIFF
--- a/crates/sdk/src/workflow_context/options.rs
+++ b/crates/sdk/src/workflow_context/options.rs
@@ -108,10 +108,13 @@ impl ActivityOptions {
                 }
                 .into(),
             ),
-            user_metadata: self.summary.map(|s| UserMetadata {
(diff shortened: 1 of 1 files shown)


chris-olszewski:
This has been addressed now.

PR #1132, opened 2026-03-05
feat(sdk): add schedule handle

chris-olszewski:
## What was changed
Add client methods for interacting with schedules.

## Why?
TSIA

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/555

2. How was this tested:
New unit & integration tests

3. Any docs updates needed?
N/A


crates/client/src/lib.rs
```diff
@@ -696,6 +706,130 @@ impl Client {
     }
 }
 
+// Schedule operations on Client.
+impl Client {
+    /// Create a schedule and return a handle to it.
+    pub async fn create_schedule(
```
Sushisource: Kinda inclined to move all this (really, the tests is the main thing I'd want to move, but may as well bring this along too) into a dedicated `schedules.rs` which could include the handle too?

PR #1135, opened 2026-03-07
fix: evict terminated/timed-out workflows from sticky cache

gibbonjj:
## Summary

- Terminated and timed-out workflows are never evicted from the sticky cache, causing unbounded cache growth
- When the server terminates a workflow, `have_seen_terminal_event` is set but `workflow_end_time` is never set (because the workflow code never issues a terminal command). The existing eviction check requires both `workflow_is_finished()` (checks `workflow_end_time`) and `WFTReportStatus::Reported` — both are false for terminated workflows.
- Added a second eviction check in `process_post_activation` that fires when `have_seen_terminal_event()` is true but `workflow_is_finished()` is false, covering the termination/timeout case
- Added test `terminated_workflow_with_pending_activity_evicts_from_cache` that verifies the eviction occurs (hangs without the fix)

## Impact

Any workload that terminates workflows at scale (e.g. via `WorkflowIdConflictPolicy.TERMINATE_EXISTING`) will see `temporal_sticky_cache_size` grow unboundedly until LRU eviction kicks in when the cache is full. This causes unnecessary memory pressure on workers.

### DIFF
--- a/crates/sdk-core/src/core_tests/workflow_tasks.rs
+++ b/crates/sdk-core/src/core_tests/workflow_tasks.rs
@@ -3127,3 +3127,46 @@ async fn grpc_message_too_large_doesnt_spam_task_fails() {
     core.drain_pollers_and_shutdown().await;
     // Mock only expects 1 task failure, and would fail here if we spammed
 }
+
+/// When a workflow is terminated by the server while an activity is in-flight, the worker receives
+/// a WFT containing a `WorkflowExecutionTerminated` event. The workflow code never issues a
+/// terminal command, but the run should still be evicted from the sticky cache.
+#[tokio::test]
+async fn terminated_workflow_with_pending_activity_evicts_from_cache() {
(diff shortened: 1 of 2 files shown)


crates/sdk-core/src/core_tests/workflow_tasks.rs
```diff
@@ -3127,3 +3127,46 @@ async fn grpc_message_too_large_doesnt_spam_task_fails() {
     core.drain_pollers_and_shutdown().await;
     // Mock only expects 1 task failure, and would fail here if we spammed
 }
+
+/// When a workflow is terminated by the server while an activity is in-flight, the worker receives
+/// a WFT containing a `WorkflowExecutionTerminated` event. The workflow code never issues a
```
Sushisource: This statement isn't true - the server will never send a workflow task to the worker when a workflow is terminated.

PR #1153, opened 2026-03-13
fix(sdk): re-poll update futures after workflow state changes

mfateev:
## Summary
- **Bug:** When both `run` and an update handler use `wait_condition` on each other's state, they deadlock. Update futures are polled before `poll_wf_future`, so the update sees stale state. After `poll_wf_future` sets the flag via `state_mut`, update futures are never re-polled within the same activation.
- **Fix:** Re-poll update futures after `poll_wf_future` so that newly-unblocked `wait_condition` predicates are observed within the same activation. The inline update-polling code is extracted into `poll_update_futures` to avoid duplication.
- **Test:** Adds a regression test (`update_wait_condition_unblocked_by_run_state_change`) that exercises the cross-future `state_mut`/`wait_condition` pattern.

## Test plan
- [x] `cargo check --tests` — zero warnings in changed code
- [x] `cargo test` — all 450 tests pass
- [x] New regression test deadlocks without the fix, passes with it

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "temporalio-client"
-version = "0.1.0"
+version = "0.2.0"
 edition = "2024"
 authors = ["Temporal Technologies Inc. <sdk@temporal.io>"]
 license-file = { workspace = true }
@@ -44,7 +44,7 @@ rand = "0.9.2"
 
 [dependencies.temporalio-common]
 path = "../common"
-version = "0.1"
+version = "0.2"
 
 [dev-dependencies]
 assert_matches = "1"
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "temporalio-common"
-version = "0.1.0"
+version = "0.2.0"
 edition = "2024"
 authors = ["Temporal Technologies Inc. <sdk@temporal.io>"]
 license-file = { workspace = true }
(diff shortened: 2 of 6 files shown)


crates/sdk-core/tests/integ_tests/update_tests.rs
```diff
@@ -1379,3 +1379,261 @@ async fn update_lost_on_activity_mismatch() {
...
+//
+// Each "step" is (Option<flag_to_wait_for>, vec_of_flags_to_set).
+// A handler's script is a Vec of steps executed sequentially.
+// The workflow input contains scripts for run, each update, and each signal.
+
+/// A single step: optionally wait for a flag, then set flags.
+type Step = (Option<usize>, Vec<usize>);
+
+fn wait(flag: usize) -> Step { (Some(flag), vec![]) }
+fn set(flag: usize) -> Step { (None, vec![flag]) }
```
Sushisource: This should be an enum since the wait and set cases are mutually exclusive

crates/sdk-core/tests/integ_tests/update_tests.rs
```diff
@@ -1379,3 +1379,261 @@ async fn update_lost_on_activity_mismatch() {
         .await
         .unwrap();
 }
+
+// ---------------------------------------------------------------------------
+// Scripted convergence-loop test harness
+// ---------------------------------------------------------------------------
+//
```
Sushisource: This whole thing could be in its own file. Not really specifically an update test at this point.

Sushisource (approved):
Thanks! Looks like it just needs a format.

PR #1159, opened 2026-03-16
chore(ci): avoid caching rust on PR

chris-olszewski:
## What was changed
Two things:
 - Skip rust cache save in PRs
 - Add an extra 5 minutes to main job to make sure we populate the cache

## Why?
The cache artifacts created on branches other than main don't get reused by other branches. Not sure saving these artifacts has much usage compared to just making sure the main cache is present.

For whatever reason creating the cache artifact takes ~1min on `macos-intel` runners which leads to the job just timing out consistently and the cache never getting populated. Bump up the timeout on main to avoid this.

## Checklist

1. Closes N/A

2. How was this tested:

CI

3. Any docs updates needed?

N/A

### DIFF
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -17,7 +17,7 @@ concurrency:
 jobs:
   build-and-lint:
     name: "Format, docs, and lint"
-    timeout-minutes: 10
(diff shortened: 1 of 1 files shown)


Sushisource (approved):
Nice find!

PR #1160, opened 2026-03-17
Add activity cancels from nexus

Sushisource:
Creates a worker which polls on the appropriately-named worker commands nexus task queue, and handles and activity cancellations sent down it.

The worker is not started unless the server supports this capability.


crates/sdk-core/src/worker/heartbeat.rs
```diff
@@ -239,4 +381,163 @@ mod tests {
 
         assert_eq!(3, heartbeat_count.load(Ordering::Relaxed));
     }
+
+    fn make_execute_commands_nexus_response(
+        task_token: Vec<u8>,
+        commands: Vec<WorkerCommand>,
+    ) -> PollNexusTaskQueueResponse {
+        let exec_req = ExecuteCommandsRequest { commands };
```
rkannan82: Please incorporate the following changes:
1. Set task_kind to worker_commands. https://github.com/temporalio/api/pull/763
2. Guard new code (creating nexus queue, polling, processing) with worker_commands capability: https://github.com/temporalio/api/pull/771

crates/common/src/protos/mod.rs
```diff
@@ -2755,6 +2755,13 @@ pub mod temporal {
                 }
             }
         }
+        pub mod nexusservices {
```
rkannan82: Does the nexus worker send heartbeat? If it does, then I need to find a way to exclude this from the ListWorkers result.

Sushisource: ~Good question. It shouldn't, if it does I'll make sure it doesn't. ~

Changed my mind. We are already making sure it gets excluded from metrics (since we saw it show up and had that whole discussion re: Nexus usage), and being able to diagnose possible issues with it (getting overloaded, not enough slots, whatever) is useful. Also being dealt with here: https://github.com/temporalio/cli/pull/1019

crates/sdk-core/tests/shared_tests/mod.rs
```diff
@@ -222,3 +230,85 @@ pub(crate) async fn shutdown_during_active_timer_activity_workflows() {
...
+                ActivityOptions::with_start_to_close_timeout(Duration::from_secs(30))
+                    .retry_policy(RetryPolicy {
+                        maximum_attempts: 1,
+                        ..Default::default()
+                    })
+                    .cancellation_type(ActivityCancellationType::WaitCancellationCompleted)
+                    .build(),
+            );
+            // Timer needed to avoid cancel-before-sent
+            ctx.timer(Duration::from_millis(10)).await;
```
yuandrew, line 289: This doesn't actually guarantee the activity has started, right? Only that it's been scheduled? Can we better guarantee that the activity has started?

crates/sdk-core/tests/integ_tests/worker_heartbeat_tests.rs
```diff
@@ -271,13 +274,15 @@ async fn docker_worker_heartbeat_basic(#[values("otel", "prom", "no_metrics")] b
...
                             query: String::new(),
+                            include_system_workers: false,
                         }
                         .into_request(),
                     )
                     .await
                     .unwrap()
                     .into_inner();
                     #[allow(deprecated)]
+                    #[allow(deprecated)]
```
yuandrew: ?

crates/sdk-core/src/worker/client.rs
```diff
@@ -373,19 +387,26 @@ impl WorkerClient for WorkerClientBag {
...
             task_queue: Some(TaskQueue {
                 name: poll_options.task_queue,
-                kind: TaskQueueKind::Normal as i32,
+                kind: kind as i32,
                 normal_name: "".to_string(),
             }),
             identity: self.identity(),
             worker_version_capabilities: self.worker_version_capabilities(),
             deployment_options: self.deployment_options(),
+            // TODO: Piggyback worker heartbeats here if this is the system nexus worker and reset
```
yuandrew, line 408: We have something tracking this work? Worried this will get lost. If not, i can make one and assign it to myself, I'll need to dig up my old impl for this, I remember doing this like a year ago haha

crates/sdk-core/tests/shared_tests/mod.rs
```diff
@@ -222,3 +230,85 @@ pub(crate) async fn shutdown_during_active_timer_activity_workflows() {
...
+            act_fut.cancel();
```
yuandrew, line 290: It never clicked from me until now that this is all it takes to cancel the activity. Some reason I thought there would need to be special actions you'd have to initiate from the server, or something. Very nice

yuandrew (commented):
A few minor things, biggest question is should we update any public/user facing docstrings to mention activity cancellation can now be delivered without activity heartbeats when worker heartbeating is enabled?

crates/sdk-core/src/worker/client.rs
Sushisource: We should as part of the overall workstream - but yes please make sure we are somewhere

crates/sdk-core/tests/integ_tests/worker_heartbeat_tests.rs
Sushisource: Lol what. Thanks. I think it was a merge issue.

crates/sdk-core/tests/shared_tests/mod.rs
Sushisource: It doesn't but that doesn't matter, it only needs to be scheduled here

Sushisource:
> A few minor things, biggest question is should we update any public/user facing docstrings to mention activity cancellation can now be delivered without activity heartbeats when worker heartbeating is enabled?

We certainly will in lang layers but, I don't think there's actually much in core that will need to change. I'll look for any and change them if so (Rust SDK itself perhaps, but, I'm not sure it ever mentioned this limitation anywhere).

yuandrew (approved):

Issue #692, opened 2024-02-23
[Bug] Panic when failing to start worker due to being unauthorized


djc:
### What are you really trying to do?

Trying to bring up a worker process for the first time in our production environment.

### Describe the bug

There's a panic here. There probably shouldn't be? (It's probably not super important, but would be nice to handle in a better way.)

```
2024-02-23T16:29:13.189124Z ERROR temporal_sdk_core::worker::workflow::workflow_stream: Workflow processing encountered fatal error and must shut down TonicError(Status { code: PermissionDenied, message: "Request unauthorized.", details: b"\x08\x07\x12\x15Request unauthorized.\x1a\x8c\x01\nHtype.googleapis.com/temporal.api.errordetails.v1.PermissionDeniedFailure\x12@\n>caller does not have required permissions for namespace 'prod'", metadata: MetadataMap { headers: {"content-type": "application/grpc", "server": "temporal", "date": "Fri, 23 Feb 2024 16:29:12 GMT"} }, source: None })
2024-02-23T16:29:13.190812Z ERROR panic: shared::tracers: panic: panicked at /Users/djc/.cargo/git/checkouts/sdk-core-3d5ba50f52602518/42e2a94/core/src/worker/workflow/mod.rs:231:34:
Activation processor channel not dropped: SendError { .. } location="/Users/djc/.cargo/git/checkouts/sdk-core-3d5ba50f52602518/42e2a94/core/src/worker/workflow/mod.rs:231:34" payload="unknown"
thread 'workflow-processing' panicked at /Users/djc/.cargo/git/checkouts/sdk-core-3d5ba50f52602518/42e2a94/core/src/worker/workflow/mod.rs:231:34:
Activation processor channel not dropped: SendError { .. }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: Unhandled grpc error when activity polling: Status { code: PermissionDenied, message: "Request unauthorized.", details: b"\x08\x07\x12\x15Request unauthorized.\x1a\x8c\x01\nHtype.googleapis.com/temporal.api.errordetails.v1.PermissionDeniedFailure\x12@\n>caller does not have required permissions for namespace 'prod'", metadata: MetadataMap { headers: {"content-type": "application/grpc", "server": "temporal", "date": "Fri, 23 Feb 2024 16:29:12 GMT"} }, source: None }
```

### Minimal Reproduction

[...]

PR #1165, opened 2026-03-18
fix(core): avoid panic when workflow channels close

sachinsharma3191:
## Summary
- avoid panics in workflow processing when fetch/activation channels are already dropped
- stop workflow processing loop gracefully in that case and log instead of panicking
- add focused regression tests for channel-forwarding behavior

## Test plan
- [x] `cargo test -p temporalio-sdk-core forwarding_output_stops_when_activation_channel_dropped -- --nocapture`
- [x] `cargo test -p temporalio-sdk-core forwarding_output_succeeds_with_live_channels -- --nocapture`
- [x] `cargo test -p temporalio-sdk-core core_tests::workers::worker_can_shutdown_after_never_polling_ok -- --nocapture`
- [x] `cargo test -p temporalio-sdk-core core_tests::workers::shutdown -- --nocapture`
- [x] `cargo fmt --all --check`


chris-olszewski (changes requested):
We do not want to ignore cases where receivers get dropped unexpectedly. This is unexpected behavior and should result in a panic.

~~Were you encountering these panics in your own code? If so, please open an issue.~~

Ah, I see the branch references https://github.com/temporalio/sdk-core/issues/692. I think we can solve this more gracefully than just ignoring unexpected receiver drops.Ah, I see the branch references https://github.com/temporalio/sdk-core/issues/692. I think we can solve this more gracefully than just ignoring unexpected receiver drops.

sachinsharma3191:
Thanks for the review, @chris-olszewski — you're right that silently ignoring dropped receivers hides real bugs.

I've reworked this so we only tolerate a dropped receiver **when shutdown is already underway**, checked via the worker's shutdown token (which `poll_workflow_activation` cancels before the receiver in `Workflows` is dropped — the exact #692 path). A receiver dropping while the worker is still running now **panics**, since it would otherwise silently discard activations.

I also pared the PR back to just this change — the unrelated WFT-sequencing tweaks and the other `dbg_panic!`→log conversions have been removed, so the diff is now limited to `worker/workflow/mod.rs`. Added unit tests for all three cases: drop-during-shutdown stops processing, unexpected drop panics, and live channels forward normally.

Mind taking another look?

sachinsharma3191:
@chris-olszewski Friendly ping — the PR has been reworked to only tolerate a dropped receiver during shutdown (checked via the worker's shutdown token), while an unexpected drop during normal operation now panics. The diff is scoped to just `worker/workflow/mod.rs` with unit tests for all three cases. Would you be able to take another look when you get a chance? Thanks!

PR #1177, opened 2026-03-24
feat: add proto json wrapper type

### LINKED ISSUE #1142: [Feature Request] Protobuf binary payload converter
Only JSON payload converter is implemented. Need protobuf binary and protobuf JSON payload converters, plus composite converter support to try multiple converters in sequence.

chris-olszewski:
## What was changed
Add `ProstJsonSerializable` wrapper type that will take mark it as using `json/proto`.

Also ensure that messageType gets set for `ProstSerializable` and `ProstJsonSerializable`

Note, that this current implementation does have risk of misuse as we cannot verify that the `Serialization` implementation is ProtobufJSON complaint. There is nothing stopping a `#[derive(serde::Serialize)]` from using this and possibly not producing a proper JSON representation. This only presents a real risk for cross-lang situations.

Also, as noted before: due to type erasure the wrapper types do lead to a more invasive user experience compared to other SDKs for specifying non-standard encoding of payloads.

## Why?
Primarily bumping this to avoid hand rolling in #1208 

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/1142

2. How was this tested:
Additional unit & integration tests

3. Any docs updates needed?
Samples for how to use these is probably required.

### DIFF
--- a/crates/common/src/data_converters.rs
+++ b/crates/common/src/data_converters.rs
@@ -158,9 +158,6 @@ impl PayloadConverter {
     pub fn serde_json() -> Self {
         Self::Serde(Arc::new(SerdeJsonPayloadConverter))
     }
-    // Proto binary and JSON protobuf encoding are handled via wrapper types
-    // (`ProstSerializable<T>` and `ProstJsonSerializable<T>`) through the `UseWrappers`
-    // variant, not through dedicated converter variants.
 }
 
 impl Default for PayloadConverter {
@@ -628,6 +625,7 @@ where
     T: prost::Message + prost::Name + Default + 'static,
 {
     fn to_payload(&self, _: &SerializationContext<'_>) -> Result<Payload, PayloadConversionError> {
+        let as_proto = prost::Message::encode_to_vec(&self.0);
         Ok(Payload {
             metadata: {
                 let mut hm = HashMap::new();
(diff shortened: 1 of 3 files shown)


chris-olszewski:
@claude review

crates/common/src/data_converters.rs
```diff
@@ -656,6 +658,74 @@ where
...
+/// Use this when you want human-readable JSON protobuf format (proto3 JSON mapping with
+/// camelCase field names and string enums).
+pub struct ProstJsonSerializable<T: prost::Message>(pub T);
+impl<T> TemporalSerializable for ProstJsonSerializable<T>
+where
+    T: prost::Message + prost::Name + serde::Serialize + Default + 'static,
+{
+    fn to_payload(&self, _: &SerializationContext<'_>) -> Result<Payload, PayloadConversionError> {
+        let as_json = serde_json::to_vec(&self.0)
+            .map_err(|e| PayloadConversionError::EncodingError(e.into()))?;
```
Sushisource, line 685: I'm confused about how this will reliably use the proto JSON encoding vs. normal serde-derive based encodings. In build.rs we apply serde derive to basically everything, and then pbjson is used for 

```
            ".temporal.api.failure",
            ".temporal.api.common",
            ".temporal.api.enums",
```

Currently, but not everything else?

PR #1208, opened 2026-04-09
Support `static_summary`, `static_details` and current_details` in the Rust sdk

### LINKED ISSUE #1141: [Feature Request] Workflow current details and static metadata
No `set_current_details()` / `current_details()` for dynamically updating workflow metadata during execution. No `static_summary` / `static_details` at workflow start.

GregoryTravis:
- Implements setting `static_summary` and `static_details` at workflow start
- Implements setting and getting `current_details` during workflow execution
- Implements the `__temporal_workflow_metadata` query
- Integration tests for all three fields

Closes #1141


crates/sdk/src/workflow_context.rs
```diff
@@ -847,6 +852,14 @@ impl<W> SyncWorkflowContext<W> {
         ))
     }
 
+    /// Set the current details string for this workflow execution.
+    ///
+    /// The value is surfaced to the Temporal server UI in real time via the
+    /// `__temporal_workflow_metadata` built-in query.
```
Sushisource: I don't think we really want to refer to the funky built-in query name in the public docstring.

crates/client/src/lib.rs
```diff
@@ -1030,6 +1031,20 @@ where
...
+        let user_metadata = if options.static_summary.is_some() || options.static_details.is_some()
+        {
+            Some(UserMetadata {
+                summary: options
+                    .static_summary
+                    .map(|s| s.as_json_payload().expect("String-to-JSON payload serialization is infallible")),
+                details: options
+                    .static_details
+                    .map(|s| s.as_json_payload().expect("String-to-JSON payload serialization is infallible")),
+            })
```
Sushisource, line 1045: These will need to be serialized using the user's data converter

crates/sdk-core/tests/integ_tests/workflow_tests/client_interactions.rs
```diff
@@ -620,3 +620,75 @@ async fn test_typed_signal_query_update() {
...
+        .describe(WorkflowDescribeOptions::default())
+        .await
+        .unwrap();
+
+    let user_metadata = description
+        .raw_description
+        .execution_config
+        .expect("execution_config present")
+        .user_metadata
+        .expect("user_metadata present");
```
chris-olszewski: Users shouldn't have to reach into raw proto to access these fields.

crates/sdk/src/workflow_context.rs
GregoryTravis: Removed.

crates/client/src/lib.rs
GregoryTravis: As @mjameswh pointed out, we do not use the user's converter, because "because the metadata must absolutely be serialized as JSON". We *do* use the user-specified codec, if any.

crates/sdk/src/workflow_context/options.rs
```diff
@@ -108,10 +108,13 @@ impl ActivityOptions {
...
-                summary: Some(s.into()),
-                details: None,
-            }),
+            user_metadata: self
+                .summary
+                .map(|s| s.as_json_payload().expect("String-to-JSON payload serialization is infallible"))
+                .map(|summary| UserMetadata {
+                    summary: Some(summary),
+                    details: None,
+                }),
```
chris-olszewski, line 122: As mentioned by Spencer [over here](https://github.com/temporalio/sdk-core/pull/1129#issuecomment-3947271936), this should go through a user payload converter and not hard-coded to json encoding.

crates/sdk/src/workflow_context/options.rs
```diff
@@ -200,7 +203,7 @@ impl LocalActivityOptions {
             ),
             user_metadata: self
                 .summary
-                .and_then(|summary| summary.as_json_payload().ok())
+                .map(|summary| summary.as_json_payload().expect("String-to-JSON payload serialization is infallible"))
```
chris-olszewski: Same as above

crates/sdk/src/workflow_context/options.rs
```diff
@@ -251,8 +254,8 @@ impl ChildWorkflowOptions {
     ) -> WorkflowCommand {
         let user_metadata = if self.static_summary.is_some() || self.static_details.is_some() {
             Some(UserMetadata {
-                summary: self.static_summary.map(Into::into),
-                details: self.static_details.map(Into::into),
+                summary: self.static_summary.map(|s| s.as_json_payload().expect("String-to-JSON payload serialization is infallible")),
+                details: self.static_details.map(|s| s.as_json_payload().expect("String-to-JSON payload serialization is infallible")),
```
chris-olszewski: Same as above

crates/client/src/lib.rs
```diff
@@ -1030,6 +1031,22 @@ where
         let workflow_id = options.workflow_id.clone();
         let task_queue_name = options.task_queue.clone();
 
+        let user_metadata = if options.static_summary.is_some() || options.static_details.is_some()
```
chris-olszewski, line 1034: This should go through the user payload converter: e.g. [ruby](https://github.com/temporalio/sdk-ruby/blob/ddb2762486fd9602080ba377742e63f6f27e08c8/temporalio/lib/temporalio/internal/proto_utils.rb#L127)

crates/sdk-core/tests/integ_tests/workflow_tests/queries.rs
```diff
@@ -412,3 +412,182 @@ async fn query_returns_workflow_context_view_info() {
...
+#[workflow]
+#[derive(Default)]
+struct CurrentDetailsWf;
+
+#[workflow_methods]
+impl CurrentDetailsWf {
+    #[run(name = DEFAULT_WORKFLOW_TYPE)]
+    async fn run(ctx: &mut WorkflowContext<Self>) -> WorkflowResult<()> {
+        ctx.set_current_details("details from workflow");
+        std::future::pending::<()>().await;
```
chris-olszewski: Just to keep integration tests closer to end user DX
```suggestion
        ctx.wait_condition(|_| false).await;
```

crates/sdk-core/tests/integ_tests/workflow_tests/queries.rs
```diff
@@ -412,3 +412,182 @@ async fn query_returns_workflow_context_view_info() {
...
+/// Workflow that blocks indefinitely without ever setting current_details.
+#[workflow]
+#[derive(Default)]
+struct NoCurrentDetailsWf;
+
+#[workflow_methods]
+impl NoCurrentDetailsWf {
+    #[run(name = DEFAULT_WORKFLOW_TYPE)]
+    async fn run(_ctx: &mut WorkflowContext<Self>) -> WorkflowResult<()> {
+        std::future::pending::<()>().await;
```
chris-olszewski: Same as above
```suggestion
        ctx.wait_condition(|_| false).await;
```

crates/sdk-core/tests/integ_tests/workflow_tests/queries.rs
```diff
@@ -412,3 +412,182 @@ async fn query_returns_workflow_context_view_info() {
...
+    t.add_full_wf_task();
+
+    let tasks = [
+        // First task: workflow starts, sets current_details, and blocks on pending.
+        hist_to_poll_resp(&t, wfid.to_owned(), ResponseType::ToTaskNum(1)),
+        // Second task: legacy query for __temporal_workflow_metadata while workflow is blocked.
+        {
+            let mut pr = hist_to_poll_resp(&t, wfid.to_owned(), ResponseType::ToTaskNum(1));
+            pr.query = Some(WorkflowQuery {
+                query_type: "__temporal_workflow_metadata".to_string(),
```
chris-olszewski, line 449: I would like to see how end users would use this query with both typed and untyped workflow handles.

You will probably need to update our procmacro to inject a `workflow_metadata` marker struct to make the typed workflow handle work.

crates/client/src/lib.rs
Sushisource: Aaaaah yeah ok cool makes sense

crates/sdk/src/workflow_future.rs
```diff
@@ -263,15 +263,45 @@ impl WorkflowFuture {
...
-                        Ok(r) => r,
-                        Err(e) => Some(Err(anyhow!(
-                            "Panic in query handler: {}",
-                            panic_formatter(e)
-                        )
-                        .into())),
+                    let dispatch_result = if query_type == "__temporal_workflow_metadata" {
+                        // Mirror the proto JSON shape of temporal.api.sdk.v1.WorkflowMetadata.
+                        #[derive(serde::Serialize)]
+                        struct WorkflowMetadataJson {
```
chris-olszewski, line 272: Is there a reason we aren't using the proto here and instead redefining it?

crates/sdk/src/workflow_future.rs
```diff
@@ -263,15 +263,45 @@ impl WorkflowFuture {
...
+                                    "messageType".to_string(),
+                                    b"temporal.api.sdk.v1.WorkflowMetadata".to_vec(),
+                                ),
+                            ]
+                            .into_iter()
+                            .collect(),
+                            data: json_bytes,
+                            ..Default::default()
+                        };
+                        Some(Ok(payload))
```
chris-olszewski: We should be using the default payload converter here instead of special casing the logic. We currently don't have a proto json wrapper type, but it should be fine for this to be encoded as `json/plain`. You could leave a TODO for us to come back and reevaluate once we have proto json support.

chris-olszewski (changes requested):
Overall this makes sense to me just a few things which are mostly issues the pre-existed the PR.

- We should be running through the static user metadata through the payload converter. This might be some plumbing so feel free to do in a precursor PR if that would be eaiser.
- Instead of handrolling the `json/proto` encoding, it is fine to reuse the default converter and use `json/plain`.
- I think this query is currently unusable with a typed workflow, this could be done as a follow up PR.

crates/client/src/lib.rs
Sushisource: Hmm, I wonder if Go isn't doing this properly then? https://github.com/temporalio/sdk-go/blob/25039756e1f5c603466641064b0bd04dfb492d73/internal/internal_workflow_client.go#L3027 -- it sure looks like the path to here is the user's full data converter

It does make sense to me it'd have to be JSON, otherwise the UI/CLI can't render it.

chris-olszewski: IIUC that was for the query for current details. User metadata on workflow start/activity summaries seem to run through user payload converters. [TS](https://github.com/temporalio/sdk-typescript/blob/b782f914fe87739c8cfb20852e316e2ef7df0a95/packages/workflow/src/workflow.ts#L209) [Ruby](https://github.com/temporalio/sdk-ruby/blob/ddb2762486fd9602080ba377742e63f6f27e08c8/temporalio/lib/temporalio/internal/proto_utils.rb#L127)

Sushisource: Looks like Java and Python do it as well. Arguably this doesn't make much sense, but, I suppose I can see (at least for current details which is user-gettable) why we might allow a custom converter. We'll have to decide if we want to diverge from the others. My inclination would be no, even though maybe what is being done right now doesn't totally make sense. We might need to go see if we can dig up what the rationale was.

```
Java SDK — makeUserMetaData (WorkflowExecutionUtils.java:224) takes a DataConverter dc and calls dc.toPayload(). Every call site passes either:
  - dataConverterWithCurrentWorkflowContext (workflow-side, SyncWorkflowContext.java:121-123) — the user's full DataConverter with serialization context
  - dataConverterWithWorkflowContext (client-side, RootWorkflowClientInvoker.java:62-67) — clientOptions.getDataConverter().withContext(...) — again the user's full data converter

  Go SDK — buildUserMetadata (internal_workflow_client.go:3016) takes a converter.DataConverter and calls dataConverter.ToPayload(). Call sites pass w.client.dataConverter (client-side)
  or wc.dataConverter (workflow-side) — the user's full data converter.
  
  
  Here's the full picture for Python:

  _encode_user_metadata (client.py:9741) takes a DataConverter and calls converter.encode() — which runs through both the payload converter and the codec.

  Client-side call sites:
  - _populate_start_workflow_execution_request (client.py:8111-8115): data_converter = self._client.data_converter.with_context(WorkflowSerializationContext(...)) — the user's full data
  converter with context
  - Schedule action _to_proto (client.py:6164-6169): same — client.data_converter.with_context(...)

  Workflow-side call sites (activities, child workflows, timers):
  - Activities (_workflow_instance.py:3058): self._payload_converter.to_payload(summary) — the user's payload converter with ActivitySerializationContext
  - Child workflows (_workflow_instance.py:3230): self._payload_converter.to_payload(...) — user's payload converter with WorkflowSerializationContext
  - Timers (_workflow_instance.py:1746): self._workflow_context_payload_converter.to_payload(summary) — user's payload converter with workflow context
```

crates/sdk/src/workflow_future.rs
GregoryTravis: Using the proto here seems to require some changes to build.rs (see [TODO](https://github.com/temporalio/sdk-core/blob/master/crates/common/build.rs#L180)) which seemed a bit out of scope for this PR, but I agree that duplicating it here is bad, so I can go ahead with those changes.

Sushisource: _normal_ JSON serialization should actually already work right now just fine - it's the proto-compatible JSON serialization that's annoying, but, you don't need that for this.

crates/client/src/lib.rs
GregoryTravis: Conclusion after discussion: all three fields should *not* go through a user converter, and the other SDKs should do this as well.

crates/sdk/src/workflow_future.rs
GregoryTravis: Keeping as-is, after discussion with @chris-olszewski -- "the protobuf's Serialize would follow the json/protobuf style not json/plain."

crates/sdk-core/tests/integ_tests/workflow_tests/queries.rs
GregoryTravis: Will do in a followup PR.

crates/sdk/src/workflow_future.rs
GregoryTravis: Using default converter.

crates/sdk/Cargo.toml
```diff
@@ -24,6 +24,7 @@ gethostname = "1.0.2"
 parking_lot = { version = "0.12", features = ["send_guard"] }
 prost-types = { workspace = true }
 serde = "1.0"
+serde_json = { workspace = true }
```
chris-olszewski: I don't believe this is still necessary

crates/sdk-core/tests/integ_tests/workflow_tests/queries.rs
```diff
@@ -412,3 +412,182 @@ async fn query_returns_workflow_context_view_info() {
...
+            panic!("Expected Succeeded query result variant");
+        };
+        let payload = success
+            .response
+            .as_ref()
+            .expect("Expected a response payload");
+
+        assert_eq!(
+            payload.metadata.get("encoding").map(|v| v.as_slice()),
+            Some(b"json/protobuf".as_slice()),
```
chris-olszewski:
```suggestion
            Some(b"json/plain".as_slice()),
```

crates/sdk-core/tests/integ_tests/workflow_tests/queries.rs
```diff
@@ -412,3 +412,182 @@ async fn query_returns_workflow_context_view_info() {
...
+            "Expected json/protobuf encoding"
+        );
+        assert_eq!(
+            payload.metadata.get("messageType").map(|v| v.as_slice()),
```
chris-olszewski: I don't think will be present anymore

crates/sdk-core/tests/integ_tests/workflow_tests/queries.rs
```diff
@@ -412,3 +412,182 @@ async fn query_returns_workflow_context_view_info() {
...
+            panic!("Expected Succeeded query result variant");
+        };
+        let payload = success
+            .response
+            .as_ref()
+            .expect("Expected a response payload in __temporal_workflow_metadata response");
+
+        assert_eq!(
+            payload.metadata.get("encoding").map(|v| v.as_slice()),
+            Some(b"json/protobuf".as_slice()),
```
chris-olszewski:
```suggestion
            Some(b"json/plain".as_slice()),
```

crates/sdk-core/tests/integ_tests/workflow_tests/queries.rs
```diff
@@ -412,3 +412,182 @@ async fn query_returns_workflow_context_view_info() {
...
+            "Expected json/protobuf encoding"
+        );
+        assert_eq!(
+            payload.metadata.get("messageType").map(|v| v.as_slice()),
```
chris-olszewski: No longer present

crates/sdk/src/workflow_future.rs
```diff
@@ -263,15 +263,40 @@ impl WorkflowFuture {
...
+                            data: &SerializationContextData::Workflow,
+                            converter: &converter,
+                        };
+                        let payload = converter.to_payload(
+                            &ctx,
+                            &WorkflowMetadataJson {
+                                current_details: self.base_ctx.current_details(),
+                            },
+                        );
+                        Some(payload.map_err(|e| anyhow::Error::from(e).into()))
```
chris-olszewski: Same behavior, just slightly more legible IMO
```suggestion
                        Some(Ok(payload?))
```

chris-olszewski (approved):

PR #1209, opened 2026-04-10
feat(sdk): add typed continue as new

chris-olszewski:
## What was changed
Added `continue_as_new` helper method to workflow context.

## Why?
More ergonomic and enforces correct input type for the workflow.

## Checklist

1. Closes N/A

2. How was this tested:
Added small unit tests to verify serialization error gets populated and input threaded through. Updated existing integration tests to use helper method.

3. Any docs updates needed?
Updated READMEs

### DIFF
--- a/crates/sdk-core/tests/integ_tests/workflow_tests/continue_as_new.rs
+++ b/crates/sdk-core/tests/integ_tests/workflow_tests/continue_as_new.rs
@@ -4,7 +4,7 @@ use temporalio_client::WorkflowStartOptions;
 use temporalio_common::{
     protos::{
         DEFAULT_WORKFLOW_TYPE, canned_histories,
-        coresdk::{AsJsonPayloadExt, workflow_commands::ContinueAsNewWorkflowExecution},
+        coresdk::workflow_commands::ContinueAsNewWorkflowExecution,
         temporal::api::{
             command::v1::command::Attributes,
             enums::v1::{CommandType, ContinueAsNewVersioningBehavior},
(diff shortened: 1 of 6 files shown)


crates/sdk/src/workflow_context.rs
```diff
@@ -726,6 +726,31 @@ impl<W> SyncWorkflowContext<W> {
...
 
+    /// Signal that this workflow should continue as a new workflow execution with the given input and
+    /// options.
+    ///
+    /// This always returns an `Err` which should be propigated.
+    pub fn continue_as_new(
+        &self,
+        input: &<W::Run as WorkflowDefinition>::Input,
+        opts: ContinueAsNewOptions,
+    ) -> Result<std::convert::Infallible, WorkflowTermination>
```
Sushisource, line 737: Fun to see Infallible in the OK spot

crates/sdk-core/tests/integ_tests/workflow_tests/continue_as_new.rs
```diff
@@ -145,12 +139,8 @@ impl ContinueAsNewSuggestedWf {
...
         // Second WFT: flag should be true (set on WFT started event 8)
         assert!(ctx.continue_as_new_suggested());
-        Err(WorkflowTermination::continue_as_new(
-            ContinueAsNewWorkflowExecution {
-                arguments: vec![[1].into()],
-                ..Default::default()
-            },
-        ))
+        ctx.continue_as_new(&(), ContinueAsNewOptions::default())?;
+        Ok(())
```
Sushisource, line 143: Will the compiler let you remove the `Ok(())` here because the OK path of CaN is infallible? 

Alternatively, writing the can call without the `?` as the final statement would make sense to me here

Sushisource (approved):

crates/sdk-core/tests/integ_tests/workflow_tests/continue_as_new.rs
chris-olszewski: Sadly, no

PR #1212, opened 2026-04-10
Add DNS load balancing

Sushisource:
## What was changed
Title

## Why?
Matches behavior of Go & Java, theoretically improves availability.

## Checklist

1. Closes 

2. How was this tested:
Added some unit tests, virtually all existing tests run through this right now anyway, importantly including cloud tests.

3. Any docs updates needed?

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -35,7 +35,7 @@ hyper-util = "0.1.16"
 opentelemetry = { workspace = true, features = ["metrics"], optional = true }
 parking_lot = "0.12"
 thiserror = { workspace = true }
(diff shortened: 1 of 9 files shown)


chris-olszewski (changes requested):
Example failures are real. I think we should disable DNS load balancing if the host is localhost.

Sushisource:
> Example failures are real. I think we should disable DNS load balancing if the host is localhost.

Yeah, we probably should. I fixed it in a few places but it requires whatever's serving on localhost to bind properly which users may not do.

chris-olszewski (approved):

PR #1221, opened 2026-04-17
💥 feat(sdk): enforce some activity to close timeout is set

chris-olszewski:
## What was changed
- Move to a builder for `ActivityOptions`
- Combine `schedule_to_close_timeout` and `start_to_close_timeout` into a combined field of `close_timeouts: ActivityCloseTimeouts` to enforce that some timeout to close is set.

Aside from the vanilla builder we add a few methods to help with ergonomics by letting them skip importing the new enum:
 - `ActivityOptions::schedule_to_close_timeout`/`ActivityOptions::start_to_close_timeout`: Directly construct activity options with just this timeout set
 - `ActivityOptions::with_schedule_to_close_timeout`/`ActivityOptions::with_start_to_close_timeout`: Start an options builder with one of these timeouts set
There is no both variant for these since it would take 2 `Duration` objects and there isn't a clear way to indicate which one is for which timeout without a struct, which we already have this with the enum.

## Why?
Collapsing the close timeouts allows us to enforce that all activities must have a timeout to close.

I ended up combining this with moving `ActivityOptions` to a builder as the close timeout change was invasive enough and at the end of the day I wanted to see the end state of how we want users to construct these options.

### Alternatives Considered

There is a nice design where we continue to expose the activity timeout setting only via the builder with `.start_to_close_timeout` and `schedule_to_close_timeout` setters that both flow into a single required field.
https://github.com/elastio/bon/issues/149

## Checklist

1. Closes N/A

2. How was this tested:
Migrated existing tests/examples to use new activity options API

3. Any docs updates needed?
README was updated

### DIFF
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -66,6 +66,7 @@ opentelemetry-otlp = { version = "0.31", default-features = false, features = [
   "tls-roots",
   "http-proto",
   "grpc-tonic",
+  "reqwest-blocking-client",
   "reqwest-rustls",
 ], optional = true }
 parking_lot = { version = "0.12" }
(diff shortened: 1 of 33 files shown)


Sushisource (approved):
Nice! This is a big improvement.

PR #1226, opened 2026-04-21
💥 feat(sdk): failure converter + rich error types

### LINKED ISSUE #1137: [Feature Request] Failure Converter implementation
`FailureConverter` trait is defined but `DefaultFailureConverter` is a no-op. Implement proper conversion of Rust errors to/from Temporal `Failure` protobuf with data converter support for encryption.

chris-olszewski:
### Richer Error Types
We no longer return `Failure` protos directly as our error types, instead providing more specific error variants for each method in order to simplify error handling.

Due to `FailureConverter` living in common, we also move the error types defined in `temporalio-sdk` to common while still reexporting these for end users.

Nexus and some less common failures (e.g. workflow reset) are not enriched with first class accessors at this time.

### Failure Converter
We now define a default failure converter and use at the relevant callsites.

We redefine the `FailureConverter` trait definition to the following:
```
pub trait FailureConverter {
    /// Convert an error into a Temporal failure protobuf.
    fn to_failure(
        &self,
        error: OutgoingError,
        payload_converter: &PayloadConverter,
        context: &SerializationContextData,
    ) -> Result<Failure, PayloadConversionError>;

    /// Convert a Temporal failure protobuf back into a Rust error.
    fn to_error(
        &self,
        failure: Failure,
        payload_converter: &PayloadConverter,
        context: &SerializationContextData,
    ) -> Result<IncomingError, PayloadConversionError>;
}
```
This change focuses on 2 things: 
 - errors that we convert to a failure come from known call sites and thus the shape can be restricted compared to `Box<dyn Error>`
 - we intentionally limit what errors can be produced from `to_error` in order to ensure that we can provide a structured error experience for end users instead of forcing them to cast a `Box<dyn Error>`

Instead of giving this `IncomingError` directly to user, we further restrict the known types by using an error type hint at the callsite to indicate what the expected shape is:
```
/// Adapts a normalized incoming failure into a caller-facing error surface.
pub trait FailureDecodeHint {
    /// The caller-facing error type produced by this hint.
    type Output;

    /// Adapt a normalized incoming error to the caller-facing output.
    fn adapt(self, normalized: IncomingError) -> Self::Output;
}
```
The type hint is used to "force" a specific error type so the user does not have to match against all possible failures that server could send even though many are not well formed. The "adapt" is an infallible action so type hints are expected to provide a "default" if the failure info they receive is not expected.

Used via  `DataConverter` helper method:
```
pub fn to_error<H: FailureDecodeHint>(
        &self,
        context: &SerializationContextData,
        failure: crate::protos::temporal::api::failure::v1::Failure,
        hint: H,
    ) -> Result<H::Output, PayloadConversionError> {
        let normalized =
            self.failure_converter
                .to_error(failure, &self.payload_converter, context)?;
        Ok(hint.adapt(normalized))
    }
```

### Notes

This PR does not remove `anyhow` from our public API, that will be a follow up. This PR is already sprawling enough and the anyhow removal would muddy the waters. This PR should make that change easier as our exposure to anyhow is lessened.

I am not especially tied to `OutgoingError` or `IncomingError` naming.

## Why?
- Provide the ability for users to customize the behavior of how errors get encoded as failures
- Provide a more Rust-y way to deal with failures we receive from server as opposed to raw `Failure` protos or generic `Box<dyn Error>`

## Checklist

1. Closes #1137 

2. How was this tested:
Many integration tests & new unit tests

3. Any docs updates needed?
Rust docs


crates/common/src/data_converters/failure_converter.rs
```diff
@@ -0,0 +1,1187 @@
...
+impl FailureDecodeHint for ActivityExecutionDecodeHint {
+    type Output = ActivityExecutionError;
+
+    fn adapt(self, normalized: IncomingError) -> Self::Output {
+        match normalized {
+            IncomingError::Activity(activity) => {
+                if self.cancelled {
+                    // We collapse to the inner cancellation error so callers do not see a cancel
+                    // caused by another cancel.
+                    if matches!(activity.cause(), Some(IncomingError::Cancelled(_))) {
```
Sushisource: I think these could be collapsed and save on the duplicate `Failed` branches?

crates/common/src/error.rs
```diff
@@ -0,0 +1,865 @@
...
+    }
+}
+
+/// A typed outbound error surface used before encoding to a Temporal failure proto.
+#[derive(Debug, thiserror::Error)]
+pub enum OutgoingError {
+    /// An error produced while completing an activity.
+    #[error(transparent)]
+    Activity(#[from] OutgoingActivityError),
+    /// An error produced while failing a workflow execution.
```
Sushisource: Does this necessarily involve failing the workflow execution or could it be an error that was caught and then sent elsewhere?

crates/common/src/data_converters/failure_converter.rs
```diff
@@ -0,0 +1,1187 @@
...
+        )))
+    }
+}
+
+impl FailureConverter for DefaultFailureConverter {
+    fn to_failure(
+        &self,
+        error: OutgoingError,
+        _: &PayloadConverter,
+        _: &SerializationContextData,
```
Sushisource: Don't we need to use the `PayloadConverter` to encode the `details` fields of errors? Or are they always pass-through in other SDKs?

crates/common/src/data_converters/failure_converter.rs
```diff
@@ -0,0 +1,1187 @@
...
+    ChildWorkflowSignal(&'a ChildWorkflowSignalError),
+    Generic(&'a (dyn std::error::Error + 'static)),
+}
+
+impl<'a> ClassifiedFailure<'a> {
+    fn from_error(err: &'a (dyn std::error::Error + 'static)) -> Self {
+        Self::from_known_error(err).unwrap_or(Self::Generic(err))
+    }
+
+    fn from_known_error(err: &'a (dyn std::error::Error + 'static)) -> Option<Self> {
```
Sushisource: Seems like this can just be inlined?

crates/common/src/data_converters/failure_converter.rs
```diff
@@ -0,0 +1,1187 @@
...
+                    non_retryable: self.is_non_retryable(),
+                    details: self.details().cloned(),
+                    next_retry_delay: self.next_retry_delay().and_then(|d| d.try_into().ok()),
+                    category: self.category() as i32,
+                },
+            )),
+            ..Default::default()
+        };
+        if failure.message.is_empty() {
+            failure.message = "Application failure".to_owned();
```
Sushisource: Wondering what other SDKs do, if we just leave it blank here?

crates/common/src/data_converters/failure_converter.rs
```diff
@@ -0,0 +1,1187 @@
...
+    #[test]
+    fn application_failures_skip_generic_wrappers_around_known_causes() {
+        let activity_failure = Failure {
+            message: "activity failed".to_owned(),
+            failure_info: Some(FailureInfo::ActivityFailureInfo(
+                ActivityFailureInfo::default(),
+            )),
+            ..Default::default()
+        };
+        let app = ApplicationFailure::new(anyhow::Error::new(TestError::new(
```
Sushisource, line 821: Is the `anyhow::Error::new` the thing being skipped here? And I assume this is to deal with `?` bubbling things up and `Into`ing them as anyhow or std errors, etc?

The idea makes sense to me and I think it's probably a reasonable thing to do regardless, but, would be worth calling out how it compares to other SDKs. We probably want a bit of coverage in the README or wherever appropriate explaining how the squashing/wrapping works.

crates/common/src/data_converters.rs
```diff
@@ -1,7 +1,14 @@
 //! Contains traits for and default implementations of data converters, codecs, and other
 //! serialization related functionality.
 
-use crate::protos::temporal::api::{common::v1::Payload, failure::v1::Failure};
+mod failure_converter;
```
Sushisource, line 4: I kinda prefer the `data_converters.rs` to `mod.rs` naming schemes. If you have multiple mod.rs files open it's harder to scan at a glance which one is which. I'd prefer not moving the file.

crates/common/src/error.rs
```diff
@@ -0,0 +1,865 @@
...
+            IncomingError::ChildWorkflowExecution(err) => err.failure(),
+            IncomingError::NexusOperationExecution(err) => err.failure(),
+            IncomingError::NexusHandler(err) => err.failure(),
+        }
+    }
+
+    /// Returns the normalized cause, if any.
+    pub fn cause(&self) -> Option<&IncomingError> {
+        match self {
+            IncomingError::Application(err) => err.cause(),
```
Sushisource, line 426: Lol. I love these enum patterns

crates/common/src/error.rs
```diff
@@ -0,0 +1,865 @@
...
+pub enum ChildWorkflowExecutionError {
+    /// The child workflow failed.
+    #[error("Child workflow failed: {}", .0.failure().message)]
+    Failed(#[source] Box<ChildWorkflowFailureError>),
+    /// Failed to serialize input or deserialize the child workflow result payload.
+    #[error("Payload conversion failed: {0}")]
+    Serialization(#[from] PayloadConversionError),
+}
+
+impl ChildWorkflowExecutionError {
```
Sushisource, line 952: I think `is_cancelled()` and `is_timed_out()` helpers that drill into the cause on this & activity errors would be useful.

I know I had/have those on the protos themselves and they are useful.

crates/sdk-core/tests/integ_tests/workflow_tests/local_activities.rs
```diff
@@ -639,10 +640,13 @@ async fn x_to_close_timeout(#[case] is_schedule: bool) {
...
             let err = res.unwrap_err();
-            if let ActivityExecutionError::Failed(f) = &err {
+            if let ActivityExecutionError::Failed(failure) = &err
+                && let Some(temporalio_common::error::IncomingError::Timeout(timeout)) =
+                    failure.cause()
+            {
                 assert_eq!(
-                    f.is_timeout(),
-                    Some(TimeoutType::try_from(timeout_type).unwrap())
+                    timeout.timeout_type(),
```
Sushisource: Ya, per my other comment about the helpers this was easier to write before. Both ways are good, but the helper is something I think people would reach for a lot.

crates/sdk/src/lib.rs
```diff
@@ -949,42 +952,42 @@ impl ActivityHalf {
...
-                            format!("Activity function panicked: {}", panic_formatter(e)),
-                            true,
-                        )),
+                        Err(e) => ActivityExecutionResult::fail(
+                            data_converter.to_failure(
+                                &activity_context,
+                                OutgoingError::Activity(OutgoingActivityError::Application(
+                                    ApplicationFailure::non_retryable(anyhow!(
+                                        "Activity function panicked: {}",
+                                        panic_formatter(e)
```
Sushisource: Acknowledging that it was non-retryable before this change... should this actually be retryable? Looks like activity panics in Go are retryable, which makes sense to me.

crates/sdk/src/workflow_context.rs
```diff
@@ -759,7 +693,7 @@ impl<W> SyncWorkflowContext<W> {
         };
         let arguments = pc
             .to_payloads(&ctx, input)
-            .map_err(WorkflowTermination::failed)?;
+            .map_err(|err| WorkflowTermination::from(anyhow::Error::new(err)))?;
```
Sushisource: Here and a few other places were needing to wrap in anyhow first. Is this `err` just a std error? We should have a `From<std::Error>` implementation too for all these I think

crates/sdk/src/workflow_context.rs
```diff
@@ -1517,13 +1448,13 @@ impl CancellableFuture<ActivityResolution> for LATimerBackoffFut {
 enum ActivityFut<F, Output> {
     /// Immediate error (e.g., input serialization failure). Resolves on first poll.
     Errored {
-        error: Option<ActivityExecutionError>,
+        error: Box<Option<ActivityExecutionError>>,
```
Sushisource: I think this would typically be `Option<Box<xxx>>` since that doesn't always have to heap allocate

crates/sdk/src/workflow_future.rs
```diff
@@ -147,7 +148,9 @@ pub(crate) struct WorkflowFuture {
     cancel_sender: watch::Sender<Option<String>>,
     /// Base workflow context for sending commands
     base_ctx: BaseWorkflowContext,
-    /// Payload converter for serialization/deserialization
+    /// Data converter for workflow failure conversion and payload serialization.
+    data_converter: DataConverter,
+    /// Payload converter used directly by workflow command serialization paths.
     payload_converter: PayloadConverter,
```
Sushisource: Is the `payload_converter` not just always available inside the `data_converter`?

crates/common/src/error.rs
chris-olszewski: Bad doc comment, it is an error that produced while processing a workflow, not necessarily failing a workflow.

crates/common/src/data_converters/failure_converter.rs
chris-olszewski: Python/Ruby insert a default message on decode, rest leave blank. Moved to keeping it blank as that is the majority behavior.

crates/sdk/src/lib.rs
chris-olszewski: Ah yup, good catch.

crates/common/src/data_converters/failure_converter.rs
chris-olszewski: Yes, addressed this with the addition of `FailurePayloads` since these are semi-special cased since they might:
 - be coming from user land and are still types that need to be converted into payloads
 - coming from a Failure proto as a raw payload that will need to be converted before it is read.

Talked to Tim about his recent Nexus learnings and that seemed like a larger change that we don't have time to tackle right now.

crates/common/src/data_converters/failure_converter.rs
chris-olszewski: Yeah exactly, added a bit to the README.

crates/common/src/error.rs
chris-olszewski: I used a `as_timeout`/`as_cancelled` accessor similar to `serde_json::Value::as_array` etc. I think this is a little nicer in case we end up wanting to expose addition information on the timeout/cancellation errors we don't have a conflict with what piece of information `is_timed_out` should return.

Sushisource (approved):
Looking good to me. Maybe run it through Claude bot just to see if anything useful pops up

chris-olszewski:
@claude review

chris-olszewski:
@claude review

Issue #1213, opened 2026-04-14
[Feature Request] Implement WorkflowHistory as a stream


chris-olszewski:
### Is your feature request related to a problem? Please describe.

Address this [TODO](https://github.com/temporalio/sdk-core/blob/08adc4a43be357a214fe80ae64f877a0ca8c8ee2/crates/client/src/workflow_handle.rs#L78) by implementing `WorkflowHistory` as a stream to avoid fetching the entire history at once. At the moment we fetch all history eagerly.

### Describe the solution you'd like

A `Stream` implementation that fetches workflow history lazily.

### Additional context

See `Client::list_workflows` for an example of creating a stream over server calls
https://github.com/temporalio/sdk-core/blob/08adc4a43be357a214fe80ae64f877a0ca8c8ee2/crates/client/src/lib.rs#L1156-L1230

emiling:
@chris-olszewski  I’ve been going through the temporal repositories to get familiar with the codebase, and I came across this “good first issue.” Would it be okay if I worked on it?

chris-olszewski:
@emiling Yes! Any issues with that label are good candidates for new contributors.

iksuddle:
Hi, this hasn't been worked on in a while. Could I pick it up?

Sushisource:
> Hi, this hasn't been worked on in a while. Could I pick it up?

Go for it!

PR #1232, opened 2026-04-24
feat(sdk) : implement WorkflowHistory as a stream

emiling:
## What was changed

- Added `WorkflowHistoryStream` and `fetch_history_stream` to `WorkflowHandle`, enabling lazy, paginated streaming of workflow history events instead of eagerly fetching all pages into memory at once. 
- This follows the existing `ListWorkflowsStream` pattern using `stream::unfold` with and internal page buffer. 
- This change is purely addictive— existing `fetch_history` and `WorkflowHistory` APIs remain unchanged.

## Why?

- Large workflow histories can consume significant memory and introduce latency when fetched eagerly, since all pages must be retrieved before any events can be processed.
- A lazy stream allows callers to process events incrementally as they arrive, reducing peak memory usage and improving time-to-first-event 

## Checklist

1. Closes  #1213

2. How was this tested:

No new unit tests were added, matching the existing `ListWorkflowsStream` which also relies solely on integration tests. Happy to add integration tests if reviewers think it's needed


crates/client/src/workflow_handle.rs
```diff
@@ -275,7 +282,6 @@ impl WorkflowExecutionDescription {
     }
 }
 
-// TODO [rust-sdk-branch]: Could implment stream a-la ListWorkflowsStream
```
Sushisource, line 278: I think rather than having two totally separate APIs, we can just change `WorkflowHistory` to become like your new stream-based type. Since we're still in pre-release we can change the public API. 

We can keep `into_events` as a helper for people who would rather consume it all at once.

emiling: @Sushisource Originally, I separated this out to maintain backward compatibility. Since we're still in pre-release and public api changes are fine, I’ve updated it to use the stream-based WorkflowHistory as suggested. Thanks for the suggestion!

Sushisource:
@emiling Looks like formatting needs to get run and a variety of tests need updating

chris-olszewski:
Closing as PR seem abandoned and want to open this issue up to first time contributors. Feel free to reopen if this was a mistake.

PR #1236, opened 2026-04-25
chore: feature cleanup

chris-olszewski:
## What was changed
Make temporalio-sdk feature forwarding explicit so SDK users can opt out of optional lower-level dependencies with `default-features = false`. Most of this boiled down to disabling default features when declaring dependencies on in-repo crates.

## Why?
Give Rust SDK users full control of what underlying features get enabled. Building on top of `core-based-sdk` was too coarse as end users in other languages can't specify Cargo level features as they can in Rust.

## Checklist

1. Closes N/A

2. How was this tested:
CI

3. Any docs updates needed?
Updated README

### DIFF
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -14,7 +14,7 @@ exclude = ["protos/*/.github/*"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-history_builders = ["rand"]
+history_builders = []
 otel = ["dep:opentelemetry", "dep:opentelemetry_sdk", "dep:opentelemetry-otlp"]
 prometheus = [
   "dep:prometheus",
@@ -26,11 +26,14 @@ prometheus = [
 envconfig = ["dep:toml", "dep:dirs"]
 serde_serialize = []
 test-utilities = ["history_builders"]
-core-based-sdk = [
-  "prometheus",
-  "envconfig",
+core-telemetry-bridge = [
   "dep:ringbuf",
   "dep:futures-channel",
+]
+core-based-sdk = [
+  "core-telemetry-bridge",
+  "prometheus",
+  "envconfig",
 ]
 
 [dependencies]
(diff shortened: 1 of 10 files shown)


crates/common/Cargo.toml
```diff
@@ -26,12 +26,8 @@ prometheus = [
 envconfig = ["dep:toml", "dep:dirs"]
 serde_serialize = []
 test-utilities = ["history_builders"]
-core-based-sdk = [
-  "prometheus",
-  "envconfig",
-  "dep:ringbuf",
-  "dep:futures-channel",
-]
+core-telemetry-bridge = ["dep:ringbuf", "dep:futures-channel"]
```
chris-olszewski, line 29: Splitting this out so we can allow Rust SDK users to opt out of `prometheus`/`envconfig`

crates/sdk-core/Cargo.toml
```diff
@@ -10,12 +10,15 @@ repository = "https://github.com/temporalio/sdk-rust"
 keywords = ["temporal", "workflow"]
 categories = ["development-tools"]
 exclude = ["machine_coverage/*"]
-rust-version = "1.88.0" # due to cargo msrv find
+# due to cargo msrv find
```
chris-olszewski, line 13: Comment move as my TOML formatter claims that inline `#` comments should start at column 80 which looks horrible.

crates/sdk/Cargo.toml
```diff
@@ -60,9 +62,11 @@ futures = "0.3"
 rstest = "0.26"
 
 [features]
-default = []
-antithesis_assertions = ["temporalio-sdk-core/antithesis_assertions"]
```
chris-olszewski, line 64: I didn't see any real reason why this needed to be surfaced to the SDK crate.

crates/common/Cargo.toml
```diff
@@ -14,7 +14,7 @@ exclude = ["protos/*/.github/*"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-history_builders = ["rand"]
+history_builders = ["dep:rand"]
```
chris-olszewski, line 17: P sure `dep:` prefix is required here

Sushisource (approved):

PR #1237, opened 2026-04-27
Move Core-specific test utilities out of the Common crate

mjameswh:
## What changed

- Moved `HistoryBuilder`, `HistoryInfo`, `test_utils.rs` and canned histories out of the `common` create to the `sdk-core` crate.
- Removed the `history_builders` and `test-utilities` features on the `common` crate.

## Why

- The content of those files is exclusively used by the `sdk-core` crate's internal tests. They are not meant for public usage.
- Having them in the `sdk-core` crate makes it possible for `HistoryBuilder` and canned histories to have access to the `InternalFlags` enum.

### DIFF
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -14,7 +14,6 @@ exclude = ["protos/*/.github/*"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-history_builders = ["dep:rand"]
 otel = ["dep:opentelemetry", "dep:opentelemetry_sdk", "dep:opentelemetry-otlp"]
 prometheus = [
   "dep:prometheus",
@@ -25,7 +24,6 @@ prometheus = [
 ]
 envconfig = ["dep:toml", "dep:dirs"]
 serde_serialize = []
-test-utilities = ["history_builders"]
 core-telemetry-bridge = ["dep:ringbuf", "dep:futures-channel"]
 core-based-sdk = ["core-telemetry-bridge", "prometheus", "envconfig"]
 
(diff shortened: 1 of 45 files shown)


crates/sdk-core/src/core_tests/workflow_tasks.rs
```diff
@@ -2391,7 +2389,13 @@ async fn lang_internal_flag_with_update() {
     let mut t = TestHistoryBuilder::default();
     t.add_by_type(EventType::WorkflowExecutionStarted);
     t.add_full_wf_task();
-    t.set_flags_last_wft(&[1, 2], &[1]);
+    t.set_flags_last_wft(
+        &[
+            CoreInternalFlags::IdAndTypeDeterminismChecks as u32,
+            CoreInternalFlags::UpsertSearchAttributeOnPatch as u32,
```
chris-olszewski: 👍

chris-olszewski (approved):

crates/common/Cargo.toml
```diff
@@ -25,7 +24,6 @@ prometheus = [
 ]
 envconfig = ["dep:toml", "dep:dirs"]
 serde_serialize = []
-test-utilities = ["history_builders"]
 core-telemetry-bridge = ["dep:ringbuf", "dep:futures-channel"]
```
mjameswh, line 27: @chris-olszewski Note the casing inconsistency in `serde_serialize` vs other features.

crates/sdk-core/Cargo.toml
```diff
@@ -27,6 +27,9 @@ otel = [
 ]
 ephemeral-server = ["dep:flate2", "dep:reqwest", "dep:tar", "dep:zip"]
 debug-plugin = ["dep:reqwest", "dep:hyper"]
+# Exposes mocks and helpers used by sdk-core's own integration tests, benches,
+# and the `histfetch` binary. Not part of the public API surface; downstream
+# crates should not enable this feature.
 test-utilities = ["dep:assert_matches", "dep:bimap"]
```
mjameswh, line 33: Is there a Rust convention for features that are not meant to be used publicly?

chris-olszewski: Not that I know of. We could prefix with `__` to make the name convey internal usage, but I don't think that is a widely used pattern.

chris-olszewski (approved):
Looks like some casts need to be removed, but LGTM

PR #1239, opened 2026-04-29
:boom: WASM Workflows

Sushisource:
> [!TIP]
> It's easier to review the first two commits separately rather than trying to look at all of it at once. Those two commits contain almost all the main work, the ones after are smaller cleanup. 

## What was changed
Separated out the workflow APIs into their own crate (as well as the subset of common they need) and enabled the compilation of WASM workflows.

This change shouldn't be breaking for current Rust SDK users. Everything that was moved to new crates is re-exported from its old location. The execution semantics of workflows should remain the same as well, although the logic for driving the workflow future has indeed changed.

> [!NOTE]
> Protos are still [de]serialized on either side of the WASM boundary, which is not ideal but avoids adding even more heft to this already-hefty PR. I think the long-term plan would be to codegen WIT types from our Proto definitions (for every language we end up doing WASM workflows for), but I wasn't going to go there right away.

## Why?
WASM workflows unlock some very cool potential future use cases but also, at minimum, provide a viable part-compile-time part-runtime sandbox for Rust SDK users.

### Summary

  Splits the workflow runtime out of crates/sdk into a new temporalio-workflow crate and adds a WASM execution backend alongside the existing native one.

 ### New crates

  - crates/workflow — owns workflow context, options, and the runtime interface. Workflow authors compile against this crate regardless of which backend executes them.
  - crates/common-wasm — WASM-compatible subset of common (protos, data converters, priority, task token, activity/workflow definitions). The existing common crate pulls in host-only
  dependencies that won't build for wasm32, so guest-relevant code was moved/duplicated here.

###  WASM workflow execution architecture

  The runtime defines a single logical guest/host interface, expressed two ways:

  1. WIT contract (crates/workflow/wit/) — defines temporal:workflow-runtime@0.1.0:
    - workflow-guest (exported by the workflow): list-workflows, instantiate-workflow, and a workflow-instance resource with activate and poll-routine.
    - workflow-host (imported by the workflow): set-current-details, push-command, etc.
    - Combined into the workflow-module world.
  2. Rust traits in temporalio_workflow::runtime (WorkflowInstance, WorkflowHost, shared types in runtime::types) — a direct mirror of the WIT shapes, used in-process by the native backend with
   no serialization.

  Both backends plug into the same WorkflowDefinitions registry via a WorkflowExecutionFactory, so the worker is backend-agnostic.

  **Native backend:** workflows compile into the worker binary; the worker invokes the Rust traits directly in-process. This is the existing path, refactored to sit behind the new trait surface.

  **WASM backend** (crates/sdk/src/workflow_wasm.rs, crates/workflow/src/component.rs):
  - Host side uses wasmtime component-model + wasmtime::component::bindgen! to load .wasm components from a file or in-memory bytes, instantiate the guest, and register each exported workflow
  into the same registry as native workflows.
  - Guest side uses wit-bindgen and an export_workflow_module! macro that takes a list of workflow types, builds a static component implementing workflow-guest, and wires WorkflowImplementation
   impls (generated by #[workflow_methods]) into the WIT exports. Workflow authors write the same code as native — the macro is the only extra step.
  - The samples/wasm-workflows/hello crate is the reference guest setup (compiles to wasm32-wasip2, uses export_workflow_module!).

  **Polling model:** the guest interface is intentionally synchronous — activate and poll-routine return ordinary results, and the guest signals "blocked" by returning `routine-poll-result.made-progress = false`. The host re-enters poll-routine after the relevant activation lands. This matches what stable wasmtime + wit-bindgen support today; wit/README.md
  documents the planned WASIp3 / Preview-3 evolution to async func (a major-version WIT bump).

###  Other changes

  - Macro consistency — activity_definitions and workflow_definitions macros now share structure; mismatched-definition compile-fail test added.
  - New wasm_workflow_tests integration test exercises end-to-end execution of a WASM workflow against a real worker.


crates/common-wasm/build.rs
```diff
@@ -0,0 +1,182 @@
...
+    ".temporal.api.testservice",
+    ".temporal.api.update",
+    ".temporal.api.version",
+    ".temporal.api.worker",
+    ".temporal.api.workflow",
+    ".temporal.api.workflowservice",
+];
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    println!("cargo:rerun-if-changed=../common/protos");
```
eamsden: When `temporalio-common-wasm` is built from its published crate, or as a path dependency outside this workspace, `../common/protos` will not exist because Cargo packages only contain files under this crate directory. The build script then fails before generating protobufs, which also breaks `temporalio-common` because it now depends on this crate.

The protos need to be included under `common-wasm`, or the build script needs to reference a path that is actually included in the packaged crate.

**(Codex)**

crates/macros/src/workflow_definitions.rs
```diff
@@ -660,7 +625,7 @@ impl WorkflowMethodsDefinition {
         const_definitions.push(generate_const_definition(&run_method.method, &module_ident));
 
         trait_impls.push(quote! {
-            impl ::temporalio_common::WorkflowDefinition for #module_ident::#struct_ident {
+            impl ::temporalio_workflow::common::WorkflowDefinition for #module_ident::#struct_ident {
```
eamsden, line 628: For crates that depend on `temporalio-sdk` and `temporalio-macros` but have not added `temporalio-workflow` as a direct dependency, the generated code now references `::temporalio_workflow`, which is not in the extern prelude. That makes `#[workflow_methods]` and `#[workflow]` expansions fail with E0433 for downstream users.

Either generate paths through a crate the user already depends on, or make the direct `temporalio-workflow` dependency requirement explicit and update the SDK re-export path accordingly.

**(Codex)**

crates/sdk/src/workflow_future.rs
```diff
@@ -538,310 +446,197 @@ impl Future for WorkflowFuture {
...
                         }
-                        workflow_command::Variant::ScheduleNexusOperation(req) => {
-                            CommandID::NexusOpStart(req.seq)
+                        Some(result) => {
+                            match result {
+                                RoutineCompletion::Main(MainRoutineCompletion::Blocked) => {}
+                                RoutineCompletion::Main(MainRoutineCompletion::TaskFailed(
+                                    task_failure,
+                                )) => {
+                                    self.fail_wft(
```
eamsden: When a WASM guest returns `MainRoutineCompletion::TaskFailed` with `force_cause` set, this branch rebuilds an error from only the message and passes `None` to `fail_wft`. That means core loses both the requested `WorkflowTaskFailedCause` and any structured failure details.

Components using the WIT field for causes like nondeterminism or gRPC-message-too-large will be reported incorrectly.

**(Codex)**

crates/sdk/src/workflow_future.rs
```diff
@@ -538,310 +446,197 @@ impl Future for WorkflowFuture {
...
-                            failure: Some(self.data_converter.to_failure(
-                                &SerializationContextData::Workflow,
-                                OutgoingError::Workflow(e),
-                            )),
-                        })
-                    }
-                },
-            };
-            activation_cmds.push(cmd.into())
+            activation_cmds.extend(self.host.take_commands());
```
eamsden, line 638: Commands accumulated in `NativeWorkflowHost` are only drained on the successful-completion path. If a workflow task fails after workflow code has pushed a command, for example by starting a timer or activity and then panicking or calling `force_task_fail`, those commands stay buffered.

On retry, replay will push the same commands again, so stale duplicates can be sent unless the buffer is cleared when failing the workflow task.

**(Codex)**

eamsden (changes requested):

crates/macros/src/workflow_definitions.rs
Sushisource: This isn't true since for most of these (certainly the one on this particular line) since they are re-exported from their original spots, but it looks like it may be true very specifically for some futures util types we re-export from the workflow crate. Please verify these AI-suggested fixes before adding them here.

crates/sdk/src/workflow_future.rs
Sushisource: This isn't a real failure mode since if a task fails, the workflow is going to be evicted. I'll add a test to verify (though many existing tests would fail if this wasn't true) - but, same comment here, please confirm these before relaying.

crates/sdk/examples/wasm_workflows/README.md
```diff
@@ -0,0 +1,9 @@
+# WASM Hello Workflow
```
chris-olszewski, line 1: Could we move this to be in crates/temporalio-sdk/examples? If not even just moving to be `examples` so we're consistent in the repo would be beneficial.

crates/common-wasm/build.rs
Sushisource: Yeah, this one is definitely an issue. I hate having _another_ crate but the one option is just create a new `temporal-protos` crate and have both refer to that.

Alternatively, we could move everything into `common-wasm` and put the payload visitation stuff behind a feature flag. That might be the cleanest way.

@chris-olszewski Thoughts?

crates/macros/src/workflow_definitions.rs
```diff
@@ -1196,6 +1229,17 @@ impl WorkflowMethodsDefinition {
                     <#impl_type>::name()
                 }
 
+                fn definition() -> ::temporalio_workflow::runtime::types::WorkflowDefinitionDescriptor {
+                    ::temporalio_workflow::runtime::types::WorkflowDefinitionDescriptor {
+                        workflow_type: Self::name().to_string(),
+                        has_init: #has_init,
+                        init_takes_input: #init_has_input,
+                        signals: vec![#(#signal_names),*],
+                        queries: vec![#(#query_names),*],
```
chris-olszewski, line 1238: Note to self, groundwork for listing query names on missing query error

crates/sdk/src/workflow_registry.rs
chris-olszewski: Worth calling out in doc comment that we now panic on dup workflow definitions.

crates/macros/src/lib.rs
```diff
@@ -24,6 +23,19 @@ pub fn activity(_attr: TokenStream, item: TokenStream) -> TokenStream {
     item
 }
 
+/// Declares activities without providing implementations. Each method must omit the
+/// `ActivityContext` parameter and have a body of exactly `unimplemented!()`. The macro emits
+/// the same marker structs and inherent consts as `#[activities]` so call sites use the same
+/// `MyActivities::greet` form, but no execution machinery is generated.
+///
+/// Intended for workflow crates that need typed activity declarations which are implemented
+/// elsewhere by a separate worker crate (or in another language).
```
chris-olszewski: Maybe split out the references to the marker structs/consts to the bottom. I think we're burying the real usefulness of this macro with implementation details.

crates/sdk-core/tests/integ_tests/workflow_tests/continue_as_new.rs
```diff
@@ -164,3 +164,47 @@ async fn continue_as_new_suggested_flag_exposed() {
...
+            ctx.continue_as_new(&false, opts)?;
+        }
+
+        assert!(ctx.search_attributes().indexed_fields.is_empty());
+        Ok(())
+    }
+}
+
+#[tokio::test]
+async fn clear_search_attributes_on_continue_as_new() {
```
chris-olszewski, line 190: Not to self, was this bug or just asserting some behavior?

crates/sdk-core/tests/integ_tests/workflow_tests/resets.rs
```diff
@@ -250,9 +258,14 @@ async fn reset_randomseed() {
             .await
             .unwrap();
 
-        // Wait for the now-reset workflow to finish
+        // Wait for the original run to terminate and the reset run to finish.
         let result = handle.get_result(Default::default()).await;
         assert_matches!(result, Err(WorkflowGetResultError::Terminated { .. }));
+        reset_handle.get_result(Default::default()).await.unwrap();
+        assert!(
```
chris-olszewski, line 269: Note to self, was there a bug fix for this additional assertion?

crates/sdk/src/workflow_wasm.rs
```diff
@@ -0,0 +1,395 @@
...
+        component_id: impl Into<String>,
+        path: impl Into<PathBuf>,
+    ) -> Result<Self, anyhow::Error> {
+        let path = path.into();
+        if !path.exists() {
+            bail!(
+                "WASM workflow component file does not exist: {}",
+                path.display()
+            );
+        }
```
chris-olszewski, line 65: Unsure of the benefit of this check compared to waiting until it fails during workflow registration

crates/sdk/src/workflow_wasm.rs
```diff
@@ -0,0 +1,395 @@
...
+    File(PathBuf),
+    Bytes(Arc<[u8]>),
+}
+
+impl WasmWorkflowComponent {
+    /// Register a workflow component from a component file on disk.
+    pub fn from_file(
+        component_id: impl Into<String>,
+        path: impl Into<PathBuf>,
+    ) -> Result<Self, anyhow::Error> {
```
chris-olszewski: Either in PR or follow up: moving to error enum to avoid `anyhow` in the public API

crates/sdk/src/workflow_future.rs
```diff
@@ -224,256 +193,218 @@ impl WorkflowFuture {
...
+            match job
+                .variant
+                .as_mut()
+                .context("Empty activation job variant")?
+            {
                 Variant::InitializeWorkflow(_) => {
-                    // Don't do anything in here. Init workflow is looked at earlier, before
-                    // jobs are handled, and may have information taken out of it to avoid clones.
+                    should_poll_routines = true;
+                    push_context!(ActivationJobContext::Passive);
```
chris-olszewski: Seems to be a manual expansion of the macro
```suggestion
                    push_polled_context!(ActivationJobContext::Passive);
```

crates/sdk/src/workflow_future.rs
```diff
@@ -224,256 +193,218 @@ impl WorkflowFuture {
...
+                Variant::ResolveChildWorkflowExecution(attrs) => {
+                    attrs
+                        .result
+                        .as_ref()
+                        .context("Child Workflow execution must have a result")?;
+                    push_polled_context!(ActivationJobContext::Passive);
+                }
+                Variant::UpdateRandomSeed(_) => {
+                    should_poll_routines = true;
+                    push_context!(ActivationJobContext::Passive);
```
chris-olszewski: Same as above
```suggestion
                    push_polled_context!(ActivationJobContext::Passive);
```

chris-olszewski (approved):
Took awhile, but got through this. Mostly questions to help me write the release notes when this lands.

crates/common-wasm/build.rs
chris-olszewski: I lean towards `temporalio-protos` because the boundary is extremely clear.

Sushisource: Changed to make a separate crate

crates/sdk/src/workflow_wasm.rs
Sushisource: Eh, it's a bit more specific at least. The empty bytes one below is kinda silly tho.

crates/sdk/src/workflow_wasm.rs
Sushisource: Done now, just using std errors since none of these really seem useful to match on.

eamsden (approved):

PR #1244, opened 2026-05-01
Ensure waiters are always notified in Nexus worker

VegetarianOrc:
## What was changed

- Add a guard to notification in complete_nexus_task to ensure that waiters are notified even if the running future is dropped.

## Why?

I discovered a bug while testing standalone nexus operations in Python where the nexus worker shutdown would hang indefinitely in some cases. 

The issue exists when Python has called into core with a task completion at the same time that core delivers a task cancellation to Python. The order of events that cause this problem are:

1. Core delivers task for a nexus start operation
2. Python starts to send a task completion for start operation (this triggers an rpc in Core)
3. While the the task delivery is in progress, Core delivers a "cancel_task" to Python
4. Python looks up the asyncio task and cancels it.
5. Py03 bindings drop the Rust future started in step 2 when step 4 happens and before `notify_waiters()` is executed.
6. Nexus task stream goes to exit, but waits forever on notification that was missed.

I've included a fix Python side to shield the task completion future in the [standalone nexus operations PR](https://github.com/temporalio/sdk-python/pull/1461/changes#diff-ae0118ba1d5e950d7ac96e52cd92a8c3a799dee4be92d1301ef5833d84e5ba9dR219) but I think it would be a good idea to apply a fix on the Rust side as well to ensure that if a task is ever removed from the outstanding_task_map then there is always a notification that follows even if the future is dropped.

## Checklist

1. How was this tested:

Added a test that starts shutdown while a task is in flight, forces the task completion future to drop before `notify_waiters()` is executed and shows that a subsequent poll produces a `PollError::ShutDown` as expected.

### DIFF
--- a/crates/sdk-core/src/core_tests/workers.rs
+++ b/crates/sdk-core/src/core_tests/workers.rs
@@ -1333,7 +1333,7 @@ async fn graceful_shutdown_sends_shutdown_worker_rpc_during_initiate() {
 
 /// Verifies that even if the a nexus task completion is dropped, the nexus worker
 /// is able to trigger PollShutdown.
-#[tokio::test]
+#[tokio::test(start_paused = true)]
 async fn nexus_shutdown_does_not_hang_when_pending_completion_is_cancelled() {
     let telemetry = telemetry_init(
         TelemetryOptions::builder()
--- a/crates/sdk-core/src/core_tests/workers.rs
+++ b/crates/sdk-core/src/core_tests/workers.rs
@@ -1330,11 +1330,15 @@ async fn graceful_shutdown_sends_shutdown_worker_rpc_during_initiate() {
     worker.finalize_shutdown().await;
 }
 
+/// Verifies that even if the a nexus task completion is dropped, the nexus worker
+/// is able to trigger PollShutdown.
 #[tokio::test]
 async fn nexus_shutdown_does_not_hang_when_pending_completion_is_cancelled() {
     let mut client = mock_manual_worker_client();
     let completion_rpc_started = Arc::new(Barrier::new(2));
     let completion_rpc_started_clone = completion_rpc_started.clone();
+
+    // Create a client that will hang on complete_nexus_task
     client
         .expect_complete_nexus_task()
         .times(1)
@@ -1370,16 +1374,22 @@ async fn nexus_shutdown_does_not_hang_when_pending_completion_is_cancelled() {
     });
     let worker = mock_worker(mocks);
 
+    // Poll the first task from the stream and initiate shutdown
     let nexus_task = worker.poll_nexus_task().await.unwrap();
     worker.initiate_shutdown();
 
-    let mut shutdown_poll = Box::pin(worker.poll_nexus_task());
-    tokio::time::timeout(Duration::from_millis(50), shutdown_poll.as_mut())
+    // Poll the stream again and expect it to wait for the task to complete
+    let mut poll_future = Box::pin(worker.poll_nexus_task());
+    tokio::time::timeout(Duration::from_millis(50), poll_future.as_mut())
         .await
-        .expect_err("shutdown poll should wait for the outstanding Nexus task");
+        .expect_err("poll should wait for the outstanding Nexus task");
 
+    // Send completion that we know will hang indefinitely before notifying waitiers
     let mut completion =
         Box::pin(worker.complete_nexus_task(create_test_nexus_completion(nexus_task.task_token())));
+
+    // Wait for the completion to start then drop it before notify_waitiers is triggered
+    // Use select so the completion future is polled and completion actually starts
     tokio::select! {
         _ = completion_rpc_started.wait() => {}
         result = completion.as_mut() => {
@@ -1388,11 +1398,13 @@ async fn nexus_shutdown_does_not_hang_when_pending_completion_is_cancelled() {
     }
     drop(completion);
 
-    let poll_result = tokio::time::timeout(Duration::from_secs(1), shutdown_poll.as_mut())
+    // Polling again should now return PollError::ShutDown because the outstanding task map is empty
+    // and waiters should have been notified.
+    let poll_result = tokio::time::timeout(Duration::from_secs(1), poll_future.as_mut())
         .await
         .expect("shutdown poll should finish when a pending completion future is cancelled");
-    assert_matches!(poll_result.unwrap_err(), PollError::ShutDown);
-    drop(shutdown_poll);
+    assert_matches!(poll_result, Err(PollError::ShutDown));
+    drop(poll_future);
 
     worker.shutdown().await;
     worker.finalize_shutdown().await;
--- a/crates/sdk-core/src/core_tests/workers.rs
+++ b/crates/sdk-core/src/core_tests/workers.rs
@@ -1384,11 +1384,11 @@ async fn nexus_shutdown_does_not_hang_when_pending_completion_is_cancelled() {
         .await
         .expect_err("poll should wait for the outstanding Nexus task");
 
-    // Send completion that we know will hang indefinitely before notifying waitiers
+    // Send completion that we know will hang indefinitely before notifying waiters
     let mut completion =
         Box::pin(worker.complete_nexus_task(create_test_nexus_completion(nexus_task.task_token())));
 
-    // Wait for the completion to start then drop it before notify_waitiers is triggered
+    // Wait for the completion to start then drop it before notify_waiters is triggered
     // Use select so the completion future is polled and completion actually starts
     tokio::select! {
         _ = completion_rpc_started.wait() => {}
--- a/crates/sdk-core/src/core_tests/workers.rs
+++ b/crates/sdk-core/src/core_tests/workers.rs
@@ -62,6 +62,7 @@ use temporalio_common::{
             },
         },
     },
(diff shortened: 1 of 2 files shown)


crates/sdk-core/src/core_tests/workers.rs
```diff
@@ -1326,3 +1329,83 @@ async fn graceful_shutdown_sends_shutdown_worker_rpc_during_initiate() {
...
+    let nexus_task = worker.poll_nexus_task().await.unwrap();
+    worker.initiate_shutdown();
+
+    // Poll the stream again and expect it to wait for the task to complete
+    let mut poll_future = Box::pin(worker.poll_nexus_task());
+    tokio::time::timeout(Duration::from_millis(50), poll_future.as_mut())
+        .await
+        .expect_err("poll should wait for the outstanding Nexus task");
+
+    // Send completion that we know will hang indefinitely before notifying waitiers
```
VegetarianOrc:
```suggestion
    // Send completion that we know will hang indefinitely before notifying waiters
```

crates/sdk-core/src/core_tests/workers.rs
```diff
@@ -1326,3 +1329,83 @@ async fn graceful_shutdown_sends_shutdown_worker_rpc_during_initiate() {
...
+    let mut completion =
+        Box::pin(worker.complete_nexus_task(create_test_nexus_completion(nexus_task.task_token())));
+
+    // Wait for the completion to start then drop it before notify_waitiers is triggered
```
VegetarianOrc:
```suggestion
    // Wait for the completion to start then drop it before notify_waiters is triggered
```

crates/sdk-core/src/worker/nexus.rs
```diff
@@ -294,7 +296,7 @@ impl NexusManager {
                 }
             };
 
-            self.task_completed_notify.notify_waiters();
+            task_completed_notify.notify_waiters();
```
chris-olszewski, line 298:
```suggestion
            drop(task_completed_notify);
```

crates/sdk-core/src/worker/nexus.rs
```diff
@@ -326,6 +328,37 @@ impl NexusManager {
...
+    fn notify_waiters(mut self) {
+        self.notify_inner();
+    }
+}
+
+impl Drop for TaskCompletedNotify {
+    fn drop(&mut self) {
+        self.notify_inner();
+    }
+}
```
chris-olszewski: Maybe go full RAII and change to `TaskCompletionGuard` and fully rely on `Drop` as the only API.
```suggestion
struct TaskCompletionGuard {
    inner: Option<Arc<Notify>>,
}

impl TaskCompletionGuard {
    fn new(notify: Arc<Notify>) -> Self {
        Self {
            inner: Some(notify),
        }
    }
}

impl Drop for TaskCompletionGuard {
    fn drop(&mut self) {
        if let Some(notify) = self.inner.take() {
            notify.notify_waiters();
        }
    }
}
```

chris-olszewski (commented):
The code changes make sense to prevent the hang, but I'm not sure if we want to support this. IIUC the hang displayed a real issue with the Python usage of `complete_task` being dropped.

Not sure what semantics Nexus has wrt once a handler completes a task around delivering that completion.

crates/sdk-core/src/core_tests/workers.rs
```diff
@@ -1326,3 +1329,83 @@ async fn graceful_shutdown_sends_shutdown_worker_rpc_during_initiate() {
...
+    });
+    let worker = mock_worker(mocks);
+
+    // Poll the first task from the stream and initiate shutdown
+    let nexus_task = worker.poll_nexus_task().await.unwrap();
+    worker.initiate_shutdown();
+
+    // Poll the stream again and expect it to wait for the task to complete
+    let mut poll_future = Box::pin(worker.poll_nexus_task());
+    tokio::time::timeout(Duration::from_millis(50), poll_future.as_mut())
```
cconstable, line 1394: Is it possible to use something like `#[tokio::test(start_paused = true)]` to get deterministic time here?

cconstable (dismissed):
I left one small comment but I think this looks nice. I would wait for a 2nd opinion from someone more familiar with sdk-core though.

crates/sdk-core/src/worker/nexus.rs
cconstable: This will just naturally drop a few lines down right?

`drop(task_completed_notify);` is less meaningful to me than `task_completed_notify.notify_waiters();` unless it was named something like `drop(task_notify_on_drop)` (TaskNotifyOnDrop). Take what I'm saying with a grain of salt as I'm not up to date with Rust idioms.

VegetarianOrc:
> The code changes make sense to prevent the hang, but I'm not sure if we want to support this. IIUC the hang displayed a real issue with the Python usage of `complete_task` being dropped.
> 
> Not sure what semantics Nexus has wrt once a handler completes a task around delivering that completion.

The situation that causes this essentially makes delivering the task completion semi-irrelevant. The situations where Core delivers a cancel to Lang are 1) the [request deadline has been reached](https://github.com/temporalio/sdk-rust/blob/87162f5083fd657def2d77686807633d1cb291f3/crates/sdk-core/src/worker/nexus.rs#L390-L403) and the task is assumed to have failed, or 2) the [worker is shutting down](https://github.com/temporalio/sdk-rust/blob/87162f5083fd657def2d77686807633d1cb291f3/crates/sdk-core/src/worker/nexus.rs#L483-L507). In both of these cases if the completion fails to be delivered to the server, the server will reschedule the task.

During shutdown, Core has an assumption that it will be notified for any outstanding task it's sent over to Lang. The guard is to ensure that a bug in Lang can't invalidate this invariant and prevent shutdown. I could be convinced that we don't want this here and would rather have the bug manifest on the Lang side, but given the race my thought was it would be best to protect here as it might affect users before we caught it.


```rust
    async move {
                source_done_clone.cancelled().await;
                let (grace_killer, stop_grace) = futures_util::future::abortable(async {
                    if let Some(gp) = self.grace_period {
                        tokio::time::sleep(gp).await;
                        for (tt, _) in outstanding_task_clone.lock().iter() {
                            let _ = cancels_tx_clone.send(CancelNexusTask {
                                task_token: tt.0.clone(),
                                reason: NexusTaskCancelReason::WorkerShutdown.into(),
                            });
                        }
                    }
                });
                join!(
                    async {
                        // This is reached before the task completion is delivered from Lang to Core
                        // because the graceful timer starts in parallel. When the future that is dropped
                        // fails to notify, the assumption fails and we're stuck.
                        while !outstanding_task_clone.lock().is_empty() {
                            self.task_completed_notify.notified().await;
                        }
                        // If we were waiting for the grace period but everything already finished,
                        // we don't need to keep waiting.
                        stop_grace.abort();
                    },
                    grace_killer
                )
            }
```


Talking through this I do think it is worth it to have some error logging when the situation occurs so Core can both protect itself from getting stuck and the bug can be surfaced to Lang. I'll rework the guard a bit to log when drop is triggered and notify had not been called. Another way to do this would be to just set some timeout on the `notified().await` in shutdown and avoid the guard entirely. I'm not sure what a reasonable timeout would be, but worth considering that avenue as well.

crates/sdk-core/src/worker/nexus.rs
VegetarianOrc: I liked this, but after thinking through your comment about the approach, I think the best option is to keep the semantics separate and use `drop` to detect bad behavior from a caller and log it. Will rename though to reflect the purpose more clearly.

VegetarianOrc:
The guard will now emit a log like this when this when dropped without notifying:
> 2026-05-01T16:44:38.060261Z ERROR complete_nexus_task{task_queue=nexus-shutdown-cancelled-completion task_token="bmV4LXRhc2s=" status="NexusTaskCompletion(NexusResponse(StartOperation))"}: temporalio_sdk_core::worker::nexus: TaskCompletedGuard triggered notify on drop. This indicates that the caller has dropped the future for `complete_task`. This should not happen. Please file a bug report.

crates/sdk-core/src/core_tests/workers.rs
VegetarianOrc: I've added this! Good callout! Didn't need real-time to detect the hang so this works perfectly.

crates/sdk-core/src/worker/nexus.rs
```diff
@@ -326,6 +327,40 @@ impl NexusManager {
...
+        if let Some(notify) = self.inner.take() {
+            notify.notify_waiters();
+        }
+    }
+}
+
+impl Drop for TaskCompletedGuard {
+    fn drop(&mut self) {
+        if let Some(notify) = self.inner.take() {
+            error!(
```
Sushisource, line 354:
```suggestion
            dbg_panic!(
```

This will ensure tests panic but production just logs

Sushisource (commented):
One important suggestion to make it easier to detect in lang tests:

I think generally I agree with Chris, though, and I'm not sure I want to unblock here, but we should crash lang in non-production. The invariant that all Core tasks are responded to is important. However... it's also nice to not hose things up in the event this happens in the wild.

The other option would be to only allow the unblocking to happen if graceful shutdown duration is set, and it elapses. Then we can just force the conclusion.

PR #1246, opened 2026-05-01
fix(macros): drop anyhow dependency from workflow_methods expansion

### LINKED ISSUE #1245: [Bug] workflow_methods macro requires downstream crate to depend on anyhow

I am upgrading a Rust SDK user crate from `temporalio-*` `0.3.0` to `0.4.0` and using the proc-macro based workflow API.

The workflow crate intentionally does not depend on `anyhow`; it uses typed errors where needed. After the upgrade, a crate that uses `#[workflow_methods]` no longer compiles unless it adds a direct dependency on `anyhow`.

### Describe the bug

`temporalio-macros 0.4.0` expands `#[workflow_methods]` into code that directly references `::anyhow::Error` from the downstream crate's dependency graph.

The generated code appears to include this conversion:

```rust
.map_err(|e| ::temporalio_sdk::WorkflowTermination::from(::anyhow::Error::new(e)))
```

Because `anyhow` is not re-exported by `temporalio-sdk`, the downstream crate must add `anyhow` as a direct dependency even if application code does not use it.

This feels like a proc-macro hygiene / dependency boundary issue. The SDK already has `impl From<temporalio_common::data_converters::PayloadConversionError> for WorkflowTermination`, so this conversion probably does not need to go through `anyhow` at all.

### Minimal Reproduction

A workflow crate depending on `temporalio-sdk = "0.4.0"` and `temporalio-macros = "0.4.0"`, but not `anyhow`, fails to compile when using `#[workflow_methods]`.

```rust
use temporalio_macros::{workflow, workflow_methods};
use temporalio_sdk::{WorkflowContext, WorkflowResult};

#[workflow]
#[derive(Default)]
pub struct ExampleWorkflow;

#[workflow_methods]
impl ExampleWorkflow {
    #[run]
    pub async fn run(_ctx: &mut WorkflowContext<Self>) -> WorkflowResult<()> {
        Ok(())
    }
}
```

Compilation fails with an error like:

```text
error[E0433]: failed to resolve: could not find `anyhow` in the list of imported crates
  --> src/lib.rs:...
   |
... #[workflow_methods]
   | ^^^^^^^^^^^^^^^^^^^ could not find `anyhow` in the list of imported crates
```

Adding this fixes the compile error, but it should not be required by the macro expansion:

```toml
anyhow = "1"
```

### Environment/Versions

- OS and processor: macOS arm64
- Rust SDK crates: `0.4.0`
- Rust edition: 2024

### Additional context

The local workaround is to add a direct `anyhow` dependency to the workflow crate. That works, but it exposes an implementation detail of the SDK macro to user crates and is surprising for projects that intentionally avoid `anyhow` in favor of typed error handling.

pranc1ngpegasus:
## What was changed

The `#[workflow_methods]` macro previously expanded to code that referenced `::anyhow::Error` directly:

```rust
.map_err(|e| ::temporalio_sdk::WorkflowTermination::from(::anyhow::Error::new(e)))
```

This forced every downstream crate using the proc-macro workflow API to add a direct `anyhow` dependency, even when its application code did not use `anyhow`.

This PR:

1. Adds `impl From<workflows::WorkflowError> for WorkflowTermination` in `crates/sdk/src/lib.rs`. It dispatches on the existing `WorkflowError` variants and delegates to the already-present `From<PayloadConversionError>` / `From<anyhow::Error>` impls, so semantics are preserved.
2. Replaces the `anyhow::Error::new(e)` construction in the macro with the function reference `::temporalio_sdk::WorkflowTermination::from`, removing the `::anyhow::` path from the generated code.

## Why?

`anyhow` is an implementation detail of the SDK. Leaking it through macro expansion forces user crates that intentionally avoid `anyhow` (e.g. those using typed errors) to take a direct dependency on it just to get `#[workflow_methods]` to compile. The SDK already exposes the conversions we need via `WorkflowTermination`, so the macro can produce equivalent code without naming `anyhow`.

## Checklist

1. Closes #1245

2. How was this tested:
   - `cargo check --workspace --all-targets`
   - `cargo lint`
   - `cargo test -p temporalio-macros`
   - `cargo test -p temporalio-sdk --lib`

3. Any docs updates needed?
   N/A — no public API change beyond an additional `From` impl for `WorkflowTermination`.

### DIFF
--- a/crates/macros/src/workflow_definitions.rs
+++ b/crates/macros/src/workflow_definitions.rs
@@ -1012,7 +1012,7 @@ impl WorkflowMethodsDefinition {
                 let result = #run_call;
                 match result {
                     Ok(value) => ::temporalio_sdk::workflows::serialize_result(value, &ctx.payload_converter())
-                        .map_err(|e| ::temporalio_sdk::WorkflowTermination::from(::anyhow::Error::new(e))),
+                        .map_err(::temporalio_sdk::WorkflowTermination::from),
                     Err(e) => Err(e),
                 }
             }.boxed_local()
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -1281,6 +1281,15 @@ impl From<temporalio_common::data_converters::PayloadConversionError> for Workfl
     }
 }
 
+impl From<workflows::WorkflowError> for WorkflowTermination {
+    fn from(err: workflows::WorkflowError) -> Self {
+        match err {
+            workflows::WorkflowError::PayloadConversion(e) => Self::from(e),
+            workflows::WorkflowError::Execution(e) => Self::from(e),
+        }
+    }
+}
+
 impl From<ActivityExecutionError> for WorkflowTermination {
     fn from(value: ActivityExecutionError) -> Self {
         Self::Failed(value.into())


Sushisource (approved):
Thank you! This was definitely a mistake

PR #1251, opened 2026-05-06
expose dns load balancing options through c bridge

robcao:
## What was changed

Added `TemporalCoreClientDnsLoadBalancingOptions` to the C bridge, so that users of the C bridge can set these options how they need

## Why?

Users of the C bridge want to be able to change the dns load balancing options 

## Checklist

1. Closes 

https://github.com/temporalio/sdk-rust/issues/1250

3. How was this tested:

Consumed this change in the .NET SDK: https://github.com/temporalio/sdk-dotnet/pull/676

Verified all tests pass (both `cargo test` in this repository, and `dotnet test` in the dotnet-sdk`)

4. Any docs updates needed?

### DIFF
--- a/crates/sdk-core-c-bridge/include/temporal-sdk-core-c-bridge.h
+++ b/crates/sdk-core-c-bridge/include/temporal-sdk-core-c-bridge.h
@@ -145,6 +145,11 @@ typedef struct TemporalCoreClientHttpConnectProxyOptions {
 typedef void (*TemporalCoreClientGrpcOverrideCallback)(struct TemporalCoreClientGrpcOverrideRequest *request,
                                                        void *user_data);
 
+typedef struct TemporalCoreClientDnsLoadBalancingOptions {
+  bool enabled;
+  uint64_t resolution_interval_millis;
+} TemporalCoreClientDnsLoadBalancingOptions;
+
 typedef struct TemporalCoreConnectionOptions {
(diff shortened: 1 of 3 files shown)


Sushisource (approved):
Thanks!

Sushisource:
Looks like this just needs a `cargo fmt` @robcao

robcao:
woops, done

jmaeagle99:
I think this should drop the `enabled` fields in favor of breaking the default behavior for a cleaner design such that passing `null` actually means disabled (instead of default behavior) and push the default behavior up to the lang SDK layers, just like in https://github.com/temporalio/sdk-python/pull/1501. @Sushisource, what do you think?

robcao:
I'll be on vacation until June 8, so I created another commit with the alternate approach dropping the `enabled` field: https://github.com/robcao/sdk-core/commit/497a98eac6f5ccb84386af3e710855cab42dad92

Feel free to use that instead (or to re-create the change, or I could also pick this up when I return)

jmaeagle99:
Thank you for bringing the issue to our attention and proactively providing a fix.

I ended up reimplementing the change in https://github.com/temporalio/sdk-rust/pull/1267 without the `enabled` field and aligned it with the behavior of the other SDKs. I'll close out this PR at this time since it has been superseded.

PR #1258, opened 2026-05-12
docs: update sdk arch docs to include nexus.

cconstable:


arch_docs/sdks_intro.md
```diff
@@ -46,6 +46,51 @@ async def say_hello(name: str) -> str:
     return f"Hello, {name}!"
 ```
 
+### Nexus
+
+Nexus allows Temporal operations to be sent across Namespaces. Operations can be anything from starting a new workflow to updating the state of a currently running workflow. A Nexus Service is what we call the interface that abstracts sending Operations. A Nexus Service abstracts implementation details like the target Namespace and the Task Queues that the Operations run on. Like Activities, Nexus Operations get all the Temporal durable execution guarantees. Instead of interacting directly with a Nexus Service, we call a Nexus Endpoint. Nexus Endpoints are reverse proxies that take incoming Nexus Operations and map them to Nexus Services (Namespace + Task Queue). Nexus Endpoints are registered in a Nexus Registry that exposes them to your organization.
```
cconstable: please be critical of my terminology and phrasing

Sushisource: I think this is pretty accurate, but Nexus is not necessarily for  _just_ x-namespace calling - in theory users can implement Nexus handlers that aren't really meaningfully tied to a namespace in any way other than that we do require you to define the endpoints inside some namespace.

Also a "Temporal operation" isn't quite a well-defined thing (though a "Nexus operation" certainly is). I'd say more like:

"Nexus allows you to define and invoke functionality within and across namespaces by calling Nexus Operations. Nexus Operations can be anything from starting a new workflow to updating the state of a currently running workflow, or other arbitrary user-defined functionality."

It's currently in the plans to allow you to fulfill async Nexus operations totally independently of a workflow (by invoking a callback), but that's not quite a thing that exists yet.

Sushisource (approved):
Thanks for adding!

arch_docs/sdks_intro.md
cconstable: Thanks for clearing that up. That make sense. Updated.

PR #1263, opened 2026-05-13
chore(sdk): rename child_workflow -> start_child_workflow

chris-olszewski:
## What was changed
- Rename `child_workflow` -> `start_child_workflow`
- Add deprecated alias `child_workflow` for `start_child_workflow`

## Why?
`start_child_workflow` is more accurate and matches existing SDKs.

Deprecating instead of immediate removal due to Public Preview status with plans to remove after a release of deprecation.

## Checklist

1. Closes N/A

2. How was this tested:
N/A

3. Any docs updates needed?
Yes, once next version is published we can update docs.


Sushisource (approved):
I am totally in support of being consistent and I get it but man personally the long names just grind my gears lol, hence why it was this way rather than being start in the first place 🙈

PR #1269, opened 2026-05-14
fix: retry transport-sourced Cancelled errors and count them in metrics

### LINKED ISSUE #1252: Dropped activity completions on transport-level Cancelled are invisible and not retried sufficiently
**Situation**

An activity completes successfully on the worker, but during sustained transport-level network turbulence (e.g., AWS AZ outage), the SDK's RespondActivityTaskCompleted gRPC call fails with `Code::Cancelled` whose source chain is `tonic::transport::Error → hyper::Error("connection closed")`. This is semantically a transport-level failure (the wire died), not an application/caller-initiated cancellation, but the gRPC status code conflates the two.

**SDK Behavior (Currently)**

Does not include Cancelled in RETRYABLE_ERROR_CODES (correct per gRPC spec for true caller-cancel, but incorrect for transport-source Cancelled).

Has a one-shot goaway-retry workaround at that matches only on "connection closed" literal in the debug string and allows only a single retry 

After the single retry exhausts, the worker layer silently drops the completion with only a warn!`("Network error while completing activity")` log and produce no metric/error log/event in workflow history/application callback.

The request_failure metric is gated on `if let Ok(ref ok_res) = res`
it only increments when the call gets a parseable grpc-status header back from the server. Transport-killed calls bypass this counter entirely.

**Result ( Activity completion drop)**

Activity that did complete successfully on the worker silently has its completion dropped. The server eventually heartbeat-times-out the attempt and retries it from scratch. If the activity is not idempotent, the retry can fail in application specific ways and client has no observable signal that this occurred and diagnosis requires log-grepping or worker memory dumps.

_What's needed :-_ 

1. When `Code::Cancelled` has a `tonic::transport::Error → hyper::Error` source chain, SDK treat as Unavailable-class and retry on the same bounded retry budget as Unavailable (instead of the current single-retry goaway workaround) and also preserves the spec-correct default for application-level cancellations 
2. Also count transport-level failures (no server response received) in request_failure. Either add a new status_code=TRANSPORT_ERROR label, or reuse status_code=UNAVAILABLE with the gRPC message field carrying the transport detail so users can distinguish them from server-returned Unavailable on dashboards.

brucearctor:
## Summary

When `Code::Cancelled` originates from a transport-level failure (`tonic::transport::Error` → `hyper::Error`, e.g. connection closed during an AZ outage), the SDK now treats it as `Unavailable`-class and retries using the standard bounded backoff budget instead of the previous single-retry goaway workaround.

Additionally, transport-level failures that never produce a `grpc-status` HTTP header are now counted in the `request_failure` metric with a `status_code=TRANSPORT_ERROR` label.

Fixes #1252

## Problem

Three compounding issues cause silent activity completion drops during sustained network turbulence:

1. **One-shot retry limit**: The goaway-cancel workaround in `retry.rs` only allows a single retry. During sustained outages (not a single GOAWAY frame), that one retry burns immediately against a still-dead connection.

2. **Silent drop**: After the retry exhausts, `activities.rs` catches the forwarded error and logs `warn!("Network error while completing activity")` — no metric, no callback, no error propagation.

3. **Metrics blind spot**: `GrpcMetricSvc::call()` only increments `request_failure` when it can parse a `grpc-status` header from the HTTP response. Transport-killed calls return `Err(...)` at the HTTP level, so the counter is never incremented.

### `crates/client/src/retry.rs`
- Replaced the one-shot `have_retried_goaway_cancel` flag with `is_transport_cancelled()` — a helper that inspects the error source chain for `tonic::transport::Error` → `hyper::Error`
- Transport-sourced `Cancelled` errors now flow through the standard backoff budget (same as `Unavailable`)
- True application/caller-initiated cancellations (no transport error in source chain) remain non-retryable per gRPC spec

### `crates/client/src/metrics.rs`
- Added `svc_request_failed_transport()` method with `status_code=TRANSPORT_ERROR` label
- `GrpcMetricSvc::call()` now counts `Err(...)` responses (transport failures) in addition to `Ok(...)` responses with non-OK grpc-status headers

## Testing
- All 11 existing retry unit tests pass
- Clippy clean with `-D warnings`


crates/client/src/retry.rs
```diff
@@ -617,4 +617,82 @@ mod tests {
...
+        let status = Status::new(Code::Cancelled, "caller cancelled");
+        assert!(!is_transport_cancelled(&status));
+    }
+
+    // NOTE: A positive-path unit test for is_transport_cancelled (i.e., verifying it returns
+    // true for a status with tonic::transport::Error in its source chain) is not feasible
+    // because tonic::Status's public API doesn't allow constructing a Status with both a
+    // custom Code and a transport error source. In production, tonic internally builds this
+    // when a GOAWAY/connection-close kills an in-flight RPC. The detection logic is validated
+    // by the existing codebase's integration tests that exercise real gRPC connections.
```
Sushisource:
```suggestion
```

Redundant with the comment in the test below

crates/client/src/retry.rs
```diff
@@ -617,4 +617,82 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn transport_sourced_cancelled_retried_on_full_budget() {
+        // Verify the behavioral contract: when is_transport_cancelled returns true AND
+        // the code is Cancelled, the error handler retries on the full backoff budget.
+        //
```
Sushisource:
```suggestion
```

This is a little too specific I think. Referring to the exact variable name is susceptible to getting stale. The rest of the comment is plenty (a bit wordy even)

Sushisource (commented):
Just some minor comment cleanup. Thanks a lot for this, the idea is a good one.

brucearctor:
committed/accepted suggestions.

Thanks @Sushisource !

brucearctor:
I can't rerun, but seems like flaky CI test?

Sushisource (approved):

PR #1271, opened 2026-05-14
feat(client): support custom ServerCertVerifier in TlsOptions

### LINKED ISSUE #1248: [Feature Request] Accept a pre-built `rustls::ClientConfig` in client TLS options
## Is your feature request related to a problem?

`temporalio_client::TlsOptions` accepts TLS material as PEM byte strings only:

```rust
pub struct TlsOptions {
    pub server_root_ca_cert: Option<Vec<u8>>,
    pub client_cert: Option<Vec<u8>>,
    pub client_private_key: Option<Vec<u8>>,
    // ...
}
```

This works when certificates are static files, but it can't represent a `rustls::ClientConfig` built with a **custom `ResolvesClientCert` or `ServerCertVerifier`**, which is the standard rustls pattern for:

- Short-lived certificates with dynamic in-process rotation
- Custom server-certificate verification (pinning, SAN-URI trust-domain extraction, federated roots)
- TLS configuration owned by another crate that already manages the rustls setup

In these cases the caller already holds an `Arc<rustls::ClientConfig>` whose embedded resolver/verifier re-reads the cert source on every handshake. Re-serializing that to PEM bytes produces a one-time snapshot: rotation breaks, custom verification is lost.

## Describe the solution you'd like

Allow passing an `Arc<rustls::ClientConfig>` directly, bypassing PEM parsing. For example:

```rust
pub enum ClientTlsConfig {
    /// Existing path: PEM bytes parsed into a rustls config internally.
    Pem(TlsOptions),
    /// New: caller-supplied rustls config used as-is.
    Rustls(Arc<rustls::ClientConfig>),
}
```

Or equivalently, an additional field on the existing options struct that takes precedence when set.

`tonic::transport::ClientTlsConfig` already exposes `.rustls_client_config(Arc<ClientConfig>)`, so the plumbing through to the underlying channel should be minimal; the request is to expose that path through `temporalio_client`'s public surface.

## Describe alternatives you've considered

- **Extract PEM bytes from the issued config and pass those.** Works for the initial certificate's lifetime, but the connection holds the snapshot. When the source rotates, the worker keeps presenting the stale cert until process restart.
- **Reconnect on a timer.** Possible but ugly: requires guessing rotation cadence, drops in-flight long-polls, and still can't represent a custom verifier.
- **Fork/patch the client crate.** Unblocks, but maintenance burden.

## Additional context

The `ClientConfig` in question is itself stable for the process lifetime; the dynamism lives inside the `ResolvesClientCert` / `ServerCertVerifier` it was built with. So a one-time `Arc<rustls::ClientConfig>` hand-off is sufficient: there's no need for a "reloadable" hook on the `temporalio_client` side, just pass-through of the config as-is.

brucearctor:
## Summary

Adds an optional `server_cert_verifier` field to `TlsOptions` that accepts an `Arc<dyn ServerCertVerifier>`. When set, the client uses tonic's `tls_config_with_verifier()` to replace the default WebPKI certificate verification.

Closes #1248

## Motivation

The current `TlsOptions` API only supports static PEM-encoded certificates, which prevents advanced TLS patterns like:

- **Certificate pinning** — verifying a specific certificate fingerprint
- **Custom trust-domain validation** — e.g., SAN-URI extraction for SPIFFE/mesh environments
- **Federated root certificate stores** — composing multiple trust roots from different sources

## What this PR does

1. **Adds `server_cert_verifier: Option<Arc<dyn ServerCertVerifier>>`** to `TlsOptions`
2. **Updates `add_tls_to_channel`** to call `Endpoint::tls_config_with_verifier()` when the custom verifier is set
3. **Re-exports `ServerCertVerifier`** from the crate root for user convenience
4. **Handles `Debug` impl** — `dyn ServerCertVerifier` doesn't implement `Debug`, so `TlsOptions` gets a manual `Debug` impl

## What this does NOT cover

- **`ResolvesClientCert` / dynamic client cert rotation** — tonic 0.14 does not support injecting a pre-built `Arc<rustls::ClientConfig>`, so `ResolvesClientCert` cannot be passed through. This would require either:
  - An upstream change to tonic to accept `Arc<rustls::ClientConfig>` directly
  - Bypassing tonic's TLS config entirely to build the hyper/rustls connector manually

I scoped this PR to what tonic 0.14 supports today. If the maintainers want the full `Arc<ClientConfig>` passthrough, I'm happy to pursue the tonic upstream approach as a follow-up.

## Design decisions

- **Optional field vs enum:** Chose an optional field on the existing struct over an enum wrapper to avoid breaking the public API. An enum-based `ClientTlsConfig` could be a follow-up if preferred.
- **Feature-gating:** Not feature-gated since `tokio-rustls` is already a transitive dependency via tonic. The `ServerCertVerifier` trait is always available when tonic's TLS features are enabled.
- **`server_root_ca_cert` ignored when verifier is set:** This matches tonic's behavior — `tls_config_with_verifier` errors if CA certs are also configured.

## Example usage

```rust
use std::sync::Arc;
use temporalio_client::{TlsOptions, ServerCertVerifier};

// Implement your custom verifier
struct MyCertPinner { /* ... */ }
impl ServerCertVerifier for MyCertPinner { /* ... */ }

let tls = TlsOptions {
    server_cert_verifier: Some(Arc::new(MyCertPinner { /* ... */ })),
    domain: Some("my-temporal.example.com".to_string()),
    ..Default::default()
};
```

## Testing

- `cargo check` — full workspace builds clean
- `cargo fmt --all` — no formatting changes
- `cargo lint` — no warnings
- `cargo test -p temporalio-client` — 112 tests + 1 doctest pass


crates/client/src/lib.rs
```diff
@@ -410,11 +413,15 @@ async fn add_tls_to_channel(
...
         let mut tls = tonic::transport::ClientTlsConfig::new();
 
-        if let Some(root_cert) = &tls_cfg.server_root_ca_cert {
-            let server_root_ca_cert = Certificate::from_pem(root_cert);
-            tls = tls.ca_certificate(server_root_ca_cert);
-        } else {
-            tls = tls.with_native_roots();
+        // When a custom verifier is set, server_root_ca_cert is ignored because
+        // tonic's tls_config_with_verifier replaces the default WebPKI verifier entirely.
+        if tls_cfg.server_cert_verifier.is_none() {
```
Sushisource: We should probably just have these options be mutually exclusive (in the sense that this can just produce an error) rather than silently ignoring.

crates/client/src/lib.rs
```diff
@@ -1485,6 +1498,155 @@ mod tests {
...
+
+        #[test]
+        fn tls_options_debug_shows_none_when_no_verifier() {
+            let opts = TlsOptions::default();
+            let debug_str = format!("{opts:?}");
+            assert!(
+                debug_str.contains("server_cert_verifier: None"),
+                "Debug output should show None for verifier, got: {debug_str}"
+            );
+        }
```
Sushisource: I think these can all just be dropped, very obvious tests.

crates/client/src/options_structs.rs
```diff
@@ -143,6 +144,38 @@ pub struct TlsOptions {
     pub domain: Option<String>,
     /// TLS info for the client. If specified, core will attempt to use mTLS.
     pub client_tls_options: Option<ClientTlsOptions>,
+    /// Optional custom server certificate verifier. When set, this replaces the default
+    /// certificate verification and `server_root_ca_cert` is ignored.
+    ///
+    /// This is useful for:
+    /// - Certificate pinning
+    /// - Custom trust-domain validation (e.g., SAN-URI extraction)
+    /// - Federated root certificate stores
```
Sushisource, line 153: I think we need to include some warning here about the fact that this allows you to make dangerously insecure connections

crates/client/src/lib.rs
```diff
@@ -40,6 +40,9 @@ pub use metrics::{LONG_REQUEST_LATENCY_HISTOGRAM_NAME, REQUEST_LATENCY_HISTOGRAM
 pub use options_structs::*;
 pub use replaceable::SharedReplaceableClient;
 pub use retry::RetryOptions;
+/// Re-export the `ServerCertVerifier` trait so that users can implement custom TLS
+/// server certificate verification without depending on `tokio-rustls` directly.
+pub use tokio_rustls::rustls::client::danger::ServerCertVerifier;
```
Sushisource: I'm not sure I want to re-export this specifically because this import path is communicating some important information with the `danger` segment. I think a little friction here isn't a bad thing.

crates/client/src/lib.rs
```diff
@@ -40,6 +40,13 @@ pub use metrics::{LONG_REQUEST_LATENCY_HISTOGRAM_NAME, REQUEST_LATENCY_HISTOGRAM
 pub use options_structs::*;
 pub use replaceable::SharedReplaceableClient;
 pub use retry::RetryOptions;
+/// Danger-related TLS modules.
```
Sushisource:
```suggestion
/// Potentially dangerous TLS related functionality.
```

Sushisource (approved):

Sushisource:
@brucearctor Thanks again! Looks like there's one small rustfmt needed.

By the way, you've been making a lot of useful contributions and we really appreciate that. We have a program for highlighting top contributors, we can ship you some swag or something too. I didn't immediately see you in our community Slack. If you join, you can DM me and we can set it up! https://t.mp/slack

Alternatively feel free to find me on LinkedIn and send me a message there too. Thanks!

PR #1272, opened 2026-05-14
Enforce minimum DNS resolution interval

### LINKED ISSUE #1266: [Bug] DNS load balancer can be configured to run resolution interval at 0
### Describe the bug
Basically the title, it allows for basically unrestricted re-resolution of DNS

brucearctor:
## What
Adds validation to reject `dns_load_balancing.resolution_interval` values below 1 second.

## Why
Without a floor, setting `resolution_interval` to `Duration::ZERO` causes `spawn_dns_reresolution()` to run a tight loop — `tokio::time::sleep(Duration::ZERO)` yields immediately, hammering DNS with unrestricted re-resolution.

## How
- Added a `MIN_DNS_RESOLUTION_INTERVAL` (1s) constant and a check in `validate_and_get_dns_lb()`, returning `ClientConnectError::InvalidConfig` for sub-minimum values. This follows the existing pattern for conflicting option validation in the same function.
- Added two unit tests: one for `Duration::ZERO`, one for a sub-minimum value (500ms).

Fixes #1266

### DIFF
--- a/crates/client/src/dns.rs
+++ b/crates/client/src/dns.rs
@@ -11,6 +11,8 @@ use tokio::sync::mpsc;
 use tonic::transport::{Channel, Endpoint, channel::Change};
 use url::Url;
 
+const MIN_DNS_RESOLUTION_INTERVAL: Duration = Duration::from_secs(1);
+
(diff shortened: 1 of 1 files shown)


Sushisource (approved):
Thanks!

PR #1274, opened 2026-05-15
Make TLS backend feature-selectable in sdk-core and client

GregBowyer:
The hard-coded `tonic = { features = ["tls-ring", ...] }` in sdk-core and client forces ring on every consumer. Some users ban ring entirely (e.g. those standardised on aws-lc-rs for FIPS or supply-chain reasons) and cannot depend on temporalio-sdk because of this single line.

Add `tls-ring` and `tls-aws-lc` as crate features in both crates. `tls-ring` remains the default so upstream behaviour is unchanged. Downstream users can opt out:

```toml
[dependencies]
temporalio-sdk-core = {
    version = "0.4",
    default-features = false,
    features = ["envconfig", "prometheus", "tls-aws-lc"],
}
```

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -12,7 +12,9 @@ categories = ["development-tools"]
 readme = "README.md"
 
 [features]
-default = ["envconfig"]
+default = ["envconfig", "tls-ring"]
+tls-ring = ["tonic/tls-ring"]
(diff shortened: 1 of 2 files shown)


Sushisource (approved):
Thanks for the contribution!

PR #1277, opened 2026-05-15
Add activity execution interceptors

jeremy-cook:
## What was changed

Introduce an ActivityInterceptor API for wrapping Rust SDK activity execution. Interceptors receive ActivityExecutionInput, can inspect the ActivityContext and decoded typed input, call ActivityExecutionNext::run, and then inspect the typed activity output or execution error before completion is reported back to Core.

This adds:

- ActivityInterceptor, ActivityExecutionInput, ActivityExecutionNext, ActivityExecutionOutput, and ActivityExecutionValue
- ActivityInterceptorWithNext for composing interceptor chains
- Worker::set_activity_interceptor for installing an activity interceptor on a worker
- an SDK example showing typed input/output logging from an interceptor

## Why?

The existing activity hooks were too narrow for SDK-level middleware because they could not wrap the actual activity call or observe both decoded inputs and typed outputs around execution.

This change moves activity result conversion to after interceptor execution, so normal completions, failures, cancellations, WillCompleteAsync, panics, local activities, and remote activities all flow through the same execution wrapper. That gives interceptors a stable boundary for logging, metrics, tracing, policy enforcement, and other middleware without requiring activity call sites to change.

## Testing

Added integration coverage for:

- chained activity interceptor ordering
- remote activity input/output inspection
- local activity interception
- activity failure observation
- activity panic observation

Also added a runnable activity_interceptor SDK example.

### DIFF
--- a/crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
+++ b/crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
@@ -5,6 +5,7 @@ use crate::common::{
 };
 use anyhow::anyhow;
 use assert_matches::assert_matches;
+use futures_util::{FutureExt, future::BoxFuture};
 use std::{
     sync::{
         Arc, Mutex,
@@ -49,8 +50,8 @@ use temporalio_sdk::{
     WorkflowContext, WorkflowResult,
     activities::{ActivityContext, ActivityError},
     interceptors::{
-        ActivityExecutionInput, ActivityExecutionNext, ActivityInterceptor,
-        ActivityInterceptorWithNext,
+        ActivityInboundInterceptor, ActivityInboundInterceptorNext,
+        ActivityInboundInterceptorWithNext, ExecuteActivityInput, ExecuteActivityOutput,
     },
 };
 use temporalio_sdk_core::{
@@ -120,61 +121,66 @@ struct ActivityInterceptorRecords {
     records: Mutex<Vec<ActivityInterceptorRecord>>,
 }
 
-struct RecordingActivityInterceptor {
+struct RecordingActivityInboundInterceptor {
     interceptor: &'static str,
     records: Arc<ActivityInterceptorRecords>,
 }
 
-#[async_trait::async_trait]
-impl ActivityInterceptor for RecordingActivityInterceptor {
-    async fn execute_activity(
-        &self,
-        input: ActivityExecutionInput,
-        next: ActivityExecutionNext<'_>,
-    ) -> temporalio_sdk::interceptors::ActivityExecutionOutput {
-        let info = input.context().info();
-        let activity_type = info.activity_type.clone();
-        let workflow_type = info.workflow_type.clone();
-        let is_local = info.is_local;
-        self.records
-            .records
-            .lock()
-            .unwrap()
-            .push(ActivityInterceptorRecord {
-                interceptor: self.interceptor,
-                phase: ActivityInterceptorPhase::Before,
-                activity_type: activity_type.clone(),
-                workflow_type: workflow_type.clone(),
-                is_local,
-                input: input.downcast_ref::<String>().cloned(),
-                output: None,
-                status: None,
-            });
-        let result = next.run(input).await;
-        self.records
-            .records
-            .lock()
-            .unwrap()
-            .push(ActivityInterceptorRecord {
-                interceptor: self.interceptor,
-                phase: ActivityInterceptorPhase::After,
-                activity_type,
-                workflow_type,
-                is_local,
-                input: None,
-                output: result
-                    .as_ref()
-                    .ok()
-                    .and_then(|output| output.downcast_ref::<String>())
-                    .cloned(),
-                status: Some(activity_execution_output_status(&result)),
-            });
-        result
+impl ActivityInboundInterceptor for RecordingActivityInboundInterceptor {
+    fn execute_activity<'a, 'b>(
+        &'a self,
+        input: ExecuteActivityInput,
+        next: ActivityInboundInterceptorNext<'b>,
+    ) -> BoxFuture<'a, ExecuteActivityOutput>
+    where
+        'b: 'a,
+    {
+        async move {
+            let info = input.activity_info();
+            let activity_type = info.activity_type.clone();
+            let workflow_type = info.workflow_type.clone();
+            let is_local = info.is_local;
+            self.records
+                .records
+                .lock()
+                .unwrap()
+                .push(ActivityInterceptorRecord {
+                    interceptor: self.interceptor,
+                    phase: ActivityInterceptorPhase::Before,
+                    activity_type: activity_type.clone(),
+                    workflow_type: workflow_type.clone(),
+                    is_local,
+                    input: input.args_ref::<String>().cloned(),
+                    output: None,
+                    status: None,
+                });
+            let result = next.run(input).await;
+            self.records
+                .records
+                .lock()
+                .unwrap()
+                .push(ActivityInterceptorRecord {
+                    interceptor: self.interceptor,
+                    phase: ActivityInterceptorPhase::After,
+                    activity_type,
+                    workflow_type,
+                    is_local,
+                    input: None,
+                    output: result
+                        .as_ref()
+                        .ok()
+                        .and_then(|output| output.downcast_ref::<String>())
+                        .cloned(),
+                    status: Some(activity_execution_output_status(&result)),
+                });
+            result
+        }
+        .boxed()
     }
 }
 
 fn activity_execution_output_status(
-    result: &temporalio_sdk::interceptors::ActivityExecutionOutput,
+    result: &temporalio_sdk::interceptors::ExecuteActivityOutput,
 ) -> &'static str {
     match result {
         Ok(_) => "completed",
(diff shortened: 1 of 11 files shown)


crates/sdk/src/interceptors.rs
```diff
@@ -31,6 +66,102 @@ pub trait WorkerInterceptor {
...
+    /// Run the next interceptor or the activity implementation.
+    pub async fn run(self, input: ActivityExecutionInput) -> ActivityExecutionOutput {
+        (self.run)(input).await
+    }
+}
+
+/// Implementors can intercept activity execution.
+///
+/// Advanced usage only.
+#[async_trait::async_trait]
```
Sushisource: I'd prefer to avoid async_trait in our public interfaces, so we don't force it upon dependents. I know this is not universally true at the moment (ex: `WorkerInterceptor`, but that will probably go away once all other interceptors are properly implemented), but I'd like it to be for all new interfaces.

crates/sdk/src/lib.rs
```diff
@@ -752,6 +757,11 @@ impl Worker {
         self.common.worker_interceptor = Some(Box::new(interceptor));
     }
 
+    /// Set an [ActivityInterceptor].
```
Sushisource: Worth noting this replaces any existing interceptor.

crates/sdk/src/activities.rs
```diff
@@ -420,6 +433,82 @@ impl ActivityDefinitions {
...
+    AD: ActivityDefinition + ExecutableActivity,
+    AD::Input: Send + Sync,
+    AD::Output: Send + Sync,
+{
+    ActivityExecutionNext::new(move |input| {
+        catch_activity_panic(async move {
+            let (activity_context, input) = input.into_parts();
+            let input = *input
+                .downcast::<AD::Input>()
+                .expect("activity interceptor returned input with wrong concrete type");
```
Sushisource: This could happen outside of the panic catcher and simply return a future containing the appropriate error directly.

crates/sdk/src/activities.rs
```diff
@@ -420,6 +433,82 @@ impl ActivityDefinitions {
...
+            AD::execute(Some(instance), activity_context, input)
+                .await
+                .map(|output| Box::new(output) as Box<dyn ActivityExecutionValue>)
+        })
+        .boxed()
+    })
+}
+
+pub(crate) fn activity_execution_output_to_core(
```
Sushisource: I think I'd just inline this at the callsite and make `activity_error_to_core_result` `pub(crate)` instead, since this fn is only used in one place.

crates/sdk/src/activities.rs
```diff
@@ -420,6 +433,82 @@ impl ActivityDefinitions {
...
+async fn catch_activity_panic(
+    future: impl Future<Output = ActivityExecutionOutput>,
+) -> ActivityExecutionOutput {
+    match AssertUnwindSafe(future).catch_unwind().await {
+        Ok(output) => output,
+        Err(panic) => Err(activity_panic_to_execution_error(panic)),
+    }
+}
+
+fn activity_panic_to_execution_error(panic: Box<dyn std::any::Any + Send>) -> ActivityError {
```
Sushisource: This can just be inline too.

crates/sdk/src/interceptors.rs
```diff
@@ -31,6 +66,102 @@ pub trait WorkerInterceptor {
...
 }
 
+/// Activity execution data passed to [`ActivityInterceptor::execute_activity`].
+#[non_exhaustive]
+pub struct ActivityExecutionInput {
+    context: ActivityContext,
+    input: Box<dyn Any + Send + Sync>,
+}
+
+impl ActivityExecutionInput {
```
Sushisource: As it stands it looks like there's no way for the user to alter input with this API. That's one of the primary usecases for interceptors and we'd definitely want to support that.

crates/sdk/src/interceptors.rs
```diff
@@ -31,6 +66,102 @@ pub trait WorkerInterceptor {
...
+    pub(crate) fn new(context: ActivityContext, input: Box<dyn Any + Send + Sync>) -> Self {
+        Self { context, input }
+    }
+
+    pub(crate) fn into_parts(self) -> (ActivityContext, Box<dyn Any + Send + Sync>) {
+        (self.context, self.input)
+    }
+
+    /// Context for the activity execution.
+    pub fn context(&self) -> &ActivityContext {
```
Sushisource: I'm not sure we want to expose the _entire_ context here, other SDK languages are pretty much just exposing headers, args, and some info about the activity definition. EX: https://python.temporal.io/temporalio.worker.ActivityInboundInterceptor.html and https://dotnet.temporal.io/api/Temporalio.Worker.Interceptors.ActivityInboundInterceptor.html

crates/sdk/src/interceptors.rs
Sushisource: Additionally this should probably be called `ActivityInboundInterceptor` to match other SDKs

Sushisource (commented):
Thanks so much for the submission. A few minor suggestions, and then on the more substantial side, I think we want to more closely match existing SDKs in terms of naming and parity. A couple of my comments call out the specifics there.

I'm fine with this PR just being inbound interceptors, and outbound may be added later.

I'm also open to debating if the entire ActivityContext is actually reasonable to expose, since that's a much safer thing to do in Rust than in those other languages, so maybe we're fine with that. Would like @chris-olszewski to chime in.

crates/sdk/src/activities.rs
```diff
@@ -185,6 +197,10 @@ impl ActivityContext {
     pub fn headers(&self) -> &HashMap<String, Payload> {
         &self.header_fields
     }
+
+    pub(crate) fn headers_mut(&mut self) -> &mut HashMap<String, Payload> {
```
jeremy-cook, line 201: Not so sure about this one. Should anything on the context be modifiable?

crates/sdk/src/interceptors.rs
jeremy-cook: Agreed

crates/sdk/src/interceptors.rs
jeremy-cook: Definitely, admittedly I didn't put a ton of thought into that. My experience with rust is new. My background is mostly in go and my brain just glanced over the whole context thing and was like yup thats a pattern.

Adjusted. let me know your thoughts on the mutable headers.

crates/sdk/src/interceptors.rs
jeremy-cook: Doh! yup definitely lost the plot a bit on my tinkering. Adjusted to match the other SDK patterns.

crates/sdk/src/interceptors.rs
chris-olszewski: I'm in agreement that we should avoid exposing the entirety of `ActivityContext` and that `ActivityInfo` is more appropriate. Other langs have interceptors [reach out and grab](https://github.com/temporalio/sdk-python/blob/75dd80a4e4389004bc2a15152f93f9b75242ab97/temporalio/contrib/opentelemetry/_otel_interceptor.py#L338) the `ActivityInfo`, but that doesn't feel very "Rusty".

Mutable headers make sense.

crates/sdk/src/activities.rs
chris-olszewski: Not generally, but interceptors should be able to modify headers so this usecase makes sense. We could keep a separate copy of headers for modification and then reconstruct the activity context before execution, but that doesn't provide any real benefit.

crates/sdk/src/lib.rs
```diff
@@ -1376,7 +1370,7 @@ mod tests {
     #[activities]
     impl MyActivities {
         #[activity]
-        async fn my_activity(_ctx: ActivityContext) -> Result<(), ActivityError> {
+        async fn my_activity(_ctx: ActivityContext) -> Result<(), activities::ActivityError> {
```
chris-olszewski: Minor, but would rather add `use crate::activities::ActivityError` at the test module level instead of qualifying the identifier.

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -146,6 +259,386 @@ async fn one_activity_only() {
...
+    let records = Arc::new(ActivityInterceptorRecords::default());
+    let mut interceptor =
+        ActivityInboundInterceptorWithNext::new(Box::new(RecordingActivityInboundInterceptor {
+            interceptor: "outer",
+            records: records.clone(),
+        }));
+    interceptor.set_next(Box::new(RecordingActivityInboundInterceptor {
+        interceptor: "inner",
+        records: records.clone(),
+    }));
```
chris-olszewski: Having users explicitly construct the interceptor chain in this manner is really awkward and somewhat error prone. I think we should match existing SDKs and just take a `Vec` of interceptors and handle the chaining internally

chris-olszewski (changes requested):
Thanks for the PR!

I have been working on Workflow interceptors and the general pattern I came up seems to be compatible with your activity interceptor changes. I cherry picked some parts of my WIP branch and applied them to your PR here:  https://github.com/temporalio/sdk-rust/commit/a68a74cf6d9b59819c28f5211754b36270965688 (Apologies this was done as a commit, but trying to write these changes via GitHub UI is too painful)

Gist of the changes is moving away from users explicitly constructing a chain of interceptors and instead have them append interceptors to a list. I know the explicit next chaining matches our existing worker interceptor, but I think it is an awkward API that I plan on changing before 1.0. I also moved to a generic `Next<Input, Output>` over a special cased `ActivityInboundInterceptorNext`.

What do you think?

jeremy-cook:
Thanks for the feedback. 

> (Apologies this was done as a commit, but trying to write these changes via GitHub UI is too painful)

zero concerns, I get it. Also you can do whatever you want with this work. If it's easier to take it and change it how you want Im fine with that too. Ultimately Im just looking at helping in whatever way is useful so we can use rust with temporal  :) 

The list makes sense to me, that's a common pattern all over. I just copied what was currently implemented, I changed it as suggested. Thanks for the explicit examples!!

chris-olszewski (approved):
Thank you for tackling this!

PR #1279, opened 2026-05-18
refactor(examples): extract saga compensation into a reusable Saga pattern

xumaple:
Replace the hand-rolled `Booking` enum + per-variant match in the saga example with a generic `Saga<W>` struct that pairs each forward step with its compensation activity at the call site and runs compensations in reverse on failure.

### DIFF
--- a/crates/sdk/examples/saga/workflows.rs
+++ b/crates/sdk/examples/saga/workflows.rs
@@ -1,17 +1,12 @@
 #![allow(unreachable_pub)]
+use futures_util::future::LocalBoxFuture;
 use std::time::Duration;
+use temporalio_common::ActivityDefinition;
 use temporalio_macros::{activities, workflow, workflow_methods};
 use temporalio_sdk::{
     ActivityExecutionError, ActivityOptions, ApplicationFailure, WorkflowContext, WorkflowResult,
     activities::{ActivityContext, ActivityError},
 };
-
-#[derive(Debug)]
-enum Booking {
-    Hotel,
-    Flight,
-    Car,
-}
 
 #[workflow]
 #[derive(Default)]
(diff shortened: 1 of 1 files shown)


Sushisource (approved):
Cool, this looks good

crates/sdk/examples/saga/workflows.rs
```diff
@@ -24,87 +19,92 @@ impl SagaWorkflow {
...
             .await?;
-        compensations.push((Booking::Car, car));
+        Ok(vec![hotel, flight, car])
+    }
+}
 
-        Ok(())
+/// Records compensations and runs them in reverse on failure.
+struct Saga<W> {
+    ctx: WorkflowContext<W>,
```
chris-olszewski: Cool to show off that we can make this generic over workflows, but for the example I think it adds noise.
```suggestion
struct Saga {
    ctx: WorkflowContext<SagaWorkflow>,
```

crates/sdk/examples/saga/workflows.rs
```diff
@@ -24,87 +19,92 @@ impl SagaWorkflow {
...
+    async fn step<Step, Compensation>(
+        &mut self,
+        forward: Step,
+        input: impl Into<Step::Input>,
+        compensate: Compensation,
+    ) -> Result<Step::Output, ActivityExecutionError>
+    where
+        Step: ActivityDefinition,
+        Step::Output: Clone + Into<Compensation::Input>,
+        Compensation: ActivityDefinition<Output = ()> + 'static,
```
chris-olszewski, line 86: I'm a little concerned that all of the additional noise required to get the boxed futures working correctly distracts from the focus of the example.

chris-olszewski (changes requested):
I like the extraction of `Saga::step` that handles calling the activity with registring the compensation and then `Saga::compensate` to run through those.

I think this might be generic to a fault where it adds noise that distracts from pattern we are trying to highlight.

crates/sdk/examples/saga/workflows.rs
xumaple: removed

crates/sdk/examples/saga/workflows.rs
xumaple: I got rid of it. But it's a litlte fishy for me now. Not sure how you feel. We're calling the compensation activity in the step function and saving the future, and not awaiting it until compensate(). But that means that in the success case we still create all those futures. code/efficiency wise it's completely fine, just looks a bit weird to the eye. Thoughts?

chris-olszewski: I feel like this is slightly easier for me to comprehend, but if you think this makes the example less clear we can keep the closure.

chris-olszewski (approved):

crates/sdk/examples/saga/workflows.rs
xumaple: I'm okay with it.

PR #1293, opened 2026-05-25
deps: sync api_upstream subtree to temporal-api@6c8bd13

mjameswh:
## What was changed

- Sync `api_upstream` to upstream's HEAD `6c8bd13`.


crates/client/src/workflow_handle.rs
```diff
@@ -746,6 +746,9 @@ where
                         name: update.name().to_string(),
                         args: Some(Payloads { payloads }),
                     }),
+                    request_id: String::new(),
+                    completion_callbacks: vec![],
+                    links: vec![],
```
chris-olszewski: Would these values be better represented as defaults? Do we need a TODO to populate these values?
```suggestion
                    ..Default::default()
```

chris-olszewski (approved):
One small q around the new update request fields

crates/client/src/workflow_handle.rs
mjameswh: Fixed

PR #1300, opened 2026-05-28
Fix start-to-close timeout failure caused by short heatbeat timeout

GregoryTravis:
When an activity has both startToCloseTimeout and heartbeatTimeout set, and heartbeatTimeout < startToCloseTimeout, the Core SDK's local timeout timer never fires because successful heartbeats keep resetting it. The activity runs indefinitely on the worker even after the server has timed it out.

The previous implementation simply ignored the start_to_close timeout when the heartbeat timeout was shorter, which would work if the heartbeat timeout was not reset after each heartbeat ack. Instead, the heartbeat acks kept the activity alive long after the server had hit the start-to-close timeout. The workflow correctly resolved with a timeout, but the activity kept running and occupying a worker task slot.

The fix is to run both timeouts together -- the one-time start_to_close timeout and the infinitely resettable heartbeat timeout.

Tests: 3 unit, 1 integration.

Fixes #1188


crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
     assert!(was_cancelled.load(Ordering::Relaxed));
 }
 
+// Regression test for SDK-5113 / sdk-core#1188:
```
chris-olszewski:
```suggestion
// Regression test for https://github.com/temporalio/sdk-rust/issues/1188
```

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
...
+//   When start_to_close_timeout > heartbeat_timeout, continuous successful
+//   heartbeats must NOT prevent start_to_close from firing locally. With the
+//   buggy implementation, only the smaller of the two timeouts (heartbeat)
+//   has a local timer and heartbeats keep resetting it — start_to_close is
+//   never enforced locally and the activity runs indefinitely on the worker
+//   even after the server has timed it out.
```
chris-olszewski: I don't think it is useful to document previous buggy behavior. More appropriate for PR description.
```suggestion
//   heartbeats must NOT prevent start_to_close from firing locally.
```

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
...
+//
+// This is the SDK-level complement to
+// `worker::activities::tests::start_to_close_fires_when_heartbeat_timeout_shorter`
+// (a Core unit test). This one exercises the bug end-to-end through a real
+// workflow/activity against the server, observing the cancel signal at the
+// activity-function level — matching the Python repro on the ticket.
+//
```
chris-olszewski: Not particularly useful comment IMO
```suggestion
```

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
...
+// `heartbeat_timeout * 0.8` regardless of how often the activity calls
+// `record_heartbeat`. So we pick `heartbeat_timeout` just under
+// `start_to_close` to push the post-timeout heartbeat RPC well after
+// `start_to_close`:
+//   heartbeat_timeout = 4.5s  → throttle = 3.6s
+//   start_to_close     = 5.0s → server times out the activity here
+//   next heartbeat RPC firing after t=5s: at t=7.2s (NotFound, only after fix)
+// With the fix, the local start_to_close timer fires at t≈5s.
+// Without the fix, the cancel doesn't arrive until t≈7.2s (NotFound path).
+// The 6.5s assertion threshold cleanly distinguishes them.
```
chris-olszewski: This timing note seems extremely suspect. 6.5s isn't that far from 5s or 7.2s in the realm of GitHub Actions

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
...
+                    _ = ctx.cancelled() => {
+                        let elapsed_ms = started.elapsed().as_millis() as u64;
+                        // Floor at 1 so a (rare) sub-millisecond observation
+                        // isn't mistaken for "never observed".
+                        self.cancel_elapsed_ms
+                            .store(elapsed_ms.max(1), Ordering::Relaxed);
+                        return Err(ActivityError::cancelled());
+                    }
+                    _ = tokio::time::sleep(Duration::from_millis(500)) => {}
+                }
```
chris-olszewski: This might be better expressed with [tokio::time::timeout](https://github.com/temporalio/sdk-rust/issues/1188)

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
...
+#[tokio::test]
+async fn start_to_close_local_timeout_fires_with_heartbeats() {
+    let wf_name = "start_to_close_local_timeout_fires_with_heartbeats";
+    let mut starter = CoreWfStarter::new(wf_name);
+    starter
+        .set_core_cfg_mutator(|m| m.local_timeout_buffer_for_activities = Duration::from_secs(0));
+
+    // 0 means "the activity never observed a cancel". Otherwise, milliseconds
+    // from activity start to the moment cancel was observed.
+    let cancel_elapsed_ms: Arc<AtomicU64> = Arc::new(AtomicU64::new(0));
```
chris-olszewski: Note to self, why do we need to know the time elapsed since cancellation?

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
...
+                panic!("expected timeout cause, got {fail_err:?}");
+            };
+            assert_eq!(timeout.timeout_type(), TimeoutType::StartToClose);
+            // Hold the worker alive past the next throttled heartbeat RPC
+            // (~t=7.2s, which carries NotFound and would cancel the still-
+            // running activity along that alternate path). Without this
+            // trailing timer, run_until_done would shut the worker down at
+            // server-side timeout (~t=5s) and the shutdown cancel would mask
+            // the bug.
+            ctx.timer(Duration::from_secs(5)).await;
```
chris-olszewski: Would running the worker normally (e.g. `run` instead of `run_until_done`) and terminating it explicitly once we have done our assertions be better here than having a hanging timer?

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
...
+
+    starter
+        .sdk_config
+        .register_activities(HeartbeatingStuckActivities {
+            cancel_elapsed_ms: cancel_elapsed_ms.clone(),
+        });
+    let mut worker = starter.worker().await;
+
+    const START_TO_CLOSE: Duration = Duration::from_secs(5);
+    const HEARTBEAT_TIMEOUT: Duration = Duration::from_millis(4500);
```
chris-olszewski: Do these values need to be this large and this close?

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
...
+    let elapsed_ms = cancel_elapsed_ms.load(Ordering::Relaxed);
+    assert_ne!(elapsed_ms, 0, "activity never observed a cancel signal");
+    let elapsed = Duration::from_millis(elapsed_ms);
+    let threshold = START_TO_CLOSE + Duration::from_millis(1500);
+    assert!(
+        elapsed < threshold,
+        "activity ran for {elapsed:?} before observing cancel (threshold {threshold:?}) — \
+         start_to_close was not enforced locally; continuous heartbeats kept resetting the \
+         only timer, leaving the bug visible only via the much-later NotFound \
+         heartbeat-response path (SDK-5113)"
```
chris-olszewski: I don't understand the Jira ticket in this assertion

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
...
+            WorkflowStartOptions::new(task_queue.clone(), task_queue).build(),
+        )
+        .await
+        .unwrap();
+    worker.run_until_done().await.unwrap();
+
+    let elapsed_ms = cancel_elapsed_ms.load(Ordering::Relaxed);
+    assert_ne!(elapsed_ms, 0, "activity never observed a cancel signal");
+    let elapsed = Duration::from_millis(elapsed_ms);
+    let threshold = START_TO_CLOSE + Duration::from_millis(1500);
```
chris-olszewski: This timing assertion feels far too racey to admit into our CI.

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -1941,6 +1941,157 @@ async fn activity_can_be_cancelled_by_local_timeout() {
...
+// This is the SDK-level complement to
+// `worker::activities::tests::start_to_close_fires_when_heartbeat_timeout_shorter`
+// (a Core unit test). This one exercises the bug end-to-end through a real
+// workflow/activity against the server, observing the cancel signal at the
+// activity-function level — matching the Python repro on the ticket.
+//
+// Sensitivity note: there's a *second* path that can cancel a "stuck"
+// heartbeating activity — when the server has already timed it out, the
+// next heartbeat RPC returns NotFound, and the heartbeat manager (see
+// activity_heartbeat_manager.rs) turns that into a cancel. That path is
```
chris-olszewski: This doesn't seem to line up with the original issue. Was there a server side fix where server now responds with `NotFound` upon an activity timing out?

chris-olszewski (changes requested):
Haven't fully reviewed, but pretty concerned about the integration test being flakey.

crates/sdk-core/src/worker/activities.rs
```diff
@@ -1049,4 +1103,205 @@ mod tests {
...
         atm.shutdown().await;
     }
+
+    // Regression test for SDK-5113 / sdk-core#1188:
+    //   When start_to_close_timeout > heartbeat_timeout, continuous successful
+    //   heartbeats must NOT prevent start_to_close_timeout from firing. The
+    //   current implementation creates a single timer for the *minimum* of the
+    //   two timeouts (the heartbeat one here) and every heartbeat resets it,
+    //   so start_to_close is never enforced locally and the activity runs
+    //   indefinitely on the worker after the server has timed it out.
```
chris-olszewski: Nit, avoid referencing Jira tickets in public code.
```suggestion
    //  Regression test for https://github.com/temporalio/sdk-rust/issues/1188:
    //   When start_to_close_timeout > heartbeat_timeout, continuous successful
    //   heartbeats must NOT prevent start_to_close_timeout from firing.
```

crates/sdk-core/src/worker/activities.rs
```diff
@@ -1049,4 +1103,205 @@ mod tests {
...
+        };
+        // start_to_close_timeout=300ms. Allow generous slack (2s) before we
+        // declare the bug present, to keep this stable on slow CI.
+        let poller = async {
+            tokio::time::timeout(Duration::from_secs(2), atm.poll())
+                .await
+                .expect(
+                    "start_to_close_timeout was not enforced locally — \
+                     continuous heartbeats kept resetting the only timer \
+                     (SDK-5113)",
```
chris-olszewski:
```suggestion
                    "start_to_close_timeout to be enforced locally",
```

crates/sdk-core/src/worker/activities.rs
```diff
@@ -1049,4 +1103,205 @@ mod tests {
...
+                )
+                .unwrap()
+        };
+        let should_timeout = tokio::select! {
+            v = poller => v,
+            _ = heartbeater => unreachable!("heartbeat loop never exits on its own"),
+        };
+        assert!(should_timeout.is_timeout());
```
chris-olszewski: Slightly confusing var name for the type
```suggestion
        let activity_task = tokio::select! {
            v = poller => v,
            _ = heartbeater => unreachable!("heartbeat loop never exits on its own"),
        };
        assert!(activity_task.is_timeout());
```

crates/sdk-core/src/worker/activities.rs
```diff
@@ -624,43 +624,29 @@ where
...
-                                            } else {
-                                                tokio::time::sleep(sleep_time).await;
-                                            }
+                                            let timeout_type = timers.run().await;
                                             debug!(
                                                 task_token=%task_token,
-                                                "Timing out activity due to elapsed local \
-                                                 {timeout_type} timer"
+                                                "Timing out activity due to elapsed local {} timer",
+                                                timeout_type.as_str()
```
chris-olszewski:
```suggestion
                                                "Timing out activity due to elapsed local {timeout_type} timer",
```

crates/sdk-core/src/worker/activities.rs
```diff
@@ -776,18 +762,136 @@ fn worker_shutdown_failure() -> Failure {
...
+}
+
+impl ActivityLocalTimeoutKind {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::Heartbeat => "heartbeat",
+            Self::StartToClose => "start_to_close",
+        }
+    }
+}
```
chris-olszewski: Implementing the `Display` trait is more standard here
```suggestion
impl std::fmt::Display for ActivityLocalTimeoutKind {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.write_str(match self {
            Self::Heartbeat => "heartbeat",
            Self::StartToClose => "start_to_close",
        })
    }
}
```

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
GregoryTravis: After discussion, removed this test -- there's no easy way to verify the reason for the timeout and wall time is flaky.

crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
GregoryTravis: The original bug report states that it wasn't returning NotFound but I have not confirmed this. The fix is appropriate and unit-testable either way so not essential to verify.

crates/sdk-core/src/worker/activities.rs
```diff
@@ -624,43 +624,29 @@ where
...
-                                .filter_map(|(k, d)| {
-                                    d.and_then(|d| Duration::try_from(d).ok().map(|d| (k, d)))
-                                })
-                                .filter(|(_, d)| !d.is_zero())
-                                .min_by(|(_, d1), (_, d2)| d1.cmp(d2));
-                                if let Some((timeout_type, timeout_at)) = timeout_at {
-                                    let sleep_time = timeout_at + local_timeout_buffer;
+                                // There is a bug in the server that translates non-set
+                                // heartbeat timeouts into 0 duration. We treat 0 the same way
+                                // as None so we don't spawn a timer that would fire immediately.
```
chris-olszewski: I think this comment is misleading. I think we mean to say that we deserialize unset timeouts as zero so we must treat zero as unset.

crates/sdk-core/src/worker/activities.rs
```diff
@@ -624,43 +624,29 @@ where
...
+                                let to_sleep =
+                                    |d: Option<prost_types::Duration>| -> Option<Duration> {
+                                        d.and_then(|d| Duration::try_from(d).ok())
+                                            .filter(|d| !d.is_zero())
+                                            .map(|d| d + local_timeout_buffer)
+                                    };
```
chris-olszewski: I wouldn't oppose moving this logic into the `ActivityLocalTimers` constructor by passing in `local_timeout_buffer` as an arg. What do you think?

crates/sdk-core/src/worker/activities.rs
```diff
@@ -776,18 +762,160 @@ fn worker_shutdown_failure() -> Failure {
...
+            if let Some((sleep_time, rs)) = self.heartbeat {
+                loop {
+                    tokio::select! {
+                        _ = rs.notified() => continue,
+                        _ = tokio::time::sleep(sleep_time)
+                            => return ActivityLocalTimeoutKind::Heartbeat,
+                    }
+                }
+            }
+            std::future::pending().await
```
chris-olszewski: minor suggestions:
 - avoid using `return` here just because it makes the control flow a little harder to understand
 - `tokio:select` loop can be written with timeout/while loop if you think that is easier to grok
```suggestion
            if let Some((sleep_time, rs)) = self.heartbeat {
                while tokio::time::timeout(sleep_time, rs.notified())
                    .await
                    .is_ok()
                {}
                ActivityLocalTimeoutKind::Heartbeat
            } else {
                std::future::pending().await
            }
```

chris-olszewski (approved):
LGTM, very minor suggestions

crates/sdk-core/src/worker/activities.rs
GregoryTravis: Removed reference to bug.

crates/sdk-core/src/worker/activities.rs
GregoryTravis: Yes, done

PR #1301, opened 2026-05-28
fix(metrics): emit `workflow_type` attribute on `temporal_workflow_task_execution_failed`

### LINKED ISSUE #1299: [Bug] Metrics - workflow_type missing for temporal_workflow_task_execution_failed
Two separate issues here:

1. workflow_type missing from temporal_workflow_task_execution_failed for all failure reasons

https://github.com/temporalio/sdk-rust/blob/main/crates/sdk-core/src/telemetry/metrics.rs#L121

meter in this case does not use "tm" and will not include workflow_type at all

2. Once (1) is fixed i think workflow_type will still not report when failure_reason for this metric is "GrpcMessageTooLarge" added by https://github.com/temporalio/sdk-rust/pull/1069
as the path added for this failure reason 
its not using the per-run context so will not include workflow_type

chris-olszewski:
## What was changed
- Actually forward attributes passed to `MetricsContext::with_new_attributes`
- Switch task failures produced by gRPC message too large to use the run metrics instead of worker metrics

First commit adds failing test. Second commit fixes unit & first integration test. Third commit fixes gRPC message too large integration test.

## Why?
First one was straight regression, other one was oversight in initial implementation of gRPC message too large as workflow task failure.

## Checklist

1. Closes #1299

2. How was this tested:
- Added unit test to verify `MetricsContext::with_new_attributes` doesn't throw away attributes
- Integration test to verify `workflow_type` is present on `temporal_workflow_task_execution_failed`
- Integration test to verify `workflow_type` is present on `temporal_workflow_task_execution_failed` when task failure is triggered by a gRPC too large

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/sdk-core/src/telemetry/metrics.rs
+++ b/crates/sdk-core/src/telemetry/metrics.rs
@@ -118,7 +118,7 @@ impl MetricsContext {
         instruments.update_attributes(tm.get_default_attributes());
         Self {
             instruments: Arc::new(instruments),
-            meter: self.meter.clone(),
+            meter: tm,
             in_memory_metrics: self.in_memory_metrics.clone(),
         }
     }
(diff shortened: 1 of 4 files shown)


Sushisource (approved):
Thanks!

PR #1304, opened 2026-05-29
:boom: Fix possible panic due to LA completing after an eviction

### LINKED ISSUE #1230: [Bug] Flaky panic triggered by test_workflow_cancel_activity in sdk-python
Two dbg_panic! sites fire from sdk-core during a sdk-python test that rapidly cancels local activities under a 1s workflow task timeout.
* thread 'workflow-processing' panicked at crates/sdk-core/src/worker/workflow/managed_run.rs:401:17: Attempted to complete activation for run 019db732-4d78-7616-9a36-20b44eff6c09 without associated workflow task
* thread 'tokio-runtime-worker' panicked at crates/sdk-core/src/worker/workflow/mod.rs:575:13: Send half of activation complete response channel went missing
* Surfaces to Python as pyo3_async_runtimes.RustPanic: rust future panicked: unknown error from complete_workflow_activation, and subsequent QueryWorkflow RPCs return WorkflowNotReadyFailure.

Reproduction

  sdk-python (debug build):

  ```uv run pytest tests/worker/test_workflow.py::test_workflow_cancel_activity -k True -x -s```
                                                                                                                                                                                                                   
  Timing-sensitive; doesn't always reproduce.                                                                                                                                                                      
                                                                                                                                                                                                                   
  CI run: https://github.com/temporalio/sdk-python/actions/runs/24804169223

Sushisource:
## What was changed
This changes local activity eviction semantics so local activities are invalidated when their
  owning workflow run instance is evicted, rather than allowing them to survive and potentially
  be adopted by a later replay of the same RunID.

  The old behavior was timing-dependent: an in-flight LA completion could arrive after eviction
  while the workflow was replaying, but before replay had recreated the corresponding local
  activity machine. In that state, the completion no longer had a valid owner, yet Core could try
  to apply it to the new workflow-machine instance, causing nondeterminism or corrupting WFT/
  activation bookkeeping. If the completion happened to arrive later, after replay recreated the
  LA command, it could appear to work, but that relied on an accidental race rather than a sound
  ownership model.

  This also makes duplicate local activity queueing an invariant violation instead of silently
  suppressing it, and warns when eviction is requested while local activities are still in
  flight, which is especially problematic with max_cached_workflows = 0.

> [!WARNING]
> It's probably important to call this one out in release notes, as it's possible there are people out there semi-depending on the behavior that accidentally worked. This is probably only likely in situations where they have the cache off. They should not have the cache off. This is not actually a breaking change, though.

## Why?
Bugfix

## Checklist

1. Closes https://github.com/temporalio/sdk-rust/issues/1230
2. How was this tested:
Added test

3. Any docs updates needed?

### DIFF
--- a/crates/sdk-core/src/worker/activities/local_activities.rs
+++ b/crates/sdk-core/src/worker/activities/local_activities.rs
@@ -406,9 +406,9 @@ impl LocalActivityManager {
     /// Returns the next pending local-activity related action, or None if shutdown has initiated
     /// and there are no more remaining actions to take.
     pub(crate) async fn next_pending(&self) -> Option<NextPendingLAAction> {
-        let (new_or_retry, permit) = match self.rcvs.lock().await.next().await? {
-            NewOrCancel::Cancel(c) => {
-                return match c {
+        loop {
+            let (new_or_retry, permit) = match self.rcvs.lock().await.next().await? {
+                NewOrCancel::Cancel(c) => match c {
                     CancelOrTimeout::Cancel(c) => {
                         if self
                             .dat
@@ -416,11 +416,10 @@ impl LocalActivityManager {
                             .outstanding_activity_tasks
                             .contains_key(c.task_token.as_slice())
                         {
-                            Some(NextPendingLAAction::Dispatch(c))
-                        } else {
-                            // Don't dispatch cancels for things we've already stopped tracking
-                            None
+                            return Some(NextPendingLAAction::Dispatch(c));
                         }
+                        // Don't dispatch cancels for things we've already stopped tracking
+                        continue;
                     }
                     CancelOrTimeout::Timeout { run_id, resolution } => {
                         let tt = self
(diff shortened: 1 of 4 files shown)


crates/sdk-core/src/worker/workflow/managed_run.rs
```diff
@@ -904,6 +912,15 @@ impl ManagedRun {
         }
 
         if !self.activation_is_eviction() && self.trying_to_evict.is_none() {
+            let outstanding_las = self.wfm.machines.outstanding_local_activity_count();
+            if outstanding_las > 0 && self.config.max_cached_workflows == 0 {
+                warn!(
+                    run_id=%info.run_id,
+                    reason=?info.reason,
+                    outstanding_local_activities=outstanding_las,
+                    "Eviction requested while local activities are still in flight; local activities when using max_cached_workflows=0 are likely to be dropped or retried"
```
mjameswh, line 921: Interesting observation. It seems like Go and Java do not have the bug you're fixing in this PR, but this specific call out (LA vs max_cached_workflows=0) may apply there too.

yuandrew:
> Warning

> It's probably important to call this one out in release notes, as it's possible there are people out there semi-depending on the behavior that accidentally worked. This is probably only likely in situations where they have the cache off. They should not have the cache off.

worth adding a 💥 in the title so lang remembers to call this out?

Sushisource:
> > Warning
> 
> > It's probably important to call this one out in release notes, as it's possible there are people out there semi-depending on the behavior that accidentally worked. This is probably only likely in situations where they have the cache off. They should not have the cache off.
> 
> worth adding a 💥 in the title so lang remembers to call this out?

Done. It's not actually a breaking change, but, it'll call attention to this

crates/sdk-core/src/worker/activities/local_activities.rs
```diff
@@ -447,6 +465,10 @@ impl LocalActivityManager {
         let sa = new_la.schedule_cmd;
 
         let mut dat = self.dat.lock();
+        if !dat.la_info.contains_key(&id) {
+            debug!(id=?id, "Dropping invalidated local activity request");
+            return None;
+        }
```
mjameswh: I'm not sure returning `None` is correct here.

According to a comment above:

> Returns the next pending local-activity related action, or None if shutdown has initiated and there are no more remaining actions to take.

But here, we haven't confirmed there isn't more remaining actions in the queue, nor that shutdown has been initiated. Can't there be other pending LA actions after the one that has been invalidated?

Sushisource: Good point, let me see

Sushisource: Looks like this didn't _actually_ matter since the caller is also double-checking the shutdown token, but, I've changed it to loop (there were other cases that should've been loops too) and everything is more consistent w/ the docstring now.

mjameswh (approved):

PR #1307, opened 2026-06-02
fix(sdk) 💥 : no longer expose raw proto for application error category

### LINKED ISSUE #1306: ApplicationErrorCategory enum itself is not re-exported from the error module
Re-export ApplicationErrorCategory from error::* alongside ApplicationFailure

chris-olszewski:
## What was changed
Create a new SDK type that converts down to the underlying proto for `ApplicationErrorCategory`

## Why?
SDKs general stance is to avoid exposing raw protos to the user unless needed.

Note this is a breaking change for those who were setting categories on application errors.

## Checklist

1. Closes https://github.com/temporalio/sdk-rust/issues/1306

2. How was this tested:
👀 

3. Any docs updates needed?
No, did a quick check of the Rust SDK guide and we don't have a code sample for setting the error category.

### DIFF
--- a/crates/common-wasm/src/error.rs
+++ b/crates/common-wasm/src/error.rs
@@ -122,6 +122,7 @@ where
 
 /// Categorizes an [`ApplicationFailure`] to hint at how the server and tooling should treat it.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
+#[non_exhaustive]
 pub enum ApplicationErrorCategory {
     /// No category was specified.
     #[default]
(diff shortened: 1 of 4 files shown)


crates/common-wasm/src/error.rs
```diff
@@ -120,6 +120,35 @@ where
     }
 }
 
+/// Categorizes an [`ApplicationFailure`] to hint at how the server and tooling should treat it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
```
Sushisource, line 124: needs `non_exhaustive` I think

Sushisource (approved):
Just one thing

PR #1308, opened 2026-06-02
Add Event Groups in Core SDK

mjameswh:
## What was changed

Introduces support for Event Groups in Core SDK.

Surfacing that API properly in Rust SDK will be done in a distinct PR.

More specifically, this PR introduces the following two critical changes:
- Copy of the `event_group_markers` field from lang's `Command` → server's `Command`.
- Copy of the `event_id` or `update_id` field from server's `HistoryEvent` → lang's `Job`.

## Why?

Event Groups is a new form of metadata that allows regrouping of logically related Workflow Events, according to user-defined or system-inferred criteria, providing users with improved visibility, analysis, and debugging capabilities.

## Before merging

`.cargo/config.toml` pins the integ tests on `v1.7.4-standalone-nexus-operations`, since no CLI
release yet bundles a server that understands the Event Groups fields. Drop the pin once one does.

That pin exposes four failures that have nothing to do with Event Groups. Each is fixed by its own
PR, and CI here stays red until all four land:

- #1495 — `docker_metrics_with_prometheus` asserting on the wrong Prometheus series
- #1496 — the in-Docker environment check losing its offline guarantees
- #1497 — nexus `request-timeout` headers outside the Nexus duration grammar
- #1498 — activity cancellation against Server v1.32.0-158.0

## Related PRs in other repositories

- Protobuf API: https://github.com/temporalio/api/pull/793 (merged)
- Server: https://github.com/temporalio/temporal/pull/10472 (merged)

### DIFF
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -157,6 +157,10 @@ jobs:
 
   integ-tests:
     name: Integ tests
+    # TEMP(event-groups): the default cached-download dev server does not yet support Event Groups,
+    # so pin the CLI to a build that does. Remove once Event Groups support lands in a default release.
+    env:
+      CLI_VERSION_OVERRIDE: v1.7.4-standalone-nexus-operations
     timeout-minutes: ${{ github.ref == 'refs/heads/main' && 30 || 25 }}
     strategy:
       fail-fast: false
--- a/crates/protos/src/protos/mod.rs
+++ b/crates/protos/src/protos/mod.rs
@@ -89,7 +89,7 @@ pub mod coresdk {
             fn from(v: workflow_command::Variant) -> Self {
                 Self {
                     variant: Some(v),
-                    user_metadata: None,
+                    ..Default::default()
                 }
             }
         }
@@ -1307,13 +1307,16 @@ pub mod coresdk {
                 }
             }
 
-            impl From<WorkflowExecutionSignaledEventAttributes> for SignalWorkflow {
-                fn from(a: WorkflowExecutionSignaledEventAttributes) -> Self {
+            impl From<(WorkflowExecutionSignaledEventAttributes, i64)> for SignalWorkflow {
+                fn from(
+                    (a, originating_event_id): (WorkflowExecutionSignaledEventAttributes, i64),
+                ) -> Self {
                     Self {
                         signal_name: a.signal_name,
                         input: Vec::from_payloads(a.input),
                         identity: a.identity,
                         headers: a.header.map(Into::into).unwrap_or_default(),
+                        originating_event_id,
                     }
                 }
             }
@@ -1330,6 +1333,7 @@ pub mod coresdk {
                 workflow_id: String,
                 randomness_seed: u64,
                 start_time: Timestamp,
+                originating_event_id: i64,
             ) -> InitializeWorkflow {
                 InitializeWorkflow {
                     workflow_type: attrs.workflow_type.map(|wt| wt.name).unwrap_or_default(),
@@ -1363,6 +1367,8 @@ pub mod coresdk {
                     start_time: Some(start_time),
                     root_workflow: attrs.root_workflow_execution,
                     priority: attrs.priority,
+                    originating_event_id,
+                    original_execution_run_id: attrs.original_execution_run_id,
                 }
             }
         }
--- a/crates/protos/src/protos/mod.rs
+++ b/crates/protos/src/protos/mod.rs
@@ -89,7 +89,7 @@ pub mod coresdk {
             fn from(v: workflow_command::Variant) -> Self {
                 Self {
                     variant: Some(v),
-                    user_metadata: None,
+                    ..Default::default()
                 }
             }
         }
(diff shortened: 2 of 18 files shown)


crates/protos/src/protos/mod.rs
```diff
@@ -1309,6 +1309,7 @@ pub mod coresdk {
                         input: Vec::from_payloads(a.input),
                         identity: a.identity,
                         headers: a.header.map(Into::into).unwrap_or_default(),
+                        originating_event_id: 0,
```
mjameswh: This will get replaced by the actual value [by the caller](https://github.com/temporalio/sdk-rust/pull/1308/changes#diff-8a2421f3bab33549efe81c581bb2395a64467fb8d8559578e1675a562f8a2799R1028).

I'm not happy with that, but the `event_id` is stored on the _parent_ type of `WorkflowExecutionSignaledEventAttributes`. Could either add a type wrapper around `XxxAttributes`, or use a custom `FromXxx` trait that gets both the `attributes` and it's parent history event, but both approaches would be quite invasive. Not something we should do in this PR.

So I'm settling to this for now, unless anyone has objections or suggestions.

crates/workflow/src/workflow_context/options.rs
```diff
@@ -642,6 +666,10 @@ impl Signal {
             input: self.data.input,
             identity: String::new(),
             headers: self.data.headers,
+            // FIXME: I think that's wrong. This code path is for outbound commands,
+            // which means we should not have an originating event ID; we should instead
+            // have user metadata and event group markers to be attached to the command.
+            originating_event_id: 0,
```
mjameswh: @chris-olszewski Maybe I'm missing something, but it seems to me like `protos::coresdk::workflow_activation::SignalWorkflow` is not the right structure to use for this. Not important for this PR, though.

.github/workflows/per-pr.yml
```diff
@@ -123,6 +123,10 @@ jobs:
 
   integ-tests:
     name: Integ tests
+    env:
+      # TEMP(event-groups): the event groups API requires server >v1.32.0-158.0
+      # for which there's still no official CLI release.
+      CLI_VERSION_OVERRIDE: v1.7.4-standalone-nexus-operations
```
Sushisource: You want to change this in `.cargo/config.toml` instead

crates/sdk-core/src/worker/workflow/machines/workflow_machines.rs
```diff
@@ -1196,7 +1198,10 @@ impl WorkflowMachines {
                         };
                         self.add_cmd_to_wf_task(
                             new_external_cancel(0, we, attrs.child_workflow_only, attrs.reason),
+                            // FIXME: Wire metadata and group markers from lang's cancellation command,
+                            // through the state machine, into the command we issue here.
```
Sushisource, line 1202: Seems like we should probably do this in this PR too?

crates/sdk-core/src/worker/workflow/mod.rs
```diff
@@ -1409,17 +1409,19 @@ struct EmptyWorkflowCommandErr;
 
 /// [DrivenWorkflow]s respond with these when called, to indicate what they want to do next.
 /// EX: Create a new timer, complete the workflow, etc.
-#[derive(Debug, derive_more::From, derive_more::Display)]
+#[derive(Debug, Default, derive_more::From, derive_more::Display)]
```
Sushisource: I think I'd rather add a `new(variant)` method than derive `Default` here.

Sushisource (commented):
Generally looking good to me. The one thing I can't remember is if we said user converters/codecs are meant to apply to labels or not?

I'm guessing they are, in which case I think everything here makes sense, but, want to confirm.

mjameswh:
> The one thing I can't remember is if we said user converters/codecs are meant to apply to labels or not? I'm guessing they are, in which case I think everything here makes sense, but, want to confirm.

Label payload has to be a JSON string. The POC converts them using user-defined converters.

TBH, I'd prefer strictly using the default JSON converter for that, to match what UI will expect. But User Metadata went the other direction, and I don't think there's a justification for Event Groups and User Metadata to diverge on that question.

Yes, user's Payload Codecs do apply to Label payloads.

Not sure though why you're asking those questions here — payload converters/codecs are not the responsibility of Core, and this PR is not introducing the user-facing API in the Rust SDK.

Sushisource:
> > The one thing I can't remember is if we said user converters/codecs are meant to apply to labels or not? I'm guessing they are, in which case I think everything here makes sense, but, want to confirm.
> 
> Label payload has to be a JSON string. The POC converts them using user-defined converters.
> 
> TBH, I'd prefer strictly using the default JSON converter for that, to match what UI will expect. But User Metadata went the other direction, and I don't think there's a justification for Event Groups and User Metadata to diverge on that question.
> 
> Yes, user's Payload Codecs do apply to Label payloads.
> 
> Not sure though why you're asking those questions here — payload converters/codecs are not the responsibility of Core, and this PR is not introducing the user-facing API in the Rust SDK.

Ya, fair point, we only have the not-real APIs in the Rust SDK. Sounds good to me then.

crates/workflow/src/workflow_context/options.rs
mjameswh: After having a deeper look at this, the `into_invocation` method was totally dubious. I replaced that method and associated types by an `into_command()` method on `SignalExternalWorkflowOptions`, copying what we have on other command options types.

Sushisource (approved):
Looking good to me, I think you just need to rebase to fix the compile errors.

PR #1310, opened 2026-06-03
:bookmark: Add gzip compression by default with option to opt-out

Sushisource:
## What was changed
Added transport level compression by default

Lang layers will need to expose option to change compression type (should be modeled like enum as we've done here in case we add other compression kinds).

## Why?
Be efficient!

## Checklist

1. Closes 

2. How was this tested:
Existing and new tests

3. Any docs updates needed?

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -108,6 +108,7 @@ use tonic::{
     Code, IntoRequest,
     body::Body,
     client::GrpcService,
+    codec::CompressionEncoding,
     codegen::InterceptedService,
     metadata::{
         AsciiMetadataKey, AsciiMetadataValue, BinaryMetadataKey, BinaryMetadataValue, MetadataMap,
@@ -162,6 +163,13 @@ impl Connection {
     /// Connect to a Temporal service.
     pub async fn connect(options: ConnectionOptions) -> Result<Self, ClientConnectError> {
         let dns_lb_opts = dns::validate_and_get_dns_lb(&options)?.cloned();
+        // The callback-based override transport cannot decode compressed request bodies, so
+        // compression is forced off whenever a service override is in use.
+        let compression = if options.service_override.is_some() {
+            GrpcCompression::None
+        } else {
+            options.grpc_compression
+        };
         let (service, dns_task) = if let Some(service_override) = options.service_override {
             (
                 GrpcMetricSvc {
@@ -238,7 +246,7 @@ impl Connection {
             headers: headers.clone(),
         };
         let svc = InterceptedService::new(service, interceptor);
-        let mut svc_client = TemporalServiceClient::new(svc);
+        let mut svc_client = TemporalServiceClient::new(svc, compression);
 
         let capabilities = if !options.skip_get_system_info {
             match svc_client
@@ -574,7 +582,7 @@ fn get_decode_max_size() -> usize {
 }
 
 impl TemporalServiceClient {
-    fn new<T>(svc: T) -> Self
+    fn new<T>(svc: T, compression: GrpcCompression) -> Self
     where
         T: GrpcService<Body> + Send + Sync + Clone + 'static,
         T::ResponseBody: tonic::codegen::Body<Data = tonic::codegen::Bytes> + Send + 'static,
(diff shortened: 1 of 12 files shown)


mjameswh:
Is the :boom: just a precaution, or are we aware of situations where the default opt-in policy could be problematic?

Sushisource:
> Is the 💥 just a precaution, or are we aware of situations where the default opt-in policy could be problematic?

Really just a precaution... maybe we need a different emoji for "requires lang changes but isn't breaking" which is what I really need here.

crates/client/src/lib.rs
```diff
@@ -574,31 +582,33 @@ fn get_decode_max_size() -> usize {
...
+                    GrpcCompression::None => client,
+                }
+            }};
+        }
+
+        let workflow_svc_client = Box::new(configure!(WorkflowServiceClient::new(svc.clone())));
+        let operator_svc_client = Box::new(configure!(OperatorServiceClient::new(svc.clone())));
+        let cloud_svc_client = Box::new(configure!(CloudServiceClient::new(svc.clone())));
+        let test_svc_client = Box::new(configure!(TestServiceClient::new(svc.clone())));
+        let health_svc_client = Box::new(configure!(HealthClient::new(svc.clone())));
```
eamsden, line 611: I think a trait would be much clearer here, though the macro isn't horrible as macros go.

crates/sdk-core/tests/shared_tests/mod.rs
```diff
@@ -36,10 +44,177 @@ use temporalio_sdk::{
...
+/// *and* that the server actually engages compression rather than silently ignoring it: the
+/// Temporal frontend only gzip-compresses its response when the client advertises
+/// `grpc-accept-encoding: gzip` (which our toggle controls), so `grpc-encoding: gzip` on the
+/// response iff enabled proves the negotiation is live.
+///
+/// Part 2 proves the *outbound request* bytes on the wire are genuinely gzip-compressed. We cannot
+/// inspect the bytes of the real (TLS) connection, so this routes through an in-process tonic
+/// server that lets us read the raw gRPC frame and confirm the compression flag, gzip magic bytes,
+/// size reduction, and that the payload decompresses to the exact same protobuf as the
+/// uncompressed request.
```
eamsden, line 73: Still enjoying this.

eamsden (approved):

crates/sdk-core-c-bridge/src/client.rs
```diff
@@ -50,6 +50,18 @@ pub struct ConnectionOptions {
     /// refreshed. If null, DNS load balancing is disabled. Ignored (forced off) when
     /// http_connect_proxy_options is also set.
     pub dns_load_balancing_options: *const ClientDnsLoadBalancingOptions,
+    /// Selects transport-level gRPC compression. The zero value enables gzip, which is the
+    /// default. Ignored when grpc_override_callback is set, since that transport cannot decode
+    /// compressed request bodies.
+    pub grpc_compression: ClientGrpcCompression,
+}
+
+/// cbindgen:prefix-with-name=true
```
jmaeagle99, line 59: I don't see any other use of this in the repo and other enums don't have it in the C bridge. Why is this needed here?

crates/client/src/lib.rs
Sushisource: It ends up being uglier since the trait would need to be defined down in the proto crate

crates/sdk-core-c-bridge/src/client.rs
```diff
@@ -50,6 +50,18 @@ pub struct ConnectionOptions {
...
+#[derive(Clone, Copy)]
+#[repr(C)]
+pub enum ClientGrpcCompression {
+    Gzip = 0,
```
jmaeagle99, line 63: Are there realistically any options that we might want to expose in terms of compression level?

crates/sdk-core-c-bridge/src/client.rs
```diff
@@ -1576,6 +1593,29 @@ mod tests {
         assert_eq!(dns_opts.resolution_interval, Duration::from_millis(5_000));
     }
 
+    #[test]
+    fn grpc_compression_defaults_to_gzip() {
```
jmaeagle99, line 1597: Isn't this just testing the default value of a test provided connection options? If that's what you want to test, just assert that base_connection_options has the expected default value. If that's not what you want to test, then this test should be explicitly setting gzip compression on a new ConnectionsOptions instance, convert that, and assert the passed through value.

jmaeagle99 (approved):

crates/sdk-core-c-bridge/src/client.rs
jmaeagle99, line 63: Should we flip these so that None is the default to match the definition of GrpcCompression in the client?

crates/sdk-core-c-bridge/src/client.rs
Sushisource: It's because there is another enum variant called `None` that already gets generated into the C header and the names collide

crates/sdk-core-c-bridge/src/client.rs
Sushisource: Unfortunately no such thing exists in tonic

Sushisource: Gzip is the default everywhere. Where are you thinking of?

crates/sdk-core-c-bridge/src/client.rs
Sushisource: In principle this tests that the default survives conversion, and the test below tests explicitly changing it.

I agree they're not the... highest value tests though.

crates/sdk-core-c-bridge/src/client.rs
jmaeagle99: Maybe we should apply this to all enums in the C bridge as a separate change.

crates/sdk-core-c-bridge/src/client.rs
jmaeagle99: Sorry, I was confused, looks good here.

crates/sdk-core-c-bridge/src/client.rs
Sushisource: Yeah probably a good idea

jmaeagle99: https://github.com/temporalio/sdk-rust/issues/1318

PR #1316, opened 2026-06-04
upgrade on CAN

THardy98:
## What was changed
Add trampolining (upgrade on CAN) to Rust SDK

## Why?
Last SDK needed before we can go GA

2. How was this tested:
Integration tests

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,6 @@
 [env]
 # This temporarily overrides the version of the CLI used for integration tests, locally and in CI
-# CLI_VERSION_OVERRIDE = "v1.6.3-serverless"
+CLI_VERSION_OVERRIDE = "v1.7.0"
 
 [alias]
 # Not sure why --all-features doesn't work
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,6 @@
 [env]
 # This temporarily overrides the version of the CLI used for integration tests, locally and in CI
-CLI_VERSION_OVERRIDE = "v1.7.0"
+# CLI_VERSION_OVERRIDE = "v1.6.3-serverless"
 
 [alias]
 # Not sure why --all-features doesn't work
--- a/crates/sdk-core/tests/integ_tests/worker_versioning_tests.rs
+++ b/crates/sdk-core/tests/integ_tests/worker_versioning_tests.rs
@@ -333,7 +333,7 @@ impl ContinueAsNewAutoUpgradeV1 {
             .await;
         assert!(ctx.target_worker_deployment_version_changed());
         let mut options = ContinueAsNewOptions::default();
-        options.initial_versioning_behavior = ContinueAsNewVersioningBehavior::AutoUpgrade.into();
+        options.initial_versioning_behavior = Some(ContinueAsNewVersioningBehavior::AutoUpgrade);
         ctx.continue_as_new(&(attempt + 1), options)?;
         Ok("v1.0".to_string())
     }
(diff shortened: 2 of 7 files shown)


.cargo/config.toml
```diff
@@ -1,6 +1,6 @@
 [env]
 # This temporarily overrides the version of the CLI used for integration tests, locally and in CI
-# CLI_VERSION_OVERRIDE = "v1.6.3-serverless"
```
Sushisource, line 3: This shouldn't be needed (also it's old, 1.7.1 is out). It's only needed for prerelease versions

crates/sdk-core/tests/integ_tests/worker_versioning_tests.rs
```diff
@@ -308,3 +699,83 @@ impl ActivityHasDeploymentStampWf {
         Ok(())
     }
 }
+
+#[workflow]
+#[derive(Default)]
+struct ContinueAsNewAutoUpgradeV1 {
+    should_continue_as_new: bool,
+}
+
```
Sushisource: Not sure why these are added down here far away from the tests that use them

.cargo/config.toml
THardy98: removed - i was copying the same thing i had done for the other PRs...

crates/sdk-core/tests/integ_tests/worker_versioning_tests.rs
THardy98: I just put them after the existing workflows (i.e. `ActivityHasDeploymentStampWf`)

i've moved the workflows to above their respective tests

crates/sdk/src/lib.rs
```diff
@@ -87,6 +87,7 @@ pub use crate::error::{
     OutgoingWorkflowError, WorkflowRegistrationError,
 };
 pub use temporalio_client::Namespace;
+pub use temporalio_common::protos::temporal::api::enums::v1::ContinueAsNewVersioningBehavior;
```
chris-olszewski: I think we want a SDK owned type here instead of exposing the raw proto type to users. (Similar to what I just did in https://github.com/temporalio/sdk-rust/pull/1307)

crates/workflow/src/workflow_context/options.rs
```diff
@@ -518,6 +518,12 @@ pub struct ContinueAsNewOptions {
     pub retry_policy: Option<RetryPolicy>,
     /// Whether the new workflow should run on a worker with a compatible build id.
     pub versioning_intent: Option<VersioningIntent>,
+    /// Versioning behavior to use for the first workflow task of the new run.
+    ///
+    /// This experimental option is only meaningful for workers using worker deployment
+    /// versioning. `AutoUpgrade` routes the new run to the current deployment version;
+    /// `UseRampingVersion` routes it to the ramping deployment version when one is configured.
+    pub initial_versioning_behavior: Option<ContinueAsNewVersioningBehavior>,
```
chris-olszewski, line 574: I think we want a SDK owned type here instead of exposing the raw proto type to users. (Similar to what I just did in https://github.com/temporalio/sdk-rust/pull/1307)

crates/sdk/src/lib.rs
Sushisource: Not sure I want to pub-re-export proto options like this. @chris-olszewski do you think we want to just categorically re-create these enums and avoid exposing the proto versions?

THardy98: Yeah, happy to do what you're suggesting, we did the same in some of the other repos. Wasn't sure if we'd made a decision.

chris-olszewski: I'm in favor of re-creating these enums to avoid exposing the proto versions

crates/sdk-core/tests/integ_tests/worker_versioning_tests.rs
```diff
@@ -292,19 +318,464 @@ async fn versioning_off_with_custom_build_id() {
...
+                identity: client.identity(),
+                ..Default::default()
+            }
+            .into_request(),
+        )
+        .await
+        .unwrap();
+}
+
+async fn wait_for_workflow_deployment_version(
```
chris-olszewski, line 745: Just because I did this recently in Ruby: do we want to also allow specifying a ramp % to wait on?

chris-olszewski (changes requested):
Only feedback is avoiding the direct proto type.

chris-olszewski (approved):

PR #1319, opened 2026-06-06
:bookmark: Prefix enum variants with type name

### LINKED ISSUE #1318: Prefix C bridge enums with type names
Apply `/// cbindgen:prefix-with-name` to enums or `prefix_with_name = true` in toml for the c bridge crate to globally disambiguate all enum fields/consts in the C bridge to avoid collisions.

jmaeagle99:
## What was changed

Generate C bridge enum variants with type name prefix

## Why?

Avoid collisions when enums have the same variant name.

## Checklist

1. Closes #1318

### DIFF
--- a/crates/sdk-core-c-bridge/build.rs
+++ b/crates/sdk-core-c-bridge/build.rs
@@ -5,7 +5,12 @@ use std::env;
 fn main() {
     let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
 
+    // Prefix every enum's variants with the enum's type name to avoid collisions.
+    let mut config = cbindgen::Config::default();
+    config.enumeration.prefix_with_name = true;
+
     let changed = cbindgen::Builder::new()
+        .with_config(config)
         .with_cpp_compat(true)
         .with_crate(crate_dir)
         .with_pragma_once(true)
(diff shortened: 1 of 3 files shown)


Sushisource (approved):
Thanks! The WASM workflow failure I think might actually be related to this though? That one doesn't flake, whereas the other fails look like standard slow CI nonsense

jmaeagle99:
> Thanks! The WASM workflow failure I think might actually be related to this though? That one doesn't flake, whereas the other fails look like standard slow CI nonsense

I think it's unrelated because the WASM tests have no dependency on the `sdk-core-c-bridge` crate and (while not exactly the same test) [there are flakes in that suite](https://github.com/temporalio/sdk-rust/actions/runs/27044728876) on the main branch.

Sushisource:
> > Thanks! The WASM workflow failure I think might actually be related to this though? That one doesn't flake, whereas the other fails look like standard slow CI nonsense
> 
> I think it's unrelated because the WASM tests have no dependency on the `sdk-core-c-bridge` crate and (while not exactly the same test) [there are flakes in that suite](https://github.com/temporalio/sdk-rust/actions/runs/27044728876) on the main branch.

Ah, ok. Weird. I'll have to figure that out.

PR #1322, opened 2026-06-09
Fix activity_cancel_delivered_without_heartbeat test

rkannan82:
## What
- Moves `validate()` into `CoreWfStarter::worker()` so all integration tests get namespace capabilities populated. Without this, tests checking capabilities like `worker_commands` were silently skipped.

- Disable eager activity dispatch in`activity_cancel_delivered_without_heartbeat` to temporarily workaround a server bug for worker commands -- [server fix](https://github.com/temporalio/temporal/pull/10634).
- In addition, the test fails if the worker did not receive the activity cancel command. Previously, the test would be marked as success.

## How did you test it?
Updated tests

### DIFF
--- a/crates/sdk-core/tests/shared_tests/mod.rs
+++ b/crates/sdk-core/tests/shared_tests/mod.rs
@@ -15,8 +15,8 @@ use std::{
 };
 use temporalio_client::{
     Client, ClientOptions, Connection, ConnectionOptions, GrpcCompression, NamespacedClient,
-    RetryOptions, WorkflowFetchHistoryOptions, WorkflowStartOptions, WorkflowTerminateOptions,
-    grpc::WorkflowService,
+    RetryOptions, WorkflowFetchHistoryOptions, WorkflowSignalOptions, WorkflowStartOptions,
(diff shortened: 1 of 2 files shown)


crates/sdk-core/tests/shared_tests/mod.rs
```diff
@@ -460,8 +460,10 @@ pub(crate) async fn activity_cancel_delivered_without_heartbeat() {
                     .cancellation_type(ActivityCancellationType::WaitCancellationCompleted)
                     .build(),
             );
-            // Timer needed to avoid cancel-before-sent
-            ctx.timer(Duration::from_millis(10)).await;
+            // Wait for the activity to be started by a worker. The server
```
rkannan82: Temp change to debug failing test.

rkannan82: Root cause was a bug in worker commands code when eager activity dispatch happens. It is being fixed in https://github.com/temporalio/temporal/pull/10634.

Sushisource (approved):

PR #1323, opened 2026-06-10
Expose memo on WorkflowStartOptions

GregBowyer:
## What was changed

Adds a public `memo: Option<HashMap<String, Payload>>` field to the bon::Builder-generated builder for WorkflowStartOptions, and wires it through both the StartWorkflowExecutionRequest and SignalWithStartWorkflowExecutionRequest construction paths in client::start_workflow.

Memos are returned by list_workflows and describe_workflow_execution, so this lets callers stash small caller-supplied metadata on a workflow without round-tripping through a query. The values must already be serialized Payloads -- the builder does not encode for you.

Includes three unit tests covering builder default, builder set, and clone round-trip.

## Why?
So end users of the high level rust SDK can use memos ;)

## Checklist
1. How was this tested:
Basic unit tests and inhouse software

2. Any docs updates needed?
Not sure probably


crates/client/src/options_structs.rs
```diff
@@ -322,6 +322,13 @@ pub struct WorkflowStartOptions {
 
     /// Multi-line static details for the workflow, shown in the Temporal UI.
     pub static_details: Option<String>,
+
+    /// Optional memo attached to the workflow execution. Memos are returned by
+    /// `list_workflows` and `describe_workflow_execution` so callers can stash small
+    /// caller-supplied metadata on a workflow without round-tripping through a query.
+    /// Each value must already be a serialized `Payload`; the builder does not encode
+    /// values for you.
+    pub memo: Option<HashMap<String, Payload>>,
```
Sushisource, line 331: I'm sure you're just following the pattern with `search_attributes` here, but, this field (and `search_attributes`) I think both need helpers on the builder that accept `T: TemporalSerializable` types, to make the actual serialization easy for users.

Then inside this struct we can store `Box<dyn TemporalSerializable>`s instead of `Payload`s and convert them later when these options are passed to the client call.

If you'd rather not do all that we can take over the PR.

chris-olszewski:
Implemented in #1443

PR #1326, opened 2026-06-11
💥 feat(sdk): remove usage of `anyhow` from public error types

chris-olszewski:
## What was changed
- Switch `ApplicationFailure` to be backed by boxed errors
- Switch `WorkflowError` to be backed by boxed errors
- `WorkflowContext::force_task_fail` now works with boxed errors instead of `anyhow`
- Added helper for constructing a `WorkflowTermination` from `ApplicationFailure` (allows `?` to work well in workflow code)
- Cleaned up some examples/tests that still had extra `anyhow` wrapping

These are technically breaking changes, but most code should continue to work without issue as `anyhow::Error` implements `Into<Box<dyn Error + Send + Sync>>`

Each commit is reviewable on its own.

## Why?
Reduce our usage of `anyhow` in public facing APIs

## Checklist

1. Closes N/A
2. How was this tested:
Existing tests

3. Any docs updates needed?
Once released we can update docs to no longer have `ApplicationFailure::new(anyhow!("boom"))`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,7 @@ document as your quick reference when submitting pull requests.
 - If you want to format, don't bother checking first. Just run formatting, and run it by using
   `cargo +nightly fmt`, because some settings require the nightly formatter.
 - Do not extract functions for relatively simple helpers that are only used in one location.
+- Use custom error types leveraging `thiserror` for any public facing error types
 
 
 ## Repo Specific Utilities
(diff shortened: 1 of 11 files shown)


Sushisource (approved):
Niiiice!

Maybe update agents.md and/or the readme with info that we shouldn't depend on anyhow in public API of the SDK?

PR #1327, opened 2026-06-12
feat: payload limits validation visitor

jmaeagle99:
## What was changed

Add a SDK-internal, build-time generated payload/memo limit visitor:
- Generated from proto service definitions (every `temporal.api.*` rpc input message)
- Emits `PayloadLimitsValidatable` implementations to report each payload-bearing field size and classification to a `PayloadLimitsSink` implementation.
- Consumes the `PAYLOAD_LIMITS_TABLE` in `build.rs` to map each payload-bearing message and field combination to a classification. The build will fail if there are any missing or stale entries. This allows us to classify new combinations as they are added to the API. Initial classification answers built from inspection of `temporal/temporal` repository.
- Algorithm terminates at `Payload`, `Payloads`, `Memo`, `Failure`, `Any`, `SearchAttributes`, and `Header` leaf fields.
- APIs:
  - `PayloadLimitSink` receives callbacks from visitor to consume the size and classification information for each visited message and field combination.
  - `validation_payload_limits` func to log errors and warnings and returns the first error-level `PayloadLimitViolation`. Each violation carries a proto-field-name path that is relative to the root message.

Approximately 370 lines of `build.rs` is the classification table itself.
Approximately 300 lines of `payload_limits.rs` is tests.

## Why?

Create a foundational element that will be later integrated to allow clients and workers to validate payload limits and proactively issue WFT / AT failures.

## Checklist

1. How was this tested: New unit tests

### DIFF
--- a/crates/common/build.rs
+++ b/crates/common/build.rs
@@ -36,8 +36,12 @@ fn generate_payload_visitor(
     File::open(descriptor_path)?.read_to_end(&mut descriptor_bytes)?;
     let descriptor_set = FileDescriptorSet::decode(&descriptor_bytes[..])?;
 
-    let mut generator = PayloadVisitorGenerator::new();
-    generator.process_descriptors(&descriptor_set);
+    let model = PayloadModel::build(&descriptor_set);
+    let mut generator = PayloadVisitorGenerator {
+        model,
+        message_fields: HashMap::new(),
+    };
+    generator.build_field_infos();
 
     let output_path = out_dir.join("payload_visitor_impl.rs");
     let mut file = File::create(&output_path)?;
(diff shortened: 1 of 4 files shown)


crates/common/src/payload_limits.rs
```diff
@@ -0,0 +1,655 @@
...
+        m.insert("cde".to_string(), payload(&[0u8; 20]));
+        // (2 + 10) + (3 + 20) = 35
+        assert_eq!(map_payload_data_sum(m.iter()), 35);
+    }
+
+    use crate::protos::temporal::api::sdk::v1::UserMetadata;
+    use crate::protos::temporal::api::workflowservice::v1::{
+        RespondActivityTaskFailedRequest, RespondWorkflowTaskCompletedRequest,
+        StartWorkflowExecutionRequest,
+    };
```
Sushisource: These `use` statements in random locations are something LLMs like to do that drives me nuts. Can we make sure all the ones like this get consolidated at the top of the file (or `mod tests` in this case?)

crates/common/build.rs
```diff
@@ -688,3 +689,1010 @@ fn is_map_entry(options: &Option<MessageOptions>) -> bool {
         .as_ref()
         .is_some_and(|o| o.map_entry.unwrap_or(false))
 }
+
+// ===========================================================================
+// Payload-limits validator generator (sibling of the payload-visitor generator)
```
Sushisource:
```suggestion
// Payload-limits validator generator
```

Haha, k

crates/common/build.rs
```diff
@@ -688,3 +689,1010 @@ fn is_map_entry(options: &Option<MessageOptions>) -> bool {
...
+// The validated closure's roots are derived automatically from the proto service definitions (every
+// `temporal.api.*` RPC *input* message; see `service_request_roots`), so a new RPC/request cannot be
+// silently missed — its payload fields become unclassified and fail the build until added here.
+//
+// The measurement strategy is derived mechanically from the field's proto shape.
+const PAYLOAD_LIMITS_TABLE: &[(&str, LimitClass)] = &[
+    // ============================== Blob (payload) limit, warn + error ==========================
+    (
+        "temporal.api.command.v1.CompleteWorkflowExecutionCommandAttributes.result",
+        LimitClass::Blob,
```
Sushisource: I might restructure this so you list all the strings and then `zip` them with an infinite iterator of the `LimitClass` so you avoid needing to repeat the type over and over in each section.

crates/common/src/payload_limits.rs
```diff
@@ -0,0 +1,655 @@
...
+pub trait PayloadLimitSink {
+    /// Called for each validated payload field. `field_name` is the leaf field's proto name; `size`
+    /// is the field's size in bytes for the given [`LimitClass`]. When `enforce_error` is `false`,
+    /// the field may warn but must not produce an error-level violation.
+    fn check(
+        &mut self,
+        field_name: &'static str,
+        class: LimitClass,
+        size: usize,
+        enforce_error: bool,
```
Sushisource, line 84: Up here `LimitClass` only has two variants, and error enforcement is passed separately, whereas in the generator code that's all wrapped together in `LimitClass`. Would probably make sense to use the same technique in both places. I think I prefer the approach here.

crates/common/build.rs
```diff
@@ -688,3 +689,1010 @@ fn is_map_entry(options: &Option<MessageOptions>) -> bool {
...
+    /// A measurable leaf requiring a table entry at the field's proto path.
+    Leaf(LeafKind),
+    /// A payload-containing nested message to recurse into.
+    Struct(StructShape, String),
+    /// Not payload-bearing; ignored.
+    Skip,
+}
+
+#[derive(Clone, Copy)]
+enum LeafKind {
```
Sushisource, line 912: It feels like this is serving the same purpose as `PayloadFieldKind` which exists in the file already. Can we re-use that? Maybe it needs some extra variants. Broadly it seems like maybe some code could get shared here in general, AFAICT we don't re-use any of the existing visitor code but in principle I feel like we should've been able to extend that code to serve this purpose?

If it doesn't make sense, no prob, but, worth seeing if we can reduce LoC here

jmaeagle99: I tried to share that specifically but both visitors have different terminal conditions and additional semantics carried with each of these enums, so I didn't refactor that. Was able to refactor and share more of the common message traversal and classification logic though. Ended with marginally less LoC but that might be due to comment add/removal.

Sushisource (approved):
Nice, glad that zip suggestion reduced things a lot

PR #1328, opened 2026-06-12
Test: revert signal approach for cancel test

rkannan82:
Revert signal-based synchronization back to 5s timer to test if cloud test passes.


rkannan82:
Test passed after fixing dynamic config. Root cause: enableCancelActivityWorkerCommand was set per-namespace in cloud UI but the deployed server (v1.32.0-156.5) has it as a global bool, so the namespace-scoped value was silently ignored. Fixed by setting it at cell level.

PR #1329, opened 2026-06-12
💥 feat(sdk): error type for signal external workflow

chris-olszewski:
## What was changed
- Rename `ChildWorkflowSignalError` -> `WorkflowSignalError`
- Switch `ExternalWorkflowHandle::signal` to use same machinery as signaling child workflow instead of `SignalExternalFut`

## Why?
- No `Failure` as error types

## Checklist

1. Closes N/A

2. How was this tested:
Updated test expect error instead of raw failure

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/common-wasm/src/error.rs
+++ b/crates/common-wasm/src/error.rs
@@ -359,9 +359,9 @@ pub enum OutgoingWorkflowError {
     /// A workflow failure sourced from child-workflow start.
     #[error(transparent)]
     ChildWorkflowStart(#[from] Box<ChildWorkflowStartError>),
-    /// A workflow failure sourced from child-workflow signaling.
+    /// A workflow failure sourced from signaling a workflow.
     #[error(transparent)]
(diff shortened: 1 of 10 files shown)


crates/common-wasm/src/error.rs
```diff
@@ -1023,52 +1023,52 @@ impl ChildWorkflowExecutionError {
     }
 }
 
-/// Error returned when signaling a child workflow fails.
+/// Error returned when signaling a workflow fails.
 #[derive(Debug, thiserror::Error)]
-pub enum ChildWorkflowSignalError {
+pub enum WorkflowSignalError {
     /// The signal delivery failed.
     #[error("Child workflow signal failed: {}", .0.failure().message)]
```
Sushisource, line 1030:
```suggestion
    #[error("Workflow signal failed: {}", .0.failure().message)]
```

?

Sushisource (approved):

PR #1333, opened 2026-06-16
Feature-select OTLP exporter TLS backend

seanarnold:
## What was changed

[PR #1274](https://github.com/temporalio/sdk-rust/pull/1274) made the TLS crypto backend
feature-selectable (`tls-ring` / `tls-aws-lc`) for the gRPC service client. This extends
that to the OTLP metric exporter. Previously `opentelemetry-otlp` hardcoded `tls-roots` +
`reqwest-rustls`, which always pulled in `ring`. Those are removed and replaced with
feature-gated wiring (`tls-ring` keeps today's behavior; `tls-aws-lc` routes the exporter
through `aws-lc-rs`), using `opentelemetry-otlp` 0.31.1's own `tls-ring`/`tls-aws-lc`
features.

## Why?

#1274 covered the gRPC path, but the OTLP exporter still hardcoded `ring`, so a
`ring`-free, `aws-lc-rs`-only build (required for FIPS-capable deployments) wasn't
possible. Building with `--no-default-features --features tls-aws-lc,otel` now yields a
dependency tree free of `ring`.

## Checklist

1. Closes  (no issue)

2. How was this tested:

   - `cargo tree -i ring` across all targets: empty under `tls-aws-lc`, still present
     under the default `tls-ring` (no change to default builds).
   - `cargo check`/`clippy` clean for `common` (both backends) and `sdk-core` (default
     and aws-lc feature sets).

3. Any docs updates needed?

   CHANGELOG.md updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ to docs, or any other relevant information.
 * `client()` and `workflow_handle()` helpers to `ActivityContext` for easily obtaining a Temporal client
 * Exposed `backoff_start_interval` when continuing as new, which will delay the first task of the
   continued workflow by the configured interval.
+* The `tls-ring` / `tls-aws-lc` features now also select the TLS crypto backend for the OTLP metric
+  exporter (in addition to the gRPC service client). Previously the OTLP exporter hardcoded the `ring`
+  backend regardless of the selected feature, which prevented producing a `ring`-free, `aws-lc-rs`-only
+  (FIPS-capable) build. Building with `--no-default-features --features tls-aws-lc,otel` now yields a
+  dependency tree free of `ring`.
 
 ### Breaking Changes
 * The `ActivityContext` constructor now requires `ClientOptions`.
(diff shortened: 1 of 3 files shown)


seanarnold:
@Sushisource Are you able to take a look at this one please? 🙇

Sushisource (approved):
Thanks!

seanarnold:
Thanks @Sushisource. Do you think you could cut a new release, since there's quite a few changes since the last one https://github.com/temporalio/sdk-rust/compare/v0.4.0...main

We're looking to use https://github.com/temporalio/sdk-ruby/ in a FIPS environment so would appreciate a release so we can follow up with a PR against that gem.

Sushisource:
@chris-olszewski Can comment on making a release. I think there's a few things he wants to get in first but I doubt it'll be long.

chris-olszewski:
@seanarnold Are you waiting for this to make it into the Ruby SDK? In the Ruby SDK we depend on the Rust core via a Git submodule so we don't need a tagged release of the Rust crates.

PR #1340, opened 2026-06-21
Support dynamic client certificate resolution (ResolvesClientCert)

### LINKED ISSUE #1338: [Feature Request] Support dynamic/reloadable client certificates (ResolvesClientCert) in TlsOptions

We run Rust services — both clients that start workflows and long-running workers that authenticate to Temporal Cloud with short-lived mTLS certs issued by Vault and rotated on disk by a sidecar. Today the cert is read once and turned into a static `tonic::transport::Identity` in `add_tls_to_channel`, which bakes it into the `Arc<ClientConfig>` at `Connection::connect`. When the on-disk cert rotates, the running process keeps presenting the original cert; once it expires, the next (re)handshake is rejected with `CertificateExpired`,
and the only remedy is a process restart.

The Go SDK doesn't hit this: `ConnectionOptions.TLS` accepts a `*tls.Config`, and `tls.Config.GetClientCertificate` is invoked per handshake so a rotated cert is picked up transparently for both clients and pollers, no reconnect.

To be able to easily and smoothly rotate temporal certificates

### Additional context

The Go Sdk can do it

brucearctor:
Support dynamic client certificate resolution (ResolvesClientCert)

## What

Add a `client_cert_resolver` field to `TlsOptions` that accepts an `Arc<dyn ResolvesClientCert>` for dynamic, per-handshake client certificate selection. This enables transparent mTLS certificate rotation without requiring a process restart — useful for short-lived certificates managed by Vault agents, cert-manager sidecars, or HSM-backed signers.

Closes #1338

## Why

The existing `client_tls_options` requires static `(cert, key)` bytes at connection time. Users with rotating certificates (e.g., Vault-issued certs with 24h TTLs) must restart the entire Temporal client to pick up new material. The Go SDK solves this with `tls.Config.GetClientCertificate`; this PR brings equivalent functionality to the Rust SDK.

## How

Since `tonic` 0.14.6 does not expose `rustls::ClientConfig::with_client_cert_resolver()`, the implementation bypasses tonic's TLS layer when a resolver is present:

1. **`build_custom_rustls_config()`** — manually constructs a `rustls::ClientConfig` with the user's `ResolvesClientCert`, replicating tonic's security defaults (protocol versions, ALPN, native/custom CA roots)
2. **`DynamicTlsConnector`** — a `tower::Service<Uri>` that wraps TCP with TLS using `tokio_rustls::TlsConnector`, including connect timeouts and tracing
3. **`Endpoint::connect_with_connector()`** — used instead of `connect()` to wire the custom connector into tonic's channel

The resolver fires on each new TLS handshake (reconnections), not per-RPC over an existing HTTP/2 connection.

## Key design decisions

- **Mutually exclusive with static certs** — setting both `client_tls_options` and `client_cert_resolver` returns `InvalidConfig`
- **Re-exports** — `ResolvesClientCert`, `CertifiedKey`, and `SignatureScheme` are re-exported at the crate root so users don't need to depend on `tokio-rustls` directly
- **DNS load balancing** — returns a clear error (not yet supported with dynamic certs due to `balance_channel` API constraints)
- **Proxy** — returns a clear error (not yet supported; would require composing connectors)
- **C bridge** — hardcodes `client_cert_resolver: None` (dynamic resolution from C callers needs a callback design in a follow-up)

## Changes

| File | Change |
|------|--------|
| `crates/client/src/options_structs.rs` | Add `client_cert_resolver` field, update `Debug` impl |
| `crates/client/src/lib.rs` | `TlsConfigResult` enum, `add_tls_to_channel` branching, `build_custom_rustls_config`, `DynamicTlsConnector`, proxy/DNS validation, tracing |
| `crates/client/src/dns.rs` | Handle `TlsConfigResult`, reject resolver + DNS LB |
| `crates/client/src/envconfig.rs` | Add `client_cert_resolver: None` |
| `crates/client/Cargo.toml` | Add `rustls-native-certs` dep |
| `crates/sdk-core-c-bridge/src/client.rs` | Add `client_cert_resolver: None` |
| `crates/sdk-core/tests/common/mod.rs` | Update test struct literals |
| `CHANGELOG.md` | Add entry under `[Unreleased] > Added` |

## Tests (15 new, 134 total pass)

- Custom connector returned when resolver is set (explicit domain + inferred domain)
- Mutual exclusion: static + dynamic certs → error
- Resolver + custom `ServerCertVerifier` combination
- Resolver + custom CA certificate
- `build_custom_rustls_config` with/without resolver/verifier
- Debug output correctness for `TlsOptions` with resolver
- `DynamicTlsConnector` is `Clone` + `Debug`
- Default `TlsOptions` has no resolver
- No-TLS passthrough returns `Standard`
- IP host fallback for domain extraction
- Re-exports compile (`CertifiedKey`, `SignatureScheme`)

## Future work

- [ ] SDK-owned convenience trait (decouple users from rustls types) — #1341
- [ ] Connection max-lifetime for proactive cert rotation — #1342
- [ ] DNS load balancing + dynamic certs — #1343
- [ ] C bridge callback support — #1344
- [ ] `FileWatchingCertResolver` batteries-included implementation — #1345


chris-olszewski:
Will be giving this a review very soon, thanks for your patience.

brucearctor:
i'm in no rush

crates/client/src/options_structs.rs
```diff
@@ -179,6 +182,25 @@ pub struct TlsOptions {
...
+    /// - HSM-backed certificate selection
+    /// - Dynamic certificate selection based on server hints
+    ///
+    /// Mutually exclusive with [`client_tls_options`](TlsOptions::client_tls_options).
+    /// Setting both is an error.
+    ///
+    /// The resolver must implement [`ResolvesClientCert`] from the `rustls` crate.
+    /// A simple implementation that reloads certificates from disk can use
+    /// `Arc<RwLock<CertifiedKey>>` internally.
+    ///
```
chris-olszewski: I think the comment is a little lengthy as it stands.
```suggestion
    ///
    /// Mutually exclusive with [`client_tls_options`](TlsOptions::client_tls_options).
    /// Setting both is an error.
    ///
```

crates/client/Cargo.toml
```diff
@@ -46,6 +46,7 @@ tokio = { version = "1.47", default-features = false, features = [
 ] }
 tonic = { workspace = true, default-features = false, features = ["tls-native-roots", "channel", "gzip"] }
 tokio-rustls = { version = "0.26", default-features = false }
+rustls-native-certs = "0.8"
```
chris-olszewski: We should probably feature gate this.

crates/client/src/lib.rs
```diff
@@ -215,9 +240,31 @@ impl Connection {
...
+                            ClientConnectError::InvalidConfig(format!(
+                                "Invalid TLS domain name '{domain}': {e}"
+                            ))
+                        })?
+                        .to_owned();
+                let connector = DynamicTlsConnector {
+                    tls: tokio_rustls::TlsConnector::from(rustls_config),
+                    domain: Arc::new(server_name),
+                };
+                channel.connect_with_connector(connector).await?
```
chris-olszewski: Should we need to strip out the `https` here since we're already taking care of tls?

chris-olszewski (changes requested):
Overall I think this design makes sense to me. The largest thing for me is making sure we feature gate it.

crates/client/src/options_structs.rs
brucearctor: shortened

crates/client/Cargo.toml
brucearctor: feature gated

crates/client/src/lib.rs
brucearctor: this should be cleaner now

brucearctor:
@chris-olszewski  - i think addressed?

crates/client/src/lib.rs
```diff
@@ -1616,11 +1917,287 @@ mod tests {
...
+                let debug_str = format!("{:?}", connector);
+                assert!(
+                    debug_str.contains("DynamicTlsConnector"),
+                    "Debug should show struct name: {debug_str}"
+                );
+                assert!(
+                    debug_str.contains("test.temporal.io"),
+                    "Debug should show domain: {debug_str}"
+                );
+            }
```
chris-olszewski: Unsure the value of a test asserting that these traits are implemented
```suggestion
```

crates/client/src/lib.rs
```diff
@@ -1616,11 +1917,287 @@ mod tests {
...
+                // and usable from the crate root. This is a compile-time check.
+                fn _assert_types_exist(
+                    _resolver: &dyn ResolvesClientCert,
+                    _scheme: SignatureScheme,
+                ) {
+                    // This function just needs to compile
+                }
+                let _ = SignatureScheme::ECDSA_NISTP256_SHA256;
+            }
+        }
```
chris-olszewski: I don't believe the test asserts what it says since `use super::*` gets parent's imports, not just exports. Directly referencing these via `crate::` is probably what is desired, but unsure if testing this is necessary.

crates/client/src/lib.rs
```diff
@@ -416,19 +480,78 @@ impl ClientHeaders {
...
 }
 
+/// Result of TLS configuration: either standard tonic TLS was applied to the endpoint,
+/// or a custom rustls config is needed for dynamic certificate resolution.
+#[derive(Debug)]
+enum TlsConfigResult {
+    /// Standard tonic TLS was applied, endpoint is ready to connect normally.
+    Standard(Endpoint),
+    /// A custom rustls::ClientConfig is needed. The endpoint has no TLS configured;
+    /// the caller must use `connect_with_connector` with a custom TLS connector.
```
chris-olszewski, line 586:
```suggestion
    /// the caller must use `connect_with_connector` with a custom TLS connector.
    ///
    /// Experimental API subject to change
```

chris-olszewski (changes requested):
Looks good to me. Only a few questions on some of the new tests and a suggestion that we mark this variant as experimental. Thanks for your patience on the review process.

brucearctor:
@chris-olszewski — rebased onto main and addressed your outstanding comments. 


## Summary:

Removed the re_exports_are_accessible test — you were right that use super::* imports parent imports, not just public exports, so it wasn't actually testing the re-export surface. The re-exports are exercised implicitly by the other dynamic-tls tests.
The dynamic_tls_connector_is_clone_and_debug removal and "Experimental API" doc comment were already in place from earlier.


## Merge conflict:

Rebased onto current main. The only conflict was integrating #1422's connect_timeout — it slots in between Endpoint::from_shared() and add_tls_to_channel() in the dynamic-tls path.


## IPv6 fixes (new): While doing a deeper review of the connector code, caught two IPv6 bugs:

* DynamicTlsConnector::call used format!("{host}:{port}") for TcpStream::connect, which produces ambiguous addresses for IPv6 (e.g. ::1:443). Switched to the (host, port) tuple form of ToSocketAddrs.
* SNI domain extraction from uri.host() returns brackets for IPv6 literals ([::1]), which ServerName::try_from rejects. Now strips brackets before conversion.


All 166 tests pass with --features dynamic-tls, 154 pass on default features.

CHANGELOG.md
```diff
@@ -24,6 +24,57 @@ to docs, or any other relevant information.
...
+* The `ActivityContext` constructor now requires `ClientOptions`.
+* `WorkerConfig::{workflow,activity,nexus}_task_poller_behavior` and the corresponding Rust SDK
+  `WorkerOptions` fields are now `Option<PollerBehavior>`. `None` means the poller was not explicitly
+  configured and is eligible for automatic enrollment into poller autoscaling.
+### Breaking Changes
+
+- Rust SDK `ApplicationFailure` and `WorkflowError` APIs now use boxed `std::error::Error` values instead of
+  `anyhow::Error`.
+
+## [Unreleased]
```
chris-olszewski, line 122: Changelog got a little messed up.

```suggestion
## Unreleased

### Added
* Support for dynamic client certificate resolution via `TlsOptions::client_cert_resolver`, which
  accepts an `Arc<dyn ResolvesClientCert>` for per-handshake mTLS certificate selection. This enables
  transparent certificate rotation without process restarts — useful for short-lived certificates
  managed by Vault, cert-manager, or HSM-backed signers. `ResolvesClientCert`, `CertifiedKey`, and
  `SignatureScheme` are re-exported from the crate root for convenience.

### Breaking Changes :boom:
* `CancellableFuture` and `CancellableFutureWithReason` now use the inherited `Future::Output`
associated type instead of a generic output parameter.

## [0.6.0] - 2026-08-04
```

crates/client/src/lib.rs
```diff
@@ -271,13 +289,32 @@ impl Connection {
...
+                    domain,
+                } => (ep, Some((rustls_config, domain))),
             };
-            let channel = add_tls_to_channel(options.tls_options.as_ref(), channel).await?;
+            #[cfg(not(feature = "dynamic-tls"))]
+            let channel = match tls_result {
+                TlsConfigResult::Standard(ep) => ep,
+            };
+
+
```
chris-olszewski:
```suggestion
```

crates/client/src/lib.rs
```diff
@@ -1958,11 +2271,249 @@ mod tests {
...
+            #[test]
+            fn tls_options_default_has_no_resolver() {
+                let tls_opts = TlsOptions::default();
+                assert!(tls_opts.client_cert_resolver.is_none());
+                assert!(tls_opts.client_tls_options.is_none());
+                assert!(tls_opts.server_cert_verifier.is_none());
+            }
+
+
```
chris-olszewski:
```suggestion
```

crates/client/src/lib.rs
```diff
@@ -1958,11 +2271,249 @@ mod tests {
...
+                        assert_eq!(domain, "192.168.1.100");
+                    }
+                    other => panic!(
+                        "Expected CustomConnector with IP domain, got {:?}",
+                        other.err()
+                    ),
+                }
+            }
+
+
```
chris-olszewski:
```suggestion
```

brucearctor: :-/ 

Ya, that didn't look right.  

Normally handled by a pre-commit or pre-push hook.  Is there a way the repo can add something contributors can easily opt-into for these basic things like linting?  What is the automation you have locally to be able to catch these things before pushing?  

I def don't mean for you to have to clean these things up.

brucearctor:
i see some conflicts -- gonna try to resolve.

crates/client/src/lib.rs
```diff
@@ -271,13 +289,32 @@ impl Connection {
...
+                    rustls_config,
+                    domain,
+                } => (ep, Some((rustls_config, domain))),
+            };
+            #[cfg(not(feature = "dynamic-tls"))]
+            let channel = match tls_result {
+                TlsConfigResult::Standard(ep) => ep,
+            };
+
+
```
chris-olszewski:
```suggestion
```

crates/client/src/lib.rs
```diff
@@ -1957,11 +2270,249 @@ mod tests {
...
+            #[test]
+            fn tls_options_default_has_no_resolver() {
+                let tls_opts = TlsOptions::default();
+                assert!(tls_opts.client_cert_resolver.is_none());
+                assert!(tls_opts.client_tls_options.is_none());
+                assert!(tls_opts.server_cert_verifier.is_none());
+            }
+
+
```
chris-olszewski:
```suggestion
```

crates/client/src/lib.rs
```diff
@@ -1957,11 +2270,249 @@ mod tests {
...
+                        assert_eq!(domain, "192.168.1.100");
+                    }
+                    other => panic!(
+                        "Expected CustomConnector with IP domain, got {:?}",
+                        other.err()
+                    ),
+                }
+            }
+
+
```
chris-olszewski:
```suggestion
```

chris-olszewski:
```suggestion
```

crates/client/src/lib.rs
chris-olszewski: Personally I have format on save & clippy setup in my editor.

crates/client/src/lib.rs
brucearctor: 🤦 ... formatting

chris-olszewski (approved):

martinm-be:
\o/

crates/client/src/lib.rs
brucearctor: interesting -- I'll take a look, thanks for the thought!  Have only ever handled such things with pre-commit, and or lefthook pre-commit and/or pre-pushes.  

I assumed such things are per-repo configurations, since conventions vary widely, which is why I have only ever put any specific formatting in the repo.  Then everyone can opt into the same thing locally which CI also enforces.  

As long as that can be configured for just this repo, that makes sense.  I'll look at config options.

brucearctor:
Thanks for the thorough review @chris-olszewski !

PR #1346, opened 2026-06-21
feat: add typed search attributes API

### LINKED ISSUE #1337: [Feature Request] Provide well typed Search Attributes APIs

Currently you must work with the Search Attributes proto directly or encode/decode various types to payloads

API similar to other SDKs where working with Search Attributes (reading/setting/upserting) is ergonomic.

### Additional context

N/A

brucearctor:
## Summary

Adds a type-safe Search Attributes API to the Rust SDK, providing parity with Go, TypeScript, and Python SDKs.

Closes #1337

### Core Types (`crates/common-wasm/src/search_attributes.rs`)
- `SearchAttributeKey<T>` — const-constructible typed keys for all 7 Temporal indexed value types
- `SearchAttributeValue` — sealed trait implemented for `bool`, `i64`, `f64`, `String`, `prost_types::Timestamp`, `Vec<String>`
- `TypedSearchAttributes` — type-safe collection with `.get()` / `.to_proto()` / `.from_proto()`
- `SearchAttributeUpdate` — type-erased set/unset for heterogeneous upserts
- Correct wire format: `metadata["encoding"]=json/plain`, `metadata["type"]=IndexedValueType` (matches Go SDK)

### Integration
- `WorkflowContext`: `typed_search_attributes()` getter + `upsert_typed_search_attributes()`
- `WorkflowStartOptions`: `typed_search_attributes` field
- `ChildWorkflowOptions`: `typed_search_attributes` field
- `ContinueAsNewOptions`: `typed_search_attributes` field

### Design Decisions
- **No new dependencies** — uses `prost_types::Timestamp` for datetime (already in dep graph), not chrono
- **Additive** — existing raw `search_attributes` fields/methods unchanged; if both set, typed takes precedence
- **Types in `common-wasm`** — shared across workflow + client crates via re-export

## Before / After

```rust
// BEFORE: raw string keys, manual serde, no type safety
let val = ctx.search_attributes()
    .indexed_fields.get("CustomKeywordField")
    .and_then(|p| serde_json::from_slice::<String>(&p.data).ok());

ctx.upsert_search_attributes([
    ("CustomKeywordField".to_string(), "updated".as_json_payload()?),
]);
```

```rust
// AFTER: compile-time type safety
const KEYWORD: SearchAttributeKey<String> = SearchAttributeKey::keyword("CustomKeywordField");

let val = ctx.typed_search_attributes().get(&KEYWORD);

ctx.upsert_typed_search_attributes([KEYWORD.value_set("updated".into())]);
```

## Testing
- 19 unit tests covering all value type round-trips, proto conversion, unset behavior, Keyword vs Text disambiguation, graceful type mismatch handling
- 1 doc-test
- Full workspace `cargo check` — 0 errors, 0 warnings


crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,854 @@
...
+// SearchAttributeValue trait
+// ---------------------------------------------------------------------------
+
+mod private {
+    pub trait Sealed {}
+    impl Sealed for bool {}
+    impl Sealed for i64 {}
+    impl Sealed for f64 {}
+    impl Sealed for String {}
+    impl Sealed for prost_types::Timestamp {}
```
chris-olszewski: I think a better direction here would be to provide our own `Timestamp` type that we own with maybe conversions from `prost_types::Timestamp`.

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,854 @@
...
+                String::from_utf8_lossy(encoding)
+            ),
+        });
+    }
+    Ok(())
+}
+
+// --- bool ---
+
+impl SearchAttributeValue for bool {
```
chris-olszewski: For all of the value types aside from timestamp are similar enough that defining `impl SearchAttributeValue` via a macro makes sense.

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,854 @@
...
+
+fn build_payload(
+    json_bytes: Vec<u8>,
+    indexed_value_type: IndexedValueType,
+) -> Payload {
+    let mut metadata = HashMap::with_capacity(2);
+    metadata.insert(
+        ENCODING_PAYLOAD_KEY.to_string(),
+        JSON_ENCODING_VAL.as_bytes().to_vec(),
+    );
```
chris-olszewski: We should reuse the default JSON payload converter here instead of directly calling `serde_json` and building up the metadata ourselves e.g. https://github.com/temporalio/sdk-ruby/blob/11dc8dbf63a6911aa53e22cb40acf597b89e9a68/temporalio/lib/temporalio/search_attributes.rb#L149

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,854 @@
...
+        type_metadata_str(indexed_value_type).as_bytes().to_vec(),
+    );
+    Payload {
+        metadata,
+        data: json_bytes,
+        ..Default::default()
+    }
+}
+
+fn validate_encoding(payload: &Payload) -> Result<(), SearchAttributeError> {
```
chris-olszewski: I think this can mostly be removed if we decode using the default payload converter.

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,854 @@
...
+
+    fn default_indexed_value_type() -> IndexedValueType {
+        IndexedValueType::Keyword
+    }
+}
+
+// --- Timestamp ---
+// Temporal wire format expects RFC3339 strings for Datetime search attributes.
+
+fn timestamp_to_rfc3339(ts: &Timestamp) -> String {
```
chris-olszewski: I am unsure I want to inline these implementations unless we add verification. I think we already pull in `chrono` transitively so it might make sense to just use that internally to convert to/from timestamps

chris-olszewski (changes requested):
The "typed" variants of these APIs should wholesale replace the current raw payload APIs for working with search attributes.

brucearctor:
@chris-olszewski - thanks for review.  I think addressed, and found some other improvements!

crates/client/src/options_structs.rs
```diff
@@ -286,8 +287,8 @@ pub struct WorkflowStartOptions {
     /// Optionally set a cron schedule for the workflow
     pub cron_schedule: Option<String>,
 
-    /// Optionally associate extra search attributes with a workflow
-    pub search_attributes: Option<HashMap<String, Payload>>,
+    /// Typed search attributes to set when starting the workflow.
```
chris-olszewski:
```suggestion
    /// Additional search attributes for the workflow.
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1108 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// TypedSearchAttributes
+// ---------------------------------------------------------------------------
+
+/// A collection of search attribute payloads, providing type-safe access via
+/// [`SearchAttributeKey`].
+#[derive(Debug, Clone, Default)]
+pub struct TypedSearchAttributes {
```
chris-olszewski: Just since we no longer have an "untyped" version I think we can drop the `Typed` prefix.
```suggestion
pub struct SearchAttributes {
```

crates/workflow/src/workflow_context/options.rs
```diff
@@ -281,8 +282,8 @@ pub struct ChildWorkflowOptions {
     pub task_timeout: Option<Duration>,
     /// Optionally set a cron schedule for the workflow
     pub cron_schedule: Option<String>,
-    /// Optionally associate extra search attributes with a workflow
-    pub search_attributes: Option<HashMap<String, Payload>>,
+    /// Typed search attributes to set on the child workflow.
```
chris-olszewski:
```suggestion
    /// Additional search attributes to set on the child workflow.
```

crates/workflow/src/workflow_context/options.rs
```diff
@@ -561,9 +558,8 @@ pub struct ContinueAsNewOptions {
     pub memo: Option<HashMap<String, Payload>>,
     /// If set, the new workflow will have these headers.
     pub headers: Option<HashMap<String, Payload>>,
-    /// If set, the new workflow will have these search attributes. If `None`, reuses the current
-    /// search attributes.
-    pub search_attributes: Option<SearchAttributes>,
+    /// Typed search attributes for the continued workflow.
```
chris-olszewski:
```suggestion
    /// If set, the new workflow will have these search attributes. If `None`, reuses the current
    /// search attributes.
```

crates/workflow/src/workflow_context.rs
```diff
@@ -795,6 +796,11 @@ impl<W> SyncWorkflowContext<W> {
         Ref::map(self.base.inner.shared.borrow(), |s| &s.search_attributes)
     }
 
+    /// Return current values for workflow search attributes as a typed collection.
+    pub fn typed_search_attributes(&self) -> TypedSearchAttributes {
```
chris-olszewski: The typed variant should replace the raw proto accessor completely.

```suggestion
    /// Return current values for workflow search attributes.
    pub fn search_attributes(&self) -> TypedSearchAttributes {
```

crates/workflow/src/workflow_context.rs
```diff
@@ -1156,6 +1164,11 @@ impl<W> WorkflowContext<W> {
         self.sync.search_attributes()
     }
 
+    /// Return current values for workflow search attributes as a typed collection.
+    pub fn typed_search_attributes(&self) -> TypedSearchAttributes {
+        self.sync.typed_search_attributes()
+    }
```
chris-olszewski: Same as above, the typed search attributes should replace the untyped.
```suggestion
    /// Return current values for workflow search attributes.
    pub fn search_attributes(&self) -> TypedSearchAttributes {
        self.sync.typed_search_attributes()
    }
```

crates/sdk-core/tests/integ_tests/workflow_tests/upsert_search_attrs.rs
```diff
@@ -159,8 +142,9 @@ async fn upsert_search_attrs_from_workflow() {
                     let fields = &msg.search_attributes.as_ref().unwrap().indexed_fields;
                     let payload1 = fields.get(k1).unwrap();
                     let payload2 = fields.get(k2).unwrap();
-                    assert_eq!(payload1.data[0], 0x01);
-                    assert_eq!(payload2.data[0], 0x02);
+                    // Verify payloads have JSON-encoded data
+                    assert!(!payload1.data.is_empty());
+                    assert!(!payload2.data.is_empty());
```
chris-olszewski: Just for slightly better regression testing around payload conversion
```suggestion
                    assert_eq!(payload1.data, b"\"value1\"");
                    assert_eq!(payload2.data, b"\"value2\"");
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1108 @@
...
+//!
+//! Search attributes are key-value pairs attached to workflows that enable
+//! server-side filtering via visibility queries. This module provides a typed
+//! layer over the raw proto payloads so that attribute values are checked at
+//! compile time.
+//!
+//! # Example
+//!
+//! ```
+//! use temporalio_common_wasm::search_attributes::{SearchAttributeKey, Timestamp};
```
chris-olszewski: Is the `Timestamp` import needed for this doc example?

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1108 @@
...
+
+use std::collections::HashMap;
+use std::marker::PhantomData;
+
+use crate::protos::temporal::api::common::v1::{
+    Payload, SearchAttributes as ProtoSearchAttributes,
+};
+use crate::protos::temporal::api::enums::v1::IndexedValueType;
+
+/// Metadata key for the search attribute value type, matching Go SDK convention.
```
chris-olszewski:
```suggestion
/// Metadata key for the search attribute value type, kept consistent across all SDKs.
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1108 @@
...
+
+// ---------------------------------------------------------------------------
+// SDK-owned Timestamp type
+// ---------------------------------------------------------------------------
+
+/// An SDK-owned timestamp for Datetime search attributes.
+///
+/// This type decouples the public API from `prost_types::Timestamp`. Conversion
+/// traits are provided for [`prost_types::Timestamp`] and
+/// [`std::time::SystemTime`].
```
chris-olszewski, line 61:
```suggestion
/// Timestamp type for Datetime search attributes.
///
/// Conversions from [`prost_types::Timestamp`] and
/// [`std::time::SystemTime`] are provided.
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1108 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Timestamp {
+    /// Seconds since the Unix epoch (1970-01-01T00:00:00Z).
+    /// For times before the epoch, seconds is negative and `nanos` still
+    /// represents a non-negative offset within that second.
+    pub seconds: i64,
+    /// Non-negative fractions of a second at nanosecond resolution.
+    /// Must be in the range `[0, 999_999_999]` (following the protobuf
+    /// `google.protobuf.Timestamp` convention).
+    pub nanos: i32,
```
chris-olszewski: Just to better prevent misuse I think we should have these private and add a failable constructor that verifies the invariants described in the doc comments.
```suggestion
    /// Seconds since the Unix epoch (1970-01-01T00:00:00Z).
    /// For times before the epoch, seconds is negative and `nanos` still
    /// represents a non-negative offset within that second.
   seconds: i64,
    /// Non-negative fractions of a second at nanosecond resolution.
    /// Must be in the range `[0, 999_999_999]` (following the protobuf
    /// `google.protobuf.Timestamp` convention).
   nanos: i32,
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1108 @@
...
+    /// Falls back to `Debug` formatting if the timestamp is out of chrono's range.
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match timestamp_to_rfc3339(self) {
+            Ok(s) => f.write_str(&s),
+            Err(_) => write!(f, "Timestamp({}, {})", self.seconds, self.nanos),
+        }
+    }
+}
+
+impl From<prost_types::Timestamp> for Timestamp {
```
chris-olszewski, line 116: If we end up going with additional checks at construction time this should be failable conversion
```suggestion
impl TryFrom<prost_types::Timestamp> for Timestamp {
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1108 @@
...
+            indexed_fields: self.fields,
+        }
+    }
+
+    /// Construct from the proto wire representation by cloning the inner map.
+    pub fn from_proto(attrs: &ProtoSearchAttributes) -> Self {
+        Self {
+            fields: attrs.indexed_fields.clone(),
+        }
+    }
```
chris-olszewski, line 698: This should be `impl From<ProtoSearchAttributes> for TypedSearchAttributes` where the proto is consumed. Generally better to have callers do a clone to not hide the fact that there's cloning required for this conversion.

```suggestion
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1108 @@
...
+
+    /// Convert to the proto wire representation, consuming `self` to avoid
+    /// cloning.
+    pub fn into_proto(self) -> ProtoSearchAttributes {
+        ProtoSearchAttributes {
+            indexed_fields: self.fields,
+        }
+    }
```
chris-olszewski, line 691: What do you think about just providing a `impl From<TypedSearchAttributes> for ProtoSearchAttributes` instead of the special accessor methods/ad-hoc conversion methods?

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1108 @@
...
+    metadata.insert(
+        TYPE_METADATA_KEY.to_string(),
+        type_metadata_str(indexed_value_type).as_bytes().to_vec(),
+    );
+    Ok(Payload {
+        metadata,
+        data,
+        ..Default::default()
+    })
+}
```
chris-olszewski, line 260: We should use the payload converter directly, not emulate behavior
```suggestion
fn encode_json_search_attr<T: TemporalSerializable + 'static>(
    value: &T,
    indexed_value_type: IndexedValueType,
) -> Result<Payload, SearchAttributeError> {
    let pc = PayloadConverter::serde_json();
    let context = SerializationContext {
        converter: &pc,
        data: &SerializationContextData::None,
    };
    let mut payload = pc.to_payload(&context, value)?;
    payload.metadata.insert(
        TYPE_METADATA_KEY.to_string(),
        type_metadata_str(indexed_value_type).as_bytes().to_vec(),
    );
    Ok(payload)
}
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1108 @@
...
+    if encoding.as_slice() != b"json/plain" {
+        return Err(SearchAttributeError::InvalidPayload {
+            reason: format!(
+                "expected encoding 'json/plain', got '{}'",
+                String::from_utf8_lossy(encoding)
+            ),
+        });
+    }
+    Ok(serde_json::from_slice(&payload.data)?)
+}
```
chris-olszewski, line 285: Similar to above, we can just use the JSON payload converter directly
```suggestion
fn decode_json_search_attr<T: TemporalDeserializable + 'static>(
    payload: Payload,
) -> Result<T, SearchAttributeError> {
    let pc = PayloadConverter::serde_json();
    let context = SerializationContext {
        converter: &pc,
        data: &SerializationContextData::None,
    };
    Ok(pc.from_payload(&context, payload)?)
}
```

chris-olszewski (changes requested):
This is a far better API. I think it makes sense to use this new typed search attributes everywhere and remove existing methods for accessing the raw proto. If a user wants to work with the raw proto they can use the conversion and work with it directly.

Just a few other things to clean up before we merge.

brucearctor:
still workign on this.  should have it later or else soon

brucearctor:
Should be ok enough now.

I see some other improvements, but can add those later

chris-olszewski:
Please fixup the formatting and compile errors for `cargo integ-test`

brucearctor:
🤦  -- of course.  I thought was all working/clean locally.  Let me get that fixed up

chris-olszewski (changes requested):
It appears the test failure might be legitimate.

brucearctor:
Appears that the recent commits:
https://github.com/temporalio/sdk-rust/commit/0f1b94427ef35c5486ce482bcd6148301239d509
And 
https://github.com/temporalio/sdk-rust/commit/9bcc99d6e3b02838aa57af800e3e808d0518d116
Had passed.

So, seems failure introduced once pulling in https://github.com/temporalio/sdk-rust/commit/32ef2a854fef1b8053ed6ee4809e0f051d528816

When at computer with some moments, will find a resolution

brucearctor:
Everything appears to be passing locally now

fixed failure around `sends_upsert`

Also, addressed a `timestamp_from_system_time` failure that was unrelated to this work [ windows only ], but seemed wise to get fixed somewhere.  Seemed better to just fix than leave out.

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1238 @@
...
+fn type_metadata_str(ivt: IndexedValueType) -> &'static str {
+    match ivt {
+        IndexedValueType::Bool => "Bool",
+        IndexedValueType::Int => "Int",
+        IndexedValueType::Double => "Double",
+        IndexedValueType::Keyword => "Keyword",
+        IndexedValueType::Text => "Text",
+        IndexedValueType::Datetime => "Datetime",
+        IndexedValueType::KeywordList => "KeywordList",
+        IndexedValueType::Unspecified | _ => "Unspecified",
```
chris-olszewski:
```suggestion
        IndexedValueType::Unspecified => "Unspecified",
```

brucearctor:
weird -- I'll take another pass through.  Running locally didn't surface errors.  I'll work on getting my local to be able to run the same as github CI.

crates/sdk-core/tests/integ_tests/workflow_tests/upsert_search_attrs.rs
```diff
@@ -31,17 +34,12 @@ struct SearchAttrUpdater;
...
+        let typed = ctx.search_attributes();
+        let orig_val = typed.get(&SA_INT).unwrap_or(0);
+        let new_val = orig_val + 1;
         ctx.upsert_search_attributes([
-            (SEARCH_ATTR_TXT.to_string(), "goodbye".as_json_payload()?),
-            (SEARCH_ATTR_INT.to_string(), int_val),
+            SA_TXT.value_set("goodbye".into()),
+            SA_INT.value_set(new_val),
         ]);
         if orig_val == 49 {
```
chris-olszewski:
```suggestion
        if orig_val == 1 {
```

crates/sdk-core/tests/integ_tests/workflow_tests/upsert_search_attrs.rs
```diff
@@ -101,7 +96,7 @@ async fn sends_upsert() {
         "goodbye",
         String::from_json_payload(txt_attr_payload).unwrap()
     );
-    assert_eq!(3, usize::from_json_payload(int_attr_payload).unwrap());
+    assert_eq!(2, usize::from_json_payload(int_attr_payload).unwrap());
```
chris-olszewski:
```suggestion
    assert_eq!(3, usize::from_json_payload(int_attr_payload).unwrap());
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1237 @@
...
+    }
+
+    #[test]
+    fn round_trip_bool() {
+        let val = true;
+        let payload = val
+            .to_search_attribute_payload(IndexedValueType::Bool)
+            .unwrap();
+        assert_payload_metadata(&payload, "Bool");
+        assert_eq!(bool::from_search_attribute_payload(&payload).unwrap(), true);
```
chris-olszewski:
```suggestion
        assert!(bool::from_search_attribute_payload(&payload).unwrap());
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1237 @@
...
+        let payload = val
+            .to_search_attribute_payload(IndexedValueType::Int)
+            .unwrap();
+        assert_payload_metadata(&payload, "Int");
+        assert_eq!(i64::from_search_attribute_payload(&payload).unwrap(), -42);
+    }
+
+    #[test]
+    fn round_trip_double() {
+        let val: f64 = 3.14;
```
chris-olszewski:
```suggestion
        let val: f64 = 1.23;
```

crates/common-wasm/src/search_attributes.rs
```diff
@@ -0,0 +1,1237 @@
...
+        let payload = val
+            .to_search_attribute_payload(IndexedValueType::Double)
+            .unwrap();
+        assert_payload_metadata(&payload, "Double");
+        let decoded = f64::from_search_attribute_payload(&payload).unwrap();
+        assert!((decoded - 3.14).abs() < f64::EPSILON);
```
chris-olszewski:
```suggestion
        assert!((decoded - 1.23).abs() < f64::EPSILON);
```

chris-olszewski (approved):

brucearctor:
@chris-olszewski  -- thanks for the little bits to help get this over the line.

Do we think at some point it can make sense for me to be able to trigger CI.  That seems a good way to remove you from the mix, and keep development more focused [ for contributors that are eventually trusted ].

chris-olszewski:
> Do we think at some point it can make sense for me to be able to trigger CI. That seems a good way to remove you from the mix, and keep development more focused [ for contributors that are eventually trusted ].

Will raise with the team.

tconley1428:
> @chris-olszewski -- thanks for the little bits to help get this over the line.
> 
> Do we think at some point it can make sense for me to be able to trigger CI. That seems a good way to remove you from the mix, and keep development more focused [ for contributors that are eventually trusted ].

I think this isn't likely. From a security perspective, the ability to kick off CI is arbitrary code execution inside of our github environment. _Potentially_ there's a story where individuals get favored contributor status of some kind, but we'd need to work that out with our security and legal folks.

brucearctor:
Ya, @tconley1428  -- I get it.

Just highlighting where could help reduce friction, increase velocity.  

Also, raises a question of whether non-employee contributors can do anything more than contribute code.  Is there a community plan?  How is merit/seniority/longer-term [ volunteer ] efforts to the project, acknowledged and rewarded.

Often that is via sustained contributions -> commit permissions [ and/or kicking off CI ] -> something like a technical steering committee, etc etc.  

Though, temporal is not in an OSS foundation, so doesn't have as clear of conventions to follow.  Worthwhile food for thought though -- which is def distinct from this thread.

tconley1428:
Yeah, definitely a place where we need to outline practices better. Despite being open source for quite some time, the volume of external contributions has ramped up substantially in the last 6 months or so, probably due in part to increasing popularity of the platform, but also due to the rise of AI coding. 

We're starting to draft some guidance for guidance for open source contributions here: https://github.com/temporalio/sdk-python/pull/1625 (Very early ideation on the contents, feel free to comment where you have thoughts), and we're talking with our open source folks internally to figure out where we stand between "We own this, even though it is open source" and the other extreme of "Everything goes through open source committee and guidance, etc." Something that'll evolve over time I think and above my pay grade to decide.

brucearctor:
Again - completely understood.

I've worked with OSPO offices.  Well aware of the challenges.  Pushing towards hopefully making some decisions :-p

PR #1347, opened 2026-06-21
feat(client): add workflow input support to ScheduleAction::StartWorkflow

jpds:
## What was changed

Adds workflow input payload support to `ScheduleAction::StartWorkflow`.

## Why?

Feature parity with the other SDKs.

## Checklist

1. Closes 

2. How was this tested:

I have workflow which uses this to encode an environment variable as the payload and this has a new integration test.

3. Any docs updates needed?

Already included in the rustdoc.

### DIFF
--- a/crates/client/src/schedules.rs
+++ b/crates/client/src/schedules.rs
@@ -1770,7 +1770,8 @@ mod tests {
     #[tokio::test]
     async fn schedule_action_start_workflow_with_input_into_proto() {
         use temporalio_common::{
-            UntypedWorkflow, data_converters::{DataConverter, RawValue},
+            UntypedWorkflow,
+            data_converters::{DataConverter, RawValue},
             protos::temporal::api::common::v1::Payload,
         };
 
--- a/crates/client/src/schedules.rs
+++ b/crates/client/src/schedules.rs
@@ -1769,7 +1769,7 @@ mod tests {
 
     #[tokio::test]
     async fn schedule_action_start_workflow_with_input_into_proto() {
-        use temporalio_common::workflow_definition::UntypedWorkflow;
+        use temporalio_client::UntypedWorkflow;
         use temporalio_common::{
             data_converters::{DataConverter, RawValue},
             protos::temporal::api::common::v1::Payload,
--- a/crates/client/src/schedules.rs
+++ b/crates/client/src/schedules.rs
@@ -1,16 +1,25 @@
 use crate::{Client, NamespacedClient, grpc::WorkflowService};
-use futures_util::stream;
+use futures_util::{FutureExt, future::BoxFuture, stream};
 use std::{
     collections::VecDeque,
     pin::Pin,
+    sync::Arc,
     task::{Context, Poll},
     time::{Duration, SystemTime},
 };
-use temporalio_common::protos::{
-    proto_ts_to_system_time,
-    temporal::api::{
-        common::v1 as common_proto, schedule::v1 as schedule_proto,
-        taskqueue::v1 as taskqueue_proto, workflow::v1 as workflow_proto, workflowservice::v1::*,
+use temporalio_common::{
+    HasWorkflowDefinition,
+    data_converters::{
+        DataConverter, PayloadConversionError, SerializationContextData, TemporalSerializable,
+    },
+    protos::{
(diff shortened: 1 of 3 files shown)


chris-olszewski (changes requested):
Thanks for the PR! Could you replace the existing `start_workflow` with your new `schedule_action_start_workflow` completely instead of adding a new method? The previous `start_workflow` method allowed for omitting arguments that might be necessary. Your new method better mirrors the `client.start_workflow` shape.

This is a breaking change, but it offers far better end user ergonomics and we are still pre 1.0 so now is the time to do them.

jpds:
@chris-olszewski Even better - done.

chris-olszewski (changes requested):
I'm sorry, I somehow missed in my initial review that this made the schedule constructor async as it eagerly invokes the data converter.

We should delay the payload conversion and codec application until the method where we actually construct the proto message itself. There's a little song and dance to this that involves defining a private trait to store the unserialized types without leaking a generic up to `CreateScheduleOptions`. You can look at what we did when accepting `details` on errors for an example: https://github.com/temporalio/sdk-rust/blob/0689f769ccc440e8d5008d4a623dee3c82a94ff2/crates/common-wasm/src/error.rs#L22

crates/client/src/schedules.rs
```diff
@@ -23,6 +32,74 @@ pub enum ScheduleError {
...
+        }
+    }
+}
+
+impl From<common_proto::Payloads> for ScheduleWorkflowInput {
+    fn from(p: common_proto::Payloads) -> Self {
+        Self {
+            repr: ScheduleWorkflowInputRepr::Encoded(p.payloads),
+        }
+    }
```
chris-olszewski: I don't think this constructor should be made public. 
```suggestion
```

crates/sdk-core/tests/integ_tests/schedule_tests.rs
```diff
@@ -428,3 +503,83 @@ async fn describe_accessors_match_created_values() {
 
     handle.delete().await.unwrap();
 }
+
+#[tokio::test]
+async fn create_schedule_with_workflow_input() {
```
chris-olszewski, line 486: Can you repourpose this test to use `UntypedWorkflow`/`RawValue` and the standard builder instead of constructing `ScheduleAction::StartWorkflow` with payloads directly?

chris-olszewski (changes requested):
2 incredibly minor things to clean up before merge. Thanks for tackling this.

chris-olszewski (approved):

crates/client/src/schedules.rs
```diff
@@ -1677,4 +1766,34 @@ mod tests {
             ScheduleOverlapPolicy::Unspecified
         );
     }
+
+    #[tokio::test]
+    async fn schedule_action_start_workflow_with_input_into_proto() {
+        use temporalio_common::workflow_definition::UntypedWorkflow;
```
chris-olszewski:
```suggestion
        use temporalio_client::UntypedWorkflow;
```

crates/client/src/schedules.rs
```diff
@@ -1677,4 +1766,34 @@ mod tests {
             ScheduleOverlapPolicy::Unspecified
         );
     }
+
+    #[tokio::test]
+    async fn schedule_action_start_workflow_with_input_into_proto() {
+        use temporalio_client::UntypedWorkflow;
+        use temporalio_common::{
+            data_converters::{DataConverter, RawValue},
```
chris-olszewski:
```suggestion
        use temporalio_common::{
            UntypedWorkflow, data_converters::{DataConverter, RawValue},
```

crates/client/src/schedules.rs
```diff
@@ -1677,4 +1766,33 @@ mod tests {
             ScheduleOverlapPolicy::Unspecified
         );
     }
+
+    #[tokio::test]
+    async fn schedule_action_start_workflow_with_input_into_proto() {
+        use temporalio_common::{
+            UntypedWorkflow, data_converters::{DataConverter, RawValue},
```
chris-olszewski:
```suggestion
            UntypedWorkflow,
            data_converters::{DataConverter, RawValue},
```

PR #1351, opened 2026-06-23
feat(sdk): nexus operation handler support

### LINKED ISSUE #1145: [Feature Request] Nexus handler definitions
Rust SDK can invoke Nexus operations from workflows but cannot implement/handle them. Need Nexus service definition, handler registration on workers, and operation context.

jpds:
## What was changed

Added Nexus operation handing support.

## Why?

I needed this for a workflow.

## Checklist

1. Closes #1145

2. How was this tested:

Integration test with that workflow.

4. Any docs updates needed?

rustdoc all in place.

### DIFF
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -75,6 +75,7 @@ extern crate self as temporalio_sdk;
 pub mod activities;
 pub mod error;
 pub mod interceptors;
+pub mod nexus;
 mod workflow_executor;
 mod workflow_future;
 mod workflow_registry;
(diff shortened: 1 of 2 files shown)


chris-olszewski (changes requested):
Adding Nexus handler support is a pretty large lift. I haven't fully looked through the implementation, but do not see any tests. For a feature like this we need fairly extensive testing. If you still feel like taking this on, I would highly suggest looking at how these are implemented in the SDKs that already support Nexus handlers (Go, Java, Typescript, Python) and look at the test suites.

chris-olszewski:
I discussed with the team and I don't believe we are in a place to accept community contributions for this feature. We need to first figure out what the design should look like along with creating a Rust Nexus RPC SDK: https://github.com/nexus-rpc

Issue #1139, opened 2026-03-09
[Feature Request] Workflow interceptors


chris-olszewski:
No inbound/outbound workflow interceptors. The existing `WorkerInterceptor` operates at the activation level, not at the semantic level of individual workflow operations.

sachinsharma3191:
**Follow-up:** The design proposal is in the thread above. We will **not** start coding / open a PR until a maintainer approves the direction (or specifies changes).

After approval, the first PR will wire at least one real hook plus tests, aligned with the agreed design.

Please confirm when we can proceed.

PR #1352, opened 2026-06-23
feat(sdk): move signal payload decoding to happen before dispatch

chris-olszewski:
## What was changed
Moves Payload conversion to happen before dispatch to various workflow handlers. This means that dispatch now takes `Box<dyn Any>` instead of `Payloads`.

This does split dispatch method into distinct `decode` and `dispatch` steps. Instead of adding a new failure variant if the `dyn Any` produced by `decode` doesn't match what `dispatch` expects we panic. This should only happen in handwritten `WorkflowDefinition`s and is considered a programmer error.

This is a breaking change for users who don't make use of the workflow proc macros.

## Why?
This is some prework for #1139 where handler interceptors should be able to see the handler input types for args instead of raw payloads.

## Checklist

1. Closes N/A

2. How was this tested:
Exisiting tests, new test to verify missing update handler failure.

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/workflow/src/runtime/entry.rs
+++ b/crates/workflow/src/runtime/entry.rs
@@ -90,12 +90,10 @@ pub trait WorkflowImplementation: Sized + 'static {
 
     /// Dispatch a signal using an already decoded input value.
     fn dispatch_signal(
-        _ctx: WorkflowContext<Self>,
+        ctx: WorkflowContext<Self>,
         name: &str,
-        _input: Box<dyn Any>,
-    ) -> LocalBoxFuture<'static, Result<(), WorkflowError>> {
-        panic!("typed signal dispatch called for unknown signal handler '{name}'")
-    }
+        input: Box<dyn Any>,
+    ) -> LocalBoxFuture<'static, Result<(), WorkflowError>>;
 
     /// Decode a query's payloads into that query handler's concrete input type.
     fn decode_query_input(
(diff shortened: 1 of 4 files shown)


crates/workflow/src/runtime/entry.rs
```diff
@@ -72,46 +79,71 @@ pub trait WorkflowImplementation: Sized + 'static {
...
-    fn dispatch_update(
-        _ctx: WorkflowContext<Self>,
+    /// Decode a signal's payloads into that signal handler's concrete input type.
+    fn decode_signal_input(
         _name: &str,
         _payloads: Payloads,
         _converter: &PayloadConverter,
-    ) -> Option<LocalBoxFuture<'static, Result<Payload, WorkflowError>>> {
-        None
+    ) -> Result<Option<Box<dyn Any>>, WorkflowError> {
```
Sushisource, line 87: Would it make sense for these `Any`s to be `TemporalDeserializeable` instead? I suppose it doesn't really matter much but in principle they all must be, right?

crates/workflow/src/runtime/entry.rs
```diff
@@ -72,46 +79,71 @@ pub trait WorkflowImplementation: Sized + 'static {
...
-    fn validate_update(
-        &self,
-        _ctx: WorkflowContextView,
+    /// Dispatch a signal using an already decoded input value.
+    fn dispatch_signal(
+        _ctx: WorkflowContext<Self>,
+        name: &str,
+        _input: Box<dyn Any>,
+    ) -> LocalBoxFuture<'static, Result<(), WorkflowError>> {
+        panic!("typed signal dispatch called for unknown signal handler '{name}'")
```
Sushisource: These are panics now because the handlers existence is checked earlier? I wonder if these should be `unreachable` instead?

Sushisource (approved):

crates/workflow/src/runtime/entry.rs
chris-olszewski: In principal yes they are all `TemporalDeserializeable`, slight snag is that `TemporalDeserializeable` is `Sized` so it isn't dyn-safe. We could make some marker trait that makes it more clear these are deserializable types, but unsure if that provides much value.

Sushisource: Ah, yeah makes sense.

PR #1358, opened 2026-06-26
chore: bump otel minor

chris-olszewski:
## What was changed
Bump to 0.32 minor of the OTEL packages and resulting necessary code changes

## Checklist

1. Closes N/A

2. How was this tested:
Existing tests

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/common/src/telemetry/otel.rs
+++ b/crates/common/src/telemetry/otel.rs
@@ -13,9 +13,11 @@ use super::{
 };
 use crate::dbg_panic;
 use opentelemetry::{
-    self, Key, KeyValue, Value,
+    self, Key, KeyValue,
     metrics::{Meter, MeterProvider as MeterProviderT},
 };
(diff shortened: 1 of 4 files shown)


crates/common/src/telemetry/otel.rs
```diff
@@ -13,9 +13,11 @@ use super::{
 };
 use crate::dbg_panic;
 use opentelemetry::{
-    self, Key, KeyValue, Value,
+    self, Key, KeyValue,
     metrics::{Meter, MeterProvider as MeterProviderT},
 };
+#[cfg(any(feature = "tls-ring", feature = "tls-aws-lc"))]
+use opentelemetry_otlp::tonic_types::transport::ClientTlsConfig;
```
chris-olszewski, line 20: This is now only exported if `opentelemetry_otlp` is built with one of the TLS features. `tls-ring` and `tls-aws-lc` are our own features that we forward down to select a TLS impl.

tconley1428 (approved):

PR #1363, opened 2026-06-30
:boom: Payload limit enforcement

jmaeagle99:
## What was changed

Builds on the existing payload-field classification codegen to add end-to-end payload/memo
size-limit enforcement in the client and worker.

- Validate every outbound request via generated dispatcher; error-level violation fails with an invalid argument status carrying the `PayloadLimitViolation` as its source.
- Warning thresholds configured on `ConnectionOptions` as `payload_limits PayloadLimitsOptions` field. Default is to warn payloads at 512 KiB and memos at 2 KiB, same defaults as server.
- Error limits are read by workers when describing the namespace and set on the `PayloadLimitsClient` decorator that wraps the existing connection decorator. Each request has the error limits (if set) attached as a request extension.
- On error limit violation, worker converts requests to their corresponding failures and the operations are retryable:
  - WFT completion → `RespondWorkflowTaskFailed`
  - Activity completion / cancel → `RespondActivityTaskFailed`
  - Activity heartbeat → `RespondActivityTaskFailed` and report activity as cancelled
  - Nexus completion → `RespondNexusTaskFailed`
- Logs and task-failure messages render the same TMPRL1103 string.
- Richer completion-span context (namespace/workflow_id/type/attempt/worker_id) threaded through for log forwarding.
- New `FailureReason::PayloadsTooLarge` metric reason.

## :boom: Breaking Change

Payload size-limit enforcement is now on by default; existing code changes behavior without opting in: an oversized payload is now failed proactively by the worker (as a retryable task failure) before it's sent, rather than being sent and hard-failed by the server as before.

If you use a proxy between the worker and server that alters the size of payloads (e.g.
compression, encryption, external storage), it is advised that you disable size enforcement by
setting `disable_payload_error_limit` to `true` on the worker.

## Why?

Prevent server from hard-failing workflows, activities, and operations due to payloads/memo size limit violations. Allows customers to potentially update their workflows, activities, and operations to reduce the oversized payloads/memos.

## Checklist

1. How was this tested: New unit tests and integration tests
1. Any docs updates needed? Yes

### DIFF
--- a/crates/client/src/grpc.rs
+++ b/crates/client/src/grpc.rs
@@ -13,8 +13,10 @@ use crate::{
 };
 use dyn_clone::DynClone;
 use futures_util::{FutureExt, TryFutureExt, future::BoxFuture};
+use parking_lot::RwLock;
 use std::{any::Any, marker::PhantomData, sync::Arc};
 use temporalio_common::{
+    payload_limits::{PayloadLimits, validate_known_payload_limits},
     protos::{
         grpc::health::v1::{health_client::HealthClient, *},
         temporal::api::{
@@ -144,6 +146,13 @@ impl RawGrpcCaller for Connection {
         F: FnMut(Request<Req>) -> BoxFuture<'static, Result<Response<Resp>, Status>>,
         F: Send + Sync + Unpin + 'static,
     {
+        // Validate payload sizes after any request mutation but before encoding/metrics.
+        validate_request_payload_limits(
+            &req,
+            self.inner.payloads_size_warn,
+            self.inner.memo_size_warn,
+        )?;
+
         let info = self
             .inner
             .retry_options
@@ -180,6 +189,47 @@ fn req_cloner<T: Clone>(cloneme: &Request<T>) -> Request<T> {
     }
     *new_req.extensions_mut() = cloneme.extensions().clone();
     new_req
+}
+
+/// Per-call payload/memo size error limits, attached to a request's extensions by a caller that
+/// wants error-level enforcement on this call.
+#[derive(Debug, Clone, Copy)]
+pub struct PayloadErrorLimits {
+    /// Blob (payload) size error threshold, in bytes.
+    pub blob: usize,
+    /// Memo size error threshold, in bytes.
+    pub memo: usize,
+}
+
+/// `*_warn` are the connection's configured warn thresholds; per-call error limits ride a
+/// [`PayloadErrorLimits`] extension. On an error-level violation, returns a [`Status`] carrying
+/// the [`PayloadLimitViolation`] as its source (extract via [crate::payload_limit_violation_from]).
+fn validate_request_payload_limits<Req: Any>(
+    req: &Request<Req>,
+    blob_warn: Option<usize>,
+    memo_warn: Option<usize>,
+) -> Result<(), Status> {
+    let mut limits = PayloadLimits {
+        blob_warn,
+        memo_warn,
+        blob_error: None,
+        memo_error: None,
+    };
+    if let Some(error_limits) = req.extensions().get::<PayloadErrorLimits>() {
+        // A zero threshold means "no limit for this class" (how the server reports an unset limit).
+        if error_limits.blob > 0 {
+            limits.blob_error = Some(error_limits.blob);
+        }
+        if error_limits.memo > 0 {
+            limits.memo_error = Some(error_limits.memo);
+        }
+    }
+    if let Some(violation) = validate_known_payload_limits(req.get_ref(), &limits) {
+        let mut status = Status::invalid_argument(violation.to_string());
+        status.set_source(Arc::new(violation));
+        return Err(status);
+    }
+    Ok(())
 }
 
 #[async_trait::async_trait]
@@ -366,6 +416,79 @@ where
         self.inner_mut_refreshed()
             .call(call_name, callfn, req)
             .await
+    }
+}
+
+/// Wraps a client and injects a worker's configured [`PayloadErrorLimits`] (when set) as a request
+/// extension on every gRPC call, so the gRPC layer enforces error limits uniformly across all
+/// outbound requests — current and future — without each call site having to opt in.
+///
+/// The limits are shared and may be updated after construction (e.g. once a worker learns the
+/// namespace limits) via [`set_error_limits`](Self::set_error_limits); clones observe the update.
+#[derive(Clone)]
+pub struct PayloadLimitsClient<C> {
+    inner: C,
+    error_limits: Arc<RwLock<Option<PayloadErrorLimits>>>,
+}
+
+impl<C> PayloadLimitsClient<C> {
+    /// Wrap `inner`; no limits are enforced until [`set_error_limits`](Self::set_error_limits).
+    pub fn new(inner: C) -> Self {
+        Self {
+            inner,
+            error_limits: Arc::new(RwLock::new(None)),
+        }
+    }
+
+    /// Set (or clear) the error limits injected on every call. Shared across clones.
+    pub fn set_error_limits(&self, limits: Option<PayloadErrorLimits>) {
+        *self.error_limits.write() = limits;
+    }
+}
+
+impl<C: RawClientProducer> RawClientProducer for PayloadLimitsClient<C> {
+    fn get_workers_info(&self) -> Option<Arc<ClientWorkerSet>> {
+        self.inner.get_workers_info()
+    }
+    fn workflow_client(&mut self) -> Box<dyn WorkflowService> {
+        self.inner.workflow_client()
+    }
+    fn operator_client(&mut self) -> Box<dyn OperatorService> {
+        self.inner.operator_client()
+    }
+    fn cloud_client(&mut self) -> Box<dyn CloudService> {
+        self.inner.cloud_client()
+    }
+    fn test_client(&mut self) -> Box<dyn TestService> {
+        self.inner.test_client()
+    }
+    fn health_client(&mut self) -> Box<dyn HealthService> {
+        self.inner.health_client()
+    }
+}
+
+#[async_trait::async_trait]
+impl<C> RawGrpcCaller for PayloadLimitsClient<C>
+where
+    C: RawGrpcCaller + Clone + Sync + 'static,
+{
+    async fn call<F, Req, Resp>(
+        &mut self,
+        call_name: &'static str,
+        callfn: F,
+        mut req: Request<Req>,
+    ) -> Result<Response<Resp>, Status>
+    where
+        Req: Clone + Unpin + Send + Sync + 'static,
+        Resp: Send + 'static,
+        F: FnMut(Request<Req>) -> BoxFuture<'static, Result<Response<Resp>, Status>>,
+        F: Send + Sync + Unpin + 'static,
+    {
+        let limits = *self.error_limits.read();
+        if let Some(limits) = limits {
+            req.extensions_mut().insert(limits);
+        }
+        self.inner.call(call_name, callfn, req).await
     }
 }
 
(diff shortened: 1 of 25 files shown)


crates/client/src/options_structs.rs
```diff
@@ -245,6 +249,26 @@ impl Default for DnsLoadBalancingOptions {
     }
 }
 
+/// Payload size limit options for a connection.
+#[derive(Clone, Debug)]
+pub struct PayloadLimitsOptions {
+    /// Warning threshold (bytes) for the size of an outbound payload-bearing field; over-threshold
+    /// fields are logged but still sent to server. Defaults to 512 KiB. Set to `0` to disable.
+    pub payloads_size_warn: u64,
```
jmaeagle99: I might flip the naming of these fields to:
- `payloads_warn_size`
- `memo_warn_size`
I think it would allow better grouping of related settings together if we ever need to add more for each severity level e.g. `payload_warn_fancy_feature` and I think it reads better when saying them because there aren't two S's next to each other like there is in `payloads_size_warn`.

This suggestion is contrary to what we did for [Python](https://github.com/temporalio/sdk-python/blob/main/temporalio/converter/_payload_limits.py#L11) and [Go](https://github.com/temporalio/sdk-go/blob/main/internal/payload_limits.go#L24). Also note that the payloads field is plural here where it is singular in Python and Go; I intentionally did that since the sizing is applied to payload-bearing fields, which include single payload fields, repeated payload fields (a sequence), and the `Payloads` proto message.

So options are:
1. Leave as-is.
2. Match Python and Go naming for consistency sake.
3. Rename to `payloads_warn_size` and `memo_warn_size` for (IMO) maringally better naming and conceptualization.
4. Some other better suggestion.

Naming is hard.

crates/client/src/grpc.rs
```diff
@@ -144,6 +146,13 @@ impl RawGrpcCaller for Connection {
         F: FnMut(Request<Req>) -> BoxFuture<'static, Result<Response<Resp>, Status>>,
         F: Send + Sync + Unpin + 'static,
     {
+        // Validate payload sizes after any request mutation but before encoding/metrics.
+        validate_request_payload_limits(
```
jmaeagle99, line 150: This is situated right before the gRPC message is sent so that it is the last thing in the gRPC client chain that could modify payloads and memos. Things I need to check here are:
- This needs to be checked before gRPC request metrics are emitted so that we aren't double counting attempts e.g. a WFT completion that is rejected due to payload sizes should not emit a gRPC request metric for that message.
- I recall there is a grpc override callback. This probably should occur before that callback is invoked.

jmaeagle99 (commented):
Some self-review comments. If you have ideas/opinions, would love to hear them.

crates/client/src/grpc.rs
```diff
@@ -182,6 +191,47 @@ fn req_cloner<T: Clone>(cloneme: &Request<T>) -> Request<T> {
...
 
+/// Per-call payload/memo size error limits, attached to a request's extensions by a caller that
+/// wants error-level enforcement on this call.
+#[derive(Debug, Clone, Copy)]
+pub struct PayloadErrorLimits {
+    /// Blob (payload) size error threshold, in bytes.
+    pub blob: usize,
+    /// Memo size error threshold, in bytes.
+    pub memo: usize,
+}
```
Sushisource: Nit: There is `request_extensions.rs` that this could go in, but, I'm fine with it here too.

crates/client/src/grpc.rs
```diff
@@ -369,6 +419,79 @@ where
...
+/// extension on every gRPC call, so the gRPC layer enforces error limits uniformly across all
+/// outbound requests — current and future — without each call site having to opt in.
+///
+/// The limits are shared and may be updated after construction (e.g. once a worker learns the
+/// namespace limits) via [`set_error_limits`](Self::set_error_limits); clones observe the update.
+#[derive(Clone)]
+pub struct PayloadLimitsClient<C> {
+    inner: C,
+    error_limits: Arc<RwLock<Option<PayloadErrorLimits>>>,
+}
```
Sushisource, line 418: I think this should probably be an interceptor, rather than a client wrapper?

crates/client/src/lib.rs
```diff
@@ -128,6 +129,14 @@ static TEMPORAL_NAMESPACE_HEADER_KEY: &str = "temporal-namespace";
 /// Key used to communicate when a GRPC message is too large
 pub static MESSAGE_TOO_LARGE_KEY: &str = "message-too-large";
 #[doc(hidden)]
+/// The violation, if `status` is the client proactively rejecting an outbound request for exceeding a
+/// payload/memo error size limit.
```
Sushisource:
```suggestion
/// Returns the violation, if `status` is the client proactively rejecting an outbound request for exceeding a
/// payload/memo error size limit.
```

crates/client/src/options_structs.rs
Sushisource: Option 3 works for me

crates/common/src/payload_limits.rs
```diff
@@ -144,46 +139,39 @@ where
         .sum()
 }
 
-/// Warn/error thresholds for both limit classes. An error threshold of `None` disables error
-/// enforcement for that class (warnings only).
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// Warn/error thresholds for both limit classes. A `None` threshold disables that check for that
+/// class: `None` warn = no warnings, `None` error = no error enforcement (warnings only).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 pub struct PayloadLimits {
```
Sushisource, line 151: If the resolver thing earlier resolves 0 to None, I'd kinda vote for these just being 0 also means disabled? 0-but-respected seems like a bit of a nonsense state

crates/sdk-core/src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -188,6 +192,33 @@ impl ActivityHeartbeatManager {
                                             ))
                                             .expect("Receive half of heartbeat cancels not blocked");
                                     }
+                                    // Caught client-side, so we fail the activity task ourselves (the
+                                    // server otherwise would have) and stop the activity with the
+                                    // same `Cancelled`/`cancel_requested` signal the server uses.
+                                    Err(e) if payload_limit_violation_from(&e).is_some() => {
+                                        let violation = payload_limit_violation_from(&e)
+                                            .expect("violation present per guard");
```
Sushisource: I want to say this can be combined with an if-let (forget if those work in match arms)

crates/sdk-core/src/worker/workflow/mod.rs
```diff
@@ -428,6 +433,26 @@ impl Workflows {
                                 .wf_task_failed();
                             return Err(e);
                         }
+                        // Client layer rejected the completion for exceeding the worker's payload
+                        // error limit; proactively fail the WFT instead.
+                        Err(e) if payload_limit_violation_from(&e).is_some() => {
+                            let violation = payload_limit_violation_from(&e)
```
Sushisource: Same here on if-let

crates/sdk-core/src/worker/activities.rs
```diff
@@ -409,10 +440,26 @@ impl WorkerActivityTasks {
...
+                            match client
                                 .cancel_activity_task(task_token.clone(), details)
                                 .await
-                                .err()
+                            {
+                                Ok(_) => None,
+                                Err(e) => {
+                                    if let Some(violation) = payload_limit_violation_from(&e) {
+                                        act_metrics.act_execution_failed();
+                                        client
```
Sushisource, line 444: Does this make sense? Should we instead just truncate the cancel details or something?

crates/sdk-core/src/worker/client.rs
```diff
@@ -57,7 +58,13 @@ pub enum LegacyQueryResult {
 
 /// Contains everything a worker needs to interact with the server
 pub(crate) struct WorkerClientBag {
+    /// Shared connection handle, used for management operations (capabilities, identity, client
+    /// replacement, etc.).
     connection: SharedReplaceableClient<Connection>,
+    /// Issues outbound gRPC calls, automatically attaching this worker's payload/memo error limits
+    /// (set via `set_payload_error_limits`) so the gRPC layer can enforce them. Wraps a clone of
+    /// `connection`, so a client replacement on `connection` is reflected here too.
+    client: PayloadLimitsClient<SharedReplaceableClient<Connection>>,
```
Sushisource, line 66: Per my bit about interceptors, I think ideally we wouldn't need to store a separate client here..

crates/sdk-core/src/worker/mod.rs
```diff
@@ -1212,8 +1233,9 @@ impl Worker {
     /// Tell the worker that an activity has finished executing. May (and should) be freely called
     /// concurrently.
     #[instrument(skip(self, completion),
-                 fields(task_token, status,
-                        task_queue=%self.config.task_queue, workflow_id, run_id))]
+                 fields(task_token, status, namespace=%self.config.namespace,
+                        task_queue=%self.config.task_queue, worker_id=%self.client.identity(),
+                        workflow_id, run_id, workflow_type, activity_type, attempt))]
```
Sushisource: Adding these additional attributes could be nice, could also be noisy, was this to match other SDKs?

crates/client/src/grpc.rs
jmaeagle99: There are a couple of issues that I think prevent this from being an interceptor:
- tonic interceptors receive type-erased request bodies. We need the typed body to decide if it should be validated and how. I do not think there is a technical workaround as an interceptor. It could be done as a tower layer, but that would require decoding the already encoded body, which seems wasteful.
- The connection is namespace unaware. I don't think we'd be able to construct the interceptor at channel creation time and be able to plumb through namespace limit information. We'd have to mutate the interceptor later but not for the client because that'll lead to inconsistent behavior depending on when a worker is attached to the channel. We could copy of the channel for each worker, but then we lose pooling/sharing.

We could eliminate this wrapper and just have the call sites attach the error limit information per caller; that would be ripe for forgetting to do that for new callers, but its workable and could be made with a simple convenience method.

If opposed to the client wrapper, I could do the per-call site attachment.

Sushisource: Ah, yeah, good points. I forgot that there isn't really a kind of interceptor that would allow us to have all that in one place. 

I suppose my only other comment is I don't think this needs to be fully `pub`?

crates/sdk-core/src/worker/activities/activity_heartbeat_manager.rs
jmaeagle99: if-let is not stable in our version of rust (1.94). I did some consolidation of the error match arms.

crates/client/src/grpc.rs
jmaeagle99: Has to be `pub` because it lives in `client` create (due to needing to impl  `RawClientProducer` and `RawGrpcCaller`) but is used by `sdk-core` crate. Could put it behind a `pub trait` + func, but still things are public and dynamic dispatch is involved; same amount of public surface but extra overhead.

I've hidden it from docs and cleaned up some of the `use` for it.

crates/sdk-core/src/worker/activities.rs
jmaeagle99: Server fails the activity as non-retryable. We're just making it retryable. So I think this makes sense. It's consistent with how we are treating all other too-large payload sizes.

Not sure we could truncate the cancel details without knowing how to make it valid when it gets deserialized. If we did proceed with the cancellation, I think we would have to have to clear or replace the details with some completely new.

I think it's better to fail the activity as retryable (consistent explanation of how we handle the problem) and adjust it in the future if need be.

crates/sdk-core/src/worker/mod.rs
jmaeagle99: More information to actually identify which thing has the too-large payloads failure from a logging perspective. Otherwise, custom is just going to know that something in the workflow is failing, not what within the workflow. I'll pull some samples of what this looks like.

I can back this part out and we can add back later if we feel that it provides better value.

crates/sdk-core/src/worker/activities/activity_heartbeat_manager.rs
```diff
@@ -189,7 +193,34 @@ impl ActivityHeartbeatManager {
...
+                                            if let Err(fe) = sg
+                                                .fail_activity_task(
+                                                    tt.clone(),
+                                                    Some(make_payloads_too_large_failure(violation)),
+                                                )
+                                                .await
+                                            {
+                                                warn!(task_token = %tt, error = ?fe,
+                                                    "Failed to fail activity after oversized heartbeat");
+                                            }
```
Sushisource, line 210: Hmm. I feel like it would be better if we could fail lang's heartbeat call rather than failing the activity attempt. Then the user can decide if they want to fail the activity or not.

But I acknowledge we're not set up super well to do that at the moment. In fact there's a very old TODO that would also be able to be fixed if we changed that: https://github.com/temporalio/sdk-core/blob/3c605ee7e4ba10f007f89de55b67862e0e9b9ba9/crates/sdk-core/src/worker/activities.rs#L445

It'd involve changing `Worker::record_activity_heartbeat` to be async and actually return an error that lang can react to, and then I suppose in order to make it work with the whole buffering concept we'd somehow need to decide if this particular call is going to be the one that actually sends the data, and if so, block and return the error if there was one.

That's a lot, obviously, and from what I can see wouldn't even be possible to do compatibly in some languages (ie: would require changes to their heartbeat calls). So I suppose after writing all that, we probably can't do it. Sorta sucks, but this is still better than what we had before of course so I suppose it'll just need to be this way.

Sushisource (approved):
Making sense to me overall. Rubber ducked myself on the one comment here.

crates/client/src/grpc.rs
jmaeagle99: Payload limit validation does occur before both the gRPC metrics are recorded and possible hand-off to the gRPC service override.

crates/sdk-core/src/worker/mod.rs
jmaeagle99: I've backed out the instrumentation changes.

PR #1369, opened 2026-07-01
Missing initializers; put EventGroupMarker.Label under NOT_VALIDATED_FIELDS

GregoryTravis:


crates/common/build.rs
```diff
@@ -825,6 +825,9 @@ const NOT_VALIDATED_FIELDS: &[&str] = &[
     "temporal.api.sdk.v1.UserMetadata.summary",
     "temporal.api.nexus.v1.EndpointSpec.description",
     "temporal.api.cloud.nexus.v1.EndpointSpec.description",
+    // Dedicated, non-fetchable limits (not blob/memo, not in DescribeNamespace): Event group marker
+    // label: custom 400-byte server-side limit.
```
chris-olszewski, line 829: Not sure I understand the first part of this comment. Is it just saying these limits aren't returned via any endpoint?

chris-olszewski (approved):
Approving since we're merging into another branch and this raises issues with that PR

crates/common/build.rs
GregoryTravis: Yes, these don't fall into the standard categories and the server does not tell us what the limits are for them.

PR #1370, opened 2026-07-01
Respect SDK flags in history & gzip downgrade

Sushisource:
## What was changed
Title

## Why?
If `GetSystemInfo` returns `UNIMPLEMENTED` for whatever reason (recent ex: gzip compression not working specifically for this endpoint), then we will fail to respect flags already written into history. Fix that.

Additionally allow downgrading to not using compression if a the GSI RPC fails on startup

## Checklist

1. Closes 

2. How was this tested:
Upgrade existing tests and added new ones

3. Any docs updates needed?

### DIFF
--- a/crates/sdk-core/tests/integ_tests/client_tests.rs
+++ b/crates/sdk-core/tests/integ_tests/client_tests.rs
@@ -1,6 +1,6 @@
 use crate::common::{
     CoreWfStarter, NAMESPACE,
-    fake_grpc_server::{GenericService, fake_server},
+    fake_grpc_server::{FakeServer, GenericService, fake_server},
     get_integ_server_options,
     http_proxy::HttpProxy,
 };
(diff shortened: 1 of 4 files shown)


tconley1428 (approved):

Sushisource:
@claude review

crates/sdk-core/tests/integ_tests/client_tests.rs
```diff
@@ -79,6 +80,175 @@ async fn calls_get_system_info() {
...
+        let behavior = behavior.clone();
+        async move {
+            match behavior {
+                CompressionTestBehavior::UnknownMethod => Status::unimplemented(
+                    "unknown method GetSystemInfo for service \
+                     temporal.api.workflowservice.v1.WorkflowService",
+                )
+                .into_http(),
+                CompressionTestBehavior::GenericGzipUnimplemented
+                    if path.ends_with("/GetSystemInfo") =>
```
chris-olszewski: Should this only be hit if `gzip` is being used?
```suggestion
                CompressionTestBehavior::GenericGzipUnimplemented
                    if path.ends_with("/GetSystemInfo") && encoding == "gzip" =>
```

chris-olszewski (approved):

PR #1372, opened 2026-07-05
feat(schedules): provide typed accessor to schedule action

chris-olszewski:
## What was changed
Add `ScheduleDescription::action` to access `ScheduleDescriptionAction` with `ScheduleDescriptionStartWorkflowAction`. `ScheduleDescriptionStartWorkflowAction::args` provides typed access to the workflow input.

## Why?
SDK parity, ergonomics.

## Checklist

1. Closes  N/A

2. How was this tested:
Added unit tests for unhappy path. Updated integration test to verify round trip.

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/client/src/schedules.rs
+++ b/crates/client/src/schedules.rs
@@ -526,12 +526,17 @@ impl ScheduleDescription {
     }
 
     /// The action configured on this schedule.
-    pub fn action(&self) -> Option<ScheduleDescriptionAction> {
-        self.raw
+    pub fn action(&self) -> ScheduleDescriptionAction {
+        let action = self
+            .raw
             .schedule
             .as_ref()
-            .and_then(|s| s.action.as_ref())
-            .and_then(|a| ScheduleDescriptionAction::from_proto(a, self.data_converter.clone()))
+            .expect("schedule description should contain schedule")
+            .action
+            .as_ref()
+            .expect("schedule description should contain action");
+        ScheduleDescriptionAction::from_proto(action, self.data_converter.clone())
+            .expect("schedule action should contain an action variant")
     }
 
     /// Whether the schedule is paused.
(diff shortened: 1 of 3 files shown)


crates/client/src/schedules.rs
```diff
@@ -394,6 +395,97 @@ pub struct ScheduleRunningAction {
...
+
+    /// The workflow ID configured on the schedule action.
+    pub fn workflow_id(&self) -> &str {
+        &self.workflow_id
+    }
+
+    /// Returns the workflow arguments deserialized as the requested type, if present.
+    pub async fn args<T: TemporalDeserializable + 'static>(
+        &self,
+    ) -> Result<Option<T>, PayloadConversionError> {
```
Sushisource, line 481: I'm realizing `PayloadConversionError` variants I didn't really give the best names... encoding errors are smooshed together with conversion errors. We might wanna fix that.

Unrelated to this PR though.

crates/client/src/schedules.rs
```diff
@@ -417,14 +509,36 @@ impl From<&schedule_proto::ScheduleActionResult> for ScheduleRecentAction {
...
+        }
+    }
+
     /// Token used for optimistic concurrency on updates.
     pub fn conflict_token(&self) -> &[u8] {
         &self.raw.conflict_token
     }
 
+    /// The action configured on this schedule.
+    pub fn action(&self) -> ScheduleDescriptionAction {
```
Sushisource, line 564: Maybe this should return `Option` to avoid these expects?

Sushisource (approved):

crates/client/src/schedules.rs
chris-olszewski: Is it valid for a schedule to not have an action? I originally had it return an `Option`, but wasn't sure if `None` had any meaning other than a malformed schedule. We assume it is always present with in TS: https://typescript.temporal.io/api/namespaces/client#scheduledescription

Sushisource: I think it probably is invalid, I was just thinking we should probably still avoid panicking in the event that it is malformed... but yeah maybe that doesn't justify asking users to unwrap everywhere.

I'm fine keeping it the way you have it now.

PR #1378, opened 2026-07-06
fix(otel): install subscriber on PeriodicReader thread

### LINKED ISSUE #1371: [Bug] OTLP metrics export errors are silently dropped (PeriodicReader thread has no tracing subscriber)
## what happened

if the OTLP metrics exporter (`telemetryOptions.metrics.otel` from a lang sdk, or `build_otlp_metric_exporter` directly) can't export - connection refused, TLS failure, http 4xx/5xx - no error shows up anywhere, at any log filter level. metrics just silently never arrive.

## why

core installs its tracing subscriber per-thread via `set_trace_subscriber_for_current_thread` (`crates/common/src/telemetry.rs`), which is a `tracing::subscriber::set_default` guarded in a `thread_local!`. the doc comment on that function even calls out the constraint:

> This must be done in every thread which uses core stuff, otherwise traces/logs will not be collected on that thread.

but `opentelemetry_sdk`'s `PeriodicReader` spawns its own OS thread internally. core has no hook to install the subscriber there, and there's no global default, so every `otel_error!` / `otel_debug!` the opentelemetry crates emit on the export path (they ARE emitted - `internal-logs` is a default feature of opentelemetry-otlp 0.31) dispatches to a nonexistent subscriber and gets dropped.

## repro (typescript sdk 1.16.1, opentelemetry 0.31)

```ts
Runtime.install({
  logger: new DefaultLogger("TRACE", (entry) => console.log(entry.level, entry.message, entry.meta)),
  telemetryOptions: {
    logging: { filter: "warn,opentelemetry=debug,opentelemetry_sdk=debug,opentelemetry-otlp=debug", forward: {} },
    metrics: {
      otel: {
        url: "http://127.0.0.1:1/v1/metrics", // guaranteed ECONNREFUSED
        http: true,
        metricsExportInterval: 3000,
      },
    },
  },
});
// ... Worker.create + run for 15s
```

construction-time events from `opentelemetry_sdk` (`MeterProvider.Building`, `Metrics.InstrumentCreated`) DO get forwarded - those fire on a core thread that has the subscriber. export-path events: nothing, ever. not for connection refused, not for TLS failures, not for auth errors.

## how we hit this

we debugged a production outage blind because of this. our container image was missing `ca-certificates`, so the exporter (reqwest + rustls-native-certs) had zero trust roots and aborted every TLS handshake to our OTLP gateway with a fatal alert. weeks of "metrics work locally, silently absent in prod" with no error in any log at any level. we only found it with tcpdump TLS-record dissection (ClientHello -> server Certificate flight -> 19-byte fatal alert, no request bytes ever sent). a single WARN line would have made this a five minute fix.

## possible fixes

1. wrap the built `MetricExporter` in a shim whose `export()` (which runs on the reader thread) re-enters the telemetry subscriber via `tracing::subscriber::with_default(...)` around the inner call. no global state, scoped exactly to the export. this seems like the cleanest option to me
2. set the telemetry subscriber as the global default when none exists (riskier, lang might own the global)
3. at minimum, document that metric export errors aren't collectable today so people don't treat log silence as export success

happy to share the full investigation notes / packet captures if useful

chris-olszewski:
## What was changed
Wrap the exporter with a `TracingMetricExporter` shim that ensures a subscriber is used on `PeriodicReader` thread calls to the exporter.

Also emit a warning if `export` errors.

## Why?
See attatched issue.

## Checklist

1. Closes #1371 

2. How was this tested:
Added small test to verify that we subscriber is installed on `PeriodicReader` thread and that we warn if the export fails.

3. Any docs updates needed?
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,3 +50,7 @@ to docs, or any other relevant information.
   `anyhow::Error`.
 
 ## [Unreleased]
+
+### Fixed
+* OTLP metric export failures are now logged through Core telemetry when OpenTelemetry's periodic
+  metric reader reports an export error.
--- a/crates/common/src/telemetry/otel.rs
+++ b/crates/common/src/telemetry/otel.rs
@@ -414,13 +414,10 @@ pub(crate) mod tests {
         metrics::{Temporality, data::ResourceMetrics, exporter::PushMetricExporter},
     };
     use std::{
-        future::Future,
-        pin::pin,
         sync::{
             Arc,
             atomic::{AtomicBool, Ordering},
         },
-        task::{Context, Poll},
         time::Duration,
     };
     use tracing_core::{
@@ -492,8 +489,8 @@ pub(crate) mod tests {
         assert_eq!(service_name, Some(Value::from(TELEM_SERVICE_NAME)));
     }
 
-    #[test]
-    pub(crate) fn traced_exporter_enters_subscriber_while_polling_export() {
+    #[tokio::test]
+    pub(crate) async fn traced_exporter_enters_subscriber_while_polling_export() {
         let saw_export_log = Arc::new(AtomicBool::new(false));
         let saw_wrapper_warn = Arc::new(AtomicBool::new(false));
         let subscriber = Arc::new(CapturingSubscriber {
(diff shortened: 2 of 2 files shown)


Sushisource (approved):
Nice

PR #1379, opened 2026-07-06
fix(client): correctly handle grpc-js style unimplemented method message

chris-olszewski:
## What was changed
Add "not implemented" as a valid variation of "is unimplemented"

## Why?
Typescript uses grpc-js to create mock servers to assert that all methods defined on services are available on `NativeConnection`. Latest core bump fails these tests now since the `GetSystemInfo` request that always has returned `UNIMPLEMENTED` before is now treated as a failure.

## Checklist

1. Closes N/A

2. How was this tested:
New test case

3. Any docs updates needed?
N/A


Sushisource (approved):
Lovely, lol.

PR #1384, opened 2026-07-07
chore: expand mise.toml with dev tools, add lefthook, clean up CI

### LINKED ISSUE #1383: chore: expand mise.toml with remaining dev tools, add lefthook, clean up CI
## Goal

Make the repo fully self-contained for local development via `mise install` — clone, install tools, build. No hunting for binaries.

## PRs

| PR | Scope | Risk | Status |
|---|---|---|---|
| #1386 | Add tools to mise.toml + remove redundant CI installs | Low | Open |
| #1387 | Extract composite action + selective mise installs | Medium (depends on #1386) | Open |
| #1388 | Lefthook, mise-bump workflow, README update | Zero (independent) | Open |

## Supersedes

- #1384 (closed — split into the 3 PRs above)

## Context

`mise.toml` was introduced in #1382 for `protoc`. This effort expands it to cover all external dev tools:
- `temporal` CLI (replaces `temporalio/setup-temporal` action)
- `cargo-component` (replaces `cargo install --locked cargo-component`)
- `cargo-msrv` (replaces `cargo install cargo-msrv`)

Additionally:
- Extracts a composite action to DRY up the 4-step CI setup pattern across 10 jobs
- Adds selective `mise install` per job (only install what each job needs)
- Adds opt-in git hooks via lefthook
- Adds weekly automated version bump workflow for mise-managed tools

brucearctor:
## Summary

Builds on #1382 to make the repo fully self-contained for local dev via `mise install`, clean up CI redundancy, and extract a reusable composite action.

### `mise.toml` — add 3 tools to existing protoc entry

| Tool | Version | Why |
|---|---|---|
| `temporal` | `1.7.2` | Pin — replaces `temporalio/setup-temporal` in CI; gives devs `temporal server start-dev` locally |
| `cargo:cargo-component` | `0.21.1` | Pin — affects WASM compilation |
| `cargo:cargo-msrv` | `0.19.3` | Pin — per 2-week minimum age policy |

### `.github/actions/setup/action.yml` — new composite action

Extracts the 4-step setup pattern repeated across all 9 CI jobs:
1. `dtolnay/rust-toolchain` (stable)
2. `jdx/mise-action` (with selective `install_args`)
3. `Set PROTOC` env var
4. `Swatinem/rust-cache` (with `cache-bin: false`)

Inputs: `mise-install-args` (selective tool install), `rust-cache-key`, `save-cache`.

### `per-pr.yml` — major cleanup (-101 lines, +20 lines)

- **Composite action**: All 9 jobs now use `uses: ./.github/actions/setup`
- **Selective mise installs**: 6/9 jobs install only `protoc` instead of all tools
  - `msrv` → `protoc cargo:cargo-msrv`
  - `wasm-workflow-tests` → `protoc cargo:cargo-component`
  - `examples` → `protoc temporal`
  - All others → `protoc` only
- **Remove `rustup component add rustfmt clippy`** (2 places) — already in `rust-toolchain.toml`
- **Remove `cargo install cargo-msrv`** — now via mise
- **Remove `cargo install --locked cargo-component`** — now via mise
- **Remove `temporalio/setup-temporal` action** — now via mise
- **Add `cache-bin: false` to WASM job** — was missing, now consistent via composite action
- **Remove `submodules: recursive`** (2 places) — no submodules exist (removed in #194)

### `lefthook.yml` — new (opt-in only)

- **pre-commit**: `cargo +nightly fmt --all -- --check`
- **pre-push**: `cargo lint` + `cargo check`

Not auto-installed. Developers opt in via `brew install lefthook && lefthook install`.

### `.github/workflows/mise-bump.yml` — new

Weekly scheduled workflow (Monday 9am UTC) that:
1. Runs `mise outdated` to check for new versions
2. Runs `mise up --bump` if updates exist
3. Opens a PR via `peter-evans/create-pull-request`

Complements existing dependabot (which does not support `mise.toml`).

### `README.md`

- Replace manual `protoc` install instructions with `mise install`
- Add optional lefthook section with opt-in instructions

## Developer experience after this PR

```bash
git clone https://github.com/temporalio/sdk-rust
cd sdk-rust
mise install      # installs protoc, temporal, cargo-component, cargo-msrv
cargo build       # just works
```

## CI impact

- **Faster**: 6/9 jobs skip installing temporal, cargo-component, cargo-msrv
- **Faster**: cargo-msrv and cargo-component installed as pre-built binaries via mise instead of `cargo install` from source
- **Simpler**: 9 jobs share one composite action instead of duplicating 4 setup steps each
- **Consistent**: `cache-bin: false` now applied to all jobs including WASM

Closes #1383

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,2 +1,5 @@
 [tools]
 protoc = "23.4"
+temporal = "1.7.2"
+"cargo:cargo-component" = "0.21.1"
+"cargo:cargo-msrv" = "0.19.3"
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,34 @@
+name: 'Setup Rust + mise'
+description: 'Common setup: checkout, Rust toolchain, mise tools, PROTOC env, Rust cache'
+
+inputs:
+  mise-install-args:
+    description: 'Tools to install (e.g. "protoc" or "protoc temporal"). Empty installs all.'
+    required: false
+    default: ''
+  rust-cache-key:
+    description: 'Extra key segment for Swatinem/rust-cache'
+    required: false
+    default: ''
+  save-cache:
+    description: 'Expression controlling when to save the Rust cache'
+    required: false
+    default: ${{ github.ref == 'refs/heads/main' }}
+
+runs:
+  using: composite
+  steps:
+    - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
+    - name: Run mise install
+      uses: jdx/mise-action@e6a8b3978addb5a52f2b4cd9d91eafa7f0ab959d # v4.2.0
+      with:
+        version: 2026.7.1
+        install_args: ${{ inputs.mise-install-args }}
+    - name: Set PROTOC
+      shell: bash
+      run: printf 'PROTOC=%s\n' "$(mise which protoc)" >> "$GITHUB_ENV"
+    - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
+      with:
+        cache-bin: false
+        save-if: ${{ inputs.save-cache }}
+        key: ${{ inputs.rust-cache-key }}
--- a/.github/workflows/heavy.yml
+++ b/.github/workflows/heavy.yml
@@ -21,14 +21,7 @@ jobs:
     runs-on: ubuntu-latest-16-cores
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: ./.github/actions/setup
         with:
-          submodules: recursive
-      - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
-      - name: Run mise install
-        uses: jdx/mise-action@e6a8b3978addb5a52f2b4cd9d91eafa7f0ab959d # v4.2.0
-        with:
-          version: 2026.7.1
-      - name: Set PROTOC
-        shell: bash
-        run: printf 'PROTOC=%s\n' "$(mise which protoc)" >> "$GITHUB_ENV"
+          mise-install-args: protoc
       - run: cargo integ-test -c "--release" -t heavy_tests -- --test-threads 1
--- a/.github/workflows/mise-bump.yml
+++ b/.github/workflows/mise-bump.yml
@@ -1,0 +1,53 @@
+name: Bump mise tool versions
+
+on:
+  schedule:
+    - cron: '0 9 * * 1' # Weekly Monday 9am UTC
+  workflow_dispatch: {}
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  bump:
+    name: Check for mise tool updates
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - name: Run mise install
+        uses: jdx/mise-action@e6a8b3978addb5a52f2b4cd9d91eafa7f0ab959d # v4.2.0
+        with:
+          version: 2026.7.1
+
+      - name: Check for outdated tools
+        id: outdated
+        run: |
+          # mise outdated returns non-zero if updates available
+          if mise outdated 2>/dev/null | grep -q .; then
+            echo "has_updates=true" >> "$GITHUB_OUTPUT"
+            echo "### Outdated mise tools" >> "$GITHUB_STEP_SUMMARY"
+            mise outdated >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "has_updates=false" >> "$GITHUB_OUTPUT"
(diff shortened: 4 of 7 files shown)


mise.toml
```diff
@@ -1,2 +1,10 @@
 [tools]
 protoc = "23.4"
+temporal = "1.7.2"
+lefthook = "1.11.5"
+"cargo:cargo-component" = "0.21.1"
+"cargo:cargo-msrv" = "latest"
+"cargo:cargo-zigbuild" = "latest"
```
brucearctor: I was OK with latest for these two as I don't expext those to matter.  BUT, open to pinning if we want pins as requirement in repo.

brucearctor:
I'm also unclear whether CI will work.  PATH should be OK.  But running will be the definiteve test.

We might want to then use `mise x -- ` to prefix anywhere there are path issues.  There are other optoins.  Let's see whether anything fails [ I can't kickoff CI, so will wait for whoever else to, in order to test ].

mise.toml
```diff
@@ -1,2 +1,10 @@
 [tools]
 protoc = "23.4"
+temporal = "1.7.2"
+lefthook = "1.11.5"
```
chris-olszewski: I would prefer not to force commit hooks on everyone
```suggestion
```

mise.toml
```diff
@@ -1,2 +1,10 @@
 [tools]
 protoc = "23.4"
+temporal = "1.7.2"
```
chris-olszewski, line 3: We need the `temporal` installation to be managed by `setup-temporal` as sometimes we need to target prerelease CLI versions to test behavior. The `mise` registry doesn't pick up those special releases.
```suggestion
```

mise.toml
chris-olszewski: Yeah, we'll need to pin these to adhere to our minimum age policy of 2 weeks. We should have had these pinned before.
Pinning `cargo-msrv` to `0.19.3`.

chris-olszewski: Assuming this is for the `docker-cgroups-test.sh`. That script isn't invoked in CI and is only used ocassionally. I would rather not install this by default in CI/local dev machines across the board.
```suggestion
```

chris-olszewski (changes requested):

mise.toml
brucearctor: reasonable policy, makes sense.

mise.toml
brucearctor: setup-temporal only works as a github action?

How do you get temporal version defined only in one place an then usable both in GH Actions and locally [ to prevent drift ].  

It sounds like mise not the right tool, if the use case isn't solved with that choice?  Ex: I've used nix for this.  Otherwise, I haven't looked at what/how to upstream a change to accomodate this use case in mise would look like?

Or, am I missing the point of why mise [ dependencies/versions works on my machine, works on your machine, works on remote machines -- all with a single config ].  ???

mise.toml
brucearctor: I'd imagine it worth determining a guided path to success and what helps people.

Otherwise, where is the automation, in repo, that people can opt-into?

There is a diff between forcing them to use, and making it easily available because it is the reccomended path?

brucearctor: Ex: I am struggling with this.  So could probably use clarity:

* Does the repo believe people should not use some form of local hooks [ precommit, prepush, etc ]?  
^ This seems crazy.  there is lots of benefit?  Or, is there some other way to ensure that these things happen locally before push??

* Does the repo believe people should self-manage their own hooks?  
^ Seems wild to not provide sensible defaults to make it easy for contributors [ that are easy to opt into, or probably better, for power users to auto-opt-out of if they prefer something different ]?  

There are ways to not 'force' hooks, by letting them opt out or custom-configure?  Rather than not including/offering a reasonable/sensible default?  can also set it up so that it is installed but still requires user to do something [ opt-in/install? ]

mise.toml
brucearctor: @chris-olszewski : Looked into this a bit more

mise can actually handle this via the github: backend, which pulls directly from GitHub Release tags — including prereleases:

```
[ tools ]
# standard release
"github:temporalio/cli" = "1.7.2"

# prerelease / custom tag — works too
"github:temporalio/cli" = "1.6.3-serverless"
```

The github: backend downloads assets from any tag on [temporalio/cli/releases](https://github.com/temporalio/cli/releases), so tags like v1.6.3-serverless and v1.7.2-one-time-versioning-override are all installable.

Does that address your concern, so can keep full mise [ if indeed mise the choice ]?

mise.toml
brucearctor: Updated a bit and added docs as the next suggestion.  Opt-in, rather than opt-out.

mise.toml
brucearctor: .github/workflows/mise-bump.yml

Wasn't clear whether a base GH Action was preferred.

Otherwise for keeping mise up to date, see renovate.  

https://docs.renovatebot.com/bot-comparison/
https://docs.renovatebot.com/modules/manager/mise/

Since similar to dependabot, which is already in repo, was unclear whether would want to switch, would want a second bot, or just code.  I opted to create the code - to avoid the additional dependency - but renovate would be a solid option.

brucearctor:
Updated body/description/summary.  Did wider cleanup for clarity/efficiency.

brucearctor:
Splitting into 3 focused PRs for cleaner review and independent mergeability. See the follow-up PRs linked from #1383.

PR #1391, opened 2026-07-09
Fix task slots used metric

### LINKED ISSUE #1489: [Bug] temporal_worker_task_slots_used (sum across worker types) never reaches 0 during inter-activity gaps

We're using KEDA to autoscale Temporal Python SDK workers to zero replicas when no workflows are executing. We use temporal_worker_task_slots_used (summed across all worker types) as the KEDA composite metric trigger, expecting it to be 0 when the worker is idle.

### Describe the bug

sum(temporal_worker_task_slots_used{task_queue="..."}) never drops to 0 even when no workflows are running

### Minimal Reproduction

Start a Python SDK worker with any workflow that has ≥2 sequential activities
Scrape temporal_worker_task_slots_used at 15s intervals
Observe: WorkflowWorker slots = 1 during WFT scheduling between activity 1 completing and activity 2 starting — even though ActivityWorker = 0 at that exact moment
Sum across all worker_type values: the sum is always ≥1 while any workflow is in progress, including the inter-activity scheduling gaps

### Environment/Versions

OS: Linux (Kubernetes, Python 3.13)
Temporal Server: self-hosted, version does not have worker_heartbeats capability (pre-1.27)
Python SDK: latest (temporalio with sdk-core at commit 2872b53)
Worker Deployment versioning: enabled (WorkerVersioningStrategy::WorkerDeploymentBased)
Deployment: Kubernetes with KEDA autoscaler, cooldownPeriod: 600s, activationTargetValue: 0

Sushisource:
## What was changed

  Fixed worker task slot metric accounting when releasing a permit. The drop-time adjustment is now applied only to `worker_task_slots_available`, rather than also incrementing `worker_task_slots_used`.

  Added an integration test covering an idle autoscaling activity worker with reserved polling slots.

  ## Why?

  Released permits were already removed from the extant permit count, but the metric update added one back to both available and used slots. This could leave `worker_task_slots_used` at 1 after all activity work completed, until another reservation updated the gauge.

  The available-slot adjustment is still necessary because the underlying RAII permit remains alive during its drop callback; it does not apply to used-slot accounting.

## Checklist

1. Closes https://github.com/temporalio/sdk-python/issues/1489

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/crates/sdk-core/tests/integ_tests/metrics_tests.rs
+++ b/crates/sdk-core/tests/integ_tests/metrics_tests.rs
@@ -12,7 +12,10 @@ use std::{
     collections::HashMap,
     env,
     string::ToString,
-    sync::{Arc, OnceLock},
+    sync::{
+        Arc, OnceLock,
+        atomic::{AtomicUsize, Ordering},
+    },
     time::Duration,
 };
 use temporalio_client::{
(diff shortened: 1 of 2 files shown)


crates/sdk-core/tests/integ_tests/metrics_tests.rs
```diff
@@ -394,6 +400,163 @@ async fn one_slot_worker_reports_available_slot() {
...
+        let body = get_text(format!("http://{addr}/metrics")).await;
+        let metric_line = body
+            .lines()
+            .find(|line| {
+                line.starts_with("temporal_worker_task_slots_used{")
+                    && line.contains(&format!("task_queue=\"{task_queue}\""))
+                    && line.contains("worker_type=\"ActivityWorker\"")
+            })
+            .expect("activity slots-used metric should exist")
+            .to_owned();
```
chris-olszewski: Should wrap this in an `eventually`

chris-olszewski (approved):

PR #1393, opened 2026-07-10
fix: use worker hearbeat interval for heartbeat system info poller

### LINKED ISSUE #1392: [Bug] Worker unconditionally runs a 100 ms sysinfo refresh thread, detrimental on Azure App Service
### Describe the bug

Every worker eagerly spawns a `RealSysInfo` background thread that refreshes host CPU/memory every 100 ms to feed worker heartbeat, regardless of tuner type or whether anything needs it at that cadence. This sampling is normally cheap and unnoticed. On Azure App Service, each refresh instead costs ~140 ms of CPU, so the thread runs back-to-back and burns ~50–60% of a CPU core on an otherwise idle worker.

### Environment/Versions

Azure App Service for Windows

#### Unconditional frequent sampling

`worker/mod.rs` creates a `RealSysInfo` for every worker:

```rust
let sys_info = sys_info.unwrap_or_else(|| Arc::new(RealSysInfo::new()));
```

`RealSysInfo::new()` spawns a thread refreshing every `REFRESH_INTERVAL = 100ms`. Its only consumer here is the worker heartbeat, which reads the cached value on a `tokio::time::interval(heartbeat_interval)` ticker. The default heartbeat internal is 60s. So it's sampled ~600× more often than consumed. (Resource-based workers get a second such thread.)

#### Detrimental on Azure App Service for Windows

On Windows, the refresh uses `sysinfo`'s PDH backend. In `sysinfo`, `windows/system.rs::refresh_cpu_specifics`, if the English counter "\Processor(_Total)\% Idle Time" fails to register (`key_used` stays `None`), it re-runs `Query::new(true)` (→ `PdhEnumObjectsW`) + `initialize_cpu_counters` (→ `PdhAddEnglishCounterW` per CPU) on every refresh, rebuilding the full perf-counter name table from the registry. On a normal Windows box, this registration succeeds and the refresh is microseconds; in the Azure App Service sandbox it fails/stalls, so every refresh pays ~140 ms rebuild. The 100 ms cadence then makes it continuous.

#### Behavior Changed in #1038 

Prior to #1038, `RealSysInfo` was created only by the resource-based tuner. After, it is unconditionally used and polled, even for fixed-size tuner (the default).

#### Evidence

CPU profile of an idle worker (no workflows) on Windows Azure App Service:

- ~28.5% of all-machine CPU in one idle worker; 97–98% of its stacks are `sdk_core_c_bridge → pdh.dll → registry/string-parsing`; managed code <2%.
- Self-time dominated by `PERF_MACHINE::BuildNameTable`/`SortStrings`; entry points `PdhAddEnglishCounterW` (72.8%), `PdhEnumObjectsW` (24.5%).
- Per `refresh_cpu_usage()` call: median 139 ms, mean 144 ms (p99 280 ms), total of ~250 calls throughout the 63.5 s trace. At ~140 ms/call and the 100 ms sleep interval → ~56% of a core sustained.

#### Potential Workarounds

Disable worker heartbeat (e.g. .NET `TemporalRuntimeOptions.WorkerHeartbeatInterval = null`), the sole consumer of `sys_info` when using the default tuner, so the thread is dropped. Note lowering/raising the interval doesn't help (sampler is fixed at 100 ms; interval capped 1–60 s).

jmaeagle99:
## What was changed

Fix worker heartbeat to sample host CPU/memory information at the heartbeat interval (if enabled).

## Why?

Host CPU/memory information doesn't need to be collected any faster than the heartbeat interval.

## Checklist

1. Closes #1392 

2. How was this tested: Will test end-to-end with .NET

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,3 +65,5 @@ to docs, or any other relevant information.
 ### Fixed
 * OTLP metric export failures are now logged through Core telemetry when OpenTelemetry's periodic
   metric reader reports an export error.
+* Worker heartbeat now samples host CPU/memory at the heartbeat interval (only when enabled) rather
+  than every 100ms.
--- a/crates/sdk-core/src/worker/mod.rs
+++ b/crates/sdk-core/src/worker/mod.rs
@@ -631,7 +631,6 @@ impl Worker {
             sys_info = tuner_builder.get_sys_info();
             Arc::new(tuner_builder.build())
         });
-        let sys_info = sys_info.unwrap_or_else(|| Arc::new(RealSysInfo::new()));
 
         metrics.worker_registered();
         let shutdown_token = CancellationToken::new();
@@ -861,6 +860,8 @@ impl Worker {
 
         let sdk_name_and_ver = client.sdk_name_and_version();
         let worker_heartbeat = worker_heartbeat_interval.map(|hb_interval| {
+            let heartbeat_sys_info =
+                sys_info.unwrap_or_else(|| Arc::new(RealSysInfo::new(hb_interval)));
             let hb_metrics = HeartbeatMetrics {
                 in_mem_metrics: metrics.in_memory_meter(),
                 wft_slots: wft_slots.clone(),
@@ -872,7 +873,7 @@ impl Worker {
                 act_last_suc_poll_time,
                 nexus_last_suc_poll_time,
                 status: worker_status.clone(),
-                sys_info,
+                sys_info: heartbeat_sys_info,
             };
             WorkerHeartbeatManager::new(
                 config.clone(),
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -65,3 +65,5 @@ to docs, or any other relevant information.
 ### Fixed
 * OTLP metric export failures are now logged through Core telemetry when OpenTelemetry's periodic
   metric reader reports an export error.
+* Worker heartbeat now samples host CPU/memory at the heartbeat interval (only when enabled) rather
+  than every 100ms.
```
Sushisource, line 69:
```suggestion
* Worker heartbeat now samples host CPU/memory at the heartbeat interval (only when enabled) rather
  than every 100ms. If resource tuner is also enabled, the sampling remains more frequent.
```

Sushisource (approved):

CHANGELOG.md
jmaeagle99: Discussed elsewhere. This is actually not the case, even before this change. If a resource-based tuner was installed, there would be two sys_info sources, one for the resource-based tuner and one for worker heartbeat. Checking how much work it would be to unify these.

PR #1395, opened 2026-07-11
feat(sdk): gate experimental APIs behind feature

chris-olszewski:
## What was changed
- Gate any API and required code marked as "experimental" behind a new `experimental` feature gate.
- Build docs with new feature enabled so APIs are discoverable
- Enable doc_cfg for docsrs so APIs that are gated behind a feature will get the little "feature" badge
- For fields that are on Bon builders we need to a little song and dance around making the bon setters crate public, and then conditionally compiling public methods that call these setters.

## Why?
Make it explicit at a language level what APIs are experimental compared to just a comment. Will allow us to do breaking changes on these APIs with a clear consciousness. 

## Checklist

1. Closes N/A
2. How was this tested:
Existing test suite, ensured we still check that crates compile without feature enabled in CI

3. Any docs updates needed?
Worth a mention on the Rust docs site since this is a different approach than other SDKs.

### DIFF
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -31,6 +31,7 @@ jobs:
       - run: cargo lint
       - run: cargo test-lint
       - run: cargo check
+      - run: cargo check --features experimental
       - run: git diff --exit-code
 
   test:
@@ -67,12 +68,12 @@ jobs:
         with:
           script: |
             core.exportVariable('RUSTFLAGS', '-Csymbol-mangling-version=v0');
-      - run: cargo test -- --include-ignored --nocapture
+      - run: cargo test --features experimental -- --include-ignored --nocapture
       - name: Find test executable for cgroup tests
         id: find-cgroup-test
         if: runner.os == 'Linux' && runner.arch == 'X64'
         run: |
-          test_executable=$(cargo build --tests --message-format json | jq -r 'select(.profile?.test == true and .target?.name == "temporalio_sdk_core" and .executable) | .executable')
+          test_executable=$(cargo build --tests --features experimental --message-format json | jq -r 'select(.profile?.test == true and .target?.name == "temporalio_sdk_core" and .executable) | .executable')
           cp $test_executable ./core-tests
       - name: Upload cgroup test executable
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
@@ -186,7 +187,7 @@ jobs:
       - uses: ./.github/actions/setup
         with:
           mise-install-args: protoc
-      - run: cargo test --features=test-utilities --test cloud_tests
+      - run: cargo test --features=test-utilities,temporalio-sdk/experimental --test cloud_tests
       - name: Generate Cloud test certificates
         run: |
           umask 077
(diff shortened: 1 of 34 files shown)


crates/client/src/options_structs.rs
```diff
@@ -292,6 +292,7 @@ pub struct WorkflowStartOptions {
 
     /// Optionally enable Eager Workflow Start, a latency optimization using local workers
     /// NOTE: Experimental
+    #[cfg(feature = "experimental")]
```
chris-olszewski: Is this really still experimental?

Sushisource: I don't think so

Sushisource:
Still kinda on the fence about if I want to do it this way with actual gates. Feels like it'll make it harder to get feedback on new stuff. Perhaps just experimental in the docstring is sufficient?

.github/workflows/per-pr.yml
```diff
@@ -67,12 +68,12 @@ jobs:
         with:
           script: |
             core.exportVariable('RUSTFLAGS', '-Csymbol-mangling-version=v0');
-      - run: cargo test -- --include-ignored --nocapture
+      - run: cargo test --features experimental -- --include-ignored --nocapture
```
chris-olszewski, line 71: Eventually, I want to also run tests w/o experimental just to have extra coverage on that codepath, but want to get CI in a better space before adding another full run blindly. Will be more important if we ever have a behavior addition with experimental instead of just these additive changes.

crates/sdk/src/interceptors.rs
```diff
@@ -46,126 +44,144 @@ mod activity_execution_value {
...
 
-/// Input to [`WorkerInterceptor::with_workflow_replay_worker`].
-#[derive(Debug)]
-#[non_exhaustive]
-pub struct WithWorkflowReplayWorkerInput<'a> {
-    /// The worker created for this replay operation.
-    pub worker: &'a mut Worker,
-}
+#[cfg_attr(not(feature = "experimental"), allow(unreachable_pub))]
+mod worker_lifecycle {
```
Sushisource, line 66: Should this be pub or re-exported?

crates/sdk/src/lib.rs
```diff
@@ -312,9 +331,70 @@ pub struct WorkerOptions {
...
+        builder(setters(
+            some_fn(name = patch_activation_callback_impl, vis = "pub(crate)"),
+            option_fn(name = maybe_patch_activation_callback_impl, vis = "pub(crate)")
+        ))
+    )]
     pub patch_activation_callback: Option<PatchActivationCallback>,
 }
 
+// Bon does not propagate `doc(cfg)` to generated setters, so these docs-only methods forward to
+// renamed generated implementations.
```
Sushisource, line 346: Oh man this is annoying. Maybe we can make a PR there to be able to un-do this?

Sushisource (approved):

crates/sdk/src/lib.rs
chris-olszewski: I'll look into that

crates/sdk/src/interceptors.rs
chris-olszewski: Gets lost in the diff, but we publicly reexport members it if experimental is enabled. If disabled we just `pub(crate)` export the members.

PR #1396, opened 2026-07-12
Fix workflow executor cooperative budget livelock

japz:
## What was changed

- Restored the Tokio cooperative-budget exemption around the workflow future at the narrow scope used before #1185.
- Added a regression test that resolves 256 pre-completed oneshots inside an unconstrained workflow while keeping the outer select and custom executor constrained.
- Added a changelog entry.

## Why?

PR #1185 replaced spawn_local with the custom WorkflowExecutor and inadvertently dropped the existing unconstrained wrapper around workflow futures.

WorkflowExecutor drains self-woken tasks synchronously. If a workflow resolves enough already-ready Tokio operations to exhaust the cooperative budget, the workflow returns Pending and self-wakes. The executor then immediately polls it again without returning to Tokio, so the budget is never replenished and the worker livelocks.

Restoring the previous narrow exemption lets the workflow drain ready resolutions without the cooperative cutoff while retaining the custom executor, its wake tracking, and nondeterministic-future detection.

## Checklist

1. How was this tested:

- cargo test -p temporalio-sdk
- timeout 180 cargo integ-test nondeterministic_future_detection_fails_wft
- cargo lint
- cargo test-lint
- cargo +nightly fmt
- git diff --check

2. Any docs updates needed?

No documentation changes. Added an Unreleased changelog entry.

### DIFF
--- a/crates/sdk/src/workflow_executor.rs
+++ b/crates/sdk/src/workflow_executor.rs
@@ -331,7 +331,6 @@ mod tests {
             .await;
     }
 
-
     #[test]
     fn sdk_wake_guard_nesting() {
         assert!(!is_sdk_wake());
--- a/crates/sdk/src/workflow_executor.rs
+++ b/crates/sdk/src/workflow_executor.rs
@@ -331,6 +331,39 @@ mod tests {
             .await;
     }
 
+    #[tokio::test]
+    async fn executor_drains_unconstrained_future_past_cooperative_budget() {
+        let local = tokio::task::LocalSet::new();
+        local
+            .run_until(async {
+                let executor = WorkflowExecutor::new();
+                let receivers = (0..256)
+                    .map(|_| {
+                        let (tx, rx) = oneshot::channel();
+                        tx.send(()).unwrap();
+                        rx
+                    })
+                    .collect::<Vec<_>>();
+                let workflow = async move {
+                    for receiver in receivers {
+                        receiver.await.unwrap();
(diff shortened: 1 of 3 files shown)


crates/sdk/src/workflow_executor.rs
```diff
@@ -331,6 +331,39 @@ mod tests {
...
+                        _ = std::future::pending::<()>() => {}
+                    }
+                });
+
+                executor.shutdown().await;
+
+                handle.await.unwrap();
+            })
+            .await;
+    }
```
chris-olszewski: This test doesn't assert anything about how we spawn and drive workflow futures forward. It can be removed.
```suggestion
```

chris-olszewski (changes requested):
The change is sound, but I am curious how you found this. Did you encounter this in a real workflow?

japz:
> The change is sound, but I am curious how you found this. Did you encounter this in a real workflow?

Yes, I am building an astronomy data processing pipeline on top of Temporal, and I ran into this when I attempted to process a large batch of data.

crates/sdk/src/workflow_executor.rs
```diff
@@ -331,6 +331,7 @@ mod tests {
             .await;
     }
 
+
```
chris-olszewski:
```suggestion
```

chris-olszewski (approved):

PR #1398, opened 2026-07-13
feat(sdk): remove remaining raw protos in APIs

### LINKED ISSUE #1403: [Feature Request] Remove raw protobuf messages in public APIs

Currently we leak raw protobuf messages in multiple public APIs where it is not necessary.

Provide ergonomic Rust types instead.

chris-olszewski:
## What was changed
See changelog changes. TL;DR removed most* raw protos in the APIs.

This PR was generated by:
 - first building up a list of public fields/methods outside of `temporalio-protos`
 - filtered out methods with `raw` as these indicated protos were expected
 - filtered out those that took `Payloads` in an expected manner e.g. headers or payload converters
 - a few methods where we explicitly surface protos e.g. workflow history events

Review each commit individually.

* There are 2 more on the list that should be addressed: `WorkflowCountAggregationGroup` and `start_nexus_operation`. Both have some design work that deserves independent PR.

## Why?
In all other SDKs we prefer custom types over the raw ones that translate directly to protobuf messages. This PR does the same for the remaining Rust API surfaces. These changes were grouped in a single PR as many of the changes are purely mechanical. If we feel these should be independent PRs I can break it up.

## Checklist

1. Closes https://github.com/temporalio/sdk-rust/issues/1403
2. How was this tested:
Added tests.

3. Any docs updates needed?
Yes, will have to update a few usages in the docsite where we were referencing raw protos.

### DIFF
--- a/crates/client/src/async_activity_handle.rs
+++ b/crates/client/src/async_activity_handle.rs
@@ -1,17 +1,21 @@
 //! Handle for completing activities asynchronously via a client.
 
 use crate::{NamespacedClient, errors::AsyncActivityError, grpc::WorkflowService};
-use temporalio_common::protos::{
-    TaskToken,
-    temporal::api::{
-        common::v1::Payloads,
-        failure::v1::Failure,
-        workflowservice::v1::{
-            RecordActivityTaskHeartbeatByIdRequest, RecordActivityTaskHeartbeatByIdResponse,
-            RecordActivityTaskHeartbeatRequest, RecordActivityTaskHeartbeatResponse,
-            RespondActivityTaskCanceledByIdRequest, RespondActivityTaskCanceledRequest,
-            RespondActivityTaskCompletedByIdRequest, RespondActivityTaskCompletedRequest,
-            RespondActivityTaskFailedByIdRequest, RespondActivityTaskFailedRequest,
+use temporalio_common::{
+    data_converters::{SerializationContextData, TemporalSerializable},
+    error::{ApplicationFailure, OutgoingActivityError, OutgoingError},
+    payload_visitor::encode_payloads,
+    protos::{
+        TaskToken,
+        temporal::api::{
+            common::v1::Payloads,
+            workflowservice::v1::{
+                RecordActivityTaskHeartbeatByIdRequest, RecordActivityTaskHeartbeatByIdResponse,
+                RecordActivityTaskHeartbeatRequest, RecordActivityTaskHeartbeatResponse,
+                RespondActivityTaskCanceledByIdRequest, RespondActivityTaskCanceledRequest,
+                RespondActivityTaskCompletedByIdRequest, RespondActivityTaskCompletedRequest,
+                RespondActivityTaskFailedByIdRequest, RespondActivityTaskFailedRequest,
+            },
         },
     },
 };
@@ -79,7 +83,21 @@ impl<CT> AsyncActivityHandle<CT> {
 
 impl<CT: WorkflowService + NamespacedClient + Clone> AsyncActivityHandle<CT> {
     /// Complete the activity with a successful result.
-    pub async fn complete(&self, result: Option<Payloads>) -> Result<(), AsyncActivityError> {
+    pub async fn complete<T>(&self, result: Option<T>) -> Result<(), AsyncActivityError>
+    where
+        T: TemporalSerializable + 'static,
+    {
+        let result = match result {
+            Some(result) => Some(Payloads {
+                payloads: vec![
+                    self.client
+                        .data_converter()
+                        .to_payload(&SerializationContextData::Activity, &result)
+                        .await?,
+                ],
+            }),
+            None => None,
+        };
         match &self.identifier {
             ActivityIdentifier::TaskToken(token) => {
                 WorkflowService::respond_activity_task_completed(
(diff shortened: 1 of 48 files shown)


crates/client/src/workflow_handle.rs
```diff
@@ -76,23 +76,63 @@ fn decode_user_metadata(
...
+            payloads: DecodablePayloads::new(
+                payloads,
+                data_converter.payload_converter().clone(),
+                SerializationContextData::Workflow,
+            ),
+        }
+    }
+
+    /// Deserialize the details into a typed value using the client's payload converter.
+    pub fn deserialize<T: temporalio_common::data_converters::TemporalDeserializable + 'static>(
```
chris-olszewski, line 102: I don't love this name, but it is an honest name and aligns with `serde` vocabulary. Open to other suggestions.

crates/workflow/src/workflow_context.rs
```diff
@@ -1035,16 +1059,44 @@ impl<W> SyncWorkflowContext<W> {
         );
     }
 
-    /// Add or create a set of search attributes
-    pub fn upsert_memo(&self, attr_iter: impl IntoIterator<Item = (String, Payload)>) {
+    /// Add or replace memo values with `Some`; remove memo keys with `None`.
+    pub fn upsert_memo<K>(
+        &self,
+        updates: impl IntoIterator<Item = (K, Option<MemoValue>)>,
```
chris-olszewski, line 931: Mirroring Python API of `("key", None)` indicating a deletion.

crates/client/src/lib.rs
```diff
@@ -821,21 +821,6 @@ pub enum Namespace {
     Id(String),
 }
 
-impl Namespace {
-    /// Convert into grpc request
-    pub fn into_describe_namespace_request(self) -> DescribeNamespaceRequest {
```
chris-olszewski, line 822: I don't think this is something users would want to do and was just present as a test helper

crates/common/src/worker.rs
```diff
@@ -115,21 +113,6 @@ impl WorkerTaskTypes {
             || (self.enable_remote_activities && other.enable_remote_activities)
             || (self.enable_nexus && other.enable_nexus)
     }
-
-    /// Converts the enabled task types into the corresponding [`TaskQueueType`] values.
-    pub fn to_task_queue_types(&self) -> Vec<TaskQueueType> {
```
chris-olszewski, line 85: Test helper that doesn't need to be a public facing API

crates/client/src/async_activity_handle.rs
```diff
@@ -267,6 +308,25 @@ impl<CT: WorkflowService + NamespacedClient + Clone> AsyncActivityHandle<CT> {
             }
         }
     }
+
+    async fn encode_optional_payloads<T>(
```
Sushisource: Seems like this should go somewhere more generic?

crates/client/src/schedules.rs
```diff
@@ -1328,6 +1371,43 @@ mod tests {
     };
     use tonic::{Request, Response};
 
+    struct XorCodec;
+
+    impl PayloadCodec for XorCodec {
```
Sushisource: This is a copy of the one here? https://github.com/temporalio/sdk-rust/pull/1398/changes#diff-7556b09f7a3ac4aee6cf768974fbf5bd86a399b8efb4909d1dcdf89a9b68b297R1759

crates/client/src/workflow_handle.rs
Sushisource: Seems reasonable

crates/client/src/workflow_handle.rs
```diff
@@ -1024,14 +1083,117 @@ where
...
+            common::v1::{Memo, SearchAttributes},
+            enums::v1::WorkflowExecutionStatus as ProtoWorkflowExecutionStatus,
+            sdk::v1::UserMetadata,
+            workflow::v1::WorkflowExecutionConfig,
+        },
     };
 
+    struct XorCodec;
+
+    impl PayloadCodec for XorCodec {
```
Sushisource: More copies of this codec lol

crates/common-wasm/src/retry_policy.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub non_retryable_error_types: Vec<String>,
+    #[builder(skip = ProtoRetryPolicy {
+        initial_interval: initial_interval.try_into().ok(),
+        backoff_coefficient,
+        maximum_interval: maximum_interval.and_then(|duration| duration.try_into().ok()),
+        maximum_attempts,
+        non_retryable_error_types: non_retryable_error_types.clone(),
+    })]
+    raw: ProtoRetryPolicy,
+}
```
Sushisource, line 23: I'm not sure I like storing the raw and un-raw versions at the same time.

Maybe we just only store raw and provide everything else as builders/accessors?

crates/common-wasm/src/workflow_execution.rs
```diff
@@ -0,0 +1,88 @@
...
+pub struct WorkflowExecution {
+    /// The workflow ID.
+    pub workflow_id: String,
+    /// The run ID, or an empty string when targeting the latest run.
+    pub run_id: String,
+    #[builder(skip = ProtoWorkflowExecution {
+        workflow_id: workflow_id.clone(),
+        run_id: run_id.clone(),
+    })]
+    raw: ProtoWorkflowExecution,
```
Sushisource, line 7: Same comment about duplicative storage

crates/workflow/src/memo.rs
```diff
@@ -0,0 +1,124 @@
...
+        )
+    }
+}
+
+/// A typed value used in a workflow memo update.
+#[derive(Clone, derive_more::Debug)]
+#[non_exhaustive]
+pub struct MemoValue {
+    #[debug(skip)]
+    value: Rc<dyn SerializableMemoValue>,
```
Sushisource, line 41: Seemingly can just be a box?

crates/workflow/src/workflow_context.rs
```diff
@@ -309,14 +317,16 @@ pub struct WorkflowContextView {
...
     pub cron_schedule: Option<String>,
-    /// User-defined memo
-    pub memo: Option<Memo>,
+    /// Priority and fairness configuration for this workflow execution.
+    pub priority: Priority,
+    /// User-defined memo values.
+    pub memo: Memo,
     /// Initial search attributes as a typed collection.
     pub search_attributes: Option<SearchAttributes>,
+    raw: InitializeWorkflow,
```
Sushisource: Same as other comments but maybe it's not worth fixing here since we'd have to accessor-ify a million things. I'd be OK with just being a bit duplicative here.

crates/workflow/src/memo.rs
chris-olszewski: It allows us to offer up `Clone` without adding that bound to the underlying value, but could easily see requiring a `Clone` bound for a `MemoValue`.

Sushisource: Ah, nm I just didn't see the clone.

crates/client/src/async_activity_handle.rs
chris-olszewski: Ended up inlining this helper, not sure making this a public API in `common` is warranted

crates/workflow/src/workflow_context.rs
chris-olszewski: Since this is read only it seemed reasonable

Sushisource (approved):
noice

PR #1401, opened 2026-07-14
feat(sdk): workflow interceptors

### LINKED ISSUE #1139: [Feature Request] Workflow interceptors

chris-olszewski:
## What was changed
- Workflow implementation no longer handles the conversion of outputs to payloads and instead returns `WorkflowOutputValue` that allows interceptors to inspect/modify outputs before serialization.
- `continue_as_new` no longer takes reference to input since interceptors might modify/swap the input. In retrospect taking a reference was also awkward IMO.
- Add `WorkflowInterceptor` trait and update workflow machinery to use interceptors.
- `add_workflow_interceptor_factory`/`WorkflowInterceptorFactory` which constructs an named array of workflow interceptors.
- `WorkflowInterceptorFuture` a named future that is the expected future to be produced by interceptor futures. We could expose `LocalBoxFuture` directly instead if we're willing to expose that 3rd party type as an external API type.

### `WorkflowInterceptor`
Interceptors look like the following:
```
fn handle_signal<'a>(
        &'a self,
        ctx: WorkflowInterceptorContext,
        input: HandleSignalInput,
        next: WorkflowNext<
            'a,
            HandleSignalInput,
            WorkflowInterceptorFuture<'a, HandleSignalResult>,
        >,
    ) -> WorkflowInterceptorFuture<'a, HandleSignalResult> {
        next.run(input)
    }
```
They take a context with a subset of functionality of a workflow context, input for the operation and a `next` continuation that calls the next interceptor in the chain. They are always sync forcing users to do any async work in a `WorkflowInterceptorFuture`.

## Why?

The changes to the workflow machinery and the "construction" polling are to allow the interceptor chains to not delay command creation/handler execution if there isn't any async work. This was especially apparent with sync handlers that were eagerly executed previously. Without this special handling, the existence of an interceptor chain would at least one yield point before the execution of the handler/command creation. The construction poll is our way of ensuring we drive the chain as forward as possible.

An alternative for this would be to make the interceptor API where `Next` would have some explicit `.then` methods to remove any yield points. I felt this diverged too much from all other SDKs and was generally pretty awkward to work with.

A few other decisions
- We have a singular workflow interceptor trait over 2 separate ones to make registration a singular method instead of spliting inbound/outbound.
- WASM interceptors have to be defined inside the component

## Checklist

1. Closes #1139 

2. How was this tested:
See integration tests.

3. Any docs updates needed?
Once released should add interceptors page for the Rust SDK docs.


crates/sdk-core/tests/wasm_workflow_tests.rs
```diff
@@ -174,7 +174,9 @@ async fn build_wasm_hello_component() -> PathBuf {
             "--release",
             "--target",
             "wasm32-unknown-unknown",
+            "--target-dir",
```
chris-olszewski, line 413: Not strictly necessary, but I have a custom target dir setup on my machine and allows me to run WASM tests locally without overriding it.

crates/workflow/src/runtime/entry.rs
```diff
@@ -261,17 +260,16 @@ pub trait ExecutableAsyncUpdate<U: UpdateDefinition>: WorkflowImplementation {
...
+        let future = async move {
             let output = Self::handle(ctx, input)
                 .await
                 .map_err(WorkflowError::Execution)?;
-            serialize_output(&output, &converter)
-        }
-        .boxed_local()
+            Ok(Box::new(output) as Box<dyn WorkflowOutputValue>)
+        };
+        ConstructionBlockedFuture::new(base_ctx, future.boxed_local()).boxed_local()
```
chris-olszewski: We want to ensure that if there is a sync handler with no interceptors, it is driven to completed before starting to poll an async handler.

crates/sdk/examples/wasm_workflows/src/lib.rs
```diff
@@ -29,6 +35,32 @@ impl HelloWorkflow {
     }
 }
 
+struct WasmWorkflowInterceptor;
```
Sushisource: I'm not entirely opposed to putting this here, but, maybe follow what I did in https://github.com/temporalio/sdk-rust/pull/1394 by adding this as a separate fixture?

crates/workflow/src/runtime/instance.rs
```diff
@@ -397,19 +762,22 @@ where
...
+                };
+            }
+
+            let poll = future.poll_unpin(cx);
+            if let Some(failure) = Self::take_forced_wft_failure(base_ctx) {
                 return RoutinePollState::ForcedFailure {
                     failure,
                     made_progress,
                 };
             }
```
Sushisource, line 808: Did this fix a different bug you found while working on this?

crates/workflow/src/runtime/mod.rs
```diff
@@ -58,3 +59,28 @@ impl<F: Future + Unpin> Future for SdkGuardedFuture<F> {
         Pin::new(&mut self.0).poll(cx)
     }
 }
+
+pub(crate) struct ConstructionBlockedFuture<F> {
```
Sushisource: Some comments explaining the purpose of this guy would be helpful, something like what you have in you PR description. Probably here, as well as where it's actually used in some meaningful spots.

crates/workflow/src/workflow_context.rs
```diff
@@ -99,6 +167,113 @@ impl BaseWorkflowContext {
...
+            .inner
+            .poll_phase
+            .replace(WorkflowPollPhase::Construction);
+        ConstructionPollGuard {
+            phase: &self.inner.poll_phase,
+            previous,
+        }
+    }
+
+    pub(crate) fn is_construction_poll(&self) -> bool {
```
Sushisource:
```suggestion
    pub(crate) fn in_construction_phase(&self) -> bool {
```

crates/workflow/src/runtime/mod.rs
```diff
@@ -58,3 +59,28 @@ impl<F: Future + Unpin> Future for SdkGuardedFuture<F> {
...
+    }
+}
+
+impl<F: Future + Unpin> Future for ConstructionBlockedFuture<F> {
+    type Output = F::Output;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        if self.base_ctx.is_construction_poll() {
+            // Every newly started routine receives a normal poll after activation, so this barrier
+            // does not need to register the construction waker with the async handler.
```
Sushisource: I can't really understand this comment in isolation 😆

crates/workflow/src/workflow_context.rs
```diff
@@ -82,10 +99,61 @@ use temporalio_common_wasm::{
 pub struct BaseWorkflowContext {
     inner: Rc<WorkflowContextInner>,
 }
+
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+enum WorkflowPollPhase {
```
Sushisource:
```suggestion
enum WorkflowFuturePollPhase {
```

?

Sushisource: Some comments on the variants would be good to I think to help reinforce the other suggested comments

crates/workflow/src/workflow_context.rs
```diff
@@ -545,28 +758,105 @@ impl BaseWorkflowContext {
...
+        let future = call_start_timer(
+            interceptors,
+            WorkflowInterceptorContext::new(self.clone()),
+            input,
+            next,
+        );
+        self.prepare_cancellable_outbound_future(future)
+    }
+
+    fn start_timer_unintercepted(
```
Sushisource: Can be inlined? Looks like it may apply to all these?

crates/workflow/src/workflow_context.rs
```diff
@@ -808,6 +1403,11 @@ impl<W> SyncWorkflowContext<W> {
         self.base.inner.shared.borrow().activation.is_replaying
     }
 
+    /// Returns true if the current work is replaying history events
```
Sushisource, line 1342: "current work" here is unusual language. I think just "if history events are currently being replayed" would work, or can refer to other SDKs

crates/workflow/src/workflow_context.rs
```diff
@@ -1858,13 +2494,54 @@ pub(crate) struct PendingChildWorkflow<WD: WorkflowDefinition> {
     pub(crate) _phantom: PhantomData<WD>,
 }
 
+/// Output produced when an intercepted child workflow successfully starts.
+#[derive(derive_more::Debug)]
+pub struct StartChildWorkflowOutput {
```
Sushisource, line 2578: Probably needs `non_exhaustive`

(as well as, I'm sure, a bunch of other stuff - maybe we wanna do that all in one PR)

crates/workflow/src/workflow_context.rs
```diff
@@ -2243,103 +2922,105 @@ impl ExternalWorkflowHandle {
...
-    schedule_seq: u32,
-    base_ctx: BaseWorkflowContext,
+    pub(crate) result_future: Shared<WFCommandFut<NexusOperationResult, ()>>,
+    pub(crate) schedule_seq: u32,
+    pub(crate) base_ctx: BaseWorkflowContext,
+}
+
+pub(crate) enum NexusResultFuture {
+    Raw(Shared<WFCommandFut<NexusOperationResult, ()>>),
+    Intercepted(Shared<LocalBoxFuture<'static, NexusOperationResult>>),
```
Sushisource: Is an intercepted one ever constructed? Why do we distinguish here at all when we don't for other commands?

crates/workflow/src/workflow_interceptors.rs
```diff
@@ -0,0 +1,1597 @@
...
+
+/// Result of an intercepted query handler.
+pub type HandleQueryResult = Result<Box<dyn WorkflowOutputValue>, WorkflowError>;
+
+/// Result of an intercepted update validator.
+pub type ValidateUpdateResult = Result<(), WorkflowError>;
+
+/// Future produced by workflow interceptors.
+///
+/// The SDK polls a newly created interceptor future once while processing its activation. This
```
Sushisource, line 206: This is useful but I'm not sure this type is the most discoverable place for this docstring.

IMO this, and any other user-consequential stuff should go on `WorkflowInterceptor` itself, or the module-level docs, or some combination of both.

crates/workflow/src/workflow_interceptors.rs
```diff
@@ -0,0 +1,1597 @@
...
+        self.base.search_attributes()
+    }
+
+    /// Returns true if the current workflow task is happening under replay.
+    pub fn is_replaying(&self) -> bool {
+        self.base.is_replaying()
+    }
+
+    /// Returns true if the current work is replaying history events.
+    pub fn is_replaying_history_events(&self) -> bool {
```
Sushisource: I think we need to more finely delineate the difference between these two

crates/workflow/src/workflow_interceptors.rs
```diff
@@ -0,0 +1,1597 @@
...
+    }
+
+    fn into_parts(self) -> (Option<Box<dyn Any>>, HashMap<String, Payload>) {
+        (self.value, self.headers)
+    }
+}
+
+/// Input passed to [`WorkflowInterceptor::initialize_workflow`].
+///
+/// The decoded input is present when the workflow's `#[init]` method accepts the workflow start
```
Sushisource: Should have the mirror of:

```
Inputs consumed by `#[init]` are instead passed to
/// [`WorkflowInterceptor::initialize_workflow`].
```

crates/workflow/src/workflow_interceptors.rs
```diff
@@ -0,0 +1,1597 @@
...
+        Self::new(|_| {})
+    }
+
+    /// Cancel without a reason.
+    pub fn cancel(&self) {
+        (self.cancel)(None);
+    }
+
+    /// Cancel with a reason.
+    pub fn cancel_with_reason(&self, reason: String) {
```
Sushisource: Why not just have one with optional reason?

crates/workflow/src/workflow_interceptors.rs
```diff
@@ -0,0 +1,1597 @@
...
+    }
+
+    #[doc(hidden)]
+    pub fn into_inner(self) -> Vec<Arc<dyn WorkflowInterceptor>> {
+        self.interceptors
+    }
+}
+
+/// Creates a fresh interceptor collection for each workflow instance.
+pub trait WorkflowInterceptorFactory: Send + Sync + 'static {
```
Sushisource: Not totally sure I follow the purpose of this trait? Why don't we just accept the vec of them directly? If this took some param that provided context about the workflow I would get it since then you could alter you interception chain per workflow but we don't (and shouldn't, since we don't in other SDKs).

crates/workflow/src/workflow_context.rs
```diff
@@ -614,17 +970,21 @@ impl BaseWorkflowContext {
...
-                return ActivityFut::eager(e.into());
+                return CancellableWorkflowOutboundFuture::new(
+                    ActivityFut::<future::Ready<ActivityResolution>, AD::Output>::eager(e.into()),
+                    WorkflowCancellationHandle::noop(),
+                );
             }
         };
-        ActivityFut::running(
-            LATimerBackoffFut::new(activity.name().to_string(), payloads, opts, self.clone()),
+        let future = LATimerBackoffFut::new(activity_type, payloads, headers, opts, self.clone());
```
Sushisource: I think we may need to special case this guy so the timers it creates are always unintercepted? Otherwise someone can blow up that behavior. OTOH they can blow up all sorts of shit already so maybe that's not an issue but we probably need some way for them to know it's a "built-in" timer. Not sure what other SDKs do here.

crates/workflow/src/workflow_interceptors.rs
```diff
@@ -0,0 +1,1597 @@
...
+
+impl<T> CancellableFuture<T> for CancellableWorkflowOutboundFuture<T> {
+    fn cancel(&self) {
+        self.cancellation.cancel();
+    }
+}
+
+impl<T> CancellableFutureWithReason<T> for CancellableWorkflowOutboundFuture<T> {
+    fn cancel_with_reason(&self, reason: String) {
+        self.cancellation.cancel_with_reason(reason);
```
Sushisource: I think this needs to avoid sending the cancel if the future is already terminated?

crates/workflow/src/workflow_context.rs
```diff
@@ -463,6 +656,15 @@ impl BaseWorkflowContext {
     pub(crate) fn take_forced_wft_failure(
         &self,
     ) -> Option<Box<dyn std::error::Error + Send + Sync>> {
+        if self
```
Sushisource: Needs to check if nondeterministic future checking is enabled or not

crates/workflow/src/workflow_interceptors.rs
```diff
@@ -0,0 +1,1597 @@
...
+#[non_exhaustive]
+pub struct ContinueAsNewInput {
+    decoded: DecodedInput,
+    options: ContinueAsNewOptions,
+}
+
+impl ContinueAsNewInput {
+    pub(crate) fn new(input: Box<dyn Any>, options: ContinueAsNewOptions) -> Self {
+        Self {
+            decoded: DecodedInput::new(Some(input), HashMap::new()),
```
Sushisource, line 1270: How can user specify headers? Needs to be extracted from options

crates/sdk/src/workflow_wasm.rs
```diff
@@ -186,6 +186,9 @@ impl CompiledWasmWorkflowModule {
         &self,
         input: WorkflowExecutionInput,
     ) -> Result<Box<dyn WorkflowInstance>, anyhow::Error> {
+        if !input.workflow_interceptor_factories.is_empty() {
+            bail!("Native workflow interceptors cannot be used with WASM workflow components");
+        }
```
Sushisource: This seems like the wrong place to fail - we will fail every time we try to execute a workflow defined in WASM rather than just failing on worker startup, which seemingly we could do and would be faster.

However, I'm not sure why this needs to be an error. If a user defines interceptors in the WASM bundle, I think we can use those and just document that native ones do not apply to WASM workflows?

I'm also not 100% clear on _why_ native ones can't be applied to WASM workflows. I guess just extra back-and-forth glue? (To be clear, I don't object, just want to clarify).

crates/workflow/src/workflow_context.rs
```diff
@@ -441,10 +619,25 @@ impl BaseWorkflowContext {
...
                 data_converter,
                 state_mutated: Cell::new(false),
+                poll_phase: Cell::new(WorkflowPollPhase::Routine),
+                construction_non_sdk_wake: Arc::new(AtomicBool::new(false)),
+                workflow_interceptors: RefCell::new(Vec::new()),
             }),
         }
     }
 
+    pub(crate) fn set_workflow_interceptors(
```
Sushisource: I think this could be eliminated if we construct the context in a slightly smarter order. That would also allow us to just use an `Rc<[Arc<dyn WorkflowInterceptor>]>` which would avoid needing to clone the Vec repeatedly on every call which is a good savings.

Sushisource (commented):
Woof that was a doozy

crates/workflow/src/workflow_interceptors.rs
chris-olszewski: Removed the headers field from continue as new opts so headers must be set in an interceptor now. Aligns with TS/.NET.

crates/workflow/src/workflow_context.rs
chris-olszewski: On second thought I removed this entirely since the final shape of calling nexus operations hasn't stabilized.

crates/workflow/src/workflow_interceptors.rs
chris-olszewski: > If this took some param that provided context about the workflow

Meant to do that, but totally forgot which did make it pointless.

Updated to now take a vec of `WorkflowInterceptorConstructor`s which are wrappers around `Fn(&WorkflowContextView) -> WorkflowInterceptor`. Only have a named type so end users don't need to deal with the boxing of closures and the constructed interceptor.

crates/workflow/src/workflow_context.rs
chris-olszewski: Did a quick check of [Ruby](https://github.com/temporalio/sdk-ruby/blob/0d1a37f9c946a374ac204cba7d854fb14b2c1968/temporalio/lib/temporalio/internal/worker/workflow_instance/outbound_implementation.rb#L148) and [TS](https://github.com/temporalio/sdk-typescript/blob/112d3925fae9d9f850e83210f84207badeab64d9/packages/workflow/src/workflow.ts#L390) it seems that a regular timer with interceptors is used in both cases. We could set a timer summary that users could check?

crates/sdk/src/workflow_wasm.rs
chris-olszewski: > However, I'm not sure why this needs to be an error. If a user defines interceptors in the WASM bundle, I think we can use those and just document that native ones do not apply to WASM workflows?

You're right, no need to error in this scenario.

> I guess just extra back-and-forth glue?

Correct, this has been in development for long enough and the PR is large enough that I didn't feel tackling this here was necessary.

crates/workflow/src/workflow_context.rs
Sushisource: OK, interesting. Yeah probably setting a summary is a decent idea

crates/sdk/src/lib.rs
```diff
@@ -174,7 +174,7 @@ pub struct WorkerOptions {
     workflows: WorkflowDefinitions,
 
     #[builder(field)]
-    workflow_interceptor_factories: Vec<Arc<dyn WorkflowInterceptorFactory>>,
+    workflow_interceptor_constructors: Vec<WorkflowInterceptorConstructor>,
```
Sushisource, line 179: We probably should add a `workflow_interceptor_constructors_mut` that can be used to mutate this list if they want to append etc

crates/sdk/src/workflow_executor.rs
```diff
@@ -311,22 +320,22 @@ mod tests {
...
-                let result = handle.await.unwrap();
-                assert_eq!(result, 42);
+                tokio::select! {
+                    _ = executor.drive() => unreachable!("executor driver cannot finish"),
+                    result = async {
+                        polled_rx.await.unwrap();
+                        tx.send(()).unwrap();
+                        handle.await.unwrap()
+                    } => assert_eq!(result, 42),
+                }
```
Sushisource, line 341: Keeping some comments here would be good, hard to immediately track what's going on

crates/workflow/src/runtime/mod.rs
```diff
@@ -20,6 +20,171 @@ pub mod types;
...
+    handler_result_ready: bool,
+}
+
+/// Shares poll progress between the outer chain, its terminal handler boundary, and SDK command
+/// futures nested anywhere inside the chain.
+///
+/// This state lives outside [`crate::workflow_interceptors::WorkflowInterceptorFuture`] because
+/// interceptors may replace that public wrapper while composing the chain.
+#[derive(Clone)]
+pub(crate) struct InterceptedFutureStatus(Rc<Cell<InterceptedFutureStatusInner>>);
```
Sushisource, line 90: This whole thing looks to me like it should be a state machine really. Certain combinations of the internal fields should be unrepresentable seems like?

crates/workflow/src/runtime/types.rs
```diff
@@ -119,10 +119,18 @@ pub enum RoutineCompletion {
     Update(UpdateRoutineCompletion),
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum RoutinePendingState {
+    Handler,
+    Interceptor,
+    InterceptorWithActivation,
+}
```
Sushisource, line 138: Why dupe this enum?

Sushisource (commented):
I find the complexity of the whole intercepted futures system to be a bit concerning but I don't have any good alternative suggestions off the top of my head

crates/workflow/src/runtime/types.rs
chris-olszewski: Was thinking to allow `InterceptedFutureStatus` to change in the future without changing the other WIT expectations, but thinking now that doesn't make much sense.

Sushisource (approved):

PR #1402, opened 2026-07-14
💥 feat(sdk): Support Standalone Activities

### LINKED ISSUE #1070: [Feature Request] Support non-workflow activities

Allow executing activities without a workflow. Upstream server API needs to be updated to include the new activity RPCs (not yet merged as of now). Worker must support running activities outside of workflow context.

### Additional context

Server API PR (WIP): https://github.com/temporalio/api/pull/640

maciejdudko:
## 💥 Breaking changes

- `ActivityInfo` fields related to the containing workflow are made optional.
- `async_activity_handle::ActivityIdentifier::ById` is split into `ByIdWorkflow` and `ByIdStandalone`.

## What was changed

- Added support for executing Standalone Activities in SDK Worker.
- Added high-level client interfaces for starting and interacting with Standalone Activities.

## Why?

Feature request: https://github.com/temporalio/features/issues/706

## Checklist

1. Closes #1070

2. How was this tested:

Added tests in `crates/sdk-core/tests/integ_tests/standalone_activity_tests.rs`

3. Any docs updates needed?

SAA guide for Rust needs to be added here: https://docs.temporal.io/standalone-activity


crates/client/src/activity/activity_execution_info.rs
```diff
@@ -0,0 +1,570 @@
...
+        &self.raw
+    }
+}
+
+/// Contains the current state of the activity execution.
+/// Obtained from [`ActivityHandle::describe`](crate::ActivityHandle::describe).
+///
+/// To support deserialization of transmitted payloads, the object internally stores a reference
+/// to the client used to make the request. For this reason, this type is parametrized with client
+/// type. The client reference can be dropped by calling [`simple`](Self::untyped).
```
Sushisource: This comment looks like it's not accurate any more since just a dataconverter, not a whole client, is stored (which is good).

crates/client/src/activity/activity_execution_info.rs
```diff
@@ -0,0 +1,570 @@
...
+    pub fn raw_info(&self) -> &RawInfo {
+        &self.raw_info
+    }
+
+    /// True if activity input is present.
+    /// See [`ActivityDescribeOptions::include_input`](crate::ActivityDescribeOptions::include_input).
+    /// Use [`input`](Self::input) or [`raw_input`](Self::raw_input) to retrieve it.
+    pub fn has_input(&self) -> bool {
+        self.raw_input.is_some()
+    }
```
Sushisource, line 230: Nit: IMO we don't need this - easy enough to just call `raw_input().is_some()`

crates/client/src/activity/activity_execution_info.rs
```diff
@@ -0,0 +1,570 @@
...
+        let Some(input) = &self.raw_input else {
+            return Ok(None);
+        };
+        Ok(Some(self.convert_payloads(input).await?))
+    }
+
+    /// True if activity outcome is present.
+    /// See [`ActivityDescribeOptions::include_outcome`](crate::ActivityDescribeOptions::include_outcome).
+    /// Use [`outcome`](Self::outcome) or [`raw_outcome`](Self::outcome) to retrieve it.
+    pub fn has_outcome(&self) -> bool {
```
Sushisource, line 249: Same here

crates/client/src/errors.rs
```diff
@@ -311,3 +317,172 @@ impl AsyncActivityError {
...
+}
+
+/// Errors returned by [`crate::ActivityHandle::result`].
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum ActivityResultError {
+    /// Activity execution did not complete successfully.
+    #[error("Activity failed: {0}")]
+    ActivityFailed(#[source] IncomingError),
```
Sushisource, line 420: With `WorkflowGetResultError` we distinguish Failed / timedout / cancelled etc at the top level, probably makes sense to do that here too.

crates/client/src/activity/activity_handle.rs
```diff
@@ -0,0 +1,191 @@
...
+                        namespace: client.namespace(),
+                        activity_id: self.activity_id.clone(),
+                        run_id: self.run_id.clone().unwrap_or_default(),
+                    }
+                    .into_request(),
+                )
+                .await?
+                .into_inner();
+
+            // If resp.outcome.value is None, poll again
```
Sushisource, line 91: Will this hot-loop?

crates/client/src/activity.rs
```diff
@@ -0,0 +1,141 @@
...
+        },
+    },
+    search_attributes::{SearchAttributeError, SearchAttributeValue},
+};
+
+/// A stream of activity executions from a list query.
+/// Internally paginates through results from the server.
+pub struct ListActivitiesStream {
+    inner: Pin<Box<dyn Stream<Item = Result<Vec<ActivityExecutionListInfo>, ClientError>> + Send>>,
+    buffer: <Vec<ActivityExecutionListInfo> as IntoIterator>::IntoIter,
```
Sushisource: I think this could just be stored as a Vec more simply?

crates/client/src/async_activity_handle.rs
```diff
@@ -43,19 +51,56 @@ impl ActivityIdentifier {
...
 
-    /// Create an identifier from workflow and activity IDs. Use an empty run id to target the
-    /// latest workflow execution.
-    pub fn by_id(
+    /// Create an identifier of a workflow activity from workflow and activity IDs. Set run_id to
+    /// None to target the latest workflow execution.
+    pub fn by_id_workflow(
         workflow_id: impl Into<String>,
-        run_id: impl Into<String>,
+        run_id: Option<impl Into<String>>,
```
Sushisource: So the reason I didn't store it as an optional before was because it makes calling this function slightly awkward where you have to pass `Some("run_id")` instead of just `"run_id"` with the tradeoff that `None` is slightly better than `""` maybe, but, I think the empty string isn't a bad option here. Plus, now `None` and `Some("")` encode the same state.

My preference would be to keep just the string, and say empty means unset, but curious to hear @chris-olszewski  thoughts.

crates/client/src/lib.rs
```diff
@@ -830,12 +836,108 @@ impl Client {
...
+    where
+        A: ActivityDefinition,
+    {
+        WorkflowClientTrait::start_activity(self, activity, input, options).await
+    }
+
+    /// Start a standalone activity and wait for its result.
+    ///
+    /// Equivalent to `start_activity(...).await?.result()`.
+    pub async fn execute_activity<A>(
```
Sushisource: We don't have `execute_workflow`, personally not a big fan of these simple sugars.

crates/client/src/lib.rs
```diff
@@ -830,12 +836,108 @@ impl Client {
...
+    /// specified, the handle always targets the latest execution with matching ID.
+    ///
+    /// Note that the validity of the handle is not checked until a method is called on it.
+    /// If invalid ID or run ID is used, the method will return `NotFound` error.
+    ///
+    /// To get a typed handle, use [`get_activity_handle`](Self::get_activity_handle).
+    ///
+    /// To get a handle that can be used to complete an activity asynchronously,
+    /// see [`get_async_activity_handle`](Self::get_async_activity_handle).
+    pub fn get_untyped_activity_handle(
```
Sushisource, line 1252: Same here, I think just calling `get_activity_handle<UntypedActivity>(...)` is fine and we can say how to do that in the docstring.

crates/common-wasm/src/lib.rs
```diff
@@ -48,3 +50,55 @@ macro_rules! dbg_panic {
 }
 #[allow(unused_imports)]
 pub(crate) use dbg_panic;
+
+/// Represents Activity schedule-to-close and start-to-close timeouts for the purposes of specifying
+/// Activity options. Specifying at least one of them is required, but specifying both is also
+/// allowed. Note that this type does not cover all available timeout options for an Activity.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ActivityCloseTimeoutOptions {
```
Sushisource: IMO no reason to add the "options" postfix here

crates/client/src/options_structs.rs
```diff
@@ -568,3 +570,191 @@ pub struct WorkflowListOptions {
...
+pub struct ActivityStartOptions {
+    /// Task queue to run this activity on.
+    #[builder(start_fn)]
+    pub task_queue: String,
+    /// Activity ID of the started activity. It's recommended to use a meaningful business ID.
+    #[builder(start_fn)]
+    pub id: String,
+    /// Combined schedule-to-close and start-to-close timeout options. At least one of them is
+    /// required.
+    #[builder(setters(name = combined_close_timeouts, vis = ""))]
```
Sushisource: This is part of the `start_fn` when starting from a workflow, and it should be here too IMO, that simplifies a lot of other stuff.

The "combined" wording in the docstring / setter (which can go away if we make it part of constructor) diverges from the workflow version too, I think we can keep it the same.

crates/client/src/utils.rs
```diff
@@ -0,0 +1,15 @@
+use std::error::Error;
+
+pub(crate) fn try_into_or_box_err<A, B, E, MapErr>(
```
Sushisource: Since it's only used from client lib, can probably just stick it in there as private.

Sushisource (commented):
Awesome to see this ready!

crates/client/src/activity/activity_handle.rs
maciejdudko: The server uses long-poll for this request.

crates/client/src/activity.rs
maciejdudko: `buffer` is stored as an iterator so that `poll_next` can take the next element efficiently. Storing as a reversed Vec or a VecDeque would require copying out of the original Vec inside the response - using `into_iter` avoids it.

crates/client/src/lib.rs
maciejdudko: Going to remove it for now since it can always be added later.

crates/client/src/lib.rs
maciejdudko: `get_activity_handle` requires the activity as argument. `get_untyped_activity_handle` does not. `get_activity_handle::<UntypedActivity>` would not work.

crates/client/src/options_structs.rs
maciejdudko: This setter is private. The idea is that the user does not use `combined_close_timeouts` or `ActivityCloseTimeoutOptions` directly - instead, they use either `schedule_to_close_timeout` or `start_to_close_timeout` to set one of them, or `close_timeouts` to set both.

crates/client/src/activity.rs
Sushisource: I don't think that's right. You just `pop_front()` the `VecDeque` in `poll_next()`. No copy involved. Next stream element just becomes `self.buffer = items.into();` (which is constant time and doesn't reallocate)

crates/client/src/options_structs.rs
Sushisource: Ah. I didn't see it was private. Regardless... I still think it makes sense to just do exactly what the workflow version does. Passing the right variant to `close_timeouts` is easy

crates/client/src/lib.rs
Sushisource: Ah, yeah, I just wrote it wrong though - you can still just pass `UntypedActivity` as the arg.

...though I guess that does require you to pointlessly specify a name, so, that being the case I'm fine with this then

crates/client/src/activity/activity_execution_info.rs
maciejdudko: I'd keep these methods for symmetry with heartbeat_details/last_failure, also I don't like making users call `raw_*` methods when they're not interested in raw payloads.

crates/client/src/activity.rs
maciejdudko: Didn't know the last part. Going to change to VecDeque.

crates/client/src/async_activity_handle.rs
maciejdudko: That's actually old code that I forgot to change. `Option<impl Into<String>>` doesn't actually work, since passing `None` fails type inference. In other places I use `Option<String>`.

It's true that `None` is redundant with an empty string, but I think an explicit `Option` is a nicer API to use.

crates/client/src/activity.rs
```diff
@@ -45,12 +46,13 @@ impl Stream for ListActivitiesStream {
...
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         loop {
-            if let Some(info) = self.buffer.next() {
+            if let Some(info) = self.buffer.pop_front() {
                 return Poll::Ready(Some(Ok(info.into())));
             }
             match self.inner.poll_next_unpin(cx) {
                 Poll::Ready(Some(Ok(items))) => {
-                    self.buffer = items.into_iter();
+                    // Using into() rather than extend() to avoid moving items in memory
```
Sushisource: Uninteresting comment

Sushisource (approved):
Please get approval from @chris-olszewski too

crates/sdk/src/activities.rs
```diff
@@ -225,18 +230,22 @@ impl ActivityContext {
...
-        Some(WorkflowHandle::new(
-            self.client(),
-            WorkflowExecutionInfo {
-                namespace: self.client_options.namespace.clone(),
-                workflow_id: workflow_execution.workflow_id().to_owned(),
-                run_id: run_id.clone(),
-                first_execution_run_id: run_id,
-            },
-        ))
+        self.info().workflow_id.clone().map(|workflow_id| {
```
chris-olszewski: I do slightly prefer `?` usage here over `map` just to reduce nesting. Not blocking obviously.

crates/client/src/async_activity_handle.rs
```diff
@@ -73,17 +80,42 @@ impl ActivityIdentifier {
...
+    /// latest activity execution.
+    pub fn by_id_standalone(activity_id: impl Into<String>, run_id: impl Into<String>) -> Self {
+        Self::ByIdStandalone {
+            activity_id: activity_id.into(),
+            run_id: run_id.into(),
+        }
+    }
+
+    /// Returns tuple of (workflow_id, run_id, activity_id). Panics if identifier is a token.
+    fn id_tuple(self) -> (String, String, String) {
```
chris-olszewski: `into_parts` is more Rust-y, would also prefer if callers made did unwrap instead of hiding it in the function
```suggestion
    fn into_parts(self) -> Option<(String, String, String)> {
```

crates/common-wasm/src/lib.rs
```diff
@@ -48,3 +50,55 @@ macro_rules! dbg_panic {
...
+}
+
+/// For internal use. Obtained from [`ActivityCloseTimeouts`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive] // disallow direct construction
+#[doc(hidden)]
+pub struct ActivityCloseTimeoutValues {
+    pub schedule_to_close: Option<Duration>,
+    pub start_to_close: Option<Duration>,
+}
```
chris-olszewski, line 97: Unsure why we need a public hidden `into_values` here and its return type. I think providing public documented getters `start_to_close(&self) -> Option<Duration>`/`schedule_to_close(&self) -> Option<Duration>`. `Duration` is small enough that I'm not worried about the extra copy this introduces.

crates/client/src/options_structs.rs
```diff
@@ -688,3 +690,167 @@ pub struct WorkflowCountOptions {
...
+
+/// Controls which optional fields will be requested in
+/// [`ActivityHandle::describe`](crate::ActivityHandle::describe) operation. The fields will be
+/// present in returned [`ActivityExecutionDescription`](crate::ActivityExecutionDescription),
+/// subject to data availability and server support.
+///
+/// Note that these fields contain payloads that can be arbitrarily large. It's recommended not to
+/// include them unless they're needed.
+#[derive(Debug, Clone, Default)]
+pub struct ActivityDescribeOptions {
```
chris-olszewski: For future proofing
```suggestion
#[derive(Debug, Clone, Default, bon::Builder)]
#[non_exhaustive]
pub struct ActivityDescribeOptions {
```

crates/client/src/options_structs.rs
```diff
@@ -688,3 +690,167 @@ pub struct WorkflowCountOptions {
...
+    pub include_outcome: bool,
+    /// If `true` and the activity sent heartbeat details, the heartbeat details will be included.
+    pub include_heartbeat_details: bool,
+    /// If `true` and the activity has a failed attempt, the last failure will be included.
+    pub include_last_failure: bool,
+}
+
+/// Options for [`ActivityHandle::cancel`](crate::ActivityHandle::cancel).
+#[derive(Debug, Clone, Default)]
+pub struct ActivityCancelOptions {
```
chris-olszewski:
```suggestion
#[derive(Debug, Clone, Default, bon::Builder)]
#[non_exhaustive]
pub struct ActivityCancelOptions {
```

crates/client/src/options_structs.rs
```diff
@@ -688,3 +690,167 @@ pub struct WorkflowCountOptions {
...
+    /// Reason for cancellation. Can be empty.
+    pub reason: String,
+}
+
+/// Options for [`ActivityHandle::terminate`](crate::ActivityHandle::terminate).
+#[derive(Debug, Clone, Default)]
+pub struct ActivityTerminateOptions {
```
chris-olszewski:
```suggestion
#[derive(Debug, Clone, Default, bon::Builder)]
#[non_exhaustive]
pub struct ActivityTerminateOptions {
```

chris-olszewski (approved):
CHANGELOG.md entry would be appreciated especially for the breaking changes.

Only comments I feel strongly about are avoiding the `#[doc(hidden)]`s and ensuring all options are marked with `#[non_exhaustive]` so we can expand them if needed.

PR #1405, opened 2026-07-15
Detect Nexus operation changes during replay

### LINKED ISSUE #1231: [Bug] ScheduleNexusOperation doesn't raise NDE on Operation or Service mismatch
### Describe the bug

Core's `ScheduleNexusOperation` state machine does not raise a non-determinism error on replay when the command's fields don't match the corresponding `NexusOperationScheduled` history event.

This is inconsistent with Go and Java (which both check service and operation), and with Core's own `ScheduleActivityTask` and `StartChildWorkflowExecution` machines (which check ID + type under the `IdAndTypeDeterminismChecks` flag).

### Proposed fix

- In `crates/sdk-core/src/worker/workflow/machines/nexus_operation_state_machine.rs`, extend on_scheduled to compare the command's service and operation against the event attributes and return `nondeterminism!()` on mismatch.

- We might need to gate the extra NDE checks with an SDK flag so that we don't suddenly start raising NDE for changes that happened in the past.

JSap0914:
## What

- compare the service and operation in `NexusOperationScheduled` history events with the command produced during replay
- report nondeterminism when either value changes
- gate the new checks behind a dedicated Core internal flag so existing histories remain compatible
- add regression coverage and an Unreleased changelog entry

## Why

The Nexus operation state machine accepted scheduled history events without validating their service or operation. This allowed workflow code to change either field during replay without raising the non-determinism error used by comparable activity and child-workflow state machines.

Fixes #1231

## Tests

- `cargo test -p temporalio-sdk-core nexus_operation_state_machine::tests --lib`
- `cargo test -p temporalio-sdk-core --lib` (367 passed, 1 ignored)
- `cargo clippy -p temporalio-sdk-core --lib --tests --all-features -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`


Sushisource (commented):
Thanks for this! Makes perfect sense. I think adding an integration test would be great and then we can merge.

Sushisource:
@JSap0914 Looks like you need to run the formatter

Sushisource (approved):

Sushisource:
@JSap0914 We'll need you to sign the CLA too :)

Sushisource:
@JSap0914 - we can merge this if you sign the CLA. Thanks!

PR #1406, opened 2026-07-15
Auto-enroll pollers into autoscaling on PollerAutoscalingAutoEnroll (Rust SDK) 

veeral-patel:
Ports [temporalio/sdk-go#2442](https://github.com/temporalio/sdk-go/pull/2442) to `sdk-core`, so it's shared by every SDK on top of it (Rust directly; TS/Python/.NET/Ruby via the c-bridge).

## What

When a namespace advertises the `poller_autoscaling_auto_enroll` capability, workers automatically switch their workflow, activity, and nexus pollers to **autoscaling** — but only for poller types the user left at their default. Pollers with an explicitly configured count or behavior are left alone.

## Why it's implemented this way

The capability is only known after `Worker::validate()` (an async namespace lookup), but pollers are built earlier, synchronously. Instead of reordering that, the `PollScaler` resolves its behavior **lazily**, when polling starts — which always happens after `validate()`. Alternatives (blocking network I/O during construction, or rebuilding pollers afterward) were more invasive.

## Key changes

- **`Worker::validate()`** reads the new capability and also enables `poller_autoscaling` so enrolled pollers can scale down.
- **`WorkerConfig`** gains three per-type "eligibility" flags (default `false` — nothing auto-enrolls unless a layer above opts in).
- **`PollScaler`** defers its setup into `activate()` and, if the poller is eligible and the capability is present, uses `Autoscaling { min: 1, max: 100, initial: 5 }` (matching Go).
- **High-level SDK / c-bridge** signal eligibility by leaving the poller behavior unset (SDK: `Option::None`; c-bridge: both FFI pointers null).

## Notes

- Two intentional differences from the Go PR: no session-worker case and no heartbeat change — neither exists in `sdk-core`.
- The SDK `WorkerOptions` poller fields become `Option<PollerBehavior>` (minor pre-1.0 break; existing `.foo(x)` calls still compile).
- Other lang SDKs need a small follow-up to pass the unset signal through the FFI.

## Testing

New unit tests for behavior resolution, `validate()` plumbing, and the c-bridge conversion. Full `sdk-core` lib suite passes (368 tests); fmt + clippy clean.

### DIFF
--- a/crates/sdk-core-c-bridge/src/worker.rs
+++ b/crates/sdk-core-c-bridge/src/worker.rs
@@ -115,6 +115,30 @@ impl TryFrom<&PollerBehavior> for temporalio_sdk_core::PollerBehavior {
     }
 }
 
+/// Converts an FFI poller behavior into an optional core poller behavior. Both fields null means
+/// the poller was left unset by lang: the caller should apply the default behavior and treat the
+/// poller as eligible for automatic enrollment into poller autoscaling.
+fn ffi_poller_behavior_opt(
+    value: &PollerBehavior,
+) -> anyhow::Result<Option<temporalio_sdk_core::PollerBehavior>> {
+    if !value.simple_maximum.is_null() && !value.autoscaling.is_null() {
+        bail!("simple_maximum and autoscaling cannot both be non-null values");
+    }
+    if let Some(sm) = unsafe { value.simple_maximum.as_ref() } {
+        Ok(Some(temporalio_sdk_core::PollerBehavior::SimpleMaximum(
+            sm.simple_maximum,
+        )))
+    } else if let Some(a) = unsafe { value.autoscaling.as_ref() } {
+        Ok(Some(temporalio_sdk_core::PollerBehavior::Autoscaling {
+            minimum: a.minimum,
+            maximum: a.maximum,
+            initial: a.initial,
+        }))
+    } else {
+        Ok(None)
+    }
+}
+
 #[repr(C)]
 pub enum WorkerVersioningStrategy {
     None(WorkerVersioningNone),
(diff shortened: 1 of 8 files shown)


crates/sdk-core-c-bridge/src/worker.rs
```diff
@@ -115,6 +115,30 @@ impl TryFrom<&PollerBehavior> for temporalio_sdk_core::PollerBehavior {
     }
 }
 
+/// Converts an FFI poller behavior into an optional core poller behavior. Both fields null means
+/// the poller was left unset by lang: the caller should apply the default behavior and treat the
+/// poller as eligible for automatic enrollment into poller autoscaling.
+fn ffi_poller_behavior_opt(
```
Sushisource, line 121: This should probably just be an `Into` impl

crates/sdk-core/src/pollers/poll_buffer.rs
```diff
@@ -494,11 +519,44 @@ async fn handle_task_panic(t: JoinHandle<()>) {
...
+    // before polling begins).
+    configured_behavior: PollerBehavior,
+    auto_enroll_eligible: bool,
+    capabilities: Arc<NamespaceCapabilities>,
+    last_successful_poll_time: Arc<AtomicCell<Option<SystemTime>>>,
+    shutdown: CancellationToken,
+    /// Shared with the poll function so `no_retry` tracks the resolved (post-`activate`) behavior.
+    autoscaling_active: Arc<AtomicBool>,
+    // Resolved by `activate`, before the poll loop issues any poll.
+    report_handle: Option<Arc<PollScalerReportHandle>>,
```
Sushisource, line 559: This is a lot of additional complexity. I think things could overall become much, much simpler, if we just pass in the right PollerBehavior based on namespace capabilities rather than switching it dynamically.

We aren't expecting to dynamically flip between enabled and not enabled while the worker is running, right? I don't think the Go PR did that.

So IMO we should just delay spinning up pollers until after the `validate` call has happened (which might already be true)? And then this file doesn't need any changes at all, everything should just work as is.

crates/sdk-core/src/worker/mod.rs
```diff
@@ -542,6 +564,17 @@ impl Worker {
                             .poller_autoscaling
                             .store(true, Ordering::Relaxed);
                     }
+                    if caps.poller_autoscaling_auto_enroll {
+                        self.capabilities
+                            .poller_autoscaling_auto_enroll
+                            .store(true, Ordering::Relaxed);
+                        // Auto-enroll implies full autoscaling support (including scaling down on
```
veeral-patel, line 571: We don't need this. We can decouple these two capabilities.

crates/sdk-core/src/pollers/poll_buffer.rs
```diff
@@ -494,11 +519,44 @@ async fn handle_task_panic(t: JoinHandle<()>) {
...
+/// `poller_autoscaling_auto_enroll` capability. Otherwise the configured behavior is used as-is.
+fn resolve_effective_behavior(
+    configured: PollerBehavior,
+    auto_enroll_eligible: bool,
+    capabilities: &NamespaceCapabilities,
+) -> PollerBehavior {
+    if auto_enroll_eligible && capabilities.poller_autoscaling_auto_enroll() {
+        // Mirrors the Go SDK's auto-enroll autoscaling defaults.
+        PollerBehavior::Autoscaling {
+            minimum: 1,
```
veeral-patel, line 533: See if we can not pass in the min, max, initial so the existing default values are used. If we have to, see if we can reuse a shared constant instead of hard-coding

veeral-patel:
Closing in favor of this PR: https://github.com/temporalio/sdk-rust/pull/1425

PR #1407, opened 2026-07-15
Allow resource controllers to be shared across tuners

Sushisource:
## Summary

- expose the existing generic `ResourceController`, with `RealSysInfo` as its default sampler type
- add `ResourceBasedTunerConfig::{Options, Controller}` so `TunerHolderOptions` can either construct a controller or receive a shared `Arc<ResourceController>`
- update the C bridge's declarative resource-tuner path for the new enum

## Why

Language bridges currently pass only `ResourceBasedSlotsOptions` into `TunerHolderOptions`, which constructs a fresh resource controller, sampler, and PID state for every holder. That prevents a language SDK from sharing one controller across multiple Workers in the same process, as requested in temporalio/sdk-typescript#2159.

Making the existing controller injectable lets a bridge retain one `Arc<ResourceController>` and use it for multiple tuner holders. The enum keeps the declarative and injected-controller forms mutually exclusive by construction. Declarative options still create exactly one controller shared by all resource-based slot suppliers in that holder.

Worker heartbeat resource sampling remains independent of the tuner.

## Testing

- `cargo +nightly fmt --all`
- `cargo lint`
- `cargo test-lint`
- `cargo test -p temporalio-sdk-core worker::tuner`

### DIFF
--- a/crates/sdk-core/src/worker/tuner.rs
+++ b/crates/sdk-core/src/worker/tuner.rs
@@ -2,12 +2,11 @@ mod fixed_size;
 mod resource_based;
 
 pub use fixed_size::FixedSizeSlotSupplier;
+pub(crate) use resource_based::{RealSysInfo, SystemResourceInfo};
 pub use resource_based::{
     ResourceBasedSlotsOptions, ResourceBasedSlotsOptionsBuilder, ResourceBasedTuner,
-    ResourceSlotOptions,
+    ResourceController, ResourceSlotOptions,
 };
-
-pub(crate) use resource_based::{RealSysInfo, SystemResourceInfo};
 
 use crate::{
     WorkerConfig,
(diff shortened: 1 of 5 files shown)


crates/sdk-core/src/worker/tuner.rs
```diff
@@ -435,10 +447,39 @@ mod tests {
             result
                 .unwrap_err()
                 .to_string()
-                .contains("resource_based_options")
+                .contains("resource_based_config")
         );
     }
 
+    #[test]
+    fn tuner_holder_options_retain_explicit_controller_across_builds() {
```
chris-olszewski: Not sure what this test is supposed to demonstrate since the arc clone happens in our the `build_tuner` closure itself.

chris-olszewski (approved):
Code itself makes sense, unsure about the test.

crates/sdk-core/src/worker/tuner.rs
```diff
@@ -57,20 +56,34 @@ pub struct TunerHolderOptions {
...
+    pub resource_based_config: Option<ResourceBasedTunerConfig>,
+}
+
+/// Selects whether a tuner holder creates a controller or uses an existing shared controller.
+#[derive(Clone, Debug)]
+pub enum ResourceBasedTunerConfig {
+    /// Create a controller from these options.
+    Options(ResourceBasedSlotsOptions),
+    /// Use this controller for every resource-based slot supplier in the holder.
+    Controller(Arc<ResourceController>),
```
eamsden, line 69: Is it intentional that only the sampler and PID state are shared? Each holder still creates its own `ResourceBasedSlotsForType`, including independent minimum/maximum accounting and ramp-throttle clocks. Multiple holders can therefore act on the same permissive PID output before the 100 ms resource sample reflects the added load, multiplying the effective minimum and issuance rate and changing the loop dynamics. Could we either share the actuator-side limiting state as well, or document and behaviorally test that keeping it per holder is the intended and stable contract?

crates/sdk-core/src/worker/tuner.rs
```diff
@@ -435,10 +447,39 @@ mod tests {
...
+                )),
+            }
+            .build_tuner_holder()
+            .unwrap()
+        };
+
+        let _first = build_tuner();
+        assert_eq!(Arc::strong_count(&controller), 2);
+        let _second = build_tuner();
+        assert_eq!(Arc::strong_count(&controller), 3);
```
eamsden: These strong-count assertions prove that each holder retains the `Arc`, but not that sharing produces the intended controller behavior. They would still pass if independent per-holder minimums and ramp throttles allow both holders to issue simultaneously from one permissive PID decision. Could this be supplemented with a behavioral test using a fake resource sampler and two tuners, asserting the intended aggregate issuance and cross-tuner feedback behavior?

eamsden (commented):
Sharing a controller also shares its PID history, but each holder remains an independent actuator with its own minimums and ramp throttle. That changes the effective loop gain and sampling/actuation dynamics, so I'd like to see the intended aggregate behavior documented and validated with a behavioral test.

crates/sdk-core/src/worker/tuner.rs
Sushisource: Replaced this test with one that does that

crates/sdk-core/src/worker/tuner.rs
Sushisource: Yes, that is intentional, since they make different decisions based on the controller output that are appropriate for the slot type they're managing. You do make a good point though that we could possibly decide to allow a slot for an activity & a workflow at the same time and if we're too close to the edge that might push us over.

I don't think I want to try to resolve that right now though. I can add some docs to that effect.

Sushisource: Added a bit to the tuner dosctring

eamsden (approved):

Issue #1408, opened 2026-07-16
[Bug] ResourceBasedTuner memory metric counts reclaimable page cache (cgroup memory.current), freezing admission after cache-heavy bursts


odedkedemdreeze:
### What are you really trying to do?

Run activity workers under `ResourceBasedTuner` in a memory-limited container (Kubernetes, cgroup v2, no swap). Activities are memory-heavy and touch a lot of files (in our case terraform-based recovery: a ~700 MB provider plugin is mmap'd per run), so we gate admission on memory to avoid OOM.

### Describe the bug

`ResourceBasedTuner`'s memory measurement is effectively cgroup **`memory.current`**, which **includes reclaimable page cache**. Page cache does not fall when an activity finishes and poses no OOM risk (the kernel evicts it under pressure), but the tuner treats it as "used memory." After a cache-heavy burst, the reading stays **pinned above target on an otherwise-idle pod**, so `can_reserve` stays false and admission **freezes for minutes** — even for tiny activities that need almost no memory. Real (anonymous) memory is low the whole time.

This is a measurement bug, not a policy bug: the PID controller is fed the wrong signal.

### Root cause (source-traced)

`crates/sdk-core/src/worker/tuner/resource_based.rs`, `RealSysInfoInner::refresh()` (cgroup branch, ~L508 on `main`):

```rust
if let Some(cgroup_limits) = lock.cgroup_limits() {
    self.total_mem.store(cgroup_limits.total_memory, Ordering::Release);
    self.cur_mem_usage.store(
        cgroup_limits.total_memory - cgroup_limits.free_memory,   // == memory.current
        Ordering::Release,
    );
    ...
```

`cgroup_limits` comes from `sysinfo` (0.38). Its cgroup-v2 path (`src/unix/linux/system.rs`) computes:

```rust
read_u64("/sys/fs/cgroup/memory.current"),                    // mem_cur
read_table_key("/sys/fs/cgroup/memory.stat", "anon", ' '),    // mem_rss
...
limits.free_memory = limits.total_memory.saturating_sub(mem_cur);
rss: mem_rss,
```

So `total_memory − free_memory == memory.current` (page cache included). Notably, **`sysinfo` already parses `memory.stat`'s `anon` into `CGroupLimits.rss`, but `RealSysInfoInner` ignores it** and uses `total − free` instead. `used_mem_percent = cur_mem_usage / total_mem` then feeds the memory PID controller and the emitted usage gauge, so both track page cache.

[...]

PR #1409, opened 2026-07-16
Gate ResourceBasedTuner on anonymous memory instead of cgroup memory.current

odedkedemdreeze:
## What & why

`ResourceBasedTuner` feeds `total_memory - free_memory` into its memory PID controller. In the cgroup path that value is exactly **`memory.current`**, which **includes reclaimable page cache**.

In a memory-limited, no-swap container, a file-heavy activity (for us: a ~700 MB terraform provider mmap'd per activity) inflates page cache that **does not fall when the activity finishes** — page cache is reclaimable, so it poses no OOM risk, but it also isn't released on an idle worker. The tuner therefore stays pinned above `target_memory_usage` on an otherwise-idle pod and **freezes slot admission for minutes**, starving small activities, even though anonymous (real) memory is low the whole time.

This is a measurement bug: the controller is fed the wrong signal. Full analysis + repro in #1408.

## The change

Feed **anonymous (non-reclaimable) memory** into the tuner instead. `sysinfo` already parses `memory.stat`'s `anon` into `CGroupLimits::rss` — the code just wasn't using it:

```rust
// was: total_memory - free_memory  == memory.current (incl. page cache)
fn cgroup_used_mem(limits: &sysinfo::CGroupLimits) -> u64 {
    limits.rss
}
```

Extracted into a small pure fn so the derivation is unit-testable (it wasn't before — the existing `FakeMIS` fakes `used_mem()` directly, above this code).

## Test

Added `cgroup_used_mem_excludes_page_cache`, which constructs the pathological case (2 GiB limit, `memory.current` pinned at 1300 MiB by page cache, only 600 MiB anonymous) and asserts the tuner sees 600 MiB, not 1300 MiB. Verified locally: `cargo test -p temporalio-sdk-core` → passes; `rustfmt --check` clean.

## Repro evidence (live pod, 2 GiB limit, cgroup v2, no swap)

Reading a 674 MiB file into cache, then dropping it with `posix_fadvise(DONTNEED)`:

```
                        memory.current   anon    file(cache)
baseline                    621 MiB       573      24
after reading a file       1279 MiB       581     674   <- tuner sees +658 MiB "used", admission freezes
after fadvise DONTNEED       627 MiB       581      21   <- real usage was flat the whole time
```

## Open question for maintainers

This changes the metric for **all language SDKs**, so I kept the change minimal but I'm happy to adjust to whichever shape you prefer (also raised in #1408):

1. **Anonymous only** (this PR) — simplest; correct on no-swap; ignores swap and active file cache.
2. **Working set** (`memory.current - inactive_file`) — matches the kubelet's eviction metric; arguably the best general default; needs `inactive_file` surfaced.
3. **Configurable** `memory_metric` on `ResourceBasedTunerOptions`, defaulting to current behavior — zero behavior change for existing users, opt-in for the fix.

Glad to convert this to (2) or (3) — including gating behind a config option for backwards compatibility — if that's the direction you'd accept.

Related: companion to the deadlock fix in temporalio/sdk-python#1643 (same workload; different bug).

### DIFF
--- a/crates/sdk-core/src/worker/tuner/resource_based.rs
+++ b/crates/sdk-core/src/worker/tuner/resource_based.rs
@@ -522,10 +522,6 @@ struct RealSysInfoInner {
     cgroup_cpu_info: CGroupCpuInfo<CgroupV2CpuFileSystem>,
 }
 
-fn cgroup_used_mem(limits: &sysinfo::CGroupLimits) -> u64 {
-    limits.rss
-}
-
 impl RealSysInfoInner {
     fn refresh(&self) {
         let mut lock = self.sys.lock();
@@ -533,8 +529,11 @@ impl RealSysInfoInner {
         if let Some(cgroup_limits) = lock.cgroup_limits() {
             self.total_mem
                 .store(cgroup_limits.total_memory, Ordering::Release);
+            // Gate on anonymous memory (`rss`) rather than `memory.current`
+            // (`total - free`), which includes reclaimable page cache and can
+            // permanently starve slot admission.
             self.cur_mem_usage
-                .store(cgroup_used_mem(&cgroup_limits), Ordering::Release);
+                .store(cgroup_limits.rss, Ordering::Release);
 
             let cpu = self.cgroup_cpu_info.calc_cpu_percent().unwrap_or_else(|| {
                 // There won't be a cgroup cpu usage if there is no limit applied to the cgroup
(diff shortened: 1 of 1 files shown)


crates/sdk-core/src/worker/tuner/resource_based.rs
```diff
@@ -498,17 +498,19 @@ struct RealSysInfoInner {
     cgroup_cpu_info: CGroupCpuInfo<CgroupV2CpuFileSystem>,
 }
 
+fn cgroup_used_mem(limits: &sysinfo::CGroupLimits) -> u64 {
+    limits.rss
+}
```
Sushisource: I don't think we need this

crates/sdk-core/src/worker/tuner/resource_based.rs
```diff
@@ -786,6 +788,22 @@ mod tests {
             .build()
     }
 
+    #[test]
+    fn cgroup_used_mem_excludes_page_cache() {
```
Sushisource: This test doesn't really do much meaningful. It'd be better to see that `cur_mem_usage` is calculated properly

crates/sdk-core/src/worker/tuner/resource_based.rs
odedkedemdreeze: Removed — the helper is gone and `refresh()` now stores `cgroup_limits.rss` inline.

crates/sdk-core/src/worker/tuner/resource_based.rs
odedkedemdreeze: Dropped this test. Instead, `cgroup_realsysinfo_uses_cgroup_limits_mem` (which runs the real `refresh()` loop against an actual cgroup in the CI memory-limited container) now computes its expected usage from `rss`, so it verifies `cur_mem_usage` is calculated properly under the new semantics. Its previous expectation was still based on `total - free`, which would have disagreed with the new gating once page cache grew.

Sushisource: I meant it would be good to verify by injecting the specific response like you had in this test that `cur_mem_usage` is calculated properly (to an exact value). The existing coverage is good from an e2e perspective but the unit test can still be useful.

Sushisource (approved):

PR #1291, opened 2026-05-24
chore(deps): replace unmaintained backoff crate with backon

pranc1ngpegasus:
## What was changed

The SDK depended on the unmaintained [`backoff`](https://crates.io/crates/backoff) crate (see [RUSTSEC-2025-0012](https://osv.dev/vulnerability/RUSTSEC-2025-0012) / [RustSec advisory](https://rustsec.org/advisories/RUSTSEC-2025-0012)) for exponential retry delays in:

- `temporalio-client` — gRPC call retry (`TonicErrorHandler` in `retry.rs`)
- `temporalio-sdk-core` — poller autoscaling error backoff (`poll_buffer.rs`)

This PR replaces `backoff` with [`backon`](https://crates.io/crates/backon) in both crates:

1. **`temporalio-client`**: `RetryOptions` builds a `backon::ExponentialBuilder`; `TonicErrorHandler` drives retry delays via `Iterator::next()` on `backon::ExponentialBackoff`. Jitter is enabled when `randomization_factor > 0` (`with_jitter()`). Long-poll fatal-error grace still uses wall-clock elapsed time via `retry_started_at`.
2. **`temporalio-sdk-core`**: Poll scaler backoff uses shared `ExponentialBuilder` constants aligned with `RetryOptions::task_poll_retry_policy()` and `throttle_retry_policy()`; state resets with `builder.build()` after successful polls.

## Why?

Fixes #1292. [RUSTSEC-2025-0012](https://osv.dev/vulnerability/RUSTSEC-2025-0012) reports that `backoff` is unmaintained and recommends migrating to `backon` for exponential backoff/retry. Each crate now depends on `backon` directly so `sdk-core` does not pull retry utilities through `client`.

## Checklist

1. Closes #1292

2. How was this tested:
   - `cargo test -p temporalio-client retry::`
   - `cargo test -p temporalio-sdk-core autoscaler_applies_backoff`

3. Any docs updates needed?
   N/A — internal dependency and retry implementation change only; public `RetryOptions` fields are unchanged.

PR #1411, opened 2026-07-17
chore(deps): replace unmaintained backoff crate with backon

### LINKED ISSUE #1292: [Maintenance] Replace unmaintained `backoff` crate with `backon` (RUSTSEC-2025-0012)

Keep the Rust SDK's dependency tree free of unmaintained crates reported by `cargo audit` / RustSec.

### Describe the issue

The SDK currently depends on the [`backoff`](https://crates.io/crates/backoff) crate for exponential retry delays. [RUSTSEC-2025-0012](https://osv.dev/vulnerability/RUSTSEC-2025-0012) ([RustSec advisory](https://rustsec.org/advisories/RUSTSEC-2025-0012)) reports that `backoff` is **unmaintained** and recommends migrating to [`backon`](https://crates.io/crates/backon) for exponential backoff/retry.

`backoff` is used in:

- **`temporalio-client`** — gRPC call retry (`TonicErrorHandler` in `crates/client/src/retry.rs`)
- **`temporalio-sdk-core`** — poller autoscaling error backoff (`crates/sdk-core/src/pollers/poll_buffer.rs`)

### Proposed change

Replace `backoff` with `backon` in both crates, using `backon::ExponentialBuilder` / `ExponentialBackoff` (`Iterator::next()` for delays). Each crate should depend on `backon` directly so `sdk-core` does not need to route retry utilities through `client`.

### Additional context

- OSV: https://osv.dev/vulnerability/RUSTSEC-2025-0012
- RustSec: https://rustsec.org/advisories/RUSTSEC-2025-0012
- Suggested replacement: https://crates.io/crates/backon

boazy:
## What was changed

The SDK depended on the unmaintained [`backoff`](https://crates.io/crates/backoff) crate ([RUSTSEC-2025-0012](https://rustsec.org/advisories/RUSTSEC-2025-0012)) for exponential retry/poll backoff in:

- `temporalio-client` — gRPC call retry (`TonicErrorHandler` in `retry.rs`)
- `temporalio-sdk-core` — poller autoscaling error backoff (`poll_buffer.rs`)

Both crates now depend on [`backon`](https://crates.io/crates/backon) directly (`default-features = false` — we only need the iterator, not backon's sleep runtime). `backon`'s `ExponentialBuilder` / `ExponentialBackoff` drives the exponential sequence, and the existing retry semantics are preserved on top of it:

- **Jitter** — `randomization_factor` keeps its documented meaning of symmetric `+/-` jitter (the `0.2` default is still `+/-20%`). `backon`'s built-in `with_jitter` is intentionally **not** used because it is asymmetric (`+(0, delay)`) and ignores the configured `randomization_factor` magnitude, which would silently reduce the public field to an on/off flag; jitter is applied manually via `rand` instead. `randomization_factor == 0.0` stays fully deterministic.
- **Total retry budget** — `max_elapsed_time` is enforced against wall-clock elapsed time plus the jittered delay (as the old implementation did), rather than `backon`'s cumulative-sleep `total_delay`, so positive jitter cannot overshoot the budget.
- **Long-poll fatal grace** — still measured against wall-clock elapsed time.

`sdk-core`'s poll backoff uses module-level `const ExponentialBuilder`s that mirror `RetryOptions::task_poll_retry_policy()` / `throttle_retry_policy()`; on a successful poll the state resets by rebuilding from the consts.

## Why

Fixes #1292 and clears [RUSTSEC-2025-0012](https://rustsec.org/advisories/RUSTSEC-2025-0012) from downstream `cargo audit` runs. `cargo tree -i backoff` now reports no matching package, and `backon` is a direct dependency of both crates.

This supersedes the abandoned #1291 and addresses its review comments:

- Public `RetryOptions` docs stay dependency-neutral and unchanged (the abandoned PR's `backon` reference is dropped), since preserving the `+/-` jitter semantics keeps the original wording accurate.
- The `randomization_factor: 0.2, // +-20% jitter.` default line is left unchanged and stays correct.
- `sdk-core` no longer stores the builders as struct fields; the `const` builders are used directly.

## Checklist

1. Closes #1292

2. How was this tested:
   - `cargo test -p temporalio-client retry::`
   - `cargo test -p temporalio-sdk-core pollers::poll_buffer`
   - `cargo integ-test out_of_order_completion_doesnt_hang`
   - `cargo fmt --all --check`, `cargo lint`, `cargo test-lint`, `cargo check`, and `cargo doc --workspace --all-features --no-deps` (with `-D warnings`)

3. Any docs updates needed?
   CHANGELOG updated under a new `### Security` entry. Public API is unchanged.

### DIFF
--- a/crates/sdk-core/src/pollers/poll_buffer.rs
+++ b/crates/sdk-core/src/pollers/poll_buffer.rs
@@ -9,7 +9,7 @@ use crate::{
         },
     },
 };
-use backoff::{SystemClock, backoff::Backoff, exponential::ExponentialBackoff};
+use backon::{BackoffBuilder, ExponentialBuilder};
 use crossbeam_utils::atomic::AtomicCell;
 use futures_util::{FutureExt, StreamExt, future::BoxFuture};
 use std::{
@@ -23,7 +23,7 @@ use std::{
     time::{Duration, Instant, SystemTime},
 };
 use temporalio_client::{
-    ERROR_RETURNED_DUE_TO_SHORT_CIRCUIT, request_extensions::NoRetryOnMatching,
+    ERROR_RETURNED_DUE_TO_SHORT_CIRCUIT, jittered, request_extensions::NoRetryOnMatching,
 };
 use temporalio_common::protos::temporal::api::{
     taskqueue::v1::PollerScalingDecision,
@@ -43,6 +43,22 @@ use tokio_stream::wrappers::UnboundedReceiverStream;
 use tokio_util::sync::CancellationToken;
 use tonic::Code;
 use tracing::Instrument;
+
+// Mirror the retry client's `RetryOptions::task_poll_retry_policy` and
+// `throttle_retry_policy` so poller autoscaling backs off on the same schedule.
+// Jitter is applied via `jittered` rather than backon's `with_jitter` to keep the
+// symmetric +/- factor behavior the retry client uses.
+const POLL_BACKOFF_RANDOMIZATION: f64 = 0.2;
+const TASK_POLL_BACKOFF: ExponentialBuilder = ExponentialBuilder::new()
+    .with_min_delay(Duration::from_millis(200))
+    .with_factor(2.0)
+    .with_max_delay(Duration::from_secs(10))
+    .without_max_times();
+const THROTTLE_POLL_BACKOFF: ExponentialBuilder = ExponentialBuilder::new()
+    .with_min_delay(Duration::from_secs(1))
+    .with_factor(2.0)
+    .with_max_delay(Duration::from_secs(10))
+    .without_max_times();
 
 type PollReceiver<T, SK> =
     Mutex<UnboundedReceiver<pollers::Result<(T, OwnedMeteredSemPermit<SK>)>>>;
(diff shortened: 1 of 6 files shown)


crates/sdk-core/src/pollers/poll_buffer.rs
```diff
@@ -44,6 +44,36 @@ use tokio_util::sync::CancellationToken;
...
+    .with_min_delay(Duration::from_secs(1))
+    .with_factor(2.0)
+    .with_max_delay(Duration::from_secs(10))
+    .without_max_times();
+
+// Reproduce the `backoff` crate's documented jitter for backward compatibility:
+//   randomized interval = retry_interval * (random value in range [1 - randomization_factor, 1 + randomization_factor])
+// Docs: https://github.com/ihrwein/backoff/blob/587e2da8fb2dcfc65ca544cb9249022c51f1406e/src/lib.rs#L4
+// Algorithm (`get_random_value_from_interval`): https://github.com/ihrwein/backoff/blob/587e2da8fb2dcfc65ca544cb9249022c51f1406e/src/exponential.rs#L61
+fn jittered(base: Duration, randomization_factor: f64) -> Duration {
```
Sushisource: This is duplicative of the same impl in the client. We can probably just pub this from there behind a `doc(hidden)`

Sushisource (commented):
Thanks for doing this, just one thing to de-dupe here and I think we're good.

crates/sdk-core/src/pollers/poll_buffer.rs
boazy: This makes sense. This code doesn't need to be duplicated.

boazy: Done in c0cabd3c. `temporalio-client` now owns a single `#[doc(hidden)] pub fn jittered`, re-exported from the crate root; sdk-core imports that helper and the duplicate implementation/comment are removed. Verified with both affected test groups (14 client retry tests, 10 poll-buffer tests), `cargo lint`, and `cargo fmt --all --check`.

Sushisource (approved):

PR #1412, opened 2026-07-17
feat(sdk): add `random` and `uuid4` in workflow context

### LINKED ISSUE #1385: [Feature Request] Provide random utilities

Generating deterministic random numbers is tricky at the moment.

Provide some common utilities for deterministic random generation in workflow context.
I think mirroring [Typescript's current utilities](https://typescript.temporal.io/api/interfaces/workflow.UnsafeRandomSource) would make sense:
- `random` for number generation
- `fill` for filling a slice
- `uuid4` for generating a UUID

### Additional context

Look at other SDKs for prior art for how of how to construct `uuid` / Random number from a random seed.

See https://docs.rs/rand/latest/rand/ for an example of common random utilities.

chris-olszewski:
## What was changed
- Add `WorkflowContext::random` and `uuid4` for use in workflows
- Add `WorkflowRandomValue` as our trait to allow generating different number types. Used to avoid exposing underlying `rand` trait.
- Remove `WorkflowContext::random_seed()`
- Make specifying child workflow optional, if omitted or empty use `uuid4` to generate one

## Why?
SDK parity

## Checklist

1. Closes #1385 

2. How was this tested:
Various unit tests

3. Any docs updates needed?
N/A


crates/sdk-core/tests/integ_tests/workflow_tests/child_workflows.rs
```diff
@@ -127,7 +124,7 @@ impl AbandonedChildBugReproParent {
                 AbandonedChildBugReproChild::run,
                 (),
                 ChildWorkflowOptions {
-                    workflow_id: "abandoned-child".to_owned(),
+                    workflow_id: Some("abandoned-child".to_owned()),
```
Sushisource: This makes me kinda sad. We probably need to convert `ChildWorkflowOptions` into a builder?

crates/workflow/src/workflow_context.rs
```diff
@@ -97,10 +137,32 @@ impl BaseWorkflowContext {
...
 
+    fn random<T>(&self) -> T
+    where
+        T: WorkflowRandomValue,
+    {
+        let random = &mut self.inner.shared.borrow_mut().random;
+        <T as private::Sealed>::sample(random)
+    }
+
+    fn uuid4(&self) -> String {
```
Sushisource, line 209: Do we not have a dep already that can do this?

chris-olszewski: `uuid` was already in `temporalio-sdk`, but not in `temporalio-workflow`. Pulled it down.

crates/sdk-core/tests/integ_tests/workflow_tests/child_workflows.rs
chris-olszewski: Yeah, realized it wasn't marked as `#[non_exhaustive]` either which needed to be addressed before GA. Added a `ChildWorkflowOptions::workflow_id` constructor for the common case where only workflow id needs to get set.

Sushisource (approved):

PR #1425, opened 2026-07-24
feat(sdk): auto-enroll default pollers into autoscaling

veeral-patel:
Ports [temporalio/sdk-go#2442](https://github.com/temporalio/sdk-go/pull/2442) to `sdk-core`, so every SDK built on it (Rust directly; TS/Python/.NET/Ruby via the c-bridge) gets it.

Supersedes #1406 — same feature, reworked per review feedback (see "How" below).

## What it does

A worker's pollers can run in one of two modes: a **fixed** number, or **autoscaling** (the server tells the worker to add/remove pollers based on load).

This makes a worker automatically use **autoscaling** when the server says the namespace supports it (`poller_autoscaling_auto_enroll`) — but only for pollers the user didn't configure. If you set a poller count or behavior yourself, it's left exactly as you asked.

## How it works

The worker figures out the right mode **once**, right after it checks in with the server (`validate()`), and builds its pollers then. Previously the pollers were built before the check-in, so #1406 had to flip the mode later at runtime; this version avoids that entirely (`poll_buffer.rs` is unchanged here).

- `worker/mod.rs`: poller setup is deferred until after the server check-in, so the mode is known up front.
- The Rust SDK now calls `validate()` on startup (it never did before, so it was missing all namespace capabilities).

## Testing
Unit tests for the mode decision and the capability parsing; c-bridge conversion tests. `cargo check`/`fmt`/`clippy` clean; sdk-core lib 368 pass, c-bridge 14 pass.

### DIFF
--- a/crates/sdk-core-c-bridge/src/worker.rs
+++ b/crates/sdk-core-c-bridge/src/worker.rs
@@ -101,24 +101,27 @@ pub struct PollerBehavior {
     pub autoscaling: *const PollerBehaviorAutoscaling,
 }
 
-impl TryFrom<&PollerBehavior> for temporalio_sdk_core::PollerBehavior {
-    type Error = anyhow::Error;
-    fn try_from(value: &PollerBehavior) -> Result<Self, Self::Error> {
-        if !value.simple_maximum.is_null() && !value.autoscaling.is_null() {
+impl PollerBehavior {
+    /// Converts an FFI poller behavior into an optional core poller behavior. Both fields null
+    /// means the poller was left unset by lang: core applies the default behavior and treats the
+    /// poller as eligible for automatic enrollment into poller autoscaling.
+    fn to_core(&self) -> anyhow::Result<Option<temporalio_sdk_core::PollerBehavior>> {
+        if !self.simple_maximum.is_null() && !self.autoscaling.is_null() {
             bail!("simple_maximum and autoscaling cannot both be non-null values");
         }
-        if let Some(value) = unsafe { value.simple_maximum.as_ref() } {
-            return Ok(temporalio_sdk_core::PollerBehavior::SimpleMaximum(
-                value.simple_maximum,
-            ));
-        } else if let Some(value) = unsafe { value.autoscaling.as_ref() } {
-            return Ok(temporalio_sdk_core::PollerBehavior::Autoscaling {
-                minimum: value.minimum,
-                maximum: value.maximum,
-                initial: value.initial,
-            });
+        if let Some(sm) = unsafe { self.simple_maximum.as_ref() } {
+            Ok(Some(temporalio_sdk_core::PollerBehavior::SimpleMaximum(
+                sm.simple_maximum,
+            )))
+        } else if let Some(a) = unsafe { self.autoscaling.as_ref() } {
+            Ok(Some(temporalio_sdk_core::PollerBehavior::Autoscaling {
+                minimum: a.minimum,
+                maximum: a.maximum,
+                initial: a.initial,
+            }))
+        } else {
+            Ok(None)
         }
-        bail!("simple_maximum and autoscaling cannot both be null values");
     }
 }
 
@@ -1245,16 +1248,10 @@ impl TryFrom<&WorkerOptions> for temporalio_sdk_core::WorkerConfig {
             // auto-cancel-activity behavior or shutdown will not occur, so we
             // always set it even if 0.
             .graceful_shutdown_period(Duration::from_millis(opt.graceful_shutdown_period_millis))
-            .workflow_task_poller_behavior(temporalio_sdk_core::PollerBehavior::try_from(
-                &opt.workflow_task_poller_behavior,
-            )?)
+            .maybe_workflow_task_poller_behavior(opt.workflow_task_poller_behavior.to_core()?)
             .nonsticky_to_sticky_poll_ratio(opt.nonsticky_to_sticky_poll_ratio)
-            .activity_task_poller_behavior(temporalio_sdk_core::PollerBehavior::try_from(
-                &opt.activity_task_poller_behavior,
-            )?)
-            .nexus_task_poller_behavior(temporalio_sdk_core::PollerBehavior::try_from(
-                &opt.nexus_task_poller_behavior,
-            )?)
+            .maybe_activity_task_poller_behavior(opt.activity_task_poller_behavior.to_core()?)
+            .maybe_nexus_task_poller_behavior(opt.nexus_task_poller_behavior.to_core()?)
             .workflow_failure_errors(if opt.nondeterminism_as_workflow_fail {
                 HashSet::from([WorkflowErrorType::Nondeterminism])
             } else {
@@ -1416,6 +1413,55 @@ mod tests {
             simple_maximum: simple as *const _,
             autoscaling: std::ptr::null(),
         }
+    }
+
+    #[test]
+    fn ffi_poller_behavior_opt_maps_unset_to_none() {
+        let unset = PollerBehavior {
+            simple_maximum: std::ptr::null(),
+            autoscaling: std::ptr::null(),
+        };
+        assert!(unset.to_core().unwrap().is_none());
+    }
+
+    #[test]
+    fn ffi_poller_behavior_opt_maps_configured_behaviors() {
+        assert_eq!(
+            simple_poller_behavior(3).to_core().unwrap(),
+            Some(temporalio_sdk_core::PollerBehavior::SimpleMaximum(3)),
+        );
+        let autoscaling = Box::leak(Box::new(PollerBehaviorAutoscaling {
+            minimum: 2,
+            maximum: 20,
+            initial: 4,
+        }));
+        let pb = PollerBehavior {
+            simple_maximum: std::ptr::null(),
+            autoscaling: autoscaling as *const _,
+        };
+        assert_eq!(
+            pb.to_core().unwrap(),
+            Some(temporalio_sdk_core::PollerBehavior::Autoscaling {
+                minimum: 2,
+                maximum: 20,
+                initial: 4,
+            }),
+        );
+    }
+
+    #[test]
+    fn ffi_poller_behavior_opt_rejects_both_set() {
+        let simple = Box::leak(Box::new(PollerBehaviorSimpleMaximum { simple_maximum: 1 }));
+        let autoscaling = Box::leak(Box::new(PollerBehaviorAutoscaling {
+            minimum: 1,
+            maximum: 2,
+            initial: 1,
+        }));
+        let pb = PollerBehavior {
+            simple_maximum: simple as *const _,
+            autoscaling: autoscaling as *const _,
+        };
+        assert!(pb.to_core().is_err());
     }
 
     fn base_worker_options(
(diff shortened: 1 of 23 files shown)


crates/sdk-core/src/worker/mod.rs
```diff
@@ -151,6 +147,13 @@ pub struct WorkerConfig {
     /// If using SimpleMaximum, Must be at least 2 when `max_cached_workflows` > 0, or is an error.
     #[builder(default = PollerBehavior::SimpleMaximum(5))]
     pub workflow_task_poller_behavior: PollerBehavior,
+    /// True if [WorkerConfig::workflow_task_poller_behavior] was left at its default (the user
+    /// configured neither a fixed poller count nor an explicit behavior). Such pollers are eligible
+    /// for automatic enrollment into poller autoscaling when the namespace advertises the
+    /// `poller_autoscaling_auto_enroll` capability. Layers above core (the Rust SDK, the c-bridge)
+    /// set this; direct core users leave it `false`.
+    #[builder(default = false)]
+    pub workflow_task_poller_behavior_auto_enroll: bool,
```
Sushisource: I don't think we need these. They can just explicitly set the option for whatever they do want, rather than disabling the enrollment explicitly.

crates/sdk-core-c-bridge/src/worker.rs
```diff
@@ -98,24 +94,28 @@ pub struct PollerBehavior {
...
 }
 
-impl TryFrom<&PollerBehavior> for temporalio_sdk_core::PollerBehavior {
+/// Converts an FFI poller behavior into an optional core poller behavior. Both fields null means
+/// the poller was left unset by lang: the caller should apply the default behavior and treat the
+/// poller as eligible for automatic enrollment into poller autoscaling.
+impl TryFrom<&PollerBehavior> for Option<temporalio_sdk_core::PollerBehavior> {
     type Error = anyhow::Error;
-    fn try_from(value: &PollerBehavior) -> Result<Self, Self::Error> {
+    fn try_from(value: &PollerBehavior) -> anyhow::Result<Self> {
```
Sushisource: This won't compile (and doesn't, in CI).

crates/sdk-core/src/worker/mod.rs
```diff
@@ -414,14 +418,14 @@ pub struct Worker {
     client: Arc<dyn WorkerClient>,
     /// Worker instance key, unique identifier for this worker
     worker_instance_key: Uuid,
-    /// Manages all workflows and WFT processing. None if workflow polling is disabled
-    workflows: Option<Workflows>,
-    /// Manages activity tasks for this worker/task queue
-    at_task_mgr: Option<WorkerActivityTasks>,
+    /// The poller-dependent subsystems (workflows, activity task manager, nexus manager). Built
+    /// lazily once namespace capabilities are known so effective poller behavior can be resolved.
+    pollers: std::sync::OnceLock<PollerSubsystems>,
```
Sushisource: These really aren't (just) pollers. `TaskSubsystems` would be better I think.

crates/sdk-core/src/worker/mod.rs
```diff
@@ -866,131 +1108,85 @@ impl Worker {
...
-            client.workers().register_worker(
-                client_worker_registrator.clone(),
-                config.skip_client_worker_set_check,
-            )?;
+        // If shutdown was already initiated before the pollers were built (e.g. a worker that
+        // never ran validate() and only starts polling as part of draining during shutdown),
+        // propagate the shutdown signal to the freshly-built subsystems so they don't hang waiting
+        // for work that will never arrive. `initiate_shutdown` deliberately does not force a build,
+        // so it can't have done this for us.
+        if shutdown_token.is_cancelled() {
```
Sushisource: If the task subsystems were never built, we shouldn't need to do anything at all I think?

crates/sdk-core/src/worker/mod.rs
```diff
@@ -414,14 +418,14 @@ pub struct Worker {
...
-    /// Manages all workflows and WFT processing. None if workflow polling is disabled
-    workflows: Option<Workflows>,
-    /// Manages activity tasks for this worker/task queue
-    at_task_mgr: Option<WorkerActivityTasks>,
+    /// The poller-dependent subsystems (workflows, activity task manager, nexus manager). Built
+    /// lazily once namespace capabilities are known so effective poller behavior can be resolved.
+    pollers: std::sync::OnceLock<PollerSubsystems>,
+    /// Builds [PollerSubsystems] on first access. Taken exactly once.
+    #[allow(clippy::type_complexity)]
+    poller_builder: parking_lot::Mutex<Option<Box<dyn FnOnce() -> PollerSubsystems + Send>>>,
```
Sushisource: I think we could avoid needing this, and I think we can kill two birds with one stone and get rid of `PollerInitInputs` at the same time.

The `pollers` (now `task_subsystems`) field, should instead be a https://doc.rust-lang.org/beta/std/sync/struct.LazyLock.html

The closure given to the `LazyLock` constructor can be the same one you give now, and `build_poller_subsystems` can be inlined to eliminate the need for `PollerInitInputs`

crates/sdk-core/src/worker/mod.rs
```diff
@@ -2007,7 +2215,11 @@ struct ClientWorkerRegistrator {
     worker_instance_key: Uuid,
     slot_provider: SlotProvider,
     heartbeat_manager: Option<WorkerHeartbeatManager>,
-    cancel_activity_callback: Option<CancelActivityCallback>,
+    /// Slot filled by the deferred poller build with the activity cancellation callback. Shared
+    /// with the worker so it can be resolved lazily once the activity task manager is constructed.
+    cancel_activity_slot: Arc<std::sync::OnceLock<CancelActivityCallback>>,
+    /// Whether this worker polls for (non-local) activities at all.
+    activities_enabled: bool,
```
Sushisource: We don't need a separate bool here. Simply not initializing the OnceLock is sufficient.

veeral-patel: Done, removed activities_enabled

crates/sdk-core/src/worker/mod.rs
veeral-patel: Good point, renamed to TaskSubsystems

crates/sdk-core/src/worker/mod.rs
veeral-patel: Good point, I refactored the three poller-behavior fields on WorkerConfig to be `Option<PollerBehavior>` now

crates/sdk-core/src/worker/mod.rs
```diff
@@ -419,14 +413,16 @@ pub struct Worker {
...
-    /// Manages activity tasks for this worker/task queue
-    at_task_mgr: Option<WorkerActivityTasks>,
+    /// The task subsystems (workflows, activity task manager, nexus manager). Built lazily once
+    /// namespace capabilities are known so effective poller behavior can be resolved.
+    #[allow(clippy::type_complexity)]
+    task_subsystems:
+        std::sync::LazyLock<TaskSubsystems, Box<dyn FnOnce() -> TaskSubsystems + Send>>,
+    /// True once `task_subsystems` has been built. Lets shutdown/introspection peek without forcing
+    /// a build (MSRV-safe stand-in for `LazyLock::get`, which is only stable since 1.94).
+    task_subsystems_built: Arc<AtomicBool>,
```
Sushisource: It's fine for us to upgrade to requiring 1.94 luckily, since Rust SDK is still not GA.

I'm also not sure we actually _need_ to use `get` though. Since we force loading, it should be fine to just deref the lock where it's used.

crates/sdk-core/src/worker/mod.rs
```diff
@@ -419,14 +413,16 @@ pub struct Worker {
...
     worker_instance_key: Uuid,
-    /// Manages all workflows and WFT processing. None if workflow polling is disabled
-    workflows: Option<Workflows>,
-    /// Manages activity tasks for this worker/task queue
-    at_task_mgr: Option<WorkerActivityTasks>,
+    /// The task subsystems (workflows, activity task manager, nexus manager). Built lazily once
+    /// namespace capabilities are known so effective poller behavior can be resolved.
+    #[allow(clippy::type_complexity)]
+    task_subsystems:
+        std::sync::LazyLock<TaskSubsystems, Box<dyn FnOnce() -> TaskSubsystems + Send>>,
```
Sushisource: Nit: remove the `std::sync::` prefix and just import directly.

crates/sdk-core/src/worker/mod.rs
```diff
@@ -845,32 +758,274 @@ impl Worker {
...
+        // paths can peek without forcing a build. (`LazyLock::get` would do this, but it is only
+        // stable as of Rust 1.94, above our MSRV; this flag is the MSRV-safe equivalent.)
+        let task_subsystems_built = Arc::new(AtomicBool::new(false));
+
+        // Build the poller-dependent subsystems lazily. This closure runs once, after namespace
+        // capabilities have been fetched (see `Worker::validate`), so the effective poller behavior
+        // can be resolved with knowledge of those capabilities. Everything it needs is captured by
+        // clone (or moved) *before* the synchronous heartbeat manager below consumes the originals.
+        let task_subsystems_builder: Box<dyn FnOnce() -> TaskSubsystems + Send> = {
+            let task_subsystems_built = task_subsystems_built.clone();
```
Sushisource: Fairly sure not all of these need to actually be `clone()`d  and some of them could just be moved into the closure.

crates/sdk-core-c-bridge/src/worker.rs
veeral-patel: Fixed this

crates/sdk-core/src/worker/mod.rs
veeral-patel: Done, dropped LazyLock::get and now just using deref at the use sites

I checked and deref/force/new are all ≤1.80 so should be good to keep MSRV stays at 1.88 for now.

crates/sdk-core/src/worker/mod.rs
veeral-patel: Done, imported LazyLock/OnceLock directly and dropped the std::sync:: prefixes

crates/sdk-core/src/worker/mod.rs
veeral-patel: Done, turned task_subsystems into a LazyLock. Removed PollerInitInputs and inlined build_poller_subsystems

crates/sdk-core/src/worker/mod.rs
veeral-patel: Good point, removed this unnecessary special case handling

crates/sdk-core/src/worker/mod.rs
veeral-patel: Looked into this, tried removing the clone calls but it causes compiler errors. Unfortunately won't be possible in this case but thanks for flagging!

Sushisource (approved):

veeral-patel:
I manually tested this PR locally in all 5 Rust-based SDKs (Rust, Python, .NET, Ruby, TypeScript) successfully.

Steps:
1. I cloned each SDK locally and modified it to point to my local sdk-rust (core) repo with these changes
2. Started local Temporal server, with `frontend.pollerAutoscalingAutoEnroll` on and then off (confirmed via grpcurl)
3. Added temporary eprintln! line in `resolve_effective_behavior` (in sdk-rust) to print the resolved poller behavior (fixed or auto)

Result — ✅ all five pass

```
| SDK | 7233 (OFF) | 7234 (ON) |
|---|---|---|
| Rust / Python / TypeScript / Ruby / .NET | SimpleMaximum(5) | Autoscaling{1,100,5} |
```

Pending: We need to upgrade the Rust core version in all the Rust based SDKs. SDK team will handle this (discussed with Spencer)

PR #1426, opened 2026-07-25
feat(sdk): client interceptor

### LINKED ISSUE #1138: [Feature Request] Client-side interceptors
No interception of client-side operations (start, signal, query, update, cancel, terminate, describe, list). Only `WorkerInterceptor` exists today.

chris-olszewski:
## What was changed
Adding `ClientInterceptor` along with methods on client interceptors vec on client options to register the interceptors.

Adding `RpcOptions` to various client calls options.

Add options structs to some client calls that were missing them.

## Checklist

1. Closes #1138 

2. How was this tested:
Added some unit tests to ensure connection level rpc options get overridden by call specific ones. Very basic integration test making sure interceptor order is respected on calls.

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/client/src/interceptors.rs
+++ b/crates/client/src/interceptors.rs
@@ -1045,7 +1045,7 @@ impl HeartbeatAsyncActivityInput {
 /// }
 ///
 /// let _options = ClientOptions::new("my-namespace")
-///     .client_interceptors(vec![Arc::new(StartTimeout) as Arc<dyn ClientInterceptor>])
+///     .client_interceptors(vec![Arc::new(StartTimeout)])
 ///     .build();
 /// ```
 pub trait ClientInterceptor: Send + Sync + 'static {
--- a/crates/client/src/interceptors.rs
+++ b/crates/client/src/interceptors.rs
@@ -1,0 +1,1549 @@
+//! Interceptors for high-level client operations.
+
+use crate::{
+    ActivityHeartbeatResponse, ActivityIdentifier, WorkflowCancelOptions, WorkflowCountOptions,
+    WorkflowDescribeOptions, WorkflowFetchHistoryOptions, WorkflowQueryOptions,
+    WorkflowSignalOptions, WorkflowStartError, WorkflowStartOptions, WorkflowStartUpdateOptions,
+    WorkflowTerminateOptions,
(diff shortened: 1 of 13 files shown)


crates/client/src/workflow_handle.rs
```diff
@@ -1022,46 +1223,65 @@ impl<CT, T: 'static> WorkflowUpdateHandle<CT, T>
...
     where
         T: temporalio_common::data_converters::TemporalDeserializable,
     {
-        let outcome = if let Some(known) = &self.known_outcome {
-            known.clone()
-        } else {
-            // The server's internal long-poll timeout (~60s) may expire before the update
-            // completes, returning a response with outcome: None. Keep polling until we
-            // get an actual outcome.
-            loop {
```
Sushisource, line 1036: This comment got yeeted but is actually fairly useful

crates/client/src/interceptors.rs
```diff
@@ -0,0 +1,1549 @@
...
+            self.signal_name,
+            self.args,
+            self.options,
+        )
+    }
+
+    /// Attempt to access the signal arguments as a concrete type.
+    pub fn args_ref<T: Any>(&self) -> Option<&T> {
+        self.args.as_any().downcast_ref()
+    }
```
Sushisource: Might make sense to put these common `args` related functions into a trait?

crates/client/src/interceptors.rs
```diff
@@ -0,0 +1,1549 @@
...
+///         Box::pin(async move {
+///             input.rpc_options.timeout = Some(Duration::from_secs(10));
+///             let output = next.run(input).await?;
+///             Ok(output)
+///         })
+///     }
+/// }
+///
+/// let _options = ClientOptions::new("my-namespace")
+///     .client_interceptors(vec![Arc::new(StartTimeout) as Arc<dyn ClientInterceptor>])
```
Sushisource: This cast is slightly unfortunate. I see using `options_mut` to push one later that can avoid it, but, that feels awkward when first constructing. I wonder if there's any trick we can play to make this work w/o the cast?

Sushisource (approved):
Generally looking good to me, a few small things.

PR #1430, opened 2026-07-28
Flake fix investigation

Sushisource:


crates/sdk-core/src/core_tests/workflow_tasks.rs
```diff
@@ -1517,17 +1517,8 @@ async fn la_resolution_after_wft_not_found_during_eviction() {
             .is_err()
     );
 
-    core.complete_workflow_activation(WorkflowActivationCompletion::from_cmds(
+    core.complete_workflow_activation(WorkflowActivationCompletion::empty(
```
chris-olszewski, line 1520: Was the scheduled LA racing with the shutdown?

crates/sdk-core/src/worker/activities/local_activities.rs
```diff
@@ -466,17 +471,21 @@ impl LocalActivityManager {
...
                 continue;
             }
-            // If this request originated from a local backoff task, clear the entry for it. We
-            // don't await the handle because we know it must already be done, and there's no
-            // meaningful value.
-            dat.la_info
-                .get_mut(&id)
-                .map(|lai| lai.backing_off_task.take());
+            {
+                let la_info = dat.la_info.get_mut(&id).expect("Activity must exist");
```
chris-olszewski: Moving this to an `if let Some`/`match` with the `contains_key` check above can remove the `expect` and the `contains_key`.

crates/sdk-core/src/worker/workflow/workflow_stream.rs
```diff
@@ -176,14 +195,19 @@ impl WFStream {
 
                 activations.extend(maybe_act);
                 activations.extend(state.reconcile_buffered());
+                actions.extend(
+                    activations
+                        .into_iter()
+                        .map(WorkflowStreamAction::Activation),
+                );
```
chris-olszewski, line 202: Might be worth a comment explaining importance of ordering activations after `FailUnstoredWft` variants.

chris-olszewski (approved):

crates/sdk-core/src/worker/workflow/workflow_stream.rs
Sushisource: It actually doesn't matter really - it's just incidentally the most convenient way to combine the different types into the enum

PR #1431, opened 2026-07-29
fix(sdk): make payload codec failable

chris-olszewski:
## What was changed
- Update `PayloadCodec` trait methods to now return `Result<_, PayloadConversionError>`
- Update `AsyncPayloadVisitor::visit` to be faillable, `PayloadVisitable::visit_payloads_mut` is now responsible for failing when a visitor fails. Generated `PayloadVisitable` impls now have `?` after their `visit` calls.
- activity/workflow completions will fall back to non-encoded failure if encoding fails. From what I can tell this is behavior of the other SDKs
- Adding `?`

## Why?
Discovered this while working on client interceptors, codecs certainly can fail.

## Checklist

1. Closes N/A

2. How was this tested:
Added integration tests to verify codec failures result in failed tasks that are retried. Small unit tests to ensure we fallback to un-encoded failures if a codec operation fails along with ensuring codec failures abort streaming responses.

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/client/src/schedules.rs
+++ b/crates/client/src/schedules.rs
@@ -1088,7 +1088,7 @@ where
                 self.client.data_converter().codec(),
                 &SerializationContextData::Workflow,
             )
-            .await;
+            .await?;
         }
 
         ScheduleDescription::new(
@@ -1143,7 +1143,7 @@ where
                             handle.client.data_converter().codec(),
                             &SerializationContextData::Workflow,
                         )
-                        .await;
+                        .await?;
                         let description = ScheduleDescription::new(
                             response,
                             handle.client.data_converter().clone(),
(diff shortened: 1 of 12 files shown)


crates/client/src/schedules.rs
```diff
@@ -1530,38 +1530,45 @@ impl Client {
...
-                        ListSchedulesPageInput {
-                            maximum_page_size: page_size,
-                            query,
-                            next_page_token: next_page_token.clone(),
-                            rpc_options,
-                        },
-                        Next::new({
-                            let mut rpc_client = client.clone();
-                            move |input: ListSchedulesPageInput| -> BoxFuture<
+fn list_schedules_stream<CT>(client: CT, opts: ListSchedulesOptions) -> ListSchedulesStream
```
chris-olszewski, line 1537: Don't love this, but easiest way to allow us to hit this function with a fake client.

crates/client/src/workflow_handle.rs
```diff
@@ -1306,7 +1309,7 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_helpers::XorCodec;
+    use crate::test_helpers::{FailingCodec, XorCodec};
```
chris-olszewski:
```suggestion
    use crate::test_helpers::XorCodec;
```

Sushisource (approved):
Huh, not sure how I just skipped that lol

chris-olszewski:
> Huh, not sure how I just skipped that lol

Not sure how I only noticed it last week 🙈

PR #1433, opened 2026-07-30
feat(core): emit workflow task duration logging

jmaeagle99:
## What was changed

When a WFT takes longer than 5 seconds, emit a TMPRL1104 warn log with external storage data. Threshold can be configured via the `TEMPORAL_WORKFLOW_TASK_DURATION_WARN_SECONDS` environment variable.

## Why?

Help customers better identify causes of long running workflow tasks.

## Checklist

1. How was this tested: Unit tests, [Python SDK PR](https://github.com/temporalio/sdk-python/pull/1698)


crates/sdk-core/src/worker/workflow/managed_run.rs
```diff
@@ -1529,13 +1541,195 @@ impl From<WFMachinesError> for RunUpdateErr {
...
+                payload_upload_count = ul.map(|m| m.payload_count).unwrap_or_default(),
+                payload_upload_size = ul.map(|m| m.total_size_bytes).unwrap_or_default(),
+                payload_upload_duration = storage_millis(ul),
+                payload_upload_drivers = ?ul.map(|m| sorted(&m.driver_names)).unwrap_or_default(),
+                $msg,
+            )
+        };
+    }
+
+    if duration > Duration::from_secs(5) {
```
Sushisource: IMO we should allow this to be configurable, even if it's just via some `TEMPORAL_WFT_DURATION_WARNING` env var.

crates/sdk-core/src/worker/workflow/managed_run.rs
```diff
@@ -1529,13 +1541,195 @@ impl From<WFMachinesError> for RunUpdateErr {
...
+        emit!(
+            warn,
+            "[TMPRL1104] {run_id}:{event_id}:{attempt} Workflow task duration exceeded 5 seconds."
+        );
+    } else {
+        emit!(trace, "Workflow task duration information.");
```
Sushisource: IMO I don't think we really need this at all honestly. I'm not sure what situation this would be useful in other than very low-level debugging stuff at which point you'd probably just be instrumenting the code with specific logging anyway

jmaeagle99: I don't think customers have the ability to instrument their code along the workflow task seams in any deterministic manner. However, I buy that WFT latency is an already existing metric (not logging) that customers can use to measure the headline value; the external storage information is not included, but we do plan to have those metrics added at some point. They would be independent of the WFT latency metric, so no correlation, but it's probably sufficient without discrete per-WFT data. I'll drop the trace log.

jmaeagle99: Dropped the trace log

crates/sdk-core/src/worker/workflow/managed_run.rs
jmaeagle99: Added `TEMPORAL_WORKFLOW_TASK_DURATION_WARN_SECONDS` env var.

Sushisource (approved):

PR #1435, opened 2026-07-31
Add eager dispatch test for activity cancel via worker commands

rkannan82:
## What
Enable eager activity dispatch in the `activity_cancel_delivered_without_heartbeat` test by parameterizing it with rstest. The existing test now runs both with and without eager dispatch.

## Why
The server bug preventing eager dispatch on worker command task queues has been fixed (temporalio/temporal#10634). We should now have test coverage for cancel-via-worker-commands with eager dispatch.

## How did you test it?
Compilation check. These are integration tests that run against a real server in CI.


crates/sdk-core/tests/shared_tests/mod.rs
```diff
@@ -470,7 +470,6 @@ pub(crate) async fn activity_cancel_delivered_without_heartbeat() {
                         ..Default::default()
                     })
                     .cancellation_type(ActivityCancellationType::WaitCancellationCompleted)
-                    // TODO: enable eager dispatch once server supports it for worker commands.
                     .do_not_eagerly_execute(true)
```
Sushisource, line 474: We could just use `rstest` to make this a variant of the existing test and save a bunch of lines

Sushisource (approved):

PR #1436, opened 2026-08-01
feat(sdk): Add plugins

### LINKED ISSUE #1356: [Feature Request] Support Plugins

Provide a way for declaring and consuming Plugins.

### Additional context

https://docs.temporal.io/develop/plugins-guide

This cannot be done/designed before interceptors are completed.

chris-olszewski:
### Prefactors
- Added `Debug` implementations to `Client`/`Worker`
- Moving all interceptors to be registered via `WorkerOptions` instead of directly on the Worker. Allow multiple `WorkerInterceptor`s per Worker instead of only a single one. This is done in the first commit of the PR.

### Main Changes

Added `ClientPlugin` and `WorkerPlugin` traits which allows users to implement plugins. The stored types are `ErasedClientPlugin` and `ErasedWorkerPlugin`. The worker plugin is just a named `Arc<dyn WorkerPlugin>`, but `ErasedClientPlugin` carries `WorkerPluginData` which is essentially a named `Any` that carries worker plugins. This is needed since the client crate doesn't have access to the types necessary to define the worker plugin trait. On the actual worker plugin consumption side, we only respect the private `PropagatedWorkerPlugin`.

Client/worker only plugins will be registered via `ClientOptions::client_plugin(plugin: impl ClientPlugin)`/`WorkerOptions::worker_plugin(plugin: impl WorkerPlugin)`.

Dual plugins are registered using`ClientAndWorkerPlugin` struct which enabled propagating plugins that from a `Client` to any `Worker` that is constructed with that client. These plugins are used with `ClientOptions::plugin`.  e.g.
```
struct MyPlugin;
impl ClientPlugin for MyPlugin {..}
impl WorkerPlugin for MyPlugin {..}
let plugin = ClientAndWorkerPlugin::new(MyPlugin);
let client_options = ClientOptions::new("default").plugin(plugin).build();
let client = Client::connect(.., client_options)?;
// Worker will be configured with the worker plugin
let worker = Worker::new(client, ...)?;
```

## Why?
SDK Parity

I do not love the special type for dual plugins, but I didn't find a better solution that didn't encourage misuse.

We could get rid of the special `client_plugin`/`worker_plugin` constructors and force users to build the erased types themselves if we want to remove some surface area here. We could offer blanket implementation conversion for any plugin implementers into the erased type, but that opens the door for misuse of dual plugins e.g.
```
// Will not propagate worker plugins
ClientOptions::new(..).plugin(MyPlugin).build();
// Will propagate worker plugins
ClientOptions::new(..).plugin(ClientAndWorkerPlugin::new(MyPlugin)).build();
```

## Checklist

1. Closes #1356 
2. How was this tested:
Added unit tests around error propagation. Very basic integration tests just verifying plugin fields are respected.

3. Any docs updates needed?
Plugin docs can now be updated to include Rust samples once this is released.

### DIFF
--- a/crates/sdk/src/interceptors.rs
+++ b/crates/sdk/src/interceptors.rs
@@ -51,7 +51,7 @@ mod activity_execution_value {
 ///
 /// Advanced usage only.
 #[async_trait::async_trait(?Send)]
-pub trait WorkerInterceptor {
+pub trait WorkerInterceptor: Send + Sync {
     /// Called every time a workflow activation completes (just before sending the completion to
     /// core).
     async fn on_workflow_activation_completion(&self, _completion: &WorkflowActivationCompletion) {}
--- a/crates/sdk/src/interceptors.rs
+++ b/crates/sdk/src/interceptors.rs
@@ -50,6 +50,7 @@ mod activity_execution_value {
 /// Implementors can intercept certain actions that happen within the Worker.
 ///
 /// Advanced usage only.
+/// **Experimental:** This API may change or be removed.
 #[async_trait::async_trait(?Send)]
 pub trait WorkerInterceptor: Send + Sync {
     /// Called every time a workflow activation completes (just before sending the completion to
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -101,7 +101,10 @@ pub use crate::{
         ChildWorkflowStartError, OutgoingActivityError, OutgoingError, OutgoingWorkflowError,
         RetryState, TimeoutType, WorkerCreateError, WorkflowRegistrationError, WorkflowSignalError,
     },
-    plugins::{ClientAndWorkerPlugin, SimplePlugin, SimplePluginBuilder, WorkerPlugin},
+    plugins::{
+        ClientAndWorkerPlugin, SimplePlugin, SimplePluginBuilder, SimplePluginOption, WorkerPlugin,
+        WorkflowDefinitions,
+    },
 };
 pub use temporalio_client::Namespace;
 pub use temporalio_sdk_core::CoreRuntime as Runtime;
@@ -128,7 +131,6 @@ use crate::{
     workflow_executor::{TaskHandle, WorkflowExecutor},
     workflow_future::start_workflow,
     workflow_interceptors::WorkflowInterceptorConstructor,
-    workflow_registry::WorkflowDefinitions,
 };
 use anyhow::{Context, anyhow, bail};
 use futures_util::{FutureExt, StreamExt, TryFutureExt, TryStreamExt};
@@ -1681,11 +1683,47 @@ mod tests {
         }
     }
 
+    #[workflow]
+    #[derive(Default)]
+    struct OtherWorkflow;
+
+    #[workflow_methods]
+    impl OtherWorkflow {
+        #[run]
+        async fn run(_ctx: &mut WorkflowContext<Self>) -> WorkflowResult<()> {
+            Ok(())
+        }
+    }
+
     #[test]
     fn test_workflow_registration() {
         let _ = WorkerOptions::new("task_q")
             .register_workflow::<MyWorkflow>()
             .unwrap();
+    }
+
+    #[test]
+    fn simple_plugin_workflow_function_merges_definitions() {
+        let plugin = SimplePlugin::builder("simple")
+            .workflows(|existing: Option<WorkflowDefinitions>| {
+                assert!(existing.is_some());
+                let mut workflows = WorkflowDefinitions::new();
+                workflows.register_workflow::<OtherWorkflow>().unwrap();
+                workflows
+            })
+            .build();
+        let client_options = ClientOptions::new("namespace").build();
+        let mut worker_options = WorkerOptions::new("task_q")
+            .register_workflow::<MyWorkflow>()
+            .unwrap()
+            .worker_plugin(plugin)
+            .build();
+
+        crate::plugins::apply_worker_plugins(&client_options, &mut worker_options).unwrap();
+
+        let workflows = format!("{:?}", worker_options.workflows());
+        assert!(workflows.contains("MyWorkflow"));
+        assert!(workflows.contains("OtherWorkflow"));
     }
 
     #[test]
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -101,9 +101,7 @@ pub use crate::{
         ChildWorkflowStartError, OutgoingActivityError, OutgoingError, OutgoingWorkflowError,
         RetryState, TimeoutType, WorkerCreateError, WorkflowRegistrationError, WorkflowSignalError,
     },
-    plugins::{
-        ClientAndWorkerPlugin, ErasedWorkerPlugin, SimplePlugin, SimplePluginBuilder, WorkerPlugin,
-    },
+    plugins::{ClientAndWorkerPlugin, SimplePlugin, SimplePluginBuilder, WorkerPlugin},
 };
 pub use temporalio_client::Namespace;
 pub use temporalio_sdk_core::CoreRuntime as Runtime;
@@ -206,7 +204,7 @@ pub struct WorkerOptions {
     workflow_interceptor_constructors: Vec<WorkflowInterceptorConstructor>,
 
     #[builder(field)]
-    worker_plugins: Vec<ErasedWorkerPlugin>,
+    worker_plugins: Vec<Arc<dyn WorkerPlugin>>,
 
     #[builder(field)]
     client_plugin_names: HashSet<String>,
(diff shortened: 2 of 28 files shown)


crates/sdk/src/interceptors.rs
```diff
@@ -51,7 +51,7 @@ mod activity_execution_value {
 ///
 /// Advanced usage only.
 #[async_trait::async_trait(?Send)]
-pub trait WorkerInterceptor {
+pub trait WorkerInterceptor: Send + Sync {
```
chris-olszewski, line 55: We could omit this from plugins and keep the traits bounds as is

Sushisource: I think it's worth keeping but we should mark this guy experimental. We know we want to add a similar concept to all SDKs but the hooks are definitely gonna change a bit.

crates/sdk/src/lib.rs
```diff
@@ -175,9 +183,24 @@ pub struct WorkerOptions {
...
+    worker_interceptors: Vec<Arc<dyn WorkerInterceptor>>,
+
+    #[builder(field)]
+    activity_inbound_interceptors: Vec<Arc<dyn ActivityInboundInterceptor>>,
+
     #[builder(field)]
     workflow_interceptor_constructors: Vec<WorkflowInterceptorConstructor>,
 
+    #[builder(field)]
+    worker_plugins: Vec<ErasedWorkerPlugin>,
```
Sushisource: Does this make sense here? Would it be possible to have a worker plugin add itself to the worker options and infinite loop? 😆

crates/sdk/src/plugins.rs
```diff
@@ -0,0 +1,634 @@
...
+    }
+}
+
+/// A type-erased worker plugin.
+///
+/// **Experimental:** This API may change or be removed.
+#[derive(Clone)]
+pub struct ErasedWorkerPlugin {
+    worker: Arc<dyn WorkerPlugin>,
+}
```
Sushisource: Do we really need this? Would we ever add anything to this besides it just being an `Arc` wrapper?

crates/sdk/src/plugins.rs
```diff
@@ -0,0 +1,634 @@
...
+/// #     fn name(&self) -> &str { "my-plugin" }
+/// # }
+/// let plugin = ClientAndWorkerPlugin::new(MyPlugin);
+/// let client_options = ClientOptions::new("default").plugin(plugin).build();
+/// # let _ = client_options;
+/// ```
+///
+/// **Experimental:** This API may change or be removed.
+#[derive(Clone)]
+pub struct ClientAndWorkerPlugin {
```
Sushisource, line 69: Yeah I think in theory people could avoid needing to know about this directly by impl

`impl From<T> for ClientAndWorkerPlugin where T: ClientPlugin + WorkerPlugin`

And then accept anything that `Into<ClientAndWorkerPlugin>`

Or, maybe better, do

`trait ClientAndWorkerPlugin { ... }; impl<T> ClientAndWorkerPlugin for T where T: ClientPlugin + WorkerPlugin {...}`

crates/sdk/src/plugins.rs
```diff
@@ -0,0 +1,634 @@
...
+}
+
+/// Builds a [`SimplePlugin`] from common SDK configuration values.
+///
+/// **Experimental:** This API may change or be removed.
+pub struct SimplePluginBuilder {
+    definition: SimplePluginDefinition,
+}
+
+impl SimplePluginBuilder {
```
Sushisource: Why not bon?

crates/sdk/src/plugins.rs
chris-olszewski: It was only there for symmetry with the `ErasedClientPlugin`

crates/sdk/src/lib.rs
chris-olszewski: It is private so the only way that a plugin could mess with the list would be to rebuild from scratch via `*options = WorkerOptions::new(...).worker_plugins(vec![AnotherPlugin]).build()`. 

Even if a user did that we wouldn't recurse since we take the plugin list before we start applying it. Added a unit test verifying this along with a warning comment.

crates/sdk/src/plugins.rs
chris-olszewski: What would the be the trait bounds in the client crate for accepting a `ClientAndWorkerPlugin`? The trait would need to be defined in the client crate, but the blanket impl needs to be in the SDK so we have access to `WorkerPlugin`, but won't be able to because of the ophan rule.

We could do this via a trait extension defined in the SDK crate that adds a `combined_plugin<ClientPlugin + WorkerPlugin>` to client options? Still some concern about user misuse of passing a combined plugin to `client_plugin` not realizing the woker half of the plugin is getting dropped.

Sushisource: Ah, I see what you mean. Yeah, that being the case I can see why this makes sense.

crates/sdk/src/lib.rs
Sushisource: Well I was less concerned about the infinite loop thing and more just wondering why they actually need to get stored here. Doesn't the plugin apply the options stuff earlier on via the client? Or I guess there's another worker-only path that I probably just didn't quite grok in my review

Sushisource (approved):
My last comment isn't blocking. Might be good to have @tconley1428 do a pass.

crates/sdk/src/lib.rs
chris-olszewski: Yeah, we still allow for worker plugins to be explicitly configured on the worker directly instead of the client. Could force everything through the client and remove this though if that's desired.

crates/sdk/src/plugins.rs
```diff
@@ -0,0 +1,617 @@
...
+/// **Experimental:** This API may change or be removed.
+pub trait WorkerPlugin: Send + Sync + 'static {
+    /// Stable, unique name used to identify this plugin.
+    fn name(&self) -> &str;
+
+    /// Configure worker options.
+    ///
+    /// Worker plugin registrations are captured before this method is called. Altering plugins in
+    /// this method does not change which plugins are applied.
+    fn configure_worker_options(&self, _options: &mut WorkerOptions) -> Result<(), PluginError> {
```
tconley1428, line 32: In other languages this also provides a mechanism for running around the worker lifetime

crates/sdk/src/plugins.rs
```diff
@@ -0,0 +1,617 @@
...
+/// **Experimental:** This API may change or be removed.
+#[derive(Clone, bon::Builder)]
+#[builder(state_mod(vis = "pub"))]
+pub struct SimplePlugin {
+    #[builder(start_fn, into)]
+    name: String,
+    #[builder(field)]
+    data_converter: Option<DataConverter>,
+    #[builder(field)]
+    client_interceptors: Vec<Arc<dyn ClientInterceptor>>,
```
tconley1428: I have found separate client/worker interceptor lists to be tricky to use when you have an interceptor which does both (unless that's not a thing in rust) - instead I took both as a single list.

crates/sdk/src/plugins.rs
```diff
@@ -0,0 +1,617 @@
...
+/// # let _ = client_options;
+/// ```
+///
+/// Worker configuration declared by this combined plugin is automatically propagated through the
+/// configured client when a worker is created.
+///
+/// **Experimental:** This API may change or be removed.
+#[derive(Clone, bon::Builder)]
+#[builder(state_mod(vis = "pub"))]
+pub struct SimplePlugin {
```
tconley1428, line 196: This will be more limited than other simple plugins, which allow you to see the previous value before setting using functions. That's especially common around data converter manipulation.

crates/sdk/src/lib.rs
```diff
@@ -189,9 +195,24 @@ pub struct WorkerOptions {
     #[builder(field)]
     workflows: WorkflowDefinitions,
 
+    #[builder(field)]
```
tconley1428, line 199: I don't follow why adding plugins added new interceptor options here. How did worker interceptors work before?

crates/sdk/src/lib.rs
```diff
@@ -189,9 +195,24 @@ pub struct WorkerOptions {
...
     workflow_interceptor_constructors: Vec<WorkflowInterceptorConstructor>,
 
+    #[builder(field)]
+    worker_plugins: Vec<Arc<dyn WorkerPlugin>>,
+
+    #[builder(field)]
+    client_plugin_names: HashSet<String>,
+
+    #[builder(field)]
+    plugins_applied: bool,
```
tconley1428: What are these things for? They sound like internal state, but this seems like a public configuration object.

crates/sdk/src/plugins.rs
```diff
@@ -0,0 +1,617 @@
...
+};
+use temporalio_common::{WorkflowDefinition, data_converters::DataConverter};
+use temporalio_workflow::runtime::entry::WorkflowImplementation;
+
+/// Configures worker options before the worker is created.
+///
+/// Use [`ClientAndWorkerPlugin`] for plugins that target both clients and workers.
+///
+/// **Experimental:** This API may change or be removed.
+pub trait WorkerPlugin: Send + Sync + 'static {
```
tconley1428, line 24: Does the rust SDK have a replayer?

crates/sdk/src/lib.rs
chris-olszewski: Previously they were registered by modifying an already constructed worker. First commit of this PR switched so all interceptors are set via the `WorkerOptions` instead of modifying already created workers.

crates/sdk/src/lib.rs
chris-olszewski: Was able to remove the worker one by being a little more careful around the worker lifecycle. Still need the client one as the client in activity context is constructed. This bit of state isn't visible to end users.

crates/sdk/src/plugins.rs
chris-olszewski: A type could implement both. Hard thing is we would need to add a new enum with each variant having a different `Arc<dyn *Interceptor>` combination. End users would need to manually construct these as we couldn't provide conversions that would distinguish between types that are implement one interceptor or those that implement multiple.

PR #1438, opened 2026-08-03
feat(client): add update-with-start support

### LINKED ISSUE #1136: [Feature Request] Update with Start
Support atomic "start workflow + send update" operation.

discosultan:
Closes #1136

Adds atomic update-with-start to the client:

- `Client::start_update_with_start_workflow` and `Client::execute_update_with_start_workflow`, backed by the `ExecuteMultiOperation` RPC, with `WithStartWorkflowOperation` describing the start request.
- Multi-operation failures are decoded and attributed to the failing operation (`WorkflowUpdateWithStartError::Start`/`Update`).
- Interceptor support via `ClientInterceptor::update_with_start_workflow`.

Tested against local dev server.

### DIFF
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -61,6 +61,7 @@ pub use interceptors::{
     SignalWithStartWorkflowInput, SignalWorkflowInput, StartWorkflowInput, StartWorkflowOutput,
     StartWorkflowUpdateInput, StartWorkflowUpdateOutput, TemporalClientValue,
     TerminateWorkflowInput, TriggerScheduleInput, UnpauseScheduleInput, UpdateScheduleInput,
+    UpdateWithStartWorkflowInput, UpdateWithStartWorkflowOutput,
 };
 pub use metrics::{LONG_REQUEST_LATENCY_HISTOGRAM_NAME, REQUEST_LATENCY_HISTOGRAM_NAME};
 pub use options_structs::*;
@@ -115,7 +116,11 @@ use crate::{
     worker::ClientWorkerSet,
 };
 use errors::*;
-use futures_util::{future::BoxFuture, stream, stream::Stream};
+use futures_util::{
+    future::{BoxFuture, try_join},
+    stream,
+    stream::Stream,
+};
 use http::Uri;
 use parking_lot::RwLock;
 use std::{
@@ -129,7 +134,7 @@ use std::{
     time::{Duration, SystemTime},
 };
 use temporalio_common::{
-    ActivityDefinition, HasWorkflowDefinition, SignalDefinition, UntypedActivity,
+    ActivityDefinition, HasWorkflowDefinition, SignalDefinition, UntypedActivity, UpdateDefinition,
     data_converters::{DataConverter, SerializationContext, SerializationContextData},
     payload_visitor::decode_payloads,
     protos::{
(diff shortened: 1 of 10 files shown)


chris-olszewski (changes requested):
Looking at #1506 made me realize we should offer that level of type safety for update with start as well.

I am looking at addressing that issue immediately and would suggest waiting until that is resolved before so the same pattern can be applied for update-with-start.

crates/client/src/lib.rs
```diff
@@ -1156,6 +1162,73 @@ impl Client {
...
+    pub async fn start_update_with_start_workflow<W, U>(
+        &self,
+        update: U,
+        input: U::Input,
+        options: WorkflowStartUpdateOptions,
+        start_workflow_operation: WithStartWorkflowOperation<W>,
+    ) -> Result<
+        (
+            WorkflowHandle<Self, W>,
+            WorkflowUpdateHandle<Self, U::Output>,
```
chris-olszewski: IMO we should just return the update handle here.

crates/client/src/lib.rs
```diff
@@ -1156,6 +1162,73 @@ impl Client {
...
+    /// workflow.
+    ///
+    /// See [Client::start_update_with_start_workflow] for details on option requirements.
+    pub async fn execute_update_with_start_workflow<W, U>(
+        &self,
+        update: U,
+        input: U::Input,
+        options: WorkflowExecuteUpdateOptions,
+        start_workflow_operation: WithStartWorkflowOperation<W>,
+    ) -> Result<(WorkflowHandle<Self, W>, U::Output), WorkflowUpdateWithStartError>
```
chris-olszewski: Similar to above, I think we should match other SDKs and just return update result.

crates/client/src/lib.rs
```diff
@@ -1156,6 +1162,73 @@ impl Client {
...
+    /// update to complete and returning its result along with a [`WorkflowHandle`] for the
+    /// workflow.
+    ///
+    /// See [Client::start_update_with_start_workflow] for details on option requirements.
+    pub async fn execute_update_with_start_workflow<W, U>(
+        &self,
+        update: U,
+        input: U::Input,
+        options: WorkflowExecuteUpdateOptions,
+        start_workflow_operation: WithStartWorkflowOperation<W>,
```
chris-olszewski: Annoying, but I think we need to define a specific `WorkflowUpdateWithStartOptions` instead of reusing `WorkflowExecuteUpdateOptions` and `WorkflowStartOptions`. Both have `rpc_options`/`headers` and it is unclear which one would be used. Also allows us to ensure that id conflict policy is set at the type level instead of a runtime check.

crates/client/src/lib.rs
```diff
@@ -1156,6 +1162,73 @@ impl Client {
...
+        )
+        .await
+    }
+
+    /// Start a workflow and send it an update as a single atomic operation, waiting for the
+    /// update to complete and returning its result along with a [`WorkflowHandle`] for the
+    /// workflow.
+    ///
+    /// See [Client::start_update_with_start_workflow] for details on option requirements.
+    pub async fn execute_update_with_start_workflow<W, U>(
```
chris-olszewski, line 1223: Just to keep the same shape as `signal_with_start_workflow` I think this should take workflow and workflow input as explicit args instead of adding `WithStartWorkflowOperation`

chris-olszewski (changes requested):

crates/client/src/lib.rs
discosultan: Updated

crates/client/src/lib.rs
discosultan: Done

crates/client/src/lib.rs
discosultan: Updated

crates/client/src/lib.rs
discosultan: Done

chris-olszewski (approved):
Thank you for the contribution!

PR #1449, opened 2026-08-04
fix(sdk): CancellableFuture is no longer generic

### LINKED ISSUE #1440: CancellableFuture should remove its generic type parameter
`CancellableFuture` is currently defined as:
```rust
pub trait CancellableFuture<T>: Future<Output = T> + FusedFuture {
    fn cancel(&self);
}
```

By adding a top-level generic parameter instead of reusing `Future`'s associated `Output` type, `CancellableFuture` is not intuitive to use in other generic contexts. It should be redefined as:
```rust
trait CancellableFuture: Future + FusedFuture {
    fn cancel(&self);
}
```

<details><summary>Expanded reasoning</summary>

By adding a `T` generic parameter to `CancellableFuture` you are signaling that a type can implement `CancellableFuture` for multiple values of `T`, generically. However, this does not make sense because `CancellableFuture<T>` requires the type to implement `Future` with `Output` set to `T`, and `Future` can only be implemented once on a type.

</details>

For example, this code snippet fails to compile:
[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=35d7a59823ab4eb79d37eb2bef175186)
```rust
struct CancelWithWorkflow<A> {
    activity: A,
    cancel_requested: bool,
}

impl<A, T> Future for CancelWithWorkflow<A>
where
    A: CancellableFuture<T> + Unpin
{
    type Output = T;

    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        // Implementation is omitted for this example
        let this = self.get_mut();
        Pin::new(&mut this.activity).poll(cx)
    }
}
```
```
error[E0207]: the type parameter `T` is not constrained by the impl trait, self type, or predicates
  --> src/main.rs:20:9
   |
20 | impl<A, T> Future for CancelWithWorkflow<A>
   |         ^ unconstrained type parameter
   |
help: use the type parameter `T` in the `CancelWithWorkflow` type and use it in the type definition
   |
15 ~ struct CancelWithWorkflow<A, T> {
16 |     activity: A,
...
19 |
20 ~ impl<A, T> Future for CancelWithWorkflow<A, T>
   |
```

To fix this error requires adding a `Future<Output = T>` bound on `T`, which should be implied by `CancellableFuture<T>`:
```rust
impl<A, T> Future for CancelWithWorkflow<A>
where
    A: CancellableFuture<T> + Future<Output = T> + Unpin,
```

After redefining `CancellableFuture`, the following impl compiles:
[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8f89291ee9d16baaa3b8a2a90619fe6e)
```rust
impl<A> Future for CancelWithWorkflow<A>
where
    A: CancellableFuture + Unpin
```

chris-olszewski:
## What was changed
Remove generic from `CancellableFuture` along with reason variant.

## Why?
Generic was bound to exactly the associated `Output` type of the `Future` trait

## Checklist

1. Closes #1440 
2. How was this tested:
`rustc`

3. Any docs updates needed?
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ to docs, or any other relevant information.
 
 ## Unreleased
 
+### Breaking Changes :boom:
+* `CancellableFuture` and `CancellableFutureWithReason` now use the inherited `Future::Output`
+  associated type instead of a generic output parameter.
+
 ## [0.6.0] - 2026-08-04
 
 ### Added
--- a/crates/workflow/src/workflow_context.rs
+++ b/crates/workflow/src/workflow_context.rs
@@ -720,7 +720,7 @@ impl BaseWorkflowContext {
     pub fn timer<T: Into<TimerOptions>>(
         &self,
         opts: T,
-    ) -> impl CancellableFuture<TimerResult> + use<T> {
+    ) -> impl CancellableFuture<Output = TimerResult> + use<T> {
         let input = StartTimerInput::new(opts.into());
         let base_ctx = self.clone();
         let next = WorkflowNext::new(move |input: StartTimerInput| {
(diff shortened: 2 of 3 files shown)


Sushisource (approved):
Hahaha I wonder why I did that lol

PR #1450, opened 2026-08-04
chore(sdk): mark remaining `Options` types as `non_exhaustive`

chris-olszewski:
## What was changed
Go through remaining public `Options` structs and ensure they all:
 - are marked with `#[non_exhaustive]`
 - use `bon` to provide a builder
 - adding `PartialEq` implementations where applicable

Each commit is reviewable on its own.

## Why?
Without the `non_exhaustive` attribute, any expansion to these options would be a breaking change.

## Checklist

1. Closes N/A

2. How was this tested:
👀, did add unit tests that `T::default() == T::builder().build()` to ensure those don't diverge. 

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/client/src/options_structs.rs
+++ b/crates/client/src/options_structs.rs
@@ -270,13 +270,16 @@ impl Default for DnsLoadBalancingOptions {
 
 /// Payload size limit options for a connection.
 /// NOTE: Experimental
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, bon::Builder)]
+#[non_exhaustive]
 pub struct PayloadLimitsOptions {
     /// Warning threshold (bytes) for the size of an outbound payload-bearing field; over-threshold
     /// fields are logged but still sent to server. Defaults to 512 KiB. Set to `0` to disable.
+    #[builder(default = 512 * 1024)]
     pub payloads_warn_size: u64,
     /// Warning threshold (bytes) for outbound memo sizes; over-threshold memos are logged but still
     /// sent to server. Defaults to 2 KiB. Set to `0` to disable.
+    #[builder(default = 2 * 1024)]
     pub memo_warn_size: u64,
 }
 
--- a/crates/client/src/options_structs.rs
+++ b/crates/client/src/options_structs.rs
@@ -247,10 +247,11 @@ impl Default for ClientKeepAliveOptions {
 }
 
 /// Options for DNS-based load balancing.
-#[derive(Clone, Debug)]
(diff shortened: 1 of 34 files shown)


crates/workflow/src/workflow_context/options.rs
```diff
@@ -696,7 +701,9 @@ impl TimerOptions {
 }
 
 /// Options for Nexus Operations
-#[derive(Default, Debug, Clone)]
```
chris-olszewski, line 699: Removing `Default` here as I think that was compensating for the lack of builder. There isn't really a valid "default" here since empty endpoint/service/operation isn't sensible.

crates/client/src/options_structs.rs
```diff
@@ -621,3 +631,60 @@ pub struct WorkflowCountOptions {
     #[builder(default)]
     pub rpc_options: RpcOptions,
 }
+
+#[cfg(test)]
+mod tests {
```
Sushisource: Prolly don't really need these?

crates/client/src/retry.rs
```diff
@@ -27,21 +27,28 @@ pub const RETRYABLE_ERROR_CODES: [Code; 7] = [
 const LONG_POLL_FATAL_GRACE: Duration = Duration::from_secs(60);
 
 /// Configuration for retrying requests to the server
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, bon::Builder)]
+#[non_exhaustive]
 pub struct RetryOptions {
     /// initial wait time before the first retry.
+    #[builder(default = Duration::from_millis(100))]
```
Sushisource, line 34: Should we just remove the manual default impl? (I know that's around somewhere)

crates/common/src/envconfig.rs
```diff
@@ -1049,6 +1057,30 @@ mod tests {
     use std::io::Write;
     use tempfile::NamedTempFile;
 
+    #[test]
+    fn load_client_config_options_default_matches_builder_default() {
+        assert_eq!(
```
Sushisource: Looks like this is coming up in a number of places. IMO we should either just remove the hand-rolled defaults, or, make them delegate to the builder one (and not need these tests).

Sushisource (approved):
Nice!

crates/common/src/envconfig.rs
chris-olszewski: Ah yeah that's cleaner.

PR #1451, opened 2026-08-04
feat(sdk): add cancellation token

chris-olszewski:
## What was changed
- Added `WorkflowCancellationToken` as an SDK owned cancellation token. Each workflow context will provide access to the root one. `WorkflowCancellationToken::child`. To shield an action from cancellation pass a disconnected token created via `WorkflowCancellationToken::new()`. I didn't add a method for shielding entire blocks so users will need to pass through each token individually.
- Updated `wait_condition` to take an `WaitConditionOptions` containing optional cancellation token, if omitted uses workflow token. It also now returns `Result<(), WorkflowCancellationError>` to indicate whether the condition was satisfied or cancelled.
- `WorkflowContext::cancelled` now returns `Option<String>` instead of `String` so we no longer coarse `None` to `""`.
- `ExternalWorkflowHandle::signal` takes a cancellation token via a new `SignalWorkflowOptions`

## Why?
Before this PR, developers need to directly race any future against `ctx.cancelled()` if they want to be aware of cancellations. This is awkward and easy to forget and then unintentionally ignore cancellation requests.

## Checklist

1. Closes N/A

2. How was this tested:
Additional unit tests and integration tests surrounding cancellation.

3. Any docs updates needed?
Updated readmes/examples.


crates/workflow/src/workflow_context.rs
```diff
@@ -2062,13 +2121,24 @@ impl<W> WorkflowContext<W> {
     ///
     /// The condition closure receives an immutable reference to the workflow state,
     /// which is borrowed only for the duration of each poll (not across await points).
+    /// By default, the wait inherits workflow cancellation.
     pub fn wait_condition<'a>(
         &'a self,
         mut condition: impl FnMut(&W) -> bool + 'a,
-    ) -> impl FusedFuture<Output = ()> + 'a {
+        options: WaitConditionOptions,
```
Sushisource, line 2170: I know we haven't done this for much other stuff but maybe we should do a `wait_condition_with_options` since I imagine 99.9% of the time people are just passing Default here?

CHANGELOG.md
```diff
@@ -22,6 +22,12 @@ to docs, or any other relevant information.
 ### Breaking Changes :boom:
 * `CancellableFuture` and `CancellableFutureWithReason` now use the inherited `Future::Output`
   associated type instead of a generic output parameter.
+### Added
```
Sushisource, line 73: Missing breaking changes to wait_condition (but, maybe we won't need them per my other comment)

crates/workflow/src/cancellation.rs
```diff
@@ -0,0 +1,220 @@
...
+    pub fn new() -> Self {
+        Self {
+            inner: Default::default(),
+        }
+    }
+
+    /// Create a token that is cancelled when this token is cancelled.
+    pub fn child_token(&self) -> Self {
+        let child = Self::new();
+        let weak_child = Rc::downgrade(&child.inner);
```
Sushisource, line 68: Lol weak child

crates/workflow/src/workflow_context.rs
```diff
@@ -2819,6 +2882,8 @@ where
                         })
                     }
                     ChildWorkflowStartStatus::Failed(f) => {
+                        let mut result_future = pending.common.result_future;
+                        result_future.unregister_cancellation();
```
Sushisource, line 2928: Why do we unregister here and cancelled but not completed? It also makes no sense to cancel after completed, right? Maybe this can be done more generically? Just thinking out loud here, not a big concern.

crates/workflow/src/workflow_context.rs
chris-olszewski: Eh, yeah was on the fence with this one. Okay with having this as an exception just since it is a bit of an eyesore for 99% of cases.

crates/workflow/src/workflow_context.rs
chris-olszewski: IIUC the completed here is just that the start child workflow operation completed. We still want to retain the ability to cancel the child workflow via the token.

> Maybe this can be done more generically?

As in a `Drop` impl or something of the sort?

Sushisource: Ah yes great point about it just being start. 

Yeah, I was thinking something like `Drop`, but, I don't have any real issue w/ this.

Sushisource (approved):

PR #1456, opened 2026-08-05
:boom: Use heartbeat details field on activity failure instead of force-flush

### LINKED ISSUE #293: [Feature Request] Make use of new API for including last heartbeat details when failing activity tasks

Server now supports this. Using it can simplify internal logic. Of course we still have to support the old path for backwards compat...

Sushisource:
… flush

## What was changed
Title

I chose to no longer use the old path since this field has existed for a _long_ time now. A warning is issued if pointed at a too-old server.

## Why?
Avoids unnecessary heartbeat RPC and makes final details atomic with failure.

We still force a flush when the activity is completing as cancelled, because no such field exists on the cancel response (probably it should, but, fairly niche situation).

## Checklist

1. Closes https://github.com/temporalio/sdk-rust/issues/293
2. How was this tested:
Existing & new tests

3. Any docs updates needed?

### DIFF
--- a/crates/sdk-core/src/worker/activities.rs
+++ b/crates/sdk-core/src/worker/activities.rs
@@ -46,6 +46,7 @@ use temporalio_common::{
             activity_task::{ActivityCancelReason, ActivityCancellationDetails, ActivityTask},
         },
         temporal::api::{
+            common::v1::{Payload, Payloads},
             failure::v1::{
                 ApplicationFailureInfo, CanceledFailureInfo, Failure, failure::FailureInfo,
             },
@@ -117,6 +118,8 @@ struct RemoteInFlightActInfo {
     local_timeouts_task: Option<JoinHandle<()>>,
     /// Used to reset the local heartbeat timeout every time we record a heartbeat
     timeout_resetter: Option<Arc<Notify>>,
+    /// The most recent heartbeat received from lang, independently of whether it has been sent.
+    last_heartbeat_details: Option<Vec<Payload>>,
     /// The permit from the max concurrent semaphore
     _permit: UsedMeteredSemPermit<ActivitySlotKind>,
 }
@@ -140,6 +143,7 @@ impl RemoteInFlightActInfo {
             known_not_found: false,
             local_timeouts_task: None,
             timeout_resetter: None,
+            last_heartbeat_details: None,
             _permit: permit,
         }
     }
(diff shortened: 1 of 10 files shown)


crates/sdk-core/src/worker/activities.rs
```diff
@@ -347,15 +351,22 @@ impl WorkerActivityTasks {
             if let Some(jh) = act_info.local_timeouts_task {
                 jh.abort()
             };
+            // Cancellation responses cannot carry heartbeat details, so normal cancellations must
+            // flush them separately. Worker-shutdown cancellations are reported as failures below.
             let should_flush = !known_not_found
+                && matches!(&status, aer::Status::Cancelled(_))
```
eamsden, line 357: An oversized throttled heartbeat is now skipped here and added to the failure request instead. The server drops those details, but the worker ignores that response and reports the activity’s original failure. Previously this path produced a retryable PayloadsTooLarge failure. The integration test passes on main and fails on this PR.

Before skipping the heartbeat flush, validate the pending details using the same configured limits as a heartbeat RPC. If they exceed the limit, omit them and report the existing retryable PayloadsTooLarge failure; otherwise include them in the activity-failure request. This will require exposing the payload check currently held inside `PayloadLimitsClient`.

```rust
#[tokio::test]
async fn oversized_throttled_heartbeat_failure_is_reported() {
    let mut starter =
        init_core_and_create_wf("oversized_throttled_heartbeat_failure_is_reported").await;
    let core = starter.get_worker().await;
    core.validate().await.unwrap();
    let task_q = starter.get_task_queue().to_string();

    let task = core.poll_workflow_activation().await.unwrap();
    core.complete_workflow_activation(WorkflowActivationCompletion::from_cmd(
        task.run_id,
        ScheduleActivity {
            seq: 0,
            activity_id: "act-1".to_string(),
            activity_type: "dontcare".to_string(),
            task_queue: task_q,
            schedule_to_close_timeout: Some(prost_dur!(from_secs(60))),
            heartbeat_timeout: Some(prost_dur!(from_secs(10))),
            retry_policy: Some(RetryPolicy {
                maximum_attempts: 1,
                ..Default::default()
            }),
            ..Default::default()
        }
        .into(),
    ))
    .await
    .unwrap();

    let task = core.poll_activity_task().await.unwrap();
    assert_matches!(task.variant, Some(act_task::Variant::Start(_)));

    // The first heartbeat starts the throttle window so the oversized heartbeat remains pending
    // until the activity failure is reported. Three MiB exceeds the default blob limit without
    // exceeding the gRPC transport limit.
    core.record_activity_heartbeat(ActivityHeartbeat {
        task_token: task.task_token.clone(),
        details: vec!["small".into()],
    });
    core.record_activity_heartbeat(ActivityHeartbeat {
        task_token: task.task_token.clone(),
        details: vec![vec![0_u8; 3 * 1024 * 1024].into()],
    });

    core.complete_activity_task(ActivityTaskCompletion {
        task_token: task.task_token,
        result: Some(ActivityExecutionResult::fail(Failure::application_failure(
            "original activity failure".to_string(),
            false,
        ))),
    })
    .await
    .unwrap();

    let task = core.poll_workflow_activation().await.unwrap();
    let failure = assert_matches!(
        task.jobs.as_slice(),
        [WorkflowActivationJob {
            variant: Some(workflow_activation_job::Variant::ResolveActivity(
                ResolveActivity {
                    result: Some(ActivityResolution {
                        status: Some(act_res::Status::Failed(activity_result::Failure {
                            failure: Some(failure),
                        })),
                    }),
                    ..
                }
            )),
        }] => failure
    );

    let cause = failure
        .cause
        .as_deref()
        .expect("activity failure has a cause");
    assert_matches!(
        &cause.failure_info,
        Some(FailureInfo::ApplicationFailureInfo(info))
            if info.r#type == "PayloadsTooLarge" && !info.non_retryable
    );

    core.complete_execution(&task.run_id).await;
    core.handle_eviction().await;
    drain_pollers_and_shutdown(&core).await;
}
```

eamsden (changes requested):

eamsden (approved):

PR #1457, opened 2026-08-06
Collect information about runtime environment

Sushisource:
## What was changed
Add runtime environment collection that is sent once on the first worker heartbeat.

## Why?
Allows us to gain insight into how prevalent certain versions of runtimes are, etc, so we can prioritize support properly

Opt-out flag included.

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -207,6 +207,20 @@ jobs:
         with:
           compose-file: ./etc/docker/docker-compose-ci.yaml
       - run: cargo integ-test docker_
+      - name: Verify worker environment detection inside Docker
+        run: |
+          docker run --rm \
+            --user "$(id -u):$(id -g)" \
+            --env HOME="$HOME" \
+            --env CARGO_HOME="$HOME/.cargo" \
+            --env RUSTUP_HOME="$HOME/.rustup" \
+            --env PROTOC \
+            --env TEMPORAL_TEST_EXPECT_DOCKER=true \
+            --volume "$HOME:$HOME" \
+            --volume "$GITHUB_WORKSPACE:$GITHUB_WORKSPACE" \
+            --workdir "$GITHUB_WORKSPACE" \
+            ubuntu:24.04 \
+            "$HOME/.cargo/bin/cargo" integ-test worker_heartbeat_multiple_workers
 
   examples:
     name: "Build and run examples"
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -207,20 +207,31 @@ jobs:
         with:
           compose-file: ./etc/docker/docker-compose-ci.yaml
       - run: cargo integ-test docker_
+        env:
+          TMPDIR: ${{ runner.temp }}
       - name: Verify worker environment detection inside Docker
         run: |
+          integ_test_binary="$(
+            find "$GITHUB_WORKSPACE/target/debug/deps" -maxdepth 1 -type f \
+              -name 'integ_tests-*' -perm -111 \
+              -printf '%T@ %p\n' | sort -nr | head -n 1 | cut -d' ' -f2-
+          )"
+          integ_runner_binary="$(
+            find "$GITHUB_WORKSPACE/target/debug/deps" -maxdepth 1 -type f \
+              -name 'integ_runner-*' -perm -111 \
+              -printf '%T@ %p\n' | sort -nr | head -n 1 | cut -d' ' -f2-
+          )"
+          test -x "$integ_test_binary"
+          test -x "$integ_runner_binary"
           docker run --rm \
-            --user "$(id -u):$(id -g)" \
-            --env HOME="$HOME" \
-            --env CARGO_HOME="$HOME/.cargo" \
-            --env RUSTUP_HOME="$HOME/.rustup" \
-            --env PROTOC \
+            --env TMPDIR="$RUNNER_TEMP" \
             --env TEMPORAL_TEST_EXPECT_DOCKER=true \
-            --volume "$HOME:$HOME" \
+            --volume "$RUNNER_TEMP:$RUNNER_TEMP" \
             --volume "$GITHUB_WORKSPACE:$GITHUB_WORKSPACE" \
             --workdir "$GITHUB_WORKSPACE" \
-            rust:1.94-bookworm \
-            "$HOME/.cargo/bin/cargo" integ-test worker_heartbeat_multiple_workers
+            ubuntu:24.04 \
+            "$integ_runner_binary" --test-executable "$integ_test_binary" \
+              worker_heartbeat_multiple_workers
 
   examples:
     name: "Build and run examples"
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -219,7 +219,7 @@ jobs:
             --volume "$HOME:$HOME" \
             --volume "$GITHUB_WORKSPACE:$GITHUB_WORKSPACE" \
             --workdir "$GITHUB_WORKSPACE" \
-            ubuntu:24.04 \
+            rust:1.94-bookworm \
             "$HOME/.cargo/bin/cargo" integ-test worker_heartbeat_multiple_workers
 
   examples:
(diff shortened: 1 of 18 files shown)


.github/workflows/per-pr.yml
```diff
@@ -207,6 +207,20 @@ jobs:
         with:
           compose-file: ./etc/docker/docker-compose-ci.yaml
       - run: cargo integ-test docker_
+      - name: Verify worker environment detection inside Docker
```
chris-olszewski, line 212: Can we not alter the job above to run these tests?

crates/sdk-core/src/worker/mod.rs
```diff
@@ -2388,11 +2423,15 @@ impl WorkerHeartbeatManager {
             }
             worker_heartbeat
         });
+        let heartbeat_success_callback: HeartbeatSuccessCallback = Arc::new(move || {
+            environment_info.lock().take();
```
chris-olszewski, line 2428: Very obvious in the context of this review why we take env info here, but might be helpful to have a comment for future selves to why we clear it.

crates/sdk-core/src/lib.rs
```diff
@@ -185,6 +190,22 @@ pub struct RuntimeOptions {
     /// Interval must be between 1s and 60s, inclusive.
     #[builder(required, default = Some(Duration::from_secs(60)))]
     heartbeat_interval: Option<Duration>,
+    /// Disable including runtime, hosting, and platform information in worker heartbeats.
+    #[builder(default)]
+    disable_environment_info: bool,
+    /// Runtime information supplied by language SDK bridges.
+    #[doc(hidden)]
```
chris-olszewski, line 197: Unsure I understand the benefit of doc hiding this?

crates/sdk-core/src/lib.rs
```diff
@@ -185,6 +190,22 @@ pub struct RuntimeOptions {
...
+    #[builder(skip = vec![environment::native_runtime()])]
+    runtimes: Vec<Runtime>,
+}
+
+impl RuntimeOptions {
+    /// Supplies runtime information from a language SDK bridge.
+    #[doc(hidden)]
```
chris-olszewski, line 204: Same as above, unsure we need to hide things that are part of core.

chris-olszewski (approved):
Maybe call out the lang SDKs should ensure they respect env vars that should disable this reporting when they take this commit since it is on by default.

.github/workflows/per-pr.yml
Sushisource: Which job? The cloud-tests one? I put it in here just b/ it's also like... docker related? 🤷

crates/sdk-core/src/lib.rs
Sushisource: Just b/c it's only meant to be used by language-layer implementers, not users directly specifying runtime options.

crates/sdk-core/src/lib.rs
Sushisource: The problem is that these are re-exported at the SDK level. I was considering just making a _new_ Runtime options at the SDK level and converting it into this. If you're good with that I'll do that, it's a breaking change but probably makes sense for GA

.github/workflows/per-pr.yml
chris-olszewski: 🤦 nvm, misread YAML and though `cargo integ-test docker_` was running inside a docker container already

crates/sdk-core/src/lib.rs
```diff
@@ -168,10 +173,10 @@ pub struct CoreRuntime {
     runtime: Option<tokio::runtime::Runtime>,
     runtime_handle: tokio::runtime::Handle,
     heartbeat_interval: Option<Duration>,
+    environment_info: Option<Arc<EnvironmentInfo>>,
```
jmaeagle99, line 176: Do we have any documented expectations of what the format of the version fields within this structure should be? I believe they are strings so technically anything could be put into them. But should we be more prescriptive or give guidance e.g. only n-part version numbers without suffixes or labels?

crates/sdk-core/src/lib.rs
```diff
@@ -294,16 +333,34 @@ impl CoreRuntime {
     pub fn new_assume_tokio_initialized_telem(
```
jmaeagle99, line 333: nit: Update the comments to describe that runtime telemetry is collected and that if callers want to disable they should use `new_assume_tokio_initialized_telem_with_environment` instead.

crates/sdk-core/src/environment.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+    if has_env("WEBSITE_SITE_NAME") {
+        environments.push(hosting_environment(
+            HostingEnvironmentType::AzureAppService,
+            env_value("WEBSITE_PLATFORM_VERSION"),
+        ));
+    }
+    if let Some(version) = env_value("FUNCTIONS_EXTENSION_VERSION") {
+        environments.push(hosting_environment(
+            HostingEnvironmentType::AzureFunctions,
```
jmaeagle99, line 84: Azure Functions run on Azure App Service, so something running in Azure Functions will report both environments with the current detection logic. _Probably_ okay, but raising this to make a more explicit decision. Possibly worth a comment.

crates/sdk-core/src/environment.rs
```diff
@@ -0,0 +1,211 @@
...
+            None,
+        ));
+    }
+    if has_env("WEBSITE_SITE_NAME") {
+        environments.push(hosting_environment(
+            HostingEnvironmentType::AzureAppService,
+            env_value("WEBSITE_PLATFORM_VERSION"),
+        ));
+    }
+    if let Some(version) = env_value("FUNCTIONS_EXTENSION_VERSION") {
```
jmaeagle99, line 82: This value likely has a `~` character in front of the version number, which means that anything in that version range is acceptable. See https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#functions_extension_version for example. Consider stripping the character before reporting.

Same applies to all of Azure App Service based version environment variables.

crates/sdk-core/src/environment.rs
```diff
@@ -0,0 +1,211 @@
...
+        environments.push(hosting_environment(
+            HostingEnvironmentType::AzureFunctions,
+            Some(version),
+        ));
+    }
+    if has_any_env(&["CONTAINER_APP_NAME", "CONTAINER_APP_JOB_NAME"]) {
```
jmaeagle99, line 88: Similar to https://github.com/temporalio/sdk-rust/pull/1457/changes#r3738543609, Azure Container Apps run on K8s today, so likely to report both as AzureContainerApps and K8s. Possibly worth a comment.

crates/sdk-core/src/lib.rs
Sushisource: On some level we can't really control that since they're just coming from the env vars or stuff we grab. We could try hard to parse things or strip stuff (like your other comment) but I'm sort of inclined to just let whatever the env says be passed through and we can clean things up on the analytics side.

crates/sdk-core/src/environment.rs
Sushisource: Per other comment I think we should deal with grouping / representation on the analytics backend

crates/sdk-core/src/lib.rs
Sushisource: That one isn't public (and there are already too many of these lol). In that case they'll need to just use normal `new` without a pre-initialized telem instance.

crates/sdk-core/src/environment.rs
Sushisource: It's very much the intent that environments like that should show up as both. Can add comments.

crates/sdk-core/src/lib.rs
jmaeagle99: Sure, wasn't saying that we should enforce it somehow, but provide guidance on what is more conducive for us to consume it down stream. We'll do our best in our SDKs, but if other SDK authors are building on top of our core and want to provide the data, they might want some guidance as to what is expected. But I guess if we don't enforce, then any value can be provided, and we just have to deal with it anyway.

PR #1459, opened 2026-08-07
Use typed worker run errors

chris-olszewski:
## What
Remove `anyhow` in favor of our own typed error for `Worker::run`.

## Why
Remove `anyhow` from public API.

I kept it to a very small number of variants vs each possible failure point. Other than verification, none of these are really meaningfully recoverable by the user.

### DIFF
--- a/crates/sdk/src/error.rs
+++ b/crates/sdk/src/error.rs
@@ -2,6 +2,9 @@
 
 pub use crate::workflow_registry::WorkflowRegistrationError;
 use temporalio_client::PluginApplyError;
+use temporalio_sdk_core::{
+    CompleteActivityError, CompleteWfError, PollError, WorkerValidationError,
+};
 
 /// Errors that can occur while creating a worker.
 ///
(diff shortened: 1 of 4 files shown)


crates/sdk/src/error.rs
```diff
@@ -2,6 +2,7 @@
 
 pub use crate::workflow_registry::WorkflowRegistrationError;
 use temporalio_client::PluginApplyError;
+use temporalio_sdk_core::WorkerValidationError;
```
xumaple: should this be pub? since it's being exposed by WorkerRunError

crates/sdk/src/error.rs
```diff
@@ -16,6 +17,25 @@ pub enum WorkerCreateError {
...
+
+/// Errors that can occur while running a worker.
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum WorkerRunError {
+    /// Worker validation failed before polling began.
+    #[error("worker validation failed: {0}")]
+    Validation(#[source] WorkerValidationError),
+    /// An unrecoverable error occurred while running the worker.
+    #[error("{message}: {source}")]
```
xumaple: generally source shouldn't need to be part of the display error message iirc? this way I think there might be a double display, not sure haven't tried it

crates/sdk/src/error.rs
```diff
@@ -16,6 +17,25 @@ pub enum WorkerCreateError {
...
+    Fatal {
+        /// Worker operation that could not continue.
+        message: String,
```
xumaple, line 32: I think this type can just be &'static str, considering everywhere it's currently being used?

crates/sdk/src/lib.rs
```diff
@@ -952,7 +980,10 @@ impl Worker {
                     },
                 )
                 .await
-                .context("Workflow futures encountered an error")
+                .map_err(|source| WorkerRunError::Fatal {
```
xumaple: I don't think this is needed, given that the async block above is already returning WorkerRunError?

crates/sdk/src/lib.rs
```diff
@@ -936,13 +958,19 @@ impl Worker {
...
+                                source: e.into(),
+                            })?;
                             // Eviction is normal workflow lifecycle - workflows loop waiting for
                             // eviction after completion to manage cache cleanup
                             if let Err(e) = result
                                 && !matches!(e, WorkflowTermination::Evicted)
                             {
-                                return Err(anyhow::Error::new(e));
+                                return Err(WorkerRunError::Fatal {
+                                    message: "workflow execution failed".into(),
```
xumaple, line 954: I think this codepath is executed on continueasnew? in which case "workflow execution failed" or even `WorkerRunError::Fatal` is not a very good name

crates/sdk/src/error.rs
chris-olszewski: This wouldn't double display with a standard. Wanted this to behave similar to `anyhow`'s context so end users can still see the underlying error.

crates/sdk/src/error.rs
chris-olszewski: It could, but slightly limiting for future changes if it needs to be dynamic for some reason.

crates/sdk/src/lib.rs
chris-olszewski: No, continue as new results in a command being created, it won't shut down the worker.

Sushisource (approved):

PR #1462, opened 2026-08-07
fix(sdk): reject updates when validators panic

chris-olszewski:
## What was changed
Catch panics in update validators and treat them as rejections instead of retrying them.

## Why?
Match other SDK behavior described from features

> Panics/Exceptions out of validation handlers are equivalent to rejections from a durability perspective while panics/exceptions from the main execution function are the equivalent of panicking/throwing in a workflow task.

source: https://github.com/temporalio/features/tree/main/features/update/task_failure

## Checklist

1. Closes N/A

2. How was this tested:
Updated existing unit test to verify update fails instead of being retried.
Added new unit test verifying panic during actual handler is retried. (We previously just had assertion around panic after an update)

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/workflow/src/runtime/instance.rs
+++ b/crates/workflow/src/runtime/instance.rs
@@ -28,10 +28,12 @@ use futures_util::{
     future::{Fuse, LocalBoxFuture},
 };
 use std::{
-    any::Any,
+    any::{Any, TypeId},
     cell::RefCell,
     collections::HashMap,
+    fmt::{Display, Formatter},
     future::{Future, ready},
+    panic::AssertUnwindSafe,
     pin::Pin,
     rc::Rc,
     sync::Arc,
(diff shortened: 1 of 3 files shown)


crates/workflow/src/runtime/instance.rs
```diff
@@ -1088,6 +1097,43 @@ where
     GuestWorkflowInstance::<W>::instantiate(payloads, converter, base_ctx)
 }
 
+/// Attempts to turn caught panics into something printable
```
chris-olszewski, line 1100: Copied from SDK crate so we have consistent panic formatting. Could make public and share the actual impl, but I don't want this to be a public API.

Sushisource: This is actually one of my favorite weird little pieces of code I've written in here lol.

Could be in common-wasm as doc(hidden) but I don't care much either way.

Sushisource (approved):

crates/workflow/src/runtime/instance.rs
chris-olszewski: It is a very fun one.

PR #1469, opened 2026-08-10
chore(sdk): remove composing interceptor

chris-olszewski:
## What
TSIA
## Why
Not that we take a vec of interceptors that get composed when run, there is no need for this special interceptor that does the composition internally.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


Sushisource (approved):
Love removing things

PR #1471, opened 2026-08-10
feat(sdk): infer worker task types from registrations

chris-olszewski:
## What was changed
- Remove user configuration of task types and instead infer them based on the workflows/activities registered on the worker
- Remove mutating registration from `Worker` all activities/workflows must be registered at worker creation

Vast majority of the changes in this branch are moving integration tests to register workflows before worker creation.

## Why?
No need for end SDK users to need to manually toggle these bits. Match other SDKs where we infer these.

## Checklist

1. Closes #1196

2. How was this tested:
Existing test suite

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/sdk-core/src/worker/workflow/managed_run.rs
+++ b/crates/sdk-core/src/worker/workflow/managed_run.rs
@@ -1664,7 +1664,11 @@ mod tests {
         time::Duration,
     };
     use temporalio_common::protos::coresdk::common::ExternalStorageMetrics;
-    use tracing::{Event, Level, Metadata, Subscriber, field::Field, field::Visit, span};
+    use tracing::{
+        Event, Level, Metadata, Subscriber,
+        field::{Field, Visit},
+        span,
+    };
 
     use command_utils::*;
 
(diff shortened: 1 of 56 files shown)


crates/sdk-core/src/worker/workflow/managed_run.rs
```diff
@@ -1664,7 +1664,11 @@ mod tests {
         time::Duration,
     };
     use temporalio_common::protos::coresdk::common::ExternalStorageMetrics;
-    use tracing::{Event, Level, Metadata, Subscriber, field::Field, field::Visit, span};
+    use tracing::{
```
chris-olszewski, line 1667: Nightly formatting fix

crates/sdk-core/tests/common/mod.rs
```diff
@@ -587,10 +622,22 @@ impl CoreWfStarter {
                     let client = Client::new(connection.clone(), client_opts).unwrap();
                     (connection, client)
                 };
-                let mut core_config = self
-                    .sdk_config
+                let mut sdk_config = self.sdk_config.clone();
+                if matches!(mode, WorkerInitializationMode::CoreOnly) {
```
Sushisource, line 626: Heh, didn't think that'd need to be a thing lol

Sushisource (approved):
Yasss

PR #1474, opened 2026-08-11
Fix EphemeralServer Process Leak

Snipesy:
## What was changed

Ensures EphemeralServer process does not leak by adding `kill_on_drop` and an explicit `kill` on failure.

## Why?

Currently EphemeralServer's process will leak if it fails to start within 5s as the timeout path simply does not close it.

In https://github.com/Snipesy/temporal-kt/issues/26 this leads to gradle tasks simply sitting forever as the tasks refuse to finish if a process is holding stdout, stderr open (which the leaked process holds).

```
== orphan PID 64333:
   fd 0r CHR /dev/null
   fd 1 PIPE 
   fd 2 PIPE 
== orphan PID 64334:
   fd 0r CHR /dev/null
   fd 1 PIPE 
   fd 2 PIPE 
```

So the symptom in gradle's case is hung builds, CI timing out, with zero indicator or logs as to why.

For other SDKs (Python, Ruby, Rust itself, etc..) the leak still occurs but it doesnt tend to block builds so its less impactful.

Go also has a similar issue: https://github.com/temporalio/sdk-go/blob/main/testsuite/devserver.go

## Checklist

2. How was this tested:

The bug only occurs under high load, port collisions, or shoddy CI machines. Exactly how a build system handles a leaked process varies. So, replicating the exact scenarios is difficult.

Regardless I had Mr. Claude write a test that artificially replicates it. Note its set to ignored because its timing sensitive with `sleep 60` and not really suitable for CI (doing it any better way would touch a lot more code for a minor change).

3. Any docs updates needed?

No.

### DIFF
--- a/crates/sdk-core/CHANGELOG.md
+++ b/crates/sdk-core/CHANGELOG.md
@@ -46,6 +46,7 @@ relevant information.
   are preserved on failure; workers warn when the server does not advertise support.
 
 ### Fixed
+* Ephemeral server processes no longer leak on failed start.
 * Local activity resolutions are now delivered to workflows as each activity completes instead of
   waiting for every local activity in the workflow task. This allows sequences of short local
   activities to make progress while a long-running local activity executes in parallel, while
--- a/crates/sdk-core/src/ephemeral_server/mod.rs
+++ b/crates/sdk-core/src/ephemeral_server/mod.rs
@@ -203,20 +203,18 @@ pub struct EphemeralServer {
 
 impl EphemeralServer {
     async fn start(config: EphemeralServerConfig) -> anyhow::Result<EphemeralServer> {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -87,6 +87,10 @@ to docs, or any other relevant information.
 ### Fixed
 * Panics from update validators now reject the update instead of repeatedly failing workflow
   tasks.
+* Ephemeral servers (dev server and test server) no longer leak the spawned process when startup
```
Sushisource: You'll need to move this to the changelog inside the core crate now, we just broke up the files in https://github.com/temporalio/sdk-rust/pull/1484

Sushisource (approved):

PR #1442, opened 2026-08-03
Allow LAs to wake up workflow while another is still executing

Sushisource:
### Live execution

Previously, when an activation scheduled local activities, `ManagedRun` held lang's activation
completion response until every outstanding LA finished or the workflow task heartbeat fired. LA
results were applied internally, but lang could not poll another activation, so one long-running LA
could block a concurrent sequence of short LAs.

Core now acknowledges that activation completion with `DoNothing` while keeping the WFT and its
heartbeat open. Each LA result is applied and immediately checked for a new activation, even while
other LAs remain outstanding:

`LA result arrives -> apply result -> emit activation immediately -> lang schedules more work`

When the final LA finishes, Core completes the WFT normally with the accumulated marker commands.
Heartbeat completions request another WFT when LAs remain, and zero-sized-cache workers retain the
run until this process finishes.

The live LA state-machine path is unchanged:

`Executing --Schedule--> RequestSent --HandleResult--> MarkerCommandCreated --CommandRecordMarker--> ResultNotified --MarkerRecorded--> MarkerCommandRecorded`

The state machine already produced a resolution from `HandleResult`; the live-path bug was that
`ManagedRun` prevented that resolution activation from reaching lang while another LA was running.

### Replay compatibility

Incremental live delivery can create different activation groupings, so replay must reproduce the
order recorded by LA markers. Previously, `new_local_activity` could construct an LA directly in
`ReplayingPreResolved` by looking up its sequence number, causing its `Schedule` transition to emit
the resolution immediately. Results were therefore applied as activities were scheduled rather than
strictly in marker order.

Now every replaying LA starts unresolved and follows:

`Replaying --Schedule--> WaitingResolveFromMarkerLookAhead --HandleKnownResult--> ResolvedFromMarkerLookAheadWaitingMarkerEvent --MarkerRecorded--> MarkerCommandRecorded`

`workflow_machines.rs` retains peeked results in marker order in a `VecDeque`. After each workflow
iteration, it supplies `HandleKnownResult` from the front of that queue while the corresponding LA
machine exists, stopping at the first marker for an LA the workflow has not scheduled yet. The LA
state machine still owns one activity's lifecycle and validation; `workflow_machines.rs` owns the
cross-activity ordering needed to preserve activation boundaries.

This reproduces both old batched histories and new incrementally delivered histories without an SDK
compatibility flag. The old pre-resolved states and fake-marker plumbing are no longer needed.

## Why?

This allows a sequence of short local activities to continue making progress while a long-running
[...]

PR #1482, opened 2026-08-11
fix(core): delay zero-cache eviction until LA resolutions drain

jmaeagle99:
## What was changed

The zero-sized-cache eviction in `WFStream::process_completion` now also skips runs that still have jobs queued for lang, not just runs still waiting on local activities. This defers the eviction rather than skipping it since `process_completion` is called for every completion; the last completion will not have pending jobs nor outstanding LAs, thus the eviction is queued.

## Why?

With `max_cached_workflows = 0` and local activities, a run could be evicted mid-workflow-task, stranding its LA markers and livelocking the workflow.

#1442 delivers LA resolutions incrementally, so a resolution arriving while an earlier one is still outstanding with lang is parked as a pending job. That creates a state which didn't exist before that change: no outstanding LAs, but resolutions still undelivered. So the eviction landed mid-task: the accumulated `RecordMarker` commands were
discarded with the run and the WFT was never reported. Task timeout occurs and server redelivers the WFT; the same run is replayed, same local activities, same result, repeated indefinitely.

On debug builds, like in CI, a panic is intentionally raised with the message `There should not be any outgoing commands or messages when preparing a completion response if the activation was only an eviction. This is an SDK bug.`.

## Checklist

1. How was this tested: New explicitly-driven core test and locally with sdk-python submodule update

### DIFF
--- a/crates/sdk-core/tests/integ_tests/workflow_tests/local_activities.rs
+++ b/crates/sdk-core/tests/integ_tests/workflow_tests/local_activities.rs
@@ -26,6 +26,7 @@ use temporalio_common::{
         coresdk::{
             ActivityTaskCompletion, AsJsonPayloadExt, FromJsonPayloadExt,
             activity_result::ActivityExecutionResult,
+            activity_task::activity_task as act_task,
             common::extract_local_activity_marker_data,
             workflow_activation::{
                 WorkflowActivation, WorkflowActivationJob, workflow_activation_job,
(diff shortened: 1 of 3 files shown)


crates/sdk-core/tests/integ_tests/workflow_tests/local_activities.rs
```diff
@@ -3015,6 +3015,113 @@ async fn local_activity_resolutions_are_delivered_incrementally() {
...
+    let first_la = core.poll_activity_task().await.unwrap();
+    let second_la = core.poll_activity_task().await.unwrap();
+
+    core.complete_activity_task(ActivityTaskCompletion {
+        task_token: first_la.task_token,
+        result: Some(ActivityExecutionResult::ok(vec![1].into())),
+    })
+    .await
+    .unwrap();
+    let resolve_first = core.poll_workflow_activation().await.unwrap();
```
Sushisource, line 3089: Probably should assert the job is the appropriate LA resolution

Sushisource (approved):

PR #1483, opened 2026-08-11
Run integration tests against envconfig-backed Cloud namespaces

THardy98:
## What was changed

The integration-test runner now accepts `-s envconfig` and uses the SDK's normal envconfig loader
for the server address, namespace, authentication, TLS settings, and gRPC metadata. This mode does
not start or manage the configured server. Existing Temporal CLI, test-server, and external-server
modes retain their current behavior.

The runner also exposes `cloud-namespace create` and `cloud-namespace delete` subcommands for the
explicit namespace lifecycle used by CI.

The trusted Cloud job retains its existing Cloud-specific tests and proves the new path with one
existing timer test. Each run then generates temporary mTLS credentials, creates an isolated
one-day Temporal Cloud namespace through the Cloud Operations API, runs the timer test, and deletes
the namespace. Cleanup runs after test or provisioning failure without masking the original result.

## Why?

The harness previously understood only its bespoke connection environment variables, which kept
SDK developers from using the same arbitrary environment or TOML profiles as normal SDK clients.
An isolated namespace also prevents concurrent CI runs from sharing workflow state or cleanup
boundaries.

This PR intentionally runs only one existing integration test. Categorizing and expanding the
Cloud-eligible suite will follow separately.

## Checklist

1. Part of temporalio/features#851.

2. How was this tested:

   - `cargo +1.94 integ-test cloud-namespace --help`
   - Verified the helper rejects missing required environment configuration.
   - `cargo +1.94 integ-test -s envconfig --just-build`
   - `cargo +1.94 integ-test -s envconfig -- integ_tests::workflow_tests::timers::timer_workflow_workflow_driver --exact --list`
   - `timeout 180 cargo +1.94 integ-test -- integ_tests::workflow_tests::timers::timer_workflow_workflow_driver --exact`
   - `cargo +1.94 lint`
   - `cargo +1.94 test-lint`
   - `actionlint -ignore 'SC2086' .github/workflows/per-pr.yml` (the ignored warning predates this change).

3. Any docs updates needed?

   Yes. The README documents envconfig-backed integration-test execution and server lifecycle
   ownership.

### DIFF
--- a/crates/sdk-core/tests/cloud_namespace/mod.rs
+++ b/crates/sdk-core/tests/cloud_namespace/mod.rs
@@ -1,5 +1,4 @@
 use anyhow::{Context, bail};
-use clap::{Parser, Subcommand};
 use std::{collections::HashMap, env, fs::OpenOptions, io::Write, time::Duration};
 use temporalio_client::{Connection, ConnectionOptions, TlsOptions, grpc::CloudService};
 use temporalio_common::protos::temporal::api::cloud::{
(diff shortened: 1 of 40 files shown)


crates/sdk-core/tests/common/mod.rs
```diff
@@ -91,6 +93,16 @@ pub(crate) const INTEG_CLIENT_IDENTITY: &str = "integ_tester";
 pub(crate) const INTEG_CLIENT_NAME: &str = "temporal-core";
 pub(crate) const INTEG_CLIENT_VERSION: &str = "0.1.0";
 
+// Envconfig can read TOML and TLS files. Load one immutable snapshot so concurrently-created
```
Sushisource: "TOML and TLS files" implies the existence of a .tls format which is a bit weird sounding

crates/sdk-core/tests/cloud_namespace/mod.rs
```diff
@@ -0,0 +1,224 @@
+use anyhow::{Context, bail};
```
Sushisource, line 1: I don't think this really needs to be a whole new binary. Could probably just make this part of the integ test runner.

crates/sdk-core/tests/common/mod.rs
THardy98: Updated this to “TOML profiles and TLS credentials from files” in d4e109ad.

crates/sdk-core/tests/cloud_namespace/mod.rs
THardy98: Moved namespace create/delete into the integration runner in c789b7ec. CI now calls `cargo integ-test cloud-namespace create/delete`; the standalone target and alias are removed, while explicit CI steps preserve always-run cleanup.

Sushisource (approved):

PR #1484, opened 2026-08-11
docs: split sdk-core change log from the Rust SDK change log

jmaeagle99:
## What was changed

- Split the root change log into two: one for the Rust SDK (remains at root) and one for core-based SDKs (under `crates/sdk-core`).
- Update instructions to indicate what types of changes should be written into each change log.

## Why?

Entries for the Rust SDK vs all other SDKs were interleaved in the root change log, making it difficult to discern what changes apply to which SDK.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,8 +69,9 @@ Good pull requests are focused and easy to review:
   different interaction with the server. Because the Rust SDK runs on Core too, a
   user-observable change in Core behavior normally belongs in both files; an internal Rust
   API change that a language SDK absorbs inside its own bridge is invisible to its users
-  and belongs in neither. The comment at the top of either file lists the available
-  headings.
+  and belongs in neither. Write each entry from the perspective of the file's audience,
+  wording a change that goes in both files separately for each. The comment at the top of
+  either file lists the available headings.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -61,8 +61,16 @@ Good pull requests are focused and easy to review:
 * Keep each pull request scoped to one logical change.
 * Include tests for behavior changes.
 * Update public API documentation or doc comments when public behavior changes.
-* Add a high-level changelog entry for user-facing changes according to the
-  repository's local changelog convention.
+* Add a high-level changelog entry for user-facing changes, under the `## Unreleased`
```
Sushisource, line 64: I think you might want to add a bit of this language to AGENTS.md too now

Sushisource (approved):

CONTRIBUTING.md
jmaeagle99: Most of it was already in the agents.md file before this comment was made. So if you have something specific in mind, I'm not quite understanding. However, I've added some more instructions to say that entries should be written for the audience of each SDK.

Sushisource: Ah ok I didn't realize it already had that 🤦

PR #1485, opened 2026-08-12
Add JSON logging format

### LINKED ISSUE #1464: [core] Allow configuring core tracing format
Currently all spans emitted by cores `tracing` subscriber include ANSI escape codes. Add new format option to console when constructing the core runtime that alters the tracing behavior.

For core specifically this will look like:
```
pub enum Logger {
    Console {
        filter: String,
        format: ConsoleLogFormat,   // NEW
    },
}

#[derive(Default)]
pub enum ConsoleLogFormat {
    #[default]
    Pretty,   // current behavior: ANSI colors, human-readable
    Plain,    // no ANSI, same fields, text format
    Json,     // machine-readable JSON per line
}
```

Also, probably worth respecting `NO_COLOR` etc here when deciding between `Plain`/`Pretty`

Sushisource:
## What was changed
Title

## Why?
Wanted by some customers

## Checklist

1. Closes https://github.com/temporalio/sdk-rust/issues/1464

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/crates/common/src/telemetry.rs
+++ b/crates/common/src/telemetry.rs
@@ -28,7 +28,7 @@ use std::{
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 use tracing::{Level, Subscriber};
-use tracing_subscriber::{EnvFilter, Layer, layer::SubscriberExt};
+use tracing_subscriber::{EnvFilter, Layer, fmt::MakeWriter, layer::SubscriberExt};
 use url::Url;
 
 #[cfg(feature = "core-telemetry-bridge")]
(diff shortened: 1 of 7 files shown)


crates/common/src/telemetry.rs
```diff
@@ -219,6 +221,19 @@ pub enum Logger {
     },
 }
 
+/// Controls how console logs are rendered.
+#[derive(Debug, Default, Clone, Copy)]
+pub enum LoggerFormat {
```
chris-olszewski, line 228:
```suggestion
#[non_exhaustive]
pub enum LoggerFormat {
```

crates/common/src/telemetry.rs
```diff
@@ -219,6 +221,19 @@ pub enum Logger {
...
+    /// Render logs compactly on a single line.
+    #[default]
+    Compact,
+    /// Render logs in a human-readable format across multiple lines. This format is also selected
+    /// when `TEMPORAL_CORE_PRETTY_LOGS` is set, regardless of the configured format.
```
chris-olszewski: An env var having priority over an explicit code configuration feels off to me. Should `LoggerFormat` be optional and only when it is missing we consider the env var?

chris-olszewski (changes requested):
I think we should generally respect explicit code configuration over env vars. It is usually the standard ordering ([e.g. tracing itself](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.with_ansi)).

crates/common/src/telemetry.rs
Sushisource: Ya good point

chris-olszewski (approved):

PR #1487, opened 2026-08-12
Add changelog release notes CLI

tconley1428:
## Summary
- add a reusable SDK Core changelog CLI that accepts a Core SHA range
- track changelog entries from their introducing commit while retaining final wording

## Testing
- python3 scripts/changelog_release_notes.py --from <sha> --to <sha>


crates/sdk-core/src/changelog_release_notes.rs
```diff
@@ -0,0 +1,416 @@
...
+                output.extend(entry.lines);
+            }
+            output.push(String::new());
+        }
+    }
+    Ok(output)
+}
+
+#[cfg(test)]
+mod tests {
```
Sushisource, line 273: Extremely weird layout w/ this above main fn

Sushisource (approved):

PR #1489, opened 2026-08-12
fix(sdk): payload conversion causes wft failure

chris-olszewski:
## What was changed
- Payload conversion errors now result in failing the WFT. Previously a naive `?` usage would fail the workflow.
- Remove `anyhow` conversion to `WorkflowTermination`, ensures that there is explicit `ApplicationFailure` wrapping if failing the workflow is desired.
- In order to better support the first change we added a new variant to outgoing workflow error. All outgoing error types that have a payload conversion failure map to this variant directly.
- Small fixup to ensure crypto provider is hooked up for some reqwest usage in our tests.

## Why?
Match all other SDKs, payload conversion errors should result in a task being retried, not the workflow itself failed.

## Checklist

1. Closes N/A

2. How was this tested:
Added integration test to verify these fail the WFT and will result in the task being retried.

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/sdk-core/tests/integ_tests/metrics_tests.rs
+++ b/crates/sdk-core/tests/integ_tests/metrics_tests.rs
@@ -92,6 +92,7 @@ use tonic::IntoRequest;
 use url::Url;
 
 pub(crate) async fn get_text(endpoint: String) -> String {
+    temporalio_common::telemetry::ensure_default_crypto_provider();
     reqwest::get(endpoint).await.unwrap().text().await.unwrap()
 }
 
@@ -962,6 +963,7 @@ async fn docker_metrics_with_prometheus(
     eventually(
         || async {
             // Query Prometheus API for metrics
+            temporalio_common::telemetry::ensure_default_crypto_provider();
             let client = reqwest::Client::new();
             let query = format!("temporal_sdk_{}num_pollers", test_uid.clone());
             let response = client
@@ -1526,6 +1528,7 @@ async fn test_prometheus_endpoint_integration() {
     up_down_counter.adds(-2);
 
     let url = format!("http://{addr}/metrics");
+    temporalio_common::telemetry::ensure_default_crypto_provider();
     let response = tokio::time::timeout(Duration::from_secs(10), reqwest::get(&url))
         .await
         .expect("Request timed out")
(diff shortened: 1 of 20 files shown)


crates/workflow/src/runtime/model.rs
```diff
@@ -253,9 +259,13 @@ impl From<ApplicationFailure> for WorkflowTermination {
     }
 }
 
-impl From<temporalio_common_wasm::data_converters::PayloadConversionError> for WorkflowTermination {
-    fn from(value: temporalio_common_wasm::data_converters::PayloadConversionError) -> Self {
-        Self::Failed(value.into())
+fn fail_workflow_task_on_payload_conversion(error: PayloadConversionError) -> ! {
+    panic!("Workflow payload conversion failed: {error}")
+}
```
Sushisource: Discussed in slack but to post here I think we should probably trigger an explicit task failure somehow rather than introducing a panic

crates/sdk-core/tests/integ_tests/metrics_tests.rs
```diff
@@ -92,6 +92,7 @@ use tonic::IntoRequest;
 use url::Url;
 
 pub(crate) async fn get_text(endpoint: String) -> String {
+    temporalio_common::telemetry::ensure_default_crypto_provider();
```
chris-olszewski, line 95: Fallout of #1458. If you end up having a cached ephemeral server then this function doesn't get called in the normal path. Only needed in tests where we explicitly use reqwest.

crates/workflow/src/runtime/model.rs
```diff
@@ -212,6 +213,12 @@ impl CancellableID {
 pub type WorkflowResult<T> = Result<T, WorkflowTermination>;
 
 /// Represents ways a workflow can terminate without producing a normal result.
+///
+/// Payload conversion errors returned by workflow operations may be handled at the operation
+/// boundary. Propagating one directly into `WorkflowTermination`, such as with `?`, will fail
+/// the current Workflow Task so it can be retried.
```
Sushisource:
```suggestion
/// Payload conversion errors returned by workflow operations propagated directly into `WorkflowTermination`, such as with `?`, will fail
/// the current Workflow Task so it can be retried.
```

A bit wordy

Sushisource (approved):

PR #1492, opened 2026-08-13
feat(sdk): add activity and workflow test env

chris-olszewski:
## What was changed
- Make ephemeral server return structured error types instead of just anyhow
- 

## Why?
Own our public API

## Checklist

1. Closes N/A

2. How was this tested:
Unit testing for the activity env. Workflow env is mostly wrapper around ephemeral server so just a small integration test to verify basic functionality.

3. Any docs updates needed?
N/A

### DIFF
--- a/crates/sdk-core/src/ephemeral_server/mod.rs
+++ b/crates/sdk-core/src/ephemeral_server/mod.rs
@@ -57,7 +57,9 @@ pub enum EphemeralServerError {
         source: Box<dyn Error + Send + Sync>,
     },
     /// The server process did not become available before its startup deadline.
-    #[error("ephemeral server at {target} did not start within {timeout:?}")]
+    #[error(
+        "ephemeral server at {target} did not start within {timeout:?}. Make sure another download isn't stuck and delete the temp file."
+    )]
     StartupTimeout {
         /// Server target that could not be reached.
         target: String,
--- a/crates/sdk-core/src/ephemeral_server/mod.rs
+++ b/crates/sdk-core/src/ephemeral_server/mod.rs
@@ -1,7 +1,6 @@
 //! This module implements support for downloading and running ephemeral test
 //! servers useful for testing.
 
-use anyhow::anyhow;
 use flate2::read::GzDecoder;
 use futures_util::StreamExt;
 use serde::Deserialize;
@@ -9,8 +8,9 @@ use std::{
     fs::OpenOptions,
     io,
     path::{Path, PathBuf},
+    time::SystemTimeError,
 };
-use temporalio_client::{Connection, ConnectionOptions};
+use temporalio_client::{Connection, ConnectionOptions, errors::ClientConnectError};
 use tokio::{
     task::spawn_blocking,
     time::{Duration, sleep},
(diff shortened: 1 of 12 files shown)


crates/sdk-core/src/ephemeral_server/mod.rs
```diff
@@ -23,6 +23,94 @@ use zip::read::read_zipfile_from_stream;
...
+        #[source]
+        source: Box<dyn Error + Send + Sync>,
+    },
+}
+
+fn download_error(source: impl Error + Send + Sync + 'static) -> EphemeralServerError {
+    EphemeralServerError::Download {
+        source: Box::new(source),
+    }
+}
```
Sushisource: Seems like these should all be impld on the error enum?

crates/sdk-core/src/ephemeral_server/mod.rs
```diff
@@ -44,7 +132,7 @@ pub struct TemporalDevServerConfig {
     /// Whether to enable the UI. If ui_port is set, assumes true.
     #[builder(default)]
     pub ui: bool,
-    /// Log format and level
+    /// Log format and level.
```
Sushisource, line 139: Critical change

crates/sdk-core/src/ephemeral_server/mod.rs
```diff
@@ -514,32 +619,35 @@ async fn lazy_download_exe(
...
                     // No progress for a while; assume the downloader was
                     // abandoned. Reclaim it once; if it goes stale again, fail
                     // loudly rather than looping forever.
                     if already_tried_cleaning_old {
-                        return Err(anyhow!(
-                            "Temp download file at {} made no progress for over {} \
-                            seconds. Make sure another download isn't stuck and \
-                            delete the temp file.",
+                        return Err(invalid_download(format!(
+                            "temporary file at {} made no progress for over {} seconds",
```
Sushisource, line 646: Do we want to delete the advice in the error msg?

crates/sdk/src/testing.rs
```diff
@@ -0,0 +1,655 @@
+//! Test environments for running activity code and workflow workers.
+//!
+//! Activity inputs, outputs, and outbound heartbeat details stay typed. Previous heartbeat details
```
Sushisource, line 3: Previous?

crates/sdk/src/testing.rs
```diff
@@ -0,0 +1,655 @@
...
+//! assert_eq!(
+//!     env.run(GreetingActivities::greet, "Temporal".to_owned())
+//!         .await
+//!         .unwrap(),
+//!     "Hello, Temporal!"
+//! );
+//! # }
+//! ```
+//!
+//! [`WorkflowEnvironment::start_local`] owns a Temporal CLI dev server. Its client can be passed
```
Sushisource, line 47:
```suggestion
//! [`WorkflowEnvironment::start_local`] spawns a Temporal CLI dev server. Its client can be passed
```

crates/sdk/src/testing.rs
```diff
@@ -0,0 +1,655 @@
...
+}
+
+/// Options for constructing [`ActivityInfo`] with defaults suitable for an activity test.
+#[derive(bon::Builder)]
+#[builder(
+    finish_fn(name = build_internal, vis = ""),
+    state_mod(vis = "pub"),
+    on(String, into)
+)]
+pub struct ActivityInfoOptions {
```
Sushisource: Is it worth Calling this `TestActivityInfoOptions`? Yes, it's namespaced, but, I can see it maybe being a bit confusing? I'm not strong on the opinion.

crates/sdk/src/testing.rs
```diff
@@ -0,0 +1,655 @@
...
+    /// Observe the typed details supplied to every heartbeat.
+    pub fn on_heartbeat<F>(mut self, callback: F) -> Self
+    where
+        F: Fn(Box<dyn Any>) + Send + Sync + 'static,
+    {
+        self.heartbeat_callback = Some(Arc::new(callback));
+        self
+    }
+
+    /// Supply heartbeat details from a previous activity attempt.
```
Sushisource:
```suggestion
    /// Supply heartbeat details from an activity attempt.
```

crates/sdk/src/testing.rs
```diff
@@ -0,0 +1,655 @@
...
+}
+
+impl ActivityEnvironment {
+    /// Construct an activity environment builder using the default payload converter.
+    pub fn builder() -> ActivityEnvironmentBuilder {
+        Self::builder_internal(PayloadConverter::default())
+    }
+
+    /// Construct an activity environment builder using a custom payload converter.
+    pub fn builder_with_payload_converter(
```
Sushisource: I'm wondering if we should swap these so `builder_with_default` is the wordier one

crates/sdk/src/testing.rs
```diff
@@ -0,0 +1,655 @@
...
+        payload_converter: PayloadConverter,
+    ) -> ActivityEnvironmentBuilder {
+        Self::builder_internal(payload_converter)
+    }
+
+    /// Run an activity marker with already-typed input.
```
Sushisource: Marker?

crates/sdk/src/testing.rs
chris-olszewski: If you wanted to test a retry of an activity it would be useful.

crates/sdk/src/testing.rs
chris-olszewski: Maybe "the previous"? I do think previous is more accurate than just "an" activity attempt?

crates/sdk/src/testing.rs
chris-olszewski: Eh, yeah was on the fence with this once. Probably useful to indicate in the options that we don't want this to be used to build `ActivityInfo` in general.

crates/sdk-core/src/ephemeral_server/mod.rs
chris-olszewski: Gotta get that +1/-1

crates/sdk/src/testing.rs
Sushisource: My question is if it matters that it was actually previous or not. In principle it could just be an arbitrary made up details instance, no?

Sushisource (approved):

PR #1497, opened 2026-08-14
Tolerate out-of-grammar nexus request-timeout headers

mjameswh:
## What was changed

`parse_request_timeout` now accepts the full syntax Go's `time.Duration::String` can emit —
sub-millisecond units, hours, multi-unit values such as `1m30s`, and negative values — and treats
anything non-positive as a deadline already past, i.e. a zero timeout.

## Why?

The Nexus spec limits this header to a positive decimal followed by `ms`, `s` or `m`, but the server
formats it with Go's `time.Duration::String`, which is a superset of that grammar
(temporalio/temporal#11569). Core rejected anything outside the grammar, and then ran the task with
no local timeout at all — so a handler waiting to be told the task had timed out waited forever, the
task stayed outstanding, and worker shutdown blocked on it indefinitely.

Core cannot choose what the server sends it, so it has to tolerate the wider syntax. In practice
this deadlocked `metrics_tests::nexus_metrics` in roughly half of all integ-test runs, burning the
job's entire time budget.

## Checklist

1. Closes N/A

2. How was this tested:

New `parse_request_timeout_go_duration_forms` unit test covers both the newly accepted forms and the
ones that must still be rejected.

`nexus_metrics` now drains its Nexus poller until shutdown rather than expecting an exact task
count, so stopping early cannot leave the operation it deliberately never answers outstanding.

Three commits, so CI shows both the problem and the fix: pin the integ tests on
`v1.7.4-standalone-nexus-operations`, whose server surfaces this; then the fix; then remove the pin.

3. Any docs updates needed? `crates/sdk-core/CHANGELOG.md` entry added.

### DIFF
--- a/crates/sdk-core/tests/integ_tests/metrics_tests.rs
+++ b/crates/sdk-core/tests/integ_tests/metrics_tests.rs
@@ -960,12 +960,18 @@ async fn docker_metrics_with_prometheus(
     .await
     .unwrap();
 
+    let task_queue = starter.get_task_queue().to_string();
     eventually(
         || async {
             // Query Prometheus API for metrics
             temporalio_common::telemetry::ensure_default_crypto_provider();
             let client = reqwest::Client::new();
-            let query = format!("temporal_sdk_{}num_pollers", test_uid.clone());
+            // The task queue must be matched in the query rather than asserted on afterwards: this
+            // runtime's meter is also used by the shared-namespace worker, whose pollers report
+            // against the worker-commands control queue, and the order series come back in is not
+            // ours to choose.
+            let query =
+                format!("temporal_sdk_{test_uid}num_pollers{{task_queue=\"{task_queue}\"}}");
             let response = client
                 .get(PROMETHEUS_QUERY_API)
                 .query(&[("query", query.clone())])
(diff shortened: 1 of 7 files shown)


crates/sdk-core/tests/integ_tests/metrics_tests.rs
```diff
@@ -1235,8 +1235,15 @@ async fn nexus_metrics() {
         .unwrap();
 
     let nexus_polling = async {
-        for _ in 0..5 {
-            let nt = core_worker.poll_nexus_task().await.unwrap();
+        // The number of tasks the server delivers here isn't fixed, so drain until the worker
+        // shuts down rather than counting them. Stopping early would leave the operation we
+        // deliberately never respond to outstanding, which blocks shutdown forever.
+        loop {
```
Sushisource: It should be fixed though, no? Curious why this needed to change

Sushisource (approved):

crates/sdk-core/tests/integ_tests/metrics_tests.rs
mjameswh: Yeah, it should.

That specific change was meant to ensure fail fast in the situation where a bug in the logic being tested could introduce misalignments in the task polling sequence (e.g. server redelivering a task) resulting in the process hanging on a poll that would never resolve. This was the initial symptom observed in CI, with the job getting killed after 25 minutes.

But revisiting this now, I realize the premise is incorrect in a few ways, and I'm really not convinced there's any actual scenario where that structure would be able to exit earlier compared to the original loop. And running the two versions of the test without the parsing fix actually shows the same hanging behavior, thus confirming that the changing the looping structure isn't helping in the caseOf this specific bug.

So I'm reverting that hunk.

PR #1501, opened 2026-08-14
Expose cancel details on workflow cancel & propagate them in Rust SDK

### LINKED ISSUE #1463: [Bug] CancelWorkflowExecution command drops `details` — cancellation details never reach the server (Go SDK sets them)

Modelling a legitimate business non-completion as a terminal state distinct from
both success and defect, so that alerting can treat them differently:

- `Completed` — the process succeeded
- `Canceled` — a business rule stopped it (e.g. the customer never signed the contract). Expected, not an error.
- `Failed` — something actually broke. This is what should page someone.

For that to be usable, the cancellation needs to carry a structured reason that
shows up on the workflow's terminal event, so an operator opening the workflow
sees *why* it was cancelled. `Completed` carries the workflow's return value and
`Failed` carries a full structured `failure`, but in core-based SDKs `Canceled`
can carry nothing at all.

### Describe the bug

When a workflow completes as cancelled through a core-based SDK, the resulting
`WorkflowExecutionCanceled` history event always has an empty `details` field.

The core `CancelWorkflowExecution` command has no `details` field:

`crates/protos/protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto:228`
```proto
// Indicate a workflow has completed as cancelled. Generally sent as a response to an activation
// containing a cancellation job.
message CancelWorkflowExecution {}
```

...and the conversion to the server command hardcodes `None`, discarding its input `_c`:

`crates/protos/src/protos/mod.rs:2012`
```rust
impl From<workflow_commands::CancelWorkflowExecution> for command::Attributes {
    fn from(_c: workflow_commands::CancelWorkflowExecution) -> Self {
        Self::CancelWorkflowExecutionCommandAttributes(
            CancelWorkflowExecutionCommandAttributes { details: None },
        )
    }
}
```

Three things suggest this is an oversight rather than a deliberate design choice:

**1. The server fully supports it.** Both
`CancelWorkflowExecutionCommandAttributes.details` and
`WorkflowExecutionCanceledEventAttributes.details` exist and work — see the
reproduction below, which persists and renders details by bypassing the SDK.

**2. The Go SDK populates it.** `sdk-go`, `internal/internal_task_handlers.go`:
```go
if errors.As(workflowContext.err, &canceledErr) {
    closeCommand = createNewCommand(enumspb.COMMAND_TYPE_CANCEL_WORKFLOW_EXECUTION)
    closeCommand.Attributes = &commandpb.Command_CancelWorkflowExecutionCommandAttributes{
        CancelWorkflowExecutionCommandAttributes: &commandpb.CancelWorkflowExecutionCommandAttributes{
            Details: convertErrDetailsToPayloads(canceledErr.details, wth.dataConverter),
        }}
}
```
So this is a behavioural parity gap between Go and the core-based SDKs
(TypeScript, Python, .NET, Ruby).

**3. The lang-side API already accepts details, then silently drops them.**
TypeScript's `CancelledFailure` takes a `details` argument:
```ts
export declare class CancelledFailure extends TemporalFailure {
    readonly details: unknown[];
    constructor(message: string | undefined, details?: unknown[], cause?: Error);
}
```
A user can pass details and they vanish with no error or warning.

### Minimal Reproduction

**a) Details passed from a workflow are dropped.** Workflow:

```ts
import { CancelledFailure } from '@temporalio/common';
import { isCancellation, sleep } from '@temporalio/workflow';

export async function CancelWithDetails(): Promise<void> {
  try {
    await sleep('1h');
  } catch (err) {
    if (isCancellation(err)) {
      throw new CancelledFailure('cancelled', [{ reason: 'contract_expired' }]);
    }
    throw err;
  }
}
```

Client:
```ts
const handle = await client.workflow.start(CancelWithDetails, { taskQueue, workflowId });
await handle.cancel();
const history = await handle.fetchHistory();
// WorkflowExecutionCanceled attributes:
//   { "workflowTaskCompletedEventId": "N" }      <-- no `details`
```

Status is correctly `CANCELED`, but the `[{ reason: 'contract_expired' }]` passed
to `CancelledFailure` never appears anywhere in history.

**b) The same server accepts details when the SDK is bypassed.** Impersonating a
worker for one workflow task (started on a task queue with no worker on it) and
issuing the command by hand:

```ts
const task = await connection.withDeadline(Date.now() + 60_000, () =>
  connection.workflowService.pollWorkflowTaskQueue({
    namespace, taskQueue: { name: taskQueue }, identity: 'raw-probe',
  }));

await connection.workflowService.respondWorkflowTaskCompleted({
  namespace,
  identity: 'raw-probe',
  taskToken: task.taskToken,
  commands: [{
    commandType: temporal.api.enums.v1.CommandType.COMMAND_TYPE_CANCEL_WORKFLOW_EXECUTION,
    cancelWorkflowExecutionCommandAttributes: {
      details: { payloads: [defaultPayloadConverter.toPayload({ reason: 'contract_expired' })] },
    },
  }],
});
```

The terminal event now carries the details, and the Web UI renders them:
```json
{
  "workflowTaskCompletedEventId": "4",
  "details": { "payloads": [{ "reason": "contract_expired" }] }
}
```

### Environment/Versions

- OS and processor: x86_64 Linux
- Temporal Server 1.31.0 (Docker), Web UI 2.49.1
- TypeScript SDK 1.16.1 for the reproduction; sources inspected through 1.22.0
  and `sdk-rust` `main` — the proto message and the `From` impl above are
  byte-for-byte identical in all of them
- Worker run locally; server via Docker

### Additional context

Suggested fix:

1. Add a `details` field to the core `CancelWorkflowExecution` command message.
2. Thread it through the `From` impl instead of hardcoding `None`.
3. Lang SDKs pass their cancellation-failure details — for TypeScript, that would
   be `CancelledFailure.details` in `handleWorkflowFailure`, mirroring Go's
   `CanceledError.details`.

Without this, there's no way for a core-based SDK workflow to attach any
structured outcome to its own cancellation, so users are pushed to side channels
(search attributes, `setCurrentDetails`, external storage) purely because of a
dropped field.

I couldn't find an existing issue for this in `sdk-rust` or `sdk-typescript`.
Happy to open the PR for (1) and (2) plus the TypeScript side — but since this
changes a core command message consumed by several lang SDKs, I'd rather confirm
the intended shape first, in particular whether details should be sourced from
each SDK's cancellation-failure type.

Sushisource:
## What was changed
See title

## Why?
This is fully supported on the backed but was never exposed in Core

## Checklist

1. Closes https://github.com/temporalio/sdk-rust/issues/1463
2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/crates/workflow/src/runtime/instance.rs
+++ b/crates/workflow/src/runtime/instance.rs
@@ -19,7 +19,7 @@ use crate::{
         HandleSignalInput, HandleSignalResult, HandleUpdateInput, HandleUpdateResult,
         InitializeWorkflowInput, InitializeWorkflowOutput, SyncWorkflowInterceptorContext,
         ValidateUpdateInput, ValidateUpdateResult, WorkflowInterceptor, WorkflowInterceptorContext,
-        WorkflowInterceptorFuture, WorkflowNext, serialize_workflow_output,
+        WorkflowInterceptorFuture, WorkflowNext, WorkflowOutputValue, serialize_workflow_output,
         wrong_workflow_input_type,
     },
 };
@@ -767,7 +767,26 @@ where
         match result {
             Ok(result) => Ok(TerminalOutcome::Completed(result)),
             Err(WorkflowTermination::ContinueAsNew(req)) => Ok(TerminalOutcome::ContinueAsNew(req)),
-            Err(WorkflowTermination::Cancelled) => Ok(TerminalOutcome::Cancelled),
+            Err(WorkflowTermination::Cancelled { details }) => {
+                let details = details
+                    .map(|details| {
+                        (&*details as &dyn WorkflowOutputValue)
+                            .serialize_payloads(&SerializationContext {
+                                data: &SerializationContextData::Workflow,
+                                converter: self.ctx.payload_converter(),
+                            })
+                            .map(|payloads| Payloads { payloads })
+                    })
+                    .transpose()
+                    .map_err(|err| TaskFailure {
+                        failure: Box::new(Failure {
+                            message: format!("Workflow payload conversion failed: {err}"),
+                            ..Default::default()
+                        }),
+                        force_cause: None,
+                    })?;
+                Ok(TerminalOutcome::Cancelled(details))
+            }
             Err(WorkflowTermination::Evicted) => {
                 panic!("workflow instances must not explicitly return eviction")
             }
@@ -781,9 +800,13 @@ where
                 })
             }
             Err(WorkflowTermination::Failed(err)) => {
-                if self.base_ctx.cancellation_token().is_cancelled() && err.as_cancelled().is_some()
+                if self.base_ctx.cancellation_token().is_cancelled()
+                    && let Some(cancelled) = err.as_cancelled()
                 {
-                    return Ok(TerminalOutcome::Cancelled);
+                    let details = cancelled.raw_details().map(|payloads| Payloads {
+                        payloads: payloads.to_vec(),
+                    });
+                    return Ok(TerminalOutcome::Cancelled(details));
                 }
                 let failure = self.base_ctx.data_converter().to_failure(
                     &SerializationContextData::Workflow,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ relevant information.
 ### Added
 * Support for running Standalone Activities in Rust SDK Worker.
 * Client methods for starting and managing execution of Standalone Activities. 
+* `WorkflowTermination::cancelled_with_details` for recording structured details when a Workflow
+  Execution completes as cancelled.
 * `LoggerFormat` for selecting compact, pretty, or JSON Core console log output. Configured log
   filters continue to apply to JSON output.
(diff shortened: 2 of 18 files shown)


crates/workflow/src/runtime/instance.rs
```diff
@@ -767,7 +767,26 @@ where
...
+                            .map(|payloads| Payloads { payloads })
+                    })
+                    .transpose()
+                    .map_err(|err| TaskFailure {
+                        failure: Box::new(Failure {
+                            message: format!("Workflow payload conversion failed: {err}"),
+                            ..Default::default()
+                        }),
+                        force_cause: None,
+                    })?;
```
chris-olszewski, line 787: Not blocking, but would be nice to share this logic with the standard payload conversion error task failure. Little awkward as it would probably require another intermediate representation, and not sure it is worth that.

chris-olszewski (approved):

crates/workflow/src/runtime/instance.rs
Sushisource: Yeah, I think it's net more wordy than without it

PR #1508, opened 2026-08-19
fix(sdk): typed signal_with_start_workflow

### LINKED ISSUE #1506: [Feature Request] Typed input for signal-with-start

Signal-with-start seems to be the one place in the client where you can't stay typed. `start_workflow`, `WorkflowHandle::signal` and `ScheduleAction::start_workflow` all take a definition plus its typed input and encode it with the client's data converter, but `WorkflowStartSignal::input` only accepts pre-encoded `Payloads`. So while migrating to the 0.7 typed APIs I ended up hand-rolling the encoding at every signal-with-start call site:

```rust
let payload = data_converter.to_payload(&SerializationContextData::Workflow, &signal_input).await?;
let start_signal = WorkflowStartSignal::new("mySignal")
    .input(Payloads { payloads: vec![payload] })
    .build();
```

###  Describe the solution you'd like

A constructor that takes a `SignalDefinition` and its typed input, and encodes with the client's data converter when the workflow is started, like `WorkflowHandle::signal` does:

```
  let start_signal = WorkflowStartSignal::from_definition(MySignal, signal_input);
```

`ScheduleWorkflowInput` from #1347 already does exactly this for schedule actions, so I think the same deferred-encoding shape fits here. One thing I noticed is that unlike the schedule input this field already exists and is sent to the server verbatim, so pre-encoded payloads need to keep bypassing the converter, pushing them through `RawValue` would double-encode for anyone running a payload codec. Not sure what you think about that?

I would be more than happy to open a PR.

chris-olszewski:
## What was changed
- Remove `start_signal` from workflow options and `WorkflowStartSignal` which was only used there.
- Add a `signal_with_start_workflow` instead of taking an optional signal as part of `WorkflowStartOptions`
- New `signal_with_start_workflow` client interceptor hook

## Why?
We were not providing any type safety around ensuring signals sent during workflow start were valid for the workflow type *or* that the input type matched the signal definition.

I went with a different method instead of an optional signal for a few reasons:
- Bon support for [generics is currently experimental](https://bon-rs.com/reference/builder/top-level/generics)
- If we did make `WorkflowStartOption` generic over signals, we would still need a way to capture "no signal" at the type level. This is possibly, with some `WorkflowStartOption<S = NoSignal>` and adding another trait we implement for all `SignalDefinition` and this new `NoSignal` type. Was unsure this extra type work was worth keeping the single method.
- We will be introducing `update_with_start` and `execute_update_with_start` soon, which are necessarily different methods due to the return type. Feels okay to have signal/update with start have some symmetry as opposed to one being part of start options and one being an explicit arguments.

## Checklist

1. Closes #1506 

2. How was this tested:
- Some additional client unit tests around payload conversion
- Integration test ensuring that headers set from client interceptor is visible on the workflow signal handler interceptor
- trybuild test to verify you can't use incorrect signals against a workflow

3. Any docs updates needed?

### DIFF
--- a/crates/client/tests/typed_signal_with_start/mismatched_signal_fail.stderr
+++ b/crates/client/tests/typed_signal_with_start/mismatched_signal_fail.stderr
@@ -1,0 +1,34 @@
+error[E0271]: type mismatch resolving `<SecondSignal as SignalDefinition>::Workflow == Run`
+  --> tests/typed_signal_with_start/mismatched_signal_fail.rs:39:9
+   |
+36 |     let _ = client.signal_with_start_workflow(
+   |                    -------------------------- required by a bound introduced by this call
+...
+39 |         SecondWorkflow::second_signal,
+   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type mismatch resolving `<SecondSignal as SignalDefinition>::Workflow == Run`
+   |
+note: expected this to be `first_workflow::Run`
+  --> tests/typed_signal_with_start/mismatched_signal_fail.rs:24:1
+   |
+24 | #[workflow_methods]
+   | ^^^^^^^^^^^^^^^^^^^
+   = note: `second_workflow::Run` and `first_workflow::Run` have similar names, but are actually distinct types
+note: `second_workflow::Run` is defined in module `crate::second_workflow` of the current crate
+  --> tests/typed_signal_with_start/mismatched_signal_fail.rs:24:1
+   |
+24 | #[workflow_methods]
+   | ^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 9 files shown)


crates/client/tests/typed_signal_with_start/mismatched_signal_fail.stderr
```diff
@@ -0,0 +1,34 @@
+error[E0271]: type mismatch resolving `<SecondSignal as SignalDefinition>::Workflow == Run`
```
chris-olszewski, line 1: I don't *love* this error output, but unsure if introducing a new trait just to have a better `not_implemented` message seems overkill.

Sushisource (approved):

PR #1512, opened 2026-08-20
Preserve Event Group Markers on cancellation commands

mjameswh:
## What was changed

Follow-up to #1308. Two kinds of command were reaching the server with no Event Group Markers, because they don't translate one-for-one from a command lang sent:

- **Cancellations.** The timer, activity, child workflow and Nexus machines now keep the markers and user metadata of the command that created them, and stamp them on the cancel command they issue. If lang annotates the cancel command itself, that wins field by field; anything it leaves out is inherited.
- **The `TemporalChangeVersion` upsert** Core synthesizes alongside a patch marker, which now repeats what the patch carried.

`MachineResponse::NewCoreOriginatedCommand` became a struct variant so it can carry annotations, which is what makes the child-workflow external cancel and the patch upsert work. `WFCommand`'s `metadata` and `event_group_markers` fields moved into a `CommandAnnotations` struct, since they now travel together through the machines rather than being read once at command-creation time.

Also fixes the Rust SDK's local activity retry: `LATimerBackoffFut` was starting the backoff timer with no markers, so it fell out of the group its activity belongs to.

## Why?

A cancellation is the part of a Workflow's history a user is most likely to be looking at when they reach for Event Groups, and it was the one part that had none.

Markers are inherited from the cancelled command rather than from the ambient scope of the cancelling code because most cancellations aren't issued from user code at all — a cancellation scope timing out, or a parent cancelling its children, runs where no scope is active.

## Tested

New unit tests in `crates/sdk-core/src/core_tests/event_groups.rs` cover inheritance and the lang-wins override for each cancellation kind, plus the patch upsert. Verified end to end against the TypeScript SDK's Event Groups suite.


crates/sdk-core/src/core_tests/event_groups.rs
```diff
@@ -0,0 +1,303 @@
+//! Event group markers and user metadata attached to a lang command must survive the
+//! indirections Core puts between that command and the `Command` it eventually sends to the
+//! server. Two of those indirections mean the annotations cannot simply be copied across as they
+//! are for a directly translated command: a cancellation is produced by the machine of the
+//! command being cancelled, and a patch synthesizes a search attribute upsert of Core's own
+//! making.
+//!
+//! These tests drive a bare worker with hand-built activation completions because lang, not the
+//! Rust SDK, is what annotates commands; the Rust SDK has no Event Groups API to express a
+//! cancellation carrying markers of its own.
```
Sushisource: This comment block is maybe overly specific, especially if other unit tests are added later.

Sushisource: Probably the other unit test that was on `updates.rs` should move in here

crates/sdk-core/src/worker/workflow/mod.rs
```diff
@@ -1413,16 +1415,47 @@ struct EmptyWorkflowCommandErr;
...
+    fn override_with(&mut self, other: Self) {
+        if other.metadata.is_some() {
+            self.metadata = other.metadata;
+        }
+        if !other.event_group_markers.is_empty() {
+            self.event_group_markers = other.event_group_markers;
+        }
+    }
+
+    fn into_command(self, attributes: Attributes) -> ProtoCommand {
```
Sushisource: Nit: It feels weird to me that this would be `into` a command. Usage wise it makes sense, but the command is really like the "primary" thing and this I would imagine would be more like "attached" to a command.

Maybe we want to add an `attach_annotations` impl on `Attributes` so it can be called on that?

Sushisource (approved):
Copilot comments about test coverage are reasonable too.

crates/sdk-core/src/core_tests/event_groups.rs
mjameswh: > Probably the other unit test that was on updates.rs should move in here

I really don't think that other test should be associated with event groups at all, even though that was the trigger for adding those fields. I split the existing test in two, signal and update, and moved the signal one to `core_tests/workflow_tasks.rs`.

crates/sdk-core/src/worker/workflow/mod.rs
mjameswh: Here's what I suggest:

```
trait ProtoCommandExt {
    fn new(attributes: Attributes, annotations: CommandAnnotations) -> Self;
}

impl ProtoCommandExt for ProtoCommand { ... }
```

Sushisource: Ya, that works for me. I think you don't even need to make an extension trait, since you can impl on things directly in the lib where we include the generated protos.

mjameswh: > I think you don't even need to make an extension trait, since you can impl on things directly in the lib where we include the generated protos.

I can if we're ok moving `CommandAnnotations` to that crate, which I don't think is desirable.

crates/protos/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto
```diff
@@ -211,12 +211,10 @@ message InitializeWorkflow {
     temporal.api.common.v1.WorkflowExecution root_workflow = 24;
     // Priority of this workflow execution
     temporal.api.common.v1.Priority priority = 25;
-    // Event ID of the `WORKFLOW_EXECUTION_STARTED` history event that triggered this job.
-    int64 originating_event_id = 26;
     // The run id recorded on the `WORKFLOW_EXECUTION_STARTED` event. Unlike the execution's current
     // run id, this value is preserved across workflow resets. Mirrors the `original_execution_run_id`
     // field from `WorkflowExecutionStartedEventAttributes`.
-    string original_execution_run_id = 27;
+    string original_execution_run_id = 26;
```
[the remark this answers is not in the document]
mjameswh: Not an issue. Core's workflow activation already requires that lang's protobuf be regenerated every time core is bumped.

crates/sdk-core/src/worker/workflow/mod.rs
Sushisource: Ah, sure, right, ya makes sense then

PR #1515, opened 2026-08-21
feat: workflow command pagination

jmaeagle99:
## What was changed

- Add support for paginating workflow task completions when the namespace advertises support.
- Retry pagination with exponential backoff when server report buffer loss.
- Proactively fail workflow task completion on first attempt if commands size is known to be larger than configured namespace command buffer size limit.

## Why?

Enable submitting large batches of workflow commands that would otherwise fail the gRPC data size limit.

## Checklist

1. How was this tested: Unit tests, cloud integration tests, service mocking tests.


crates/sdk-core/src/worker/client.rs
```diff
@@ -45,11 +49,132 @@ use temporalio_common::protos::{
...
+
+/// How a workflow task completion should be delivered, produced by [paginate_wft_completion].
+enum WftCompletionPages {
+    /// Send as a single request: it fits within a page, or it cannot be split.
+    Single(RespondWorkflowTaskCompletedRequest),
+    /// The server buffers only the commands of intermediate pages, so all messages and metadata
+    /// ride on the final page.
+    Paginated {
+        intermediate_pages: Vec<RespondWorkflowTaskCompletedRequest>,
+        final_page: RespondWorkflowTaskCompletedRequest,
```
Sushisource, line 94: Why isn't the final page just the last item in the vec?

crates/sdk-core/src/worker/client.rs
```diff
@@ -45,11 +49,132 @@ use temporalio_common::protos::{
...
+    },
+}
+
+/// Split a completion that exceeds `max_page_bytes` into pages that each stay under it, by
+/// distributing its commands across intermediate pages in order.
```
Sushisource:
```suggestion
/// Split a completion that may exceed `max_page_bytes` into pages that each stay under it, by
/// distributing its commands across intermediate pages in order.
```

crates/sdk-core/src/worker/client.rs
```diff
@@ -45,11 +49,132 @@ use temporalio_common::protos::{
...
+        task_token: request.task_token.clone(),
+        identity: request.identity.clone(),
+        namespace: request.namespace.clone(),
+        intermediate_page: true,
+        ..Default::default()
+    };
+    let base_len = intermediate_template.encoded_len();
+    // A command is a repeated message entry: a field tag plus a length-delimited body. Six bytes
+    // bounds the tag (1) and the length varint (up to 5) so a packed page stays under the limit.
+    let command_framing = 6;
```
Sushisource: Seems potentially fragile if the proto definitions get fields added to them? Or maybe I'm misunderstanding this.

crates/sdk-core/src/worker/client.rs
```diff
@@ -507,17 +635,94 @@ impl WorkerClient for WorkerClientBag {
...
+        }
+
+        // Buffer loss is transient, so resend the whole set from page 0 with exponential backoff.
+        // The server bounds the loop: once the task times out it starts a new attempt, and the next
+        // resend fails the token check with a non-buffer-lost error. Shutdown ends the loop sooner,
+        // so a stream of buffer losses cannot hold shutdown open until the task times out. Recovery
+        // has to live here because the client's retry layer would resend only the single failed
+        // page, which cannot rebuild the buffer the server dropped; it is told to pass buffer loss
+        // straight through (see `wft_completion_page_request`).
+        let mut backoff = WFT_COMPLETION_PAGE_RESEND_INITIAL_BACKOFF;
```
Sushisource: If you wanted, could use the `backon` crate we already use in client for backoff here, but I suppose this doesn't need much fanciness.

crates/sdk-core/src/worker/client.rs
```diff
@@ -507,17 +635,94 @@ impl WorkerClient for WorkerClientBag {
...
+                        async move {
+                            client
+                                .respond_workflow_task_completed(wft_completion_page_request(page))
+                                .await
+                        }
+                    })
+                    .buffer_unordered(MAX_CONCURRENT_WFT_COMPLETION_PAGES)
+                    .try_collect::<Vec<_>>()
+                    .await?;
+                self.client
```
Sushisource, line 723: Would be worth calling out why the final page needs to specifically be sent after all the others (if it in fact does)

crates/sdk-core/src/worker/client.rs
```diff
@@ -965,6 +1170,15 @@ pub struct WorkflowTaskCompletion {
...
     pub versioning_behavior: VersioningBehavior,
+    /// Whether the namespace permits paginating this completion across multiple page requests when
+    /// it would otherwise exceed the server's gRPC request size limit.
+    pub pagination_enabled: bool,
+    /// The namespace's limit on the recombined size of a paginated completion, if the server
+    /// advertises one. A paginated completion larger than this is rejected server-side with
+    /// `REQUEST_TOO_LARGE`, so the worker fails it proactively instead of sending doomed pages.
+    pub wft_completion_size_limit: Option<usize>,
+    /// Signaled when the worker is shutting down.
+    pub shutdown_token: CancellationToken,
```
Sushisource: It feels a little inverted to me to have the shutdown token inside the request struct which is really just a data container. I think it'd make more sense for the worker who is calling the completion to simply drop (or arbort) the future.

crates/sdk-core/src/worker/client.rs
jmaeagle99: The final page is special in that it must be sent after all of the intermediate pages are sent; will update comments to describe this requirement. So I have it carve out the final page so that it can more easily directly iterate the intermediate pages without having to know which is the final page within the vec.

crates/sdk-core/src/worker/client.rs
jmaeagle99: I don't think the caller should drop the future when the shutdown token is signaled; that would cause all outstanding work to be thrown away during graceful shutdown, making any work performed up until the WFT completion unnecessary. The intent with this is to not attempt to resubmit pages for buffer loss if we are already shutting down, otherwise it'll hold open until it receives a different error from the server. I'll adjust comments and thread the shutdown token through as a function parameter instead.

crates/sdk-core/src/worker/client.rs
jmaeagle99: Dropped the more precise per-command size calculation in favor of a large head room offset for the page size.

Sushisource (approved):

PR #1517, opened 2026-08-24
feat(metrics): emit `failure_reason` on existing activity failure metrics

### LINKED ISSUE #1516: Add failure_reason label to activity task failure metrics - Rust
Parent issue: https://github.com/temporalio/features/issues/865

jmaeagle99:
## What was changed

- Activity and local activity failure counters now carry `failure_reason`, matching workflow and Nexus task failures. Values: `ActivityError` (default), `PayloadsTooLarge`, `ExternalStorageError`, and `timeout` for local activity timeouts. `GrpcMessageTooLarge` has been deliberately left out since activity workers do not produce activity task failures when encountering gRPC message size limit error.
- Also corrects a mislabeled counter: when oversized final heartbeat details replaced the reported failure, the metric was labeled from the lang cause and suppressed entirely for benign failures, though a payloads-too-large failure reached the server.

## Why?

Better metrics observability into known activity failure reasons.

## Checklist

1. Closes #1516

2. How was this tested: unit and mock integration tests

3. Any docs updates needed? Yes

### DIFF
--- a/crates/protos/src/protos/mod.rs
+++ b/crates/protos/src/protos/mod.rs
@@ -746,7 +746,7 @@ pub mod coresdk {
                     Self {
                         status: Some(aer::Status::Failed(Failure {
                             failure: Some(fail),
-                            ..Default::default()
+                            cause: ActivityTaskFailedCause::ActivityWorkerUnhandledFailure as i32,
                         })),
                     }
                 }
@@ -833,7 +833,8 @@ pub mod coresdk {
                             Ok(p) => Some(aer::Status::Completed(Success { result: Some(p) })),
                             Err(f) => Some(aer::Status::Failed(Failure {
                                 failure: Some(f),
-                                ..Default::default()
+                                cause: ActivityTaskFailedCause::ActivityWorkerUnhandledFailure
+                                    as i32,
                             })),
                         },
                     }
--- a/crates/protos/src/protos/mod.rs
+++ b/crates/protos/src/protos/mod.rs
@@ -746,6 +746,7 @@ pub mod coresdk {
                     Self {
                         status: Some(aer::Status::Failed(Failure {
                             failure: Some(fail),
+                            ..Default::default()
                         })),
                     }
                 }
@@ -830,7 +831,10 @@ pub mod coresdk {
                     Self {
                         status: match r {
                             Ok(p) => Some(aer::Status::Completed(Success { result: Some(p) })),
-                            Err(f) => Some(aer::Status::Failed(Failure { failure: Some(f) })),
+                            Err(f) => Some(aer::Status::Failed(Failure {
+                                failure: Some(f),
+                                ..Default::default()
+                            })),
                         },
                     }
                 }
@@ -866,6 +870,7 @@ pub mod coresdk {
                     match self.status {
                         Some(activity_resolution::Status::Failed(Failure {
                             failure: Some(ref f),
+                            ..
                         })) => f.is_timeout(),
                         _ => None,
                     }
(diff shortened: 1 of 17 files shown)


crates/sdk-core/src/telemetry/metrics.rs
```diff
@@ -755,22 +758,26 @@ pub(crate) fn eager(is_eager: bool) -> MetricKeyValue {
...
     GrpcMessageTooLarge,
     PayloadsTooLarge,
+    ExternalStorageError,
 }
 impl Display for FailureReason {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let str = match self {
             FailureReason::Nondeterminism => "NonDeterminismError".to_owned(),
             FailureReason::Workflow => "WorkflowError".to_owned(),
+            FailureReason::Activity => "ActivityError".to_owned(),
```
GregoryTravis, line 774: While we're adding `Activity`, maybe also `Worker`, for `ACTIVITY_TASK_FAILED_CAUSE_ACTIVITY_WORKER_UNHANDLED_FAILURE`?

crates/protos/src/protos/mod.rs
```diff
@@ -830,7 +831,10 @@ pub mod coresdk {
                     Self {
                         status: match r {
                             Ok(p) => Some(aer::Status::Completed(Success { result: Some(p) })),
-                            Err(f) => Some(aer::Status::Failed(Failure { failure: Some(f) })),
+                            Err(f) => Some(aer::Status::Failed(Failure {
+                                failure: Some(f),
+                                ..Default::default()
```
Sushisource: Some of the situations both of the changes in this file cover may actually constitute `ACTIVITY_TASK_FAILED_CAUSE_ACTIVITY_WORKER_UNHANDLED_FAILURE` - worth tracing the call path.

crates/sdk-core/src/worker/client/mocks.rs
```diff
@@ -22,7 +22,18 @@ pub(crate) static DEFAULT_TEST_CAPABILITIES: &Capabilities = &Capabilities {
 #[cfg(any(feature = "test-utilities", test))]
 /// Create a mock client primed with basic necessary expectations
 pub fn mock_worker_client() -> MockWorkerClient {
+    mock_worker_client_with_error_limits(None)
+}
+
+#[cfg(any(feature = "test-utilities", test))]
+/// As [mock_worker_client], reporting the given payload/memo error limits. Mockall matches
+/// expectations in creation order, so limits must be set here rather than overridden afterward.
```
Sushisource: Yeah, this frustrates me about this lib: https://github.com/asomers/mockall/issues/283

Some of the other tests where we need to do special setup expectations just recreate the whole thing. Having a specific overload just for this one is slightly funky, but, I'm not hugely against it.

GregoryTravis (approved):

crates/sdk-core/src/telemetry/metrics.rs
jmaeagle99: I've added `ActivityTaskFailedCause::ActivityWorkerUnhandledFailure => FailureReason::Activity` already for the conversion. This is symmetrical with how `WorkflowTaskFailedCause.WorkflowWorkerUnhandledFailure` is mapped to `FailureReason::Workflow`. Today, there is no distinction between a worker faulting and the workflow / activity / etc faulting from a cause and metrics standpoint. While making that distinction might be good to do, I won't be doing that as part of this effort.

Sushisource (approved):

crates/protos/src/protos/mod.rs
jmaeagle99: All of the non-test situations today should default this to `ActivityTaskFailedCause::ActivityWorkerUnhandledFailure`

crates/sdk-core/src/worker/client/mocks.rs
jmaeagle99: I split the test into two so we don't have to make more testing helpers public.

PR #1518, opened 2026-08-24
Run Cloud-eligible integration tests with filtering

THardy98:
## What was changed

Integration tests are now Cloud-eligible by default. Tests that do not exercise a server, require
a local server or additional Cloud provisioning, or have a confirmed Cloud incompatibility use a
source-local `#[cloud_test_exclusion(Category, "specific reason")]` annotation. Cloud mode turns
those annotations into native ignored tests, including every case generated by `rstest`.

The classification was built by running the suite against an isolated Cloud namespace. Direct
hard-coded `default` namespace assumptions were changed to use the harness-selected namespace;
broader test fixes remain explicitly excluded.

## Why?

Cloud compatibility should be visible where each test is maintained and based on observed
behavior, without maintaining a separate regex list. Developers can inspect skipped cases with
libtest's native `--ignored --list`, while new tests enter the Cloud suite unless they declare a
concrete exclusion.

## Checklist

1. Part of temporalio/features#851.

2. How was this tested:

   - `cargo +1.94 test -q -p temporalio-macros`
   - `cargo +1.94 integ-test -s external --cloud -- --ignored --list` lists all 231 exclusions.
   - `cargo +1.94 lint`
   - `cargo +1.94 test-lint`
   - `actionlint -ignore 'SC2086' .github/workflows/per-pr.yml`
   - Cloud qualification: 326 passed, 0 failed, 231 ignored in
     [runs 33782319827](https://github.com/temporalio/sdk-rust/actions/runs/33782319827) and
     [33785575299](https://github.com/temporalio/sdk-rust/actions/runs/33785575299).

3. Any docs updates needed?

   Yes. The README documents Cloud filtering, exclusion categories, required notes, and native
   skipped-test listing.


crates/sdk-core/tests/integ_tests/workflow_tests/activities.rs
```diff
@@ -2189,6 +2188,10 @@ async fn long_activity_timeout_repro() {
     worker.run_until_done().await.unwrap();
 }
 
+#[temporalio_macros::cloud_test_exclusion(
+    DoesNotUseServer,
+    "Uses MockPollCfg and a mocked SDK worker with synthetic history; no Temporal server is contacted."
+)]
```
Sushisource, line 2194: Seems like the overwhelming majority of these use a note that is 100% redundant with the categorized reason. No need for both. We should only bother with the note if it provides additional information.

crates/macros/src/cloud_test_exclusion.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+impl Parse for CloudTestExclusion {
+    fn parse(input: ParseStream<'_>) -> syn::Result<Self> {
+        let reason = input.parse::<Ident>()?;
+        match reason.to_string().as_str() {
+            "DoesNotUseServer"
+            | "RequiresLocalServer"
+            | "RequiresCloudProvisioning"
+            | "NeedsCloudAdaptation" => {}
```
Sushisource, line 60: These reasons look like real types but aren't. These should be defined as an actual enum somewhere - it's not really possible to discover what the valid values are.

crates/sdk-core/tests/integ_tests/client_tests.rs
```diff
@@ -38,6 +38,10 @@ use tonic::{
 };
 use tracing::info;
 
+#[temporalio_macros::cloud_test_exclusion(
+    RequiresCloudProvisioning,
```
Sushisource, line 42: This is more like "RequiresOSSOnlyAPIs" or something

crates/sdk-core/tests/integ_tests/worker_tests.rs
```diff
@@ -342,6 +342,10 @@ impl PaginatedCompletionWf {
 /// server buffers and reassembles; the workflow then completes normally. Local-lane only: it needs
 /// the dev server's `history.enableWorkflowTaskCompletionPagination` and a raised
 /// `system.transactionSizeLimit`.
+#[temporalio_macros::cloud_test_exclusion(
+    RequiresLocalServer,
+    "Requires dev-server workflow-task pagination and transaction-size dynamic configuration."
```
Sushisource, line 347: This is an example of a case we'll want to enable for cloud at some point once it's available there. Not sure what the best way to track that is. Maybe something formalized isn't really worth it. Open to suggestions.

Sushisource (commented):
Overall looking good to me, thanks for this. Just a couple things to address. Main feedback here is we can remove a lot of the string reasons. Some of them are interesting but a lot of them have no value over the reason type.

crates/sdk-core/tests/integ_tests/worker_tests.rs
THardy98: I think this is something that will recur, my thinking was to periodically filter on `RequiresLocalServer` to check for new candidates that are viable for cloud

THardy98: (same for other filters)

Sushisource (approved):

PR #1531, opened 2026-08-25
Respect counters_total_suffix in Prometheus exporter

### LINKED ISSUE #1526: [Bug] Prometheus Export ignoring counters_total_suffix option

Export Prometheus counter metrics with the _total suffix.

### Describe the bug

Configure Prometheus exporter with counters_total_suffix=True
https://github.com/temporalio/sdk-python/blob/678be2b431b66ca239f2d93ead9887ce5a6046d5/temporalio/bridge/src/runtime.rs#L85

Setting that to true, we expect counter metrics such as `temporal_long_request` to be exported as `temporal_request_total` 

### Minimal Reproduction
Have checked this both the sdk-python and sdk-typescript.

In the python prometheus sample, add 
```counters_total_suffix=True```
https://github.com/temporalio/samples-python/blob/main/prometheus/worker.py#L34

Note: other options, durations_as_seconds=True, unit_suffix=True **do work** as expected.

### Environment/Versions
Reproduced on:
Python sdk 1.23.0
TS sdk 1.22.0

Assuming it affects core based sdks.

VegetarianOrc:
## What was changed

- Thread `PrometheusExporterOptions::counters_total_suffix` into the Core Prometheus meter.
- Append `_total` to counter metric names when enabled, while preserving existing names when disabled.
- Add unit and integration coverage for both option values.
- Add user-facing entries to both changelogs.

## Why?

The exporter accepted the counter suffix option but ignored it. Enabling the option therefore did not produce Prometheus-compatible `_total` counter names.

## Checklist

1. Closes #1526

2. How was this tested:

- `cargo test -p temporalio-common --features prometheus counter_total_suffix_option_controls_counter_names`
- `cargo integ-test prometheus_metrics_exported`

3. Any docs updates needed?

- No documentation updates are needed. The user-visible fix is documented in both changelogs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,8 @@ relevant information.
   `WorkflowStartOptions::start_signal` and `WorkflowStartSignal`.
 
 ### Fixed
+* The Prometheus exporter now respects `PrometheusExporterOptions::counters_total_suffix`,
+  appending `_total` to counter metric names when enabled.
 * An activity failure caused by oversized final heartbeat details is now counted in the
   `temporal_activity_execution_failed` metric as `failure_reason="PayloadsTooLarge"`. Previously it
   was counted under the reason for the failure the activity itself reported, and was not counted at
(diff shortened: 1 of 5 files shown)


Sushisource (approved):
Thanks!

PR #1534, opened 2026-08-26
Fix non-sticky WFT poller starvation with small workflow cache

DABH:
## Problem

With a small `max_cached_workflows`, a worker can intermittently fail to pick up a **new** workflow's first task for ~60s. The workflow sits with its first WFT `SCHEDULED` and `DescribeTaskQueue` on the normal queue shows no poller, until a sticky long-poll times out and it self-heals.

The way this surfaced was flaky behavior observed in sdk-typescript CI tests (https://github.com/temporalio/sdk-typescript/pull/2310) where the proposed fix was to just add retries to deal with the flake.  But ended up going down the rabbit hole to try and root-cause this and arrived here.

## Root cause

The sticky and non-sticky WFT pollers share a permit pool. `MeteredPermitDealer::acquire_owned` caps concurrency at `max_permits` = the **workflow cache size**. The balancer meant to keep at least one non-sticky poller alive (`WFTPollerShared`) was initialized with the **slot-supplier size** instead, so when the cache is smaller than the slot supplier the balancer never engages. Under startup contention the sticky poller's long-polls then take every shared permit and the non-sticky poller (which fetches new workflows' first tasks) starves at `acquire_owned`.

## Fix

Initialize the balancer with the limit `acquire_owned` actually enforces: `min(slot-supplier size, cache size)`. No-op when the cache is >= the slot-supplier size (the common case is unchanged); reserves a poll slot for the non-sticky poller when the cache is smaller.

### Why #835 didn't cover this
#835 (closing #775) fixed the same starvation class, but only in the resource-based tuner's slot supplier: it guarantees that supplier hands out at least one slot to each poller kind. It doesn't apply to the fixed-size slot supplier, where the binding concurrency limit is the workflow cache size (`max_permits`) rather than the supplier, so the shared `WFTPollerShared` balancer still gates on the slot-supplier size and a small cache lets the starvation recur. This PR closes that gap by balancing on `min(slot supplier, cache size)`, which covers both suppliers.

## Validation

Reproduced via a language-SDK worker with `max_cached_workflows: 2` under CPU oversubscription (fresh worker per iteration, new workflow each time): ~2–3% of iterations stalled ~60s with the exact signature above.

A/B on an otherwise-identical build (only this diff differs):
- without fix: 11 stalls / 400 iterations
- with fix: 0 stalls / 400; 0 / 800 under heavier load; 0 across 1600+ total
- large cache without the fix: 0 (confirming cache size is the lever)

`cargo test` for `worker::workflow::wft_poller` (incl. the added `poll_balance_limit` test) and `pollers::poll_buffer` pass; clippy clean on the touched files.

### DIFF
--- a/crates/sdk-core/src/worker/workflow/wft_poller.rs
+++ b/crates/sdk-core/src/worker/workflow/wft_poller.rs
@@ -316,15 +316,16 @@ mod tests {
     #[tokio::test]
     async fn small_cache_does_not_starve_nonsticky_poller() {
         use crate::{
+            MetricsContext,
             abstractions::MeteredPermitDealer,
             replay::TestHistoryBuilder,
             test_help::{ResponseType, hist_to_poll_resp, test_worker_cfg},
             worker::{
-                NamespaceCapabilities, PollerBehavior, client::mocks::mock_manual_worker_client,
+                NamespaceCapabilities, PollerBehavior,
+                client::{WorkerClient, mocks::mock_manual_worker_client},
                 tuner::FixedSizeSlotSupplier,
             },
         };
-        use crate::{MetricsContext, worker::client::WorkerClient};
         use futures_util::FutureExt;
         use std::sync::atomic::{AtomicUsize, Ordering};
         use temporalio_common::protos::temporal::api::enums::v1::EventType;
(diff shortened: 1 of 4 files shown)


crates/sdk-core/src/worker/workflow/wft_poller.rs
```diff
@@ -16,6 +16,12 @@ use temporalio_common::protos::temporal::api::workflowservice::v1::PollWorkflowT
 use tokio::sync::watch;
 use tokio_util::sync::CancellationToken;
 
+/// The concurrency limit the WFT poll balancer should reserve against: the smaller of the slot
+/// supplier size and the workflow cache size (both bound `acquire_owned`).
+fn poll_balance_limit(slots: Option<usize>, cache: Option<usize>) -> Option<usize> {
+    [slots, cache].into_iter().flatten().min()
+}
```
Sushisource: This is trivial enough that it should just be inlined, the unit test is uninteresting.

Sushisource (commented):
I think this makes sense, but, we need an actual integration test to (ideally deterministically, but at least stochastically) prove we're fixing something here.

It doesn't necessarily need to use real server / real tasks, could use mock ones, but it should integration test the actual poller and/or worker behavior.

crates/sdk-core/src/worker/workflow/wft_poller.rs
DABH: Done

Sushisource (approved):
Thanks!

PR #1536, opened 2026-08-26
fix(sdk): mark remaining public types with `non_exhaustive`

### LINKED ISSUE #1530: Audit public types for `#[non_exhaustive]` usage
Go through codebase and make sure all public structs/enums are marked as `#[non_exhaustive]` where appropriate. Primarily should focus on types that are currently constructible by consumers and adding a new field/variant would be a breaking change. `#[non_exhaustive]` would prevent this.

chris-olszewski:
## What was changed
Went through and got the final batch of pub types marked with `#[non_exhaustive]`. Pretty straightforward with using `bon` where appropriate.

Envconfig changes are kinda annoying as they make the c bridge conversion faillable and we need to bubble that up. I cannot image we expand beyond direct data/file so happy to keep this as exhaustive if someone thinks this is an overstep.

I did these commit by commit for the most part so each one should be reviewable. A lot of noise just from moving over `SerializationContext` construction to a constructor.

## Why?
Prevent future extensions to these types from being breaking changes

## Checklist

1. Closes #1530 

2. How was this tested:
`rustc`

3. Any docs updates needed?
I don't believe so as these are more hidden types that I don't expect end users would be constructing often.


Sushisource (approved):
LoC decrease!?!?!? 🎉

PR #1538, opened 2026-08-26
Fix worker shutdown racing in-flight activity completions

### LINKED ISSUE #1537: [Bug] Worker shutdown can permanently strand an in-flight activity completion, dropping its result
## Describe the bug

`Worker::shutdown()` can race an activity completion that is still reporting its result to the server, and in the worst interleaving that completion is stranded **forever**:

1. `WorkerActivityTasks::complete` removes the activity from the outstanding-task map *before* awaiting heartbeat eviction and the completion RPC. The removed record still owns the activity's slot permit until the function returns.
2. The activity task stream's shutdown drain treats "outstanding map empty" as "all activities finished", which becomes true the instant the entry is removed — while eviction and the RPC are still in flight.
3. If the activity's last heartbeat RPC is still in flight, eviction defers its acknowledgement until that RPC completes (`tt_needs_flush`) — but the heartbeat lifecycle loop exits via a `biased` select on the shutdown token and abandons queued messages, so the deferred ack is never delivered. The completion waits on a `Notify` that will never be signalled.

Shutdown then reaches the final slot-permit wait holding a permit that can never be released. After five seconds:

- **debug builds** panic via `dbg_panic!`: `Waiting for all slot permits to release took too long!`
- **release builds** log the invariant violation and return anyway — and the activity's result (e.g. a retryable failure) is **never sent to the server**, so the attempt only resolves once server-side timeouts fire. `Worker::shutdown()` reports success to the caller.

The window is widest when the final heartbeat RPC is slow (any high-latency `RecordActivityTaskHeartbeat`), but it is a pure ordering bug with no timing dependence required to reproduce.

## Minimal Reproduction

Deterministic, with a mocked client (no timing/sleeps):

1. Poll one activity task (with a heartbeat timeout set) and record a heartbeat, holding the mock `RecordActivityTaskHeartbeat` future open.
2. While that RPC is in flight, call `complete_activity_task` with a failure result — the completion removes the task from the outstanding map, then parks in heartbeat eviction awaiting the deferred ack.
3. Initiate shutdown and poll to `PollError::ShutDown` (which succeeds immediately because the map is empty), then await `Worker::shutdown()`.
4. Release the heartbeat RPC. Debug: the 5s `dbg_panic` fires. Release: shutdown returns with `RespondActivityTaskFailed` never issued.

An accompanying PR contains this as two regression tests (unit-level and worker-level); both fail on current `main` and on `v0.7.0` without the fix, and the release-mode result loss is demonstrated by running the worker-level test with `--release`.

## Environment/Versions

- OS and processor: any (reproduced on macOS arm64)
- Temporal Version: sdk-rust `v0.7.0` and current `main` (d9efc491) — the relevant structure is identical in both
- Server version: not relevant (reproduced against a mocked service and confirmed fixed against an ephemeral dev server)

## Additional context

A fix PR accompanies this issue: it tracks completions in flight from before the task leaves the outstanding map until the slot permit is released, requires that count to reach zero before the activity drain reports done (which also keeps the heartbeat manager alive until every deferred eviction ack is delivered), and leaves the five-second `dbg_panic` in place as the invariant tripwire.

iw:
Fixes #1537

## What was happening

`Worker::shutdown()` can race an activity completion that is still reporting its result to the server, and in the worst interleaving that completion is stranded **forever**:

1. `WorkerActivityTasks::complete` removes the activity from the outstanding-task map *before* awaiting heartbeat eviction and the completion RPC. The removed record still owns the activity's slot permit until the function returns.
2. The activity task stream's shutdown drain treats "outstanding map empty" as "all activities finished", which becomes true the instant the entry is removed — while eviction and the RPC are still in flight.
3. If the activity's last heartbeat RPC is still in flight, eviction defers its acknowledgement until that RPC completes — but the heartbeat lifecycle loop exits via a `biased` select on the shutdown token and abandons queued messages, so the deferred ack is never delivered. The completion waits on a `Notify` that will never be signalled.

Shutdown then reaches the final slot-permit wait holding a permit that can never be released. After five seconds, debug builds panic (`dbg_panic!`: `Waiting for all slot permits to release took too long!`); release builds log the invariant violation and return anyway — and the activity's result (e.g. a retryable failure) is **never sent to the server**, so the attempt only resolves once server-side timeouts fire.

The window is easiest to hit when the final heartbeat RPC is slow (a high-latency `RecordActivityTaskHeartbeat`), but it is a pure ordering bug — the regression tests reproduce it deterministically with a mocked client and no timing dependence.

## The fix

Lifecycle accounting rather than timeout inflation, keeping the five-second `dbg_panic` as the invariant tripwire:

- A completion now registers itself in a `completions_in_flight` counter *before* taking its task out of the outstanding map, via a drop-guard that releases only after the slot permit itself is dropped (panic- and cancellation-safe).
- The shutdown drain requires both an empty outstanding map **and** a zero in-flight-completion count. Since polling can no longer observe "drained" mid-completion, the heartbeat manager now outlives every deferred eviction ack, the completion RPC always finishes before `Worker::shutdown()` returns, and permits are released before the deadline.
- The drain loop is rebuilt on `Notified::enable()` (arm the waiter before checking), closing a pre-existing lost-wakeup window in the same loop.

## Tests

Two regression tests, both synchronization-driven (Notify/oneshot, no sleeps), both verified to fail before the fix and pass after it:

- `worker::activities::tests::shutdown_waits_for_completions_still_flushing_their_result` — unit-level: without the fix, `WorkerActivityTasks::shutdown` resolves while the failure was never reported.
- `core_tests::activity_tasks::worker_shutdown_awaits_activity_completion_flushing_result` — worker-level: without the fix, debug builds reproduce the exact `Waiting for all slot permits to release took too long!` panic after 5s, and `--release` builds demonstrate the silent result loss.

Validation: full sdk-core unit suite and workspace suites pass; shutdown- and activity-filtered integration tests (11 + 166) pass against a live ephemeral server; both regression tests stable across 10 repeat runs in debug and release; clippy (sdk-core, all features, tests) and `cargo fmt --check` clean. Changelog entries added to both the root and sdk-core changelogs. The change also cherry-picks cleanly onto `v0.7.0` (all tests pass there), in case a patch release is considered.


crates/sdk-core/src/worker/activities.rs
```diff
@@ -631,11 +578,38 @@ impl WorkerActivityTasks {
     }
 }
 
+/// Holds a completion's spot in [WorkerActivityTasks::completions_in_flight] from before its task
+/// leaves the outstanding map until every resource it owns (notably the slot permit) is released.
+/// A guard rather than a manual decrement so cancellation or a panic mid-completion can't leave
+/// shutdown waiting forever.
+struct InFlightCompletionGuard {
```
Sushisource: This is generic enough that it could go in `abstractions.rs` under a generic name. I thought I already had exactly this in the code somewhere but didn't find it.

crates/sdk-core/src/worker/activities.rs
```diff
@@ -1360,4 +1336,121 @@ mod tests {
...
+
+    // Regression test for a shutdown race: a completion takes its task out of the outstanding
+    // map before flushing the result to server, while still owning the activity's slot permit.
+    // If the last heartbeat's RPC is still in flight, the completion parks inside heartbeat
+    // eviction with the map already empty. Shutdown used to treat the empty map as fully
+    // drained and tear down the heartbeat manager, which dropped the pending eviction ack —
+    // stranding the completion forever, with the failure never reported and the slot permit
+    // never released.
+    #[tokio::test]
+    async fn shutdown_waits_for_completions_still_flushing_their_result() {
```
Sushisource: I think the other test that looks at things from a more complete e2e viewpoint is more valuable, and this is pretty much exactly the same one at a lower level. Can probably just get rid of it.

What might be better is to upgrade the other test so that it tests what happens not just if the heartbeat call hangs, but if the complete/fail calls hang too.

crates/sdk-core/src/worker/activities.rs
iw: Good call — it turned out to be `ActiveCounter` in `abstractions.rs` you were remembering. Switched to it in f0f78e84, which also let me drop the bespoke guard and the Notify entirely: the watch channel's versioning makes the drain loop's hand-rolled lost-wakeup handling unnecessary.

*🤖 Claude authored*

crates/sdk-core/src/worker/activities.rs
iw: Done in f0f78e84 — removed the unit-level test and parameterized the worker-level one over which RPC hangs (heartbeat vs. fail). One note: the hung-fail case passes even without the fix, since `completers_lock` already covers a completion parked in the result RPC itself — the hung-heartbeat case is the one that reproduces the original race (the eviction ack deferred behind the in-flight heartbeat). Kept both since the fail case pins the invariant independently of that lock.

*🤖 Claude authored*

crates/sdk-core/src/core_tests/activity_tasks.rs
```diff
@@ -666,40 +667,66 @@ async fn can_heartbeat_acts_during_shutdown() {
     core.drain_activity_poller_and_shutdown().await;
 }
 
+#[derive(PartialEq)]
+enum HungRpc {
+    Heartbeat,
+    Fail,
```
Sushisource, line 675: What about a successful completion, just for completeness?

iw: Added in 7b6409ec — `case_3_complete_rpc_hangs`, where the activity succeeds and the hung RPC is `RespondActivityTaskCompleted`. The reported-flag now covers whichever result RPC runs, so the assertion reads the same across all three cases.

*🤖 Claude authored*

Sushisource (approved):
Thanks! Not completely sure why CI isn't running or giving me the option to run it but we'll figure that out and get it merged.

PR #1553, opened 2026-08-31
chore(sdk): make serialization context extensible

chris-olszewski:
## What was changed
Add some empty structs to the variants on the serialization context so we can add data there in the future.

Dropped `Copy` implementation since these won't be `Copy`able once they contain real data.

Constructing these is experimental as the signature will change in the future when these need data.

## Why?
At the moment adding data here would be a breaking change.

We could remove the variants and only have `None`, but that feels a little bit like throwing the baby out with the bathwater. We already have the variants wired up in the codebase.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Sushisource (approved):
Ok you can blame me for not doing that in the first place 😬

PR #1554, opened 2026-08-31
chore(sdk): rename local activity argument inclusion option

jmaeagle99:
## What was changed

Rename `include_arguments_into_marker` to `include_arguments_in_marker`.

## Why?

More grammatically correct name.


Sushisource (approved):
Fair

PR #1563, opened 2026-09-02
chore(sdk): resolve some doc(hidden) APIs

chris-olszewski:
## What was changed
Resolving some smaller `#[doc(hidden)]` declarations that don't need full PRs to resolve.

Generally fall into a few buckets:
 - Test helpers that aren't necessary if tests using them are moved into the defining crate
 - Just hiding some types that have no real reason being public
 - Making `event_group_markers` fields on options public. They aren't indented for use and are properly marked as experimental/unstable. If anything having them `doc(hidden)` hid that they shouldn't be used.
 - Remove a small `doc(hidden)` helper function that was 1 line in favor of just defining it in each crate that used it.
 - Limit a reexport used for macros to just the identifiers used in the macros.

Highly suggest reviewing each commit individually.

## Why?
public `#[doc(hidden)]` is generally a smell we should avoid

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


crates/common-wasm/src/memo.rs
```diff
@@ -19,7 +19,6 @@ pub struct Memo {
 impl Memo {
     /// Construct a memo with the payload converter and serialization context associated with its
     /// source.
-    #[doc(hidden)]
```
chris-olszewski, line 22: Made this flat out public as I cannot imagine the API evolving from this point

Sushisource (approved):
Doing God's work 🙏

PR #1564, opened 2026-09-02
chore(sdk): limit macro exports

chris-olszewski:
## What was changed
Explicitly reexport identifiers used in macros instead of blanket reexporting entire modules.

## Why?
Prevents us accidentally leaking internal details that don't need to be used outside the crate. All "used for macro" reexports now have a single place they get accessed from.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


crates/sdk-core/tests/fixtures/wasm_patch_activation/src/lib.rs
```diff
@@ -24,34 +18,4 @@ impl PatchActivationWorkflow {
     }
 }
 
-struct WasmPatchActivationWorkflowModule;
-
-impl StaticWorkflowComponent for WasmPatchActivationWorkflowModule {
```
chris-olszewski, line 29: Unsure if there was a clear reason why we wanted to handroll this for testing?

crates/sdk-core/tests/integ_tests/workflow_tests/continue_as_new.rs
```diff
@@ -145,11 +146,13 @@ impl WfWithTimer {
     #[run(name = DEFAULT_WORKFLOW_TYPE)]
     async fn run(ctx: &mut WorkflowContext<Self>) -> WorkflowResult<()> {
         ctx.timer(Duration::from_millis(500)).await;
-        Err(WorkflowTermination::continue_as_new(ContinueAsNewRequest {
-            arguments: vec![[1].into()],
-            initial_versioning_behavior: ProtoContinueAsNewVersioningBehavior::AutoUpgrade.into(),
-            ..Default::default()
-        }))
+        ctx.continue_as_new(
```
chris-olszewski, line 149: There was no reason to be using an internal type for this test, was able to make it private

crates/sdk-core/tests/fixtures/wasm_task_failure/src/lib.rs
```diff
@@ -0,0 +1,102 @@
+use temporalio_workflow::{
```
chris-olszewski, line 1: Moved from the example. I don't think this is mostly for our own testing and not advertising to end users that they should use these types/traits directly

crates/workflow/src/runtime/instance.rs
```diff
@@ -1143,17 +1147,6 @@ where
     }
 }
 
-pub fn instantiate_workflow<W: WorkflowImplementation>(
```
chris-olszewski, line 1146: Wasn't sure what this offered over calling `GuestWorkflowInstance::instantiate` directly

crates/sdk-core/tests/fixtures/wasm_patch_activation/src/lib.rs
Sushisource: None

Sushisource (approved):
You must be tired of cleaning up my mess lol 😆 apologies

PR #1565, opened 2026-09-02
feat(sdk): expose proper cancel external error type

chris-olszewski:
## What was changed
I somehow missed this when I was doing the same for signal workflow. Making this a standard result with a proper error instead of exposing the custom types for interceptors.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,10 @@ relevant information.
 
 ## Unreleased
 
+### Added
+* `CancelExternalWorkflowError` and `workflow_interceptors::CancelExternalWorkflowResult`
+  for use in interceptors.
+
 ## [0.8.0] - 2026-09-02
 
 ### Breaking Changes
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -43,6 +43,8 @@ relevant information.
   SDK integrations.
 
 ### Added
+* `CancelExternalWorkflowError` and `workflow_interceptors::CancelExternalWorkflowResult`
```
chris-olszewski: Will fix this

crates/common-wasm/src/error.rs
```diff
@@ -1155,6 +1168,48 @@ pub enum WorkflowSignalError {
     Serialization(#[from] PayloadConversionError),
 }
 
+/// Error returned when requesting cancellation of an external workflow fails.
+#[derive(Debug, thiserror::Error)]
+pub enum CancelExternalWorkflowError {
+    /// The cancellation request failed.
+    #[error("External workflow cancellation request failed: {}", .0.failure().message)]
+    Failed(#[source] Box<IncomingError>),
```
Sushisource, line 1176: I thought we had a dedicated NotFound for similar things? If that's not true ignore me.

Sushisource (approved):

crates/common-wasm/src/error.rs
chris-olszewski: We have that on the client side, but not workflow side. Adding it in another PR up the stack 🫡

chris-olszewski: So actually Go/Java are the only SDKs with this distinction. I think it makes sense to have a special variant here, but it requires some plumbing from core.

PR #1569, opened 2026-09-03
feat(sdk): add not found variant for signal/cancel external wf

chris-olszewski:
### Rust
Added `NotFound` variant to signal/cancel external wf error types

### Core
Send the failure cause from core to lang for signal/cancel external wf

## Why?
Matches Go/Java behavior of surfacing the not found in a direct way instead of string matching.
## Checklist

1. Closes N/A

2. How was this tested:
Updated assertions in integration test to make sure correct error variant is chosen

3. Any docs updates needed?
N/A


Sushisource (approved):
noice

PR #1570, opened 2026-09-03
chore(sdk): mark pub test utilities as experimental

chris-olszewski:
## What was changed
- Mark remaining test utils as `experimental` in addition to doc hidden to reduce risk of anyone using them
- Fully remove `core_worker` accessor as it was only necessary for a few tests. Can get around its removal by just arc wrapping the core worker before the SDK worker construction.

## Why?
Will remove these test-only methods in the future, this just ensures that removal has no chance of breaking "stable" users.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/crates/sdk-core/tests/integ_tests/plugin_tests.rs
+++ b/crates/sdk-core/tests/integ_tests/plugin_tests.rs
@@ -114,13 +114,11 @@ async fn plugins_configure_client_and_worker() {
         .register_workflow::<SimplePluginWorkflow>()
         .unwrap()
         .build();
-    let worker = Worker::new(&runtime, client, worker_options).unwrap();
+    let _worker = Worker::new(&runtime, client, worker_options).unwrap();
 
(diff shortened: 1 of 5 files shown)


crates/sdk-core/tests/integ_tests/plugin_tests.rs
```diff
@@ -364,52 +362,3 @@ impl WorkerPlugin for FailingWorkerPlugin {
...
-struct WorkerOnlyMetadataPlugin;
-
-impl WorkerPlugin for WorkerOnlyMetadataPlugin {
-    fn name(&self) -> &str {
-        "worker-only-plugin"
-    }
-}
-
-#[tokio::test]
-async fn worker_metadata_includes_client_and_worker_plugin_names() {
```
chris-olszewski: This assertion is fully covered in the plugin unit tests

Sushisource (approved):

PR #1574, opened 2026-09-03
chore(client): hide jittered

chris-olszewski:
## What was changed
Dupe the code over a public export

## Why?
These 2 implementations don't need to be in lockstep for correctness. Duping the code feels better than committing to this in the public interface.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Sushisource (approved):
Heh, I asked the person who contributed this to de-dupe it.

Honestly given it's hidden, I prefer keeping it de-duped, but I don't feel strongly. Up to you.

PR #1579, opened 2026-09-03
Allow normal polling when sticky is saturated

yuandrew:
## What was changed
We now only prioritize sticky when its backlog exceeds its active pollers and its active count is **below** its autoscaling target.

## Why?
Sticky backlog could keep normal workflow pollers waiting after sticky pollers reached their autoscaling target. Because sticky could not add another poller, shared workflow slots could remain unused.

Share the sticky poller target with workflow poll admission. Sticky backlog now receives priority only while another sticky poller can start. Normal pollers may use spare capacity once sticky reaches its target.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **Medium Risk**
> Changes workflow task poll admission and sticky/non-sticky balancing in the worker hot path, which can alter throughput and latency under sticky backlog and autoscaling.
> 
> **Overview**
> Fixes a case where a large sticky workflow-task backlog could block **non-sticky** pollers even after sticky pollers were already at their configured or autoscaled limit, leaving shared poll capacity unused.
> 
> **Poll scaler target is now observable:** autoscaling target moves from an atomic counter to a `watch` channel so sticky admission and the sticky/non-sticky balancer can react when the target changes.
> 
> **Balancer rules:** sticky backlog only wins priority while another sticky poller is allowed (`active < target`). Once sticky is saturated at target, non-sticky pollers may proceed; sticky pollers stop reserving slots the scaler will not grant. Waiting pollers wake on target updates.
> 
> Changelog entries document the user-visible fix; unit tests cover target-aware priority and wake behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b2536ba29b999c57138b9bcbf4570e6bc51a959. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/sdk-core/src/worker/workflow/wft_poller.rs
+++ b/crates/sdk-core/src/worker/workflow/wft_poller.rs
@@ -9,7 +9,10 @@ use crate::{
 use crossbeam_utils::atomic::AtomicCell;
 use futures_util::{Stream, stream};
 use std::{
-    sync::{Arc, OnceLock},
+    sync::{
+        Arc, OnceLock,
+        atomic::{AtomicUsize, Ordering},
+    },
     time::SystemTime,
 };
 use temporalio_common::protos::temporal::api::workflowservice::v1::PollWorkflowTaskQueueResponse;
@@ -98,6 +101,7 @@ pub(crate) fn make_wft_poller(
 pub(crate) struct WFTPollerShared {
     last_seen_sticky_backlog: (watch::Receiver<usize>, watch::Sender<usize>),
     sticky_active: OnceLock<watch::Receiver<usize>>,
+    sticky_target: OnceLock<Arc<AtomicUsize>>,
     non_sticky_active: OnceLock<watch::Receiver<usize>>,
     max_slots: Option<usize>,
 }
@@ -107,12 +111,14 @@ impl WFTPollerShared {
         Self {
             last_seen_sticky_backlog: (rx, tx),
             sticky_active: OnceLock::new(),
+            sticky_target: OnceLock::new(),
             non_sticky_active: OnceLock::new(),
             max_slots,
         }
     }
-    pub(crate) fn set_sticky_active(&self, rx: watch::Receiver<usize>) {
+    pub(crate) fn set_sticky_active(&self, rx: watch::Receiver<usize>, target: Arc<AtomicUsize>) {
         let _ = self.sticky_active.set(rx);
+        let _ = self.sticky_target.set(target);
     }
     pub(crate) fn set_non_sticky_active(&self, rx: watch::Receiver<usize>) {
         let _ = self.non_sticky_active.set(rx);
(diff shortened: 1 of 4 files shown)


crates/sdk-core/src/worker/workflow/wft_poller.rs
```diff
@@ -107,12 +111,14 @@ impl WFTPollerShared {
         Self {
             last_seen_sticky_backlog: (rx, tx),
             sticky_active: OnceLock::new(),
+            sticky_target: OnceLock::new(),
             non_sticky_active: OnceLock::new(),
             max_slots,
         }
     }
-    pub(crate) fn set_sticky_active(&self, rx: watch::Receiver<usize>) {
+    pub(crate) fn set_sticky_active(&self, rx: watch::Receiver<usize>, target: Arc<AtomicUsize>) {
```
Sushisource: `rx` should be a more clear parameter name now

crates/sdk-core/src/worker/workflow/wft_poller.rs
```diff
@@ -124,8 +130,9 @@ impl WFTPollerShared {
         // that we won't end up using every available permit with one kind of poller. In practice
         // this is only ever likely to be an issue with very small numbers of slots.
         if let Some(max_slots) = self.max_slots
-            && let Some((sticky_active, non_sticky_active)) =
-                self.sticky_active.get().zip(self.non_sticky_active.get())
+            && let Some(sticky_active) = self.sticky_active.get()
+            && let Some(sticky_target) = self.sticky_target.get()
+            && let Some(non_sticky_active) = self.non_sticky_active.get()
```
Sushisource, line 136: This implies to me that we should see if we can find a way to just bundle up all the things that aren't max_slots into one here

crates/sdk-core/src/worker/workflow/wft_poller.rs
```diff
@@ -135,6 +142,9 @@ impl WFTPollerShared {
                 let num_sticky_active = *sticky_active.borrow_and_update();
                 let num_non_sticky_active = *non_sticky_active.borrow_and_update();
                 let num_sticky_backlog = *sticky_backlog.borrow_and_update();
+                let sticky_needs_capacity = num_sticky_backlog > 1
```
Sushisource:
```suggestion
                let sticky_should_be_chosen = num_sticky_backlog > 1
```

Seems more obvious name to me

Sushisource (approved):

crates/sdk-core/src/worker/workflow/wft_poller.rs
yuandrew: these values come from separate poller constructors. Bundling them would require a larger construction refactor, and grouping them here would only be cosmetic. I don’t think it’s worth the added complexity.

PR #1580, opened 2026-09-03
Add workflow-scoped context storage

eamsden:
## What was changed

- Added typed context storage for one Workflow run.
- Added async and synchronous context scopes for Workflow code and Workflow interceptors.
- Added read-only context lookup to `WorkflowContextView` for queries and Update validators.
- Added an example, API documentation, tests, and an Unreleased changelog entry.

The SDK restores the previous context after each Future poll. It also restores the context after completion, cancellation, or panic. Concurrent branches and handlers therefore do not share scoped values by accident.

The SDK does not serialize context values or propagate them automatically. An interceptor can read a value and put selected data in Temporal headers.

## Why?

Workflow code needs a safe way to give application context to SDK interceptors. For example, an OpenTelemetry interceptor must use an application span as the parent of an Activity, Child Workflow, Signal, Nexus Operation, or Continue-As-New span.

Process-global state and Tokio task-local state can leak between Workflow executions. This change keeps the state in the Workflow runtime. Replay rebuilds the state when Workflow code runs again.

## Checklist

1. Closes: N/A.

2. How was this tested:

   - `cargo test`
   - `cargo lint`
   - `cargo test-lint`
   - `cargo test -p temporalio-workflow --doc`
   - `cargo check -p temporalio-sdk --example workflow-context --features examples`
   - `cargo integ-test workflow_interceptors_mutate_inputs_and_replace_outputs -- --nocapture`
   - `cargo integ-test workflow_context_is_branch_and_handler_local_during_replay -- --nocapture`
   - `cargo +nightly fmt --check`

3. Any docs updates needed?

   No additional documentation update is needed. This PR adds public API documentation, an SDK example, and a changelog entry.

### DIFF
--- a/crates/workflow/src/workflow_context.rs
+++ b/crates/workflow/src/workflow_context.rs
@@ -50,8 +50,10 @@ use rand::SeedableRng;
 use rand_pcg::Pcg64Mcg;
 use siphasher::sip::SipHasher13;
 use std::{
+    any::{Any, TypeId},
     cell::{Cell, RefCell},
     collections::{HashMap, HashSet},
+    fmt,
     future::{self, Future},
     hash::Hasher,
     marker::PhantomData,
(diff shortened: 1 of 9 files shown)


crates/workflow/src/workflow_context.rs
```diff
@@ -2006,6 +2173,37 @@ impl<W> WorkflowContext<W> {
 
     // --- Delegated methods from SyncWorkflowContext ---
 
+    /// Return the value associated with key type `K` in the current workflow context scope.
+    pub fn context_value<K: WorkflowContextKey>(&self) -> Option<Rc<K::Value>> {
+        self.sync.context_value::<K>()
+    }
+
+    /// Poll `future` with `value` installed for key type `K`.
```
chris-olszewski, line 2181: I feel like the terminology of "poll" here is a little odd since this just constructs a future with this value

chris-olszewski (approved):
Great to have this addition

crates/workflow/src/workflow_context.rs
eamsden: It is a bit weird but the reason for the weirdness is that the context is only installed while polling and removed again after, thus making scoped contexts cancel-safe.

PR #1587, opened 2026-09-04
chore(sdk): add msrv to every crate

chris-olszewski:
## What was changed
Add MSRV to every crate we release and verify it.

Split this into 2 jobs for our different MSRV targets

## Why?
Best practices

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/crates/sdk/Cargo.toml
+++ b/crates/sdk/Cargo.toml
@@ -2,6 +2,7 @@
 name = "temporalio-sdk"
 version = "0.8.0"
 edition = "2024"
+rust-version = "1.92.0"
 authors = ["Spencer Judge <spencer@temporal.io>"]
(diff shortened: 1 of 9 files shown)


crates/sdk/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "temporalio-sdk"
 version = "0.8.0"
 edition = "2024"
+rust-version = "1.92.0"
```
chris-olszewski, line 5: Slightly annoying, but this is only higher than the other crates because of the wasmtime deps for `wasm-workflows` feature. No way to split MSRV by feature set so we take the higher one.

Sushisource (approved):

PR #1590, opened 2026-09-04
feat(client): allow usage of protox instead of system protoc

### LINKED ISSUE #1589: Support vendored/pure-Rust protoc for temporalio-protos (like prost-wkt-types already does)

`temporalio-protos`'s build script (`build.rs`) compiles Temporal's own `.proto` files via `tonic_prost_build`/`prost_build`, which requires a real `protoc` binary discoverable on `PATH` or via the `PROTOC` env var. There's no way to build a crate depending on `temporalio-client` without a system-installed `protoc` - it has to be present in every developer's own machine, every CI runner, and the *builder* stage of any multi-stage Docker build.

This is a real, hands-on-verified adoption friction point: I ran into it building a small bridge crate (`skilj-temporal`) on top of `temporalio-client`, and had to add a `protoc`-install step to both a CI workflow and CONTRIBUTING.md documentation just to get a clean build working. It's also somewhat related to #687 (dependency/build-time concerns for the Rust SDK), though narrower and more specifically actionable.

Notably, `prost-wkt-types` (itself a `temporalio-common` dependency) already solves exactly this for its own, smaller `.proto` compilation step - its `build.rs` supports a `vendored-protoc` feature (compiles real `protoc` from source via the `protobuf-src` crate) and a `vendored-protox` feature (uses `protox`, a pure-Rust reimplementation of the protobuf compiler, with zero system/toolchain dependencies at all). `temporalio-protos` has no equivalent option, and - because it declares `links = "temporalio_protos"` - nothing downstream can work around this via Cargo feature unification the way I could for `prost-wkt-types` itself. I tested this directly (removing `protoc` from `PATH`/`PROTOC` and unifying in `prost-wkt-types`'s own `vendored-protox` feature via a feature-only dependency) and confirmed it only fixes `prost-wkt-types`'s own build step - `temporalio-protos` still fails with "Could not find `protoc`".

Add the same `vendored-protox` (preferred - pure Rust, no C/C++ toolchain needed at all) and/or `vendored-protoc` (compiles real `protoc` from source via `protobuf-src`, needs a C++ toolchain but no separate binary) feature(s) to `temporalio-protos` that `prost-wkt-types` already offers, following the identical pattern in its own `build.rs`:

```rust
#[cfg(feature = "vendored-protoc")]
std::env::set_var("PROTOC", protobuf_src::protoc());
```

or, for `vendored-protox`, compiling the file descriptor set via `protox::compile(...)` directly and calling `.skip_protoc_run()` on the `prost_build::Config`, rather than letting `tonic_prost_build` invoke `protoc` itself.

This would let a downstream crate that already depends on `prost-wkt-types` (transitively, via `temporalio-common`) simply add a feature-only dependency edge to unify the feature workspace-wide, or - even better - let `temporalio-client`/`temporalio-common` expose their own passthrough feature (e.g. `temporalio-client = { version = "...", features = ["vendored-protox"] }`) that a consumer can turn on with no extra dependency at all.

### Additional context

Confirmed via direct testing against `temporalio-protos`/`temporalio-client`/`temporalio-common` 0.8.0 (still the latest on crates.io as of this writing). Happy to put together a PR for this if a `vendored-protox`/`vendored-protoc` feature pair along these lines would be welcome - the `prost-wkt-types` implementation is a solid, already-proven template to follow.

chris-olszewski:
## What was changed
Add `vendored-protox` feature to the protos crate and thread it through up to the client crate. Enabling the feature will switch the proto build to use `protox` instead of trying to find a system protoc install.

## Why?
Useful for avoiding setting up a protoc to use the Rust SDK.

## Checklist

1. Closes #1589 

2. How was this tested:
Verified locally, added a CI job that verifies we can build without a protoc.

3. Any docs updates needed?
Added README updates to the appropriate crates.

### DIFF
--- a/crates/protos/Cargo.toml
+++ b/crates/protos/Cargo.toml
@@ -17,6 +17,7 @@ links = "temporalio_protos"
 default = []
 serde_serialize = []
 grpc-clients = ["tonic/channel"]
+vendored-protox = ["dep:protox", "prost-types/vendored-protox"]
 
 [dependencies]
 anyhow = "1.0"
@@ -34,9 +35,9 @@ pbjson = { workspace = true }
 
 [build-dependencies]
 prost = { workspace = true }
-prost-types = "0.14"
 tonic-prost-build = { workspace = true }
 pbjson-build = { workspace = true }
+protox = { version = "0.9.1", optional = true }
 
 [lints]
 workspace = true
--- a/.github/workflows/per-pr.yml
+++ b/.github/workflows/per-pr.yml
@@ -32,6 +32,11 @@ jobs:
       - run: cargo test-lint
       - run: cargo check
       - run: cargo check --features experimental
+      - name: Check vendored proto compilation
+        run: cargo check --features temporalio-client/vendored-protox
+        env:
+          PROTOC: /does/not/exist
+          CARGO_TARGET_DIR: /tmp/vendored-protox-target
       - run: git diff --exit-code
 
   test:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ relevant information.
   `temporalio-sdk`, and Rust 1.88 for the other crates.
 
 ### Added
+* `temporalio-client` now provides a `vendored-protox` feature for compiling protobuf definitions
+  with `protox`, allowing client and Rust SDK builds without an installed `protoc`.
 * `CancelExternalWorkflowError` and `workflow_interceptors::CancelExternalWorkflowResult`
   for use in interceptors.
 * `WorkflowContextKey` and context-value scopes provide replay-safe, workflow-run-owned context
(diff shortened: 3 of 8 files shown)


crates/protos/Cargo.toml
```diff
@@ -34,9 +35,9 @@ pbjson = { workspace = true }
 
 [build-dependencies]
 prost = { workspace = true }
-prost-types = "0.14"
```
chris-olszewski, line 37: Note: our workspace aliases `prost-types` to `prost-wkt-types`. I think this dep was a mistake as it pulls in an unused and different `prost-types` crate. It was causing issues when we were trying to enable `vendored-protox` as the "real" `prost-types` crate doesn't have this feature.

Sushisource (approved):

PR #1591, opened 2026-09-04
feat(sdk): add OpenTelemetry tracing integration

eamsden:
## What was changed

- Add an optional `opentelemetry` feature to the main Rust SDK crate.
- Add a plugin that creates spans for client, Workflow, and Activity operations.
- Propagate W3C Trace Context and W3C Baggage in the cross-SDK `_tracer-data` header.
- Add replay-safe Workflow span IDs and suppress span export during replay.
- Preserve application span context across Workflow polls and calls.
- Add unit tests and a live-and-replay integration test.

This integration does not trace inbound Nexus handlers because the Rust SDK does not provide an interceptor for those handlers.

## Why?

The Rust SDK needs the same general OpenTelemetry behavior as other Temporal SDKs. Applications must be able to trace Temporal operations and propagate trace context across SDK boundaries.

## Checklist

1. Closes: N/A

2. How was this tested:

- `cargo lint`
- `cargo test-lint`
- `cargo test -p temporalio-sdk --features opentelemetry --lib`
- `cargo integ-test opentelemetry_plugin_parents_activity_to_application_workflow_span`
- `cargo doc -p temporalio-sdk --features opentelemetry --no-deps`

3. Any docs updates needed?
tbd

### DIFF
--- a/crates/sdk-core/tests/integ_tests/plugin_tests.rs
+++ b/crates/sdk-core/tests/integ_tests/plugin_tests.rs
@@ -1,8 +1,15 @@
 use crate::common::{get_integ_server_options, get_integ_telem_options, integ_namespace};
 use futures_util::future::BoxFuture;
+use opentelemetry::{
+    Context,
+    trace::{FutureExt as _, TraceContextExt, Tracer, TracerProvider as _},
+};
+use opentelemetry_sdk::trace::{
+    InMemorySpanExporter, SdkTracer, SdkTracerProvider, SimpleSpanProcessor,
+};
 use std::{
     sync::{
-        Arc,
+        Arc, OnceLock,
         atomic::{
             AtomicU8, AtomicUsize,
             Ordering::{self, Relaxed},
@@ -30,7 +37,9 @@ use temporalio_sdk::{
     WorkerPlugin, WorkflowContext, WorkflowDefinitions, WorkflowResult,
     activities::{ActivityContext, ActivityDefinitions, ActivityError},
     interceptors::WorkerInterceptor,
+    opentelemetry::{OpenTelemetryPlugin, WorkflowIdGenerator, WorkflowSpanProcessor},
     runtime::RuntimeOptions,
+    workflow_replayer::{WorkflowReplayer, WorkflowReplayerOptions},
 };
 use url::Url;
 use uuid::Uuid;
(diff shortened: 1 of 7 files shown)


crates/sdk/README.md
```diff
@@ -136,11 +136,46 @@ graph can disable defaults and opt back into the integrations they use.
...
 - `dynamic-tls`: Dynamic mTLS client-certificate resolution for transparent certificate rotation. |
 - `wasm-workflows`: Support WebAssembly workflow components through Wasmtime for workers and workflow replay. |
 
+### OpenTelemetry tracing
+
+Enable the `opentelemetry` feature. Configure an OpenTelemetry tracer provider. Add the plugin to
+the Temporal client. Workers that use the client automatically get the worker interceptors.
+
+```toml
+temporalio-sdk = { version = "1.0", features = ["opentelemetry"] }
```
chris-olszewski, line 151: Codex loves adding these to our README, but I don't think they're really helpful. Version string will be annoying to to keep up to date and I don't think it is necessary to show how to enable a feature.

crates/sdk-core/tests/integ_tests/plugin_tests.rs
```diff
@@ -186,6 +195,8 @@ struct SimplePluginWorkflow;
 
 struct SimplePluginActivities;
 
+static WORKFLOW_OTEL_TRACER: OnceLock<SdkTracer> = OnceLock::new();
+
```
Sushisource, line 199: This will end up shared across the two tests which seems wrong

crates/sdk/src/opentelemetry.rs
```diff
@@ -0,0 +1,1331 @@
...
+        });
+    }
+}
+
+/// An OpenTelemetry ID generator that uses replay-safe Workflow randomness.
+///
+/// Use this generator in the tracer provider for [`OpenTelemetryPlugin`]. This generator gives
+/// the same IDs to spans during execution and replay. Outside a Workflow context that requires
+/// replay safety, it uses the standard OpenTelemetry random ID generator.
+#[derive(Clone, Debug, Default)]
```
Sushisource, line 152: Is it possible to, when they call `with_tracer`, force the generator to be this?

crates/sdk/src/opentelemetry.rs
```diff
@@ -0,0 +1,1331 @@
...
+                .borrow()
+                .as_ref()
+                .and_then(|state| state.random.clone())
+            {
+                loop {
+                    let id = TraceId::from(random.random::<u128>());
+                    if id != TraceId::INVALID {
+                        return id;
+                    }
+                }
```
Sushisource, line 170: This is a bit odd. Why would it be invalid? Can we just clamp the random output or something if the issue is extreme values?

Sushisource (commented):
Overall this makes sense to me. I think we probably want to expand the integration test a bit to include more kinds of spans. I think there are some good examples in the TS SDK.

crates/sdk/src/opentelemetry.rs
chris-olszewski: IIRC the spec only classifies 0 as an invalid trace id

PR #1604, opened 2026-09-15
Fix PayloadsTooLarge errors spamming WFT failures

Sushisource:
## What was changed
See changelog

## Why?
This happened in the wild and was causing WFT failure spam against cloud.

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?


crates/sdk-core/src/worker/workflow/managed_run.rs
```diff
@@ -623,71 +625,66 @@ impl ManagedRun {
...
-        } else if should_report {
-            // Check if we should fail the workflow instead of the WFT because of user's preferences
-            if matches!(cause, WorkflowTaskFailedCause::NonDeterministicError)
-                && self.config.should_fail_workflow(
-                    &self.wfm.machines.workflow_type,
-                    &WorkflowErrorType::Nondeterminism,
+        let legacy_query = self.pending_work_is_legacy_query();
+        if legacy_query && is_no_report_query_fail {
+            self.reply_to_complete(ActivationCompleteOutcome::WFTFailedDontReport, resp_chan);
+            return rur;
```
jmaeagle99: This now early returns and does not increment the failure counter when it did before the change.

crates/sdk-core/src/worker/workflow/managed_run.rs
```diff
@@ -623,71 +625,66 @@ impl ManagedRun {
...
-        };
+                .unwrap_or_else(|e| {
+                    dbg_panic!("Got next page request when auto-failing workflow: {e:?}");
+                    None
+                });
+        }
 
-        self.metrics
-            .with_new_attrs([metrics::failure_reason(cause.into())])
-            .wf_task_failed();
```
jmaeagle99, line 689: This used to unconditionally emit the metrics at this point but now relies on `resp_chan` not being `None` and sending on that channel doesn't error. In both of those cases, the counter increment is dropped. The `None` case looks unreachable, but the missing receiver would cause the loss. Not necessarily a bug, but its now coupling metrics to the channel, just like the failure report.

crates/sdk-core/src/core_tests/workflow_tasks.rs
```diff
@@ -3286,3 +3287,174 @@ async fn grpc_message_too_large_doesnt_spam_task_fails() {
...
+            let mut err = tonic::Status::new(tonic::Code::InvalidArgument, violation.to_string());
+            err.set_source(Arc::new(violation));
+            times += 1;
+            Err(err)
+        } else {
+            Ok(Default::default())
+        }
+    }));
+    mh.expect_fail_wft_matcher =
+        Box::new(|_, cause, _| *cause == WorkflowTaskFailedCause::PayloadsTooLarge);
```
jmaeagle99: Should this test also atomically count and assert the attempts like the two below it?

crates/sdk-core/src/core_tests/workflow_tasks.rs
```diff
@@ -3286,3 +3287,174 @@ async fn grpc_message_too_large_doesnt_spam_task_fails() {
     core.drain_pollers_and_shutdown().await;
     // Mock only expects 1 task failure, and would fail here if we spammed
 }
+
+#[tokio::test]
+async fn payloads_too_large_doesnt_spam_task_fails() {
```
jmaeagle99, line 3303: I don't see a test that validates that the counter increments on attempts whose failure is not sent. Consider attaching a meter in this test to validate it is incremented to 6 or clone the test with a similar setup and validated it there.

crates/sdk-core/CHANGELOG.md
```diff
@@ -36,6 +36,12 @@ relevant information.
 ### Fixed
 * Workflow poll balancing now lets non-sticky pollers use capacity after sticky pollers reach their
   configured or autoscaled polling limit.
+* Every path that fails a workflow task now only reports the failure to server
```
jmaeagle99, line 39: This is also user-visible in the Rust SDK, so it should be duplicated into the root change log.

crates/sdk-core/src/worker/workflow/managed_run.rs
Sushisource: Good catch

jmaeagle99 (approved):

PR #1607, opened 2026-09-16
Update API definitions

tconley1428:
Updates the vendored temporalio/api subtree from API main, including the canonical System Nexus WIT annotation updates from temporalio/api#869.


crates/client/src/async_activity_handle.rs
```diff
@@ -282,6 +282,7 @@ impl<CT: WorkflowService + NamespacedClient + Clone> AsyncActivityHandle<CT> {
                                 identity: client.identity(),
                                 last_heartbeat_details,
                                 resource_id: Default::default(),
+                                ..Default::default()
```
Sushisource, line 285: Should this extract the cause somehow?

Sushisource (approved):

crates/client/src/async_activity_handle.rs
tconley1428: I don't _think_ so, but I have limited context.

tconley1428: @jmaeagle99 failure cause is your baby right? Any idea?

jmaeagle99: This is a deliberate call from a user to fail the activity. It cannot be a well-known failure cause, so it should be ACTIVITY_TASK_FAILED_CAUSE_UNSPECIFIED.

tconley1428: That's the default, so we good.

crates/protos/protos/api_upstream/temporal/api/enums/v1/failed_cause.proto
```diff
@@ -92,6 +92,24 @@ enum WorkflowTaskFailedCause {
     WORKFLOW_TASK_FAILED_CAUSE_REQUEST_TOO_LARGE = 40;
 }
 
+// Activity tasks can fail for various reasons. Note that some of these reasons can only originate
+// from the server, and some of them can only originate from the SDK/worker.
+enum ActivityTaskFailedCause {
```
jmaeagle99, line 97: Would you be able to replace https://github.com/temporalio/sdk-rust/blob/v1.0.0/crates/protos/protos/local/temporal/sdk/core/activity_result/activity_result.proto#L50 with this incoming type? If not, I'll do it in a separate pass.

tconley1428: Seems like a fair few changes. I'll do it in a followup just to split the PRs.

PR #1608, opened 2026-09-17
Generate payload visitor implementations in a stable order

mikek-mnx:
## What

`temporalio-common`'s build script generates `payload_visitor_impl.rs` in a non-deterministic order, so the file's content changes on every build.

Two iterations drive the output:

- `PayloadVisitorGenerator::generate` emits one implementation per entry of `PayloadModel::payload_containing`, a `HashSet<String>`.
- `build_field_infos` orders each message's fields by iterating `oneof_fields`, a `HashMap<i32, _>`.

Both use the default `RandomState`, which is seeded per process.

## Why it matters

The generated file is a compile input, so its content is part of what a compilation cache keys on. sccache misses `temporalio-common` on every build, and because each miss changes the crate's own output, it misses everything downstream too — `temporalio-client`, `temporalio-sdk-core` and `temporalio-sdk`, plus any user crate that links them. The eight crates recompile from source on every build no matter what changed.

Measured on a CI build of a workspace using the SDK: 1414 of 1422 cacheable compilations hit, and the eight that missed were exactly `temporalio-common` and its dependents — 26% of total compile time, on every run. Comparing two builds of the same commit, exactly eight of 580 hash keys differed, and `temporalio-common` was the root; every crate it depends on hashed identically.

## The change

Collect both into a `Vec` and sort before use. This matches what `generate_payload_limits_validator` in the same file already does for its roots (`roots.sort()`), its generated names (`generate_names.sort()`) and its own oneof indices (`oneof_indices.sort()`) — the visitor generator looks like it was simply missed.

## Validation

- Five consecutive build-script runs against the patched tree produce byte-identical `payload_visitor_impl.rs` and `payload_limits_impl.rs`.
- Three runs against the unpatched tree produce three different `payload_visitor_impl.rs`.
- The generated code itself is unchanged: sorting the lines of any unpatched output yields the same checksum as the patched output.

No test is included — the behaviour is a property of build-script output rather than of the crate's API, and I did not want to guess at how you would want that asserted in CI. Happy to add one if you have a preference.

I have also added a `CHANGELOG.md` entry under `Unreleased`, on the grounds that a Rust SDK user with a compilation cache observes this directly. If you read it as internal, say the word and I will drop it.


chris-olszewski:
@mikek-mnx Thanks for the fix, could you please sign the CLA so we can accept this change?

mikek-mnx:
Sure thing @chris-olszewski

chris-olszewski (approved):

